PR #1, opened 2014-04-10
Talking to synapse via the client for basic authentication actions. End-...

alxdarksage:
...to-end proof of design.


app/controllers/Authentication.java
```diff
@@ -1,40 +1,52 @@
...
+
 	public Result signIn() throws Exception {
-		Session session = synapseClient.login("", "");
-		String sessionToken = session.getSessionToken();
-		// check terms of use.
-		// set cookie with token.
+		// I cannot find a way, in Play, to convert the JSON payload to a bound Java object.
+		// serving as a form object. And I'm pretty sure the problem here is Play, not me.
+		// There are alternative code paths, but none work.
+		JsonNode body = request().body().asJson();
```
alxdarksage: Can stay with this, we can pull out what we want from the JSON into a form-like object. There are not many static objects like this.

app/controllers/Authentication.java
```diff
@@ -1,40 +1,52 @@
...
+		String email = body.get("email").asText();
+		authenticationService.resetPassword(email);
+		return jsonResult("An email has been sent allowing you to set a new password.");
+	}
+	
+	public Result getUserProfile() throws Exception {
+		String sessionToken = getSessionToken();
+		UserProfile profile = authenticationService.getUserProfile(sessionToken);
+		
+		return jsonResult(new JsonPayload<UserProfile>(profile));
```
alxdarksage, line 47: - Use unqualified name of class?
- do we need this?
- explicit good
- http status code should relate to type in a non-surprising manner. But some 500 statuses, for example, may come with  specific details from the exception that are useful to have

app/global/Global.java
```diff
@@ -17,18 +24,45 @@
...
+		} else if (throwable instanceof TermsOfUseException) {
+			// "Precondition Failed" - SynapseTermsOfUseException uses the forbidden code, not quite right I think
+			status = 412; 
+		}
+
+		JsonPayload<?> message = null;
+		if (throwable.getMessage() != null) {
+			message = new ExceptionMessage(throwable);
+		} else {
+			message = new StatusMessage("ServerError", "There has been a server error. We cannot fulfill your request at this time.");
```
alxdarksage: Why not use the throwable type here, as when the exception has a message?

conf/application.conf
```diff
@@ -61,7 +61,7 @@ application.global=global.Global
 # by providing an application-logger.xml file in the conf directory.
 
 # Root logger:
-logger.root=ERROR
```
alxdarksage, line 64: No play-based logging. Logback? Look at Heroku for any recommendations.

PR #3, opened 2014-04-18
Adding consent agreement, reset password, humane.js

alxdarksage:
There are tests, but there's more to do there. This is about all I'm going to do for authentication for the moment (the only thing missing is signing up initially. Existing accounts will see a consent page on first sign in).


app/controllers/BaseController.java
```diff
@@ -16,7 +16,7 @@
 
 	protected String getSessionToken(boolean throwException) throws Exception {
 		Cookie sessionCookie = request().cookie(BridgeConstants.SESSION_TOKEN);
-		if (sessionCookie != null && sessionCookie.value() != null) {
+		if (sessionCookie != null && sessionCookie.value() != null && !"".equals(sessionCookie.value())) {
```
alxdarksage, line 19: StringUtils.isBlank()/isEmpty()

app/global/Global.java
```diff
@@ -47,19 +48,24 @@ public void onStart(Application application) {
 		int status = 500;
 		if (throwable instanceof SynapseServerException) {
 			status = ((SynapseServerException)throwable).getStatusCode();
-		} else if (throwable instanceof TermsOfUseException) {
```
alxdarksage, line 50: Log this exception so we see it in the play framework

alxdarksage, line 50: Long term we want what amounts to a filter that delegates to a specific action method. + the annotation to add it. Also may have filters.

app/views/index.scala.html
```diff
@@ -5,17 +5,16 @@
     <meta charset="utf-8" />
     <title>Bridge: Patients &amp; Researchers in Partnership - Sage Bionetworks</title>
     <meta name="viewport" content="width=device-width" />
-    <link rel="stylesheet" href="http://fonts.googleapis.com/css?family=Archivo+Narrow" type="text/css" />
     <link rel="stylesheet" href="build/bridge.css" />
     <script>window.auth = @Html(auth);</script>
```
alxdarksage, line 9: namespace this variable?

public/app/scripts/app.js
```diff
@@ -44,9 +54,53 @@ angular.module('bridge', ['ngRoute', 'ui.bootstrap'])
...
+.directive('validateEquals', function() {
+    return {
+        restrict: 'A',
+        require: 'ngModel',
+        link: function(scope, element, attrs, controller) {
+            function getComparisonValue(expr) {
+                var comparisonModel = scope.$eval(expr);
+                return (comparisonModel && comparisonModel.$viewValue) ? comparisonModel.$viewValue : undefined;
+            }
+            // http://stackoverflow.com/questions/20982751/custom-form-validation-directive-to-compare-two-fields
```
alxdarksage, line 80: File a bug on the fact that it doesn't validate equality when you update the password field, only when you update the confirm field.

public/app/scripts/humane-modified.js
```diff
@@ -0,0 +1,234 @@
+(function (win, doc) {
```
alxdarksage, line 1: Any way to avoid this in order to fix the module imports, which seem to confuse Angular?

PR #12, opened 2014-04-25
Cleanup imports

alxdarksage:
- changed compare passwords directive so it works better 
- Added BridgeContext and environment switch to change Spring app context
- Simplified stub code based on environment facilities


app/org/sagebionetworks/bridge/services/AuthenticationServiceImpl.java
```diff
@@ -57,8 +59,14 @@ public UserSession getSession(String sessionToken) {
...
 	@Override
-	public void signOut(String sessionToken) throws Exception {
-		getSynapseClient(sessionToken).logout();
+	public void signOut(String sessionToken) {
+		// Synapse requires a session token, but it's not an error if it's missing
+		// (e.g. user has deleted the cookie).
+		try {
+			getSynapseClient(sessionToken).logout();	
+		} catch(Throwable e) {
+			logger.error(e.getMessage(), e);
```
alxdarksage: Log as a warning, it's not really an error.

PR #16, opened 2014-05-01
Deploy to Heroku from Travis CI

eric-wu:


public/Gruntfile.js
```diff
@@ -73,10 +74,8 @@ module.exports = function(grunt) {
...
-    
-    grunt.registerTask('precommit', ['jshint','clean','concat','sass','uglify','jasmine']);
-    grunt.registerTask('test', ['release']);
-    grunt.registerTask('default', ['jshint','clean','concat','sass','uglify']);
-    grunt.registerTask('release', ['jshint','clean','concat','sass','uglify','jasmine']);
-    
+
+    grunt.registerTask('test', ['build', 'jasmine']);
+    grunt.registerTask('build', ['jshint', 'clean:build', 'concat', 'sass', 'uglify']);
+    grunt.registerTask('release', ['test', 'clean:release']);
```
alxdarksage, line 80: If we just run release before deploying, we won't get the concatenated and minified build files, will we? We need those for the deployment to be functional. release might need to be ['test','build','clean:release']

PR #19, opened 2014-05-07
Health data API

alxdarksage:
This is a huge pull request, let's set aside some time to review and make notes on changes we want to make or need to make eventually. I just think this has grown too big as it is, and we have to start making smaller changes to this.


app/controllers/HealthData.java
```diff
@@ -0,0 +1,85 @@
...
+        UserSession session = getSession();
+        HealthDataKey key = new HealthDataKey(studyId, trackerId, session.getSessionToken());
+        List<HealthDataEntry> entries = healthDataService.getHealthDataByDateRange(key, new Date(date), new Date(date));
+        return jsonResult(new JsonPayload<List<HealthDataEntry>>(HealthDataEntry.class.getName(), entries));
+    }
+    
+    public Result getHealthDataByDateRange(Long studyId, Long trackerId, Long startDate, Long endDate) throws Exception {
+        UserSession session = getSession();
+        HealthDataKey key = new HealthDataKey(studyId, trackerId, session.getSessionToken());
+        List<HealthDataEntry> entries = healthDataService.getHealthDataByDateRange(key, new Date(startDate), new Date(endDate));
```
alxdarksage: What happens if you don't pass a date?

app/controllers/HealthData.java
```diff
@@ -0,0 +1,85 @@
+package controllers;
```
alxdarksage: Rename HealthDataController (maybe do all for consistency?)

app/org/sagebionetworks/bridge/context/BridgeContext.java
```diff
@@ -40,6 +55,14 @@ public String get(String key) {
 		return nameValueMap.get(key);
 	}
 	
+	public String getDecrypted(String key) {
+	    return encryptor.decrypt(nameValueMap.get(key));
+	}
+    
+    public String encryptValue(String value) {
```
alxdarksage: Remove

app/org/sagebionetworks/bridge/dynamodb/CredentialsFactory.java
```diff
@@ -0,0 +1,16 @@
+package org.sagebionetworks.bridge.dynamodb;
+
+import org.sagebionetworks.bridge.context.BridgeContext;
+
+import com.amazonaws.auth.BasicAWSCredentials;
+
+public class CredentialsFactory {
+    
+    public static BasicAWSCredentials getCredentials() {
```
alxdarksage, line 9: AWS has a provider class that does something like this: "CredentialProvider", there's one for basic credential provision.

app/org/sagebionetworks/bridge/services/HealthDataServiceImpl.java
```diff
@@ -0,0 +1,302 @@
...
+    }
+    
+    public void setBeanFactory(BeanFactory factory) {
+        this.beanFactory = factory;
+    }
+
+    @Override
+    public void afterPropertiesSet() throws Exception {
+        BridgeContext context = new BridgeContext();
+        encryptor.setPassword(context.get(BridgeContext.PASSWORD));
```
alxdarksage: And if the password were changed? If compromised? We need a way to do this that doesn't involve a password. John mentioned something like SHA256?

PR #27, opened 2014-05-16
Removing application-context-per-environment and configuring through Bri...

alxdarksage:
...dgeConfig


app/org/sagebionetworks/bridge/synapse/SynapseClientFactory.java
```diff
@@ -0,0 +1,22 @@
...
+
+import org.sagebionetworks.bridge.config.BridgeConfig;
+import org.sagebionetworks.client.SynapseClient;
+import org.sagebionetworks.client.SynapseClientImpl;
+import org.sagebionetworks.repo.model.DomainType;
+
+public class SynapseClientFactory {
+
+    public static SynapseClient createSynapseClient(String repoEndpoint, String authEndpoint, String fileEndpoint) {
+        if (new BridgeConfig().isStub()) {
```
eric-wu: I wonder if we could take BridgeConfig as the method parameter.  That will allow us to inject the 'bridgeConfig' singleton instead of creating a new one.

conf/bridge.conf
```diff
@@ -4,3 +4,13 @@ bridge.salt=fake.salt
 
 aws.key=ENC(lcFykS2Y4y4y68uUWNm+0Q==)
 aws.secret.key=ENC(lcFykS2Y4y7MEK+bwxtBTutfQyQzSMZO)
+
+# default values (ignored by stub implementation, however)
+synapse.repo.endpoint = https://repo-prod.prod.sagebase.org/repo/v1
+synapse.auth.endpoint = https://repo-prod.prod.sagebase.org/repo/v1
```
eric-wu: It seems a typo here. The auth endpoint should be https://repo-prod.prod.sagebase.org/auth/v1

PR #33, opened 2014-05-23
Set up config for Travis CI

eric-wu:
Getting an error from Synapse Prod:

org.sagebionetworks.client.exceptions.SynapseNotFoundException: Did not find a user with alias: bob (or test2)

Have we ever set up these test users in Synapse prod?


eric-wu:
I think I got it working. But we need to merge it to confirm it.  According to Travis CI, "Please note that encrypted environment variables are not available for pull requests from forks."

PR #37, opened 2014-05-29
Signin with different tests

alxdarksage:


app/views/index.scala.html
```diff
@@ -50,7 +45,7 @@
 
     <script src="//ajax.googleapis.com/ajax/libs/angularjs/1.2.15/angular.min.js"></script>
     <script src="//ajax.googleapis.com/ajax/libs/angularjs/1.2.15/angular-route.js"></script>
-    <script src="build/bridge.min.js"></script>
+    <script src="build/bridge.js"></script>
```
eric-wu: Is this what we want?

PR #72, opened 2014-06-16
Replace the default encryption with Bouncy Castle's AES

eric-wu:


eric-wu:
For Mac OS, must download and install Java Cryptography Extension (JCE) Unlimited Strength Jurisdiction Policy Files.

PR #76, opened 2014-06-19
Testinfra

alxdarksage:
- fixed authentication state and behavior between app/ and neurod/
- that led to fixing code that was copied between the apps, so this was factored out to a shared AngularJS module


public/app/scripts/app.js
```diff
@@ -1,71 +1,39 @@
-var bridge = angular.module('bridge', ['ngRoute', 'ngCookies', 'ui.bootstrap'])
+var bridge = angular.module('bridge', ['bridge.auth', 'ngRoute', 'ui.bootstrap'])
```
alxdarksage, line 1: bridge.shared is still named with the module name bridge.auth, and it includes more shared code than auth code at this point.

public/app/scripts/services/signup-service.js
```diff
@@ -11,6 +11,7 @@ bridge.service('signUpService', ['$modal', function($modal) {
             if ($scope.signUpForm.$valid) {
```
alxdarksage, line 11: We really don't do this anymore. Could remove.

public/neurod/scripts/controllers/neurod.js
```diff
@@ -10,7 +10,7 @@ if ( !window.requestAnimationFrame ) {
     } )();
 }
 
-var neurod = angular.module('neurod', ['ngRoute', 'ui.bootstrap']);
+var neurod = angular.module('neurod', ['bridge.auth', 'ngRoute', 'ui.bootstrap']);
```
alxdarksage, line 13: rename module to bridge.shared

public/shared/scripts/humane.js
```diff
@@ -1,4 +1,4 @@
-neurod.service('$humane', ['$window', function($window) {
```
alxdarksage, line 1: Rename bridgeAuth --> bridgeShared

PR #77, opened 2014-06-20
Security update

eric-wu:
Still in process but I want to show you the changes before you go away on vacation.


app/org/sagebionetworks/bridge/crypto/EncryptorUtil.java
```diff
@@ -43,12 +42,12 @@ public static PBEStringEncryptor getEncryptorOld(final String password, final St
     /**
      * Gets a password-based string encryptor.
      */
-    public static PBEStringEncryptor getEncryptor(final String password, final String salt) {
```
alxdarksage, line 46: To make a singleton, convert to Spring bean?

PR #79, opened 2014-06-26
Further implementation of the new security strategy

eric-wu:


test/org/sagebionetworks/bridge/services/HealthCodeServiceImplTest.java
```diff
@@ -0,0 +1,48 @@
...
+    @After
+    public void after() {
+        clearDynamo();
+    }
+
+    @Test
+    public void test() {
+        HealthId healthId = healthCodeService.create();
+        assertNotNull(healthId);
+        assertEquals(healthId.getCode(),
```
jay-hodgson, line 40: test that the next id is different

PR #81, opened 2014-07-08
Use the new encryption of health code

eric-wu:


app/org/sagebionetworks/bridge/BridgeConstants.java
```diff
@@ -1,14 +1,16 @@
...
-	
-	public static final String CONSENT_REQUIRED_MESSAGE = "Before you may continue, we require you to consent to participate in this study.";
+
+    public static final String SESSION_TOKEN_HEADER = "Bridge-Session";
+
+    public static final String CUSTOM_DATA_CONSENT_SUFFIX = "_consent";
+
+    public static final String CUSTOM_DATA_HEALTH_CODE_SUFFIX = "_code";
+
+    public static final String CUSTOM_DATA_VERSION = "version";
```
alxdarksage, line 11: Versioning is a bigger issue, e.g. the encryption of a health data code is scoped to a study, and this is not. So we might have a transactional issue updating multiple health codes.

PR #85, opened 2014-07-11
Consent wizard

alxdarksage:
Complete but does not have all the assents, and doesn't have a desktop design (always appears as it will appear in a mobile WebView).


public/consent/scripts/consent.js
```diff
@@ -0,0 +1,185 @@
...
+    
+    $scope.nextStep = function() {
+        var index = steps.indexOf($scope.step) + 1;
+        if (steps[index]) {
+            $scope.step = steps[index];
+            $scope.setStep($scope.step);
+        }
+    };
+
+    $scope.setStep("thankyou");
```
alxdarksage: Change this to the first card...

PR #96, opened 2014-07-18
Backfill health codes

eric-wu:


conf/routes
```diff
@@ -30,6 +30,9 @@ GET    /api/healthdata/:trackerId                       @controllers.HealthDataC
 GET    /api/trackers                   @controllers.TrackerController.getTrackers
 GET    /api/trackers/schema/:trackerId @controllers.TrackerController.getTrackerSchema(trackerId: Long)
 
+# Backfill
+GET    /api/backfill/healthId          @controllers.BackfillController.resetHealthId
```
alxdarksage: Should we put under an /admin/ namespace in case we want to put URL-based security on these at a later date?

app/org/sagebionetworks/bridge/services/HealthCodeService.java
```diff
@@ -7,4 +7,6 @@
     HealthId create();
 
     String getHealthCode(String id);
+
+    String resetHealthId(String healthCode);
```
alxdarksage, line 11: In implementation, let's make the log level ERROR if the UUID generator actually generates an existing UUID.

PR #113, opened 2014-07-31
Fix package versions

eric-wu:
This has fixed the test failures I have been having for weeks!!!

### DIFF
--- a/public/app/package.json
+++ b/public/app/package.json
@@ -9,9 +9,9 @@
   "author": "Sage Bionetworks",
   "license": "MIT",
   "devDependencies": {
+    "lodash": "^2.4.1",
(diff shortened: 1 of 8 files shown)


public/app/package.json
```diff
@@ -9,9 +9,9 @@
   "author": "Sage Bionetworks",
   "license": "MIT",
   "devDependencies": {
+    "lodash": "^2.4.1",
     "grunt-cli": "^0.1.13",
-    "tmp": "0.0.23",
-    "bower": "1.3.6",
+    "bower": "^1.3.8",
     "grunt": "^0.4.4",
```
xschildw, line 15: Should be 0.4.5

PR #121, opened 2014-08-01
Add DynamoDB implementation of consent

eric-wu:


app/org/sagebionetworks/bridge/dynamodb/DynamoUserConsentDao.java
```diff
@@ -0,0 +1,76 @@
...
+                ConsistentReads.CONSISTENT,
+                TableNameOverrideFactory.getTableNameOverride(DynamoUserConsent.class));
+        mapper = new DynamoDBMapper(client, mapperConfig);
+    }
+
+    @Override
+    public void giveConsent(String healthCode, StudyConsent studyConsent) {
+        DynamoUserConsent consent = new DynamoUserConsent();
+        consent.setHealthCodeStudy(healthCode + ":" + studyConsent.getStudyKey());
+        consent.setConsentTimestamp(studyConsent.getTimestamp());
```
jay-hodgson, line 34: If we promote the consent timestamp to be part of the hash key, we can still keep the timestamp attribute to avoid parsing.

app/org/sagebionetworks/bridge/dynamodb/DynamoUserConsentDao.java
```diff
@@ -0,0 +1,76 @@
...
+        consent.setGive(DateTime.now(DateTimeZone.UTC).getMillis());
```
jay-hodgson, line 35: When giving consent, we will need to null out the withdraw field.

app/org/sagebionetworks/bridge/dynamodb/DynamoUserConsentDao.java
```diff
@@ -0,0 +1,76 @@
...
+        mapper.save(consent);
```
jay-hodgson, line 36: What if a user resent the already consented consent?

PR #138, opened 2014-08-08
UserAdminService, mostly for tests.

alxdarksage:


app/org/sagebionetworks/bridge/services/UserAdminService.java
```diff
@@ -0,0 +1,56 @@
...
+     * 
+     * @param sessionToken
+     *            session of the admin user
+     * @param userStudy
+     *            the study of the target user
+     * @param userHref
+     *            the target user's Storm Path HREF value
+     * @throws BridgeServiceException
+     */
+    public void revokeAllConsentRecords(String sessionToken, Study userStudy, String targetUserSessionToken, String userHref) throws BridgeServiceException;
```
alxdarksage: Really do we need all of this? Can't we get storm path href from the session?

app/org/sagebionetworks/bridge/services/StormPathUserAdminService.java
```diff
@@ -0,0 +1,126 @@
...
+                return consentService.give(e.getUserSession().getSessionToken(), consent, userStudy, false);
+            } else {
+                throw e;
+            }
+        }
+        return newUserSession;
+    }
+
+    @Override
+    public void revokeAllConsentRecords(String sessionToken, Study study, String targetUserSessionToken, String userHref) throws BridgeServiceException {
```
alxdarksage: Well, if you manipulate the session here, maybe you should return it here as well.

PR #168, opened 2014-08-26
Better Vagrant performance

eric-wu:


eric-wu:
https://docs.vagrantup.com/v2/synced-folders/nfs.html

PR #192, opened 2014-09-09
Refactoring

alxdarksage:
- cleaned up exceptions, created generic exceptions for entity CRUD errors
- consolidated models reported in these exceptions under a BridgeEntity marker interface
- renamed BaseController session methods for clarity
- consolidated StudyControllerService to BaseController and moved it to be a normal service named StudyService
- deleted unused ExceptionInterceptorAction
- renamed post JSON field constants for consistency
- moved fromJson() methods to use JsonUtils for better exception/error case handling
- moved validation out of StudyConsentForm to the StudyConsentService for consistency
- removed role check from UserAdminService. Role checking will happen in controller if one is created


app/controllers/BaseController.java
```diff
@@ -103,6 +110,14 @@ protected void updateSessionUser(UserSession session, User user) {
         cacheProvider.setUserSession(session.getSessionToken(), session);
     }
 
+    protected String getHostname() {
+        String host = request().host();
```
alxdarksage, line 114: toLowerCase()

PR #194, opened 2014-09-11
Tests speed up

dr-willikers:
changed tests so that they run faster. I believe I was able to shave off at least a couple minutes off the total run time. The bad news is that I believe the largest bottleneck is refreshing the application context each time we start a test suite, so it might  be possible to speed things up further by removing that and using only mocks.

### DIFF
--- a/test/controllers/SurveyControllerTest.java
+++ b/test/controllers/SurveyControllerTest.java
@@ -3,17 +3,17 @@ package controllers;
 import static org.apache.http.HttpStatus.SC_FORBIDDEN;
 import static org.apache.http.HttpStatus.SC_OK;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;
-import static org.sagebionetworks.bridge.TestConstants.TIMEOUT;
-import static org.sagebionetworks.bridge.TestConstants.SURVEYS_URL;
 import static org.sagebionetworks.bridge.TestConstants.GET_SURVEY_URL;
 import static org.sagebionetworks.bridge.TestConstants.GET_VERSIONS_OF_SURVEY_URL;
-import static org.sagebionetworks.bridge.TestConstants.RECENT_SURVEYS_URL;
-import static org.sagebionetworks.bridge.TestConstants.RECENT_PUBLISHED_SURVEYS_URL;
-import static org.sagebionetworks.bridge.TestConstants.VERSION_SURVEY_URL;
 import static org.sagebionetworks.bridge.TestConstants.PUBLISH_SURVEY_URL;
+import static org.sagebionetworks.bridge.TestConstants.RECENT_PUBLISHED_SURVEYS_URL;
+import static org.sagebionetworks.bridge.TestConstants.RECENT_SURVEYS_URL;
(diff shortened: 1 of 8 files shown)


test/controllers/SurveyControllerTest.java
```diff
@@ -61,21 +60,23 @@
...
 
     @Before
     public void before() {
-        study = studyService.getStudyByHostname("localhost");
-        roles = Lists.newArrayList(study.getKey()+"_researcher");
-        List<Survey> surveys = surveyDao.getSurveys(study.getKey());
-        for (Survey survey : surveys) {
-            surveyDao.closeSurvey(survey.getGuid(), survey.getVersionedOn());
-            surveyDao.deleteSurvey(survey.getGuid(), survey.getVersionedOn());
+        if (!setUpComplete) {
```
alxdarksage, line 68: I think this is what @BeforeClass is for. It will avoid the boolean flag, let's use that unless you know there's an issue with it.

test/org/sagebionetworks/bridge/dynamodb/DynamoStudyConsentDaoTest.java
```diff
@@ -26,13 +25,14 @@
...
+    public static void initialSetUp() {
+        DynamoInitializer.init("org.sagebionetworks.bridge.dynamodb");
         DynamoTestUtil.clearTable(DynamoStudyConsent1.class, "active", "path", "minAge", "version");
     }
 
-    @After
-    public void after() {
+    @AfterClass
+    public static void finalCleanUp() {
         DynamoTestUtil.clearTable(DynamoStudyConsent1.class, "active", "path", "minAge", "version");
```
alxdarksage, line 36: I have wondered if we can do this once, at the beginning of the test. I think it's okay to leave stuff in the db at the end of the test...

test/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDaoTest.java
```diff
@@ -59,206 +60,206 @@ private Survey constructTestSurvey(String name) {
...
+
         survey = surveyDao.createSurvey(survey);
         surveyDao.createSurvey(survey);
     }
-    
+
     @Test
     public void crudSurvey() {
         Survey survey = constructTestSurvey();
-
```
alxdarksage, line 101: I think you have a habit of deleting Eclipses spaces on empty lines...

test/org/sagebionetworks/bridge/dynamodb/DynamoUserConsentDaoTest.java
```diff
@@ -33,8 +33,8 @@ public void before() {
                 "healthCode", "consentCreatedOn", "version");
     }
 
-    @After
-    public void after() {
+    @AfterClass
+    public static void finalCleanUp() {
```
alxdarksage, line 37: Let's check this in but I think eventually, we may not need to delete from the db at the end of the test.

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
```diff
@@ -46,14 +50,25 @@
...
     Client stormpathClient;
     
+    private static final int NUMBER_OF_TESTS = 12;
+    
+    private boolean setUpComplete = false;
+    private int testCount = 0;
+    
     @Before
     public void before() {
-        helper.createOneUser();
```
alxdarksage, line 51: Why not BeforeClass/AfterClass?

PR #197, opened 2014-09-13
Multiple values are now a separate question type...

alxdarksage:
... this simplifies everything, including the types of UI Hints there are for each type of question, and constraints are either specific to a type, or an enumeration if there are multiple values. This is simpler and doesn't model as much, but it does cover everything I've seen so far. Will probably need to add a flag to the multiple value type as to whether the user can enter another value or not ("Other" category).

### DIFF
--- a/app/org/sagebionetworks/bridge/models/surveys/MultiValueConstraints.java
+++ b/app/org/sagebionetworks/bridge/models/surveys/MultiValueConstraints.java
@@ -1,30 +1,37 @@
 package org.sagebionetworks.bridge.models.surveys;
 
+import java.util.EnumSet;
 import java.util.List;
 
-import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 
-/**
- * These are constraints that can have all their values enumerated. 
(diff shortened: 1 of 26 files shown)


app/org/sagebionetworks/bridge/models/surveys/MultiValueConstraints.java
```diff
@@ -0,0 +1,83 @@
+package org.sagebionetworks.bridge.models.surveys;
+
+import java.util.EnumSet;
+import java.util.List;
```
alxdarksage, line 4: This dictates an enumeration but can they always submit multiple values back to the server?

app/org/sagebionetworks/bridge/models/surveys/UIHint.java
```diff
@@ -1,18 +1,38 @@
...
+    TIMEPICKER(false),
+    TOGGLE(true);
+    
+    private boolean requiresEnumeration;
+    
+    private UIHint(boolean requiresEnumeration) {
+        this.requiresEnumeration = requiresEnumeration;
+    }
+    
+    public boolean requiresEnumeration(String dataType) {
```
alxdarksage, line 25: Remove this: it's not used.

PR #223, opened 2014-09-30
HTTP redirect to HTTPS; gzipping

eric-wu:


app/global/GlobalWithFilters.scala
```diff
@@ -0,0 +1,73 @@
...
+import play.libs.Json
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
+
+import scala.concurrent.Future
+
+object GlobalWithFilters extends WithFilters (
+
+    // TODO: CSRF filter
+
+    new GzipFilter(shouldGzip = (request, response) =>
```
alxdarksage, line 29: Could we add svg, html, js and css resources as well? (basically that's everything but jpg/gifs).

test/global/GlobalWithFiltersSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+
+@RunWith(classOf[JUnitRunner])
+object GlobalWithFiltersSpec extends PlaySpecification {
+
+  "gzip filter" should {
+    "do gzip on JSON response" in new WithApplication {
+      val jsonAction = GlobalWithFilters.doFilter(Action {
+          Results.Ok(Json.obj("name" -> "bob", "age" -> 31))
+        })
+      val request = FakeRequest().withHeaders(ACCEPT_ENCODING -> "gzip")
```
alxdarksage, line 20: Verify without gzip accept encoding header, no compression occurs.

PR #236, opened 2014-10-03
Read from the DynamoDB backend for user consent.

eric-wu:


app/controllers/ConsentController.java
```diff
@@ -17,44 +17,34 @@ public void setConsentService(ConsentService consentService) {
...
-        Study study = studyService.getStudyByHostname(getHostname());
-
-        User user = consentService.consentToResearch(session.getUser(), consent, study, true);
-        
+        final UserSession session = getAuthenticatedSession();
+        final ConsentSignature consent = ConsentSignature.fromJson(requestToJSON(request()));
+        final Study study = studyService.getStudyByHostname(getHostname());
+        final User user = consentService.consentToResearch(session.getUser(), consent, study, true);
         updateSessionUser(session, user);
-        setSessionToken(session.getSessionToken());
```
alxdarksage, line 27: We need to set the session token in the cookie here, so the client can redirect without losing the session, keeping the user signed in through the process of giving consent. (This is the web client)

PR #260, opened 2014-10-14
Fixing ScheduleChangeWorker; updating Jackson

alxdarksage:
- updated Jackson to version that fixed a bug where nulls were still produced on properties that used a custom (de)serializer, even if this was disabled in the mapper
- Fixed schedule change worker's locking model; now on a lock, waits and tries again without retrying via the executor, and locks are on plan and user. Much more reliable, tested by creating 1000 plans and then deleting them, as fast as possible. Schedules were all deleted correctly.


app/org/sagebionetworks/bridge/json/DateTimeJsonSerializer.java
```diff
@@ -12,11 +12,14 @@
...
-            jgen.writeString(DateUtils.convertToISODateTime(millisFromEpoch));
-        } else {
-            jgen.writeNull(); // without this errors.
-        }
+        jgen.writeString(DateUtils.convertToISODateTime(millisFromEpoch));
     }
     
+    @Override
+    public boolean isEmpty(Long millisFromEpoch) {
+        return (millisFromEpoch == null || millisFromEpoch == 0L);
```
alxdarksage, line 20: Eric is saying, 0 is a valid value, if rare, so we shouldn't treat it as "null".

app/org/sagebionetworks/bridge/redis/RedisDistributedLockDao.java
```diff
@@ -38,7 +38,7 @@ public void releaseLock(Class<?> clazz, String identifier, String lockId) {
             String key = createKey(clazz, identifier);
             String getResult = stringOps.get(key).execute();
             if (getResult != null && !getResult.equals(lockId)) {
-                throw new BridgeServiceException("Must be lock owner to release lock.");
+                return;
```
alxdarksage, line 41: You should return true/false as to whether you were able to release the lock or not.

PR #261, opened 2014-10-16
Survey fix

alxdarksage:


alxdarksage:
Fixing errors Erin encountered when trying to submit survey answers. Jackson was deserializing the objects in a way I hadn't expected or tested, so I've added a test and rewritten the classes to match what is done by Jackson, also changing types to deal with data ambiguities that can't be captured by conversion (this might extend to all the survey answers eventually, I've sent an email about this).

app/controllers/SurveyController.java
```diff
@@ -45,8 +45,7 @@ public Result getMostRecentlyPublishedSurveys() throws Exception {
     }
     
     public Result getSurveyForUser(String surveyGuid, String versionString) throws Exception {
-        Study study = studyService.getStudyByHostname(getHostname());
-        getAuthenticatedResearcherOrAdminSession(study);
+        getAuthenticatedSession();
```
alxdarksage: And consented!

alxdarksage:
Reviewed by Eric.

PR #273, opened 2014-10-21
Up the DynamoDB capacity to the newly updated free tier

eric-wu:


eric-wu:
The tables have already been updated with the new throughput values.

PR #294, opened 2014-10-31
Exposing sharing

alxdarksage:


app/org/sagebionetworks/bridge/dao/UserConsentDao.java
```diff
@@ -27,6 +28,14 @@
     boolean hasConsented(String healthCode, StudyConsent consent);
 
     /**
+     * Get the user consent record that consents the user to this study.
```
alxdarksage, line 31: Refactor entire ConsentService based on conversation with Eric.

PR #302, opened 2014-11-05
Throw error when client sends existing entity to create URL

alxdarksage:


app/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -81,4 +84,10 @@ public static Long parseLong(String value) {
         }
     }
     
+    public static void checkNewEntity(BridgeEntity entity, Object field, String message) {
+        if (field != null) {
+            throw new InvalidEntityException(entity, message);
```
alxdarksage: Throw EntityAlreadyExistsException(). Document the 409 response in the documentation.

alxdarksage: If it's an InvalidEntityException, Eric would like to log the raw JSON request payload, which we can do. So also, include this exception in logging it's not a "quiet exception"

PR #308, opened 2014-11-11
Moving study models to study package; creating Dynamo Study and DAO

alxdarksage:


test/org/sagebionetworks/bridge/dynamodb/DynamoStudyDaoTest.java
```diff
@@ -0,0 +1,126 @@
...
+        System.out.println(BridgeObjectMapper.get().writeValueAsString(study));
+    }
+    
+    private Study2 createStudy() {
+        Study2 study = new DynamoStudy();
+        study.setIdentifier(RandomStringUtils.randomAlphabetic(5));
+        study.setMaxParticipants(100);
+        study.setMinAgeOfConsent(18);
+        study.setName(RandomStringUtils.randomAlphabetic(5));
+        study.setResearcherRole("researcher");
```
eric-wu, line 120: Does the list of researcher roles also have 'admin', 'backfill'?  What are the list of research roles?

app/org/sagebionetworks/bridge/dynamodb/DynamoStudyDao.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public void setDynamoDbClient(AmazonDynamoDB client) {
+        DynamoDBMapperConfig mapperConfig = new DynamoDBMapperConfig.Builder().withSaveBehavior(SaveBehavior.UPDATE)
+                .withConsistentReads(ConsistentReads.CONSISTENT)
+                .withTableNameOverride(TableNameOverrideFactory.getTableNameOverride(DynamoStudy.class)).build();
+        mapper = new DynamoDBMapper(client, mapperConfig);
+    }
+
+    @Override
+    public boolean isStudyIdentifierUnique(String identifier) {
```
eric-wu: Not clear about the purpose of this method.  The moment you check if the identifier exists is not the same moment you set the identifier.  There is the possibility the id does not exist yet when you call this method. But when you decide to set it thinking it is unique, another thread or process has already set it for something else.  Unless you lock on the ID, calling this method does not avoid race conditions.

Without this method, you could just rely on the DynamoDB version attribute to do optimistic locking.  When you grab a new random ID, you set it right away, if the id has already existed in DynamoDB, you will get back a conflicting modification exception and you try another random ID until you get one without exceptions.  The locking on the version attribute guarantees that you are free of race conditions.

eric-wu: If we will do locking somewhere else, I don't see the need of this method.

No, I don't think DynamoDB "assumes" identity of objects.  There is no equivalent of 'INSERT' in DynamoDB.  So either a new object is created or an existing one is replaced for the PutItem API.  The Java SDK wraps it up with an additional version attribute.  Let's say you want to save an object with the hash key "123", your new object will look like {"hash": 123, "version":  null}.  If there is no hash key 123 in DynamoDB yet, a new object is created with version 1 in DynamoDB {"hash": 123, "version": 1}.  Then if you try to create a new one using the same hash {"hash": 123, "version":  null}, it will fail because the version does not match. The nice thing about the version attribute is that it is designed for optimistic locking.  If you use it for optimistic locking, it will do the work for you.

eric-wu: I'm looking at the PutItem API doc (http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_PutItem.html).  I am betting that the DynamoDB SDK uses the conditional PutItem to implement around the version attribute.  So this locking works as it happens on the DynamoDB side not the app side.

PR #310, opened 2014-11-12
Add pem utils for private keys, upgrade Play to 2.2.5

eric-wu:


app/org/sagebionetworks/bridge/crypto/PemUtils.java
```diff
@@ -0,0 +1,125 @@
...
+        StringReader stringReader = null;
+        try {
+            stringReader = new StringReader(text);
+            bufferedReader = new BufferedReader(stringReader);
+            String line = bufferedReader.readLine();
+            while (line != null) {
+                lines.add(line);
+                line = bufferedReader.readLine();
+            }
+        } finally {
```
eric-wu: Java 7 can auto-close resources for you

app/org/sagebionetworks/bridge/crypto/PemUtils.java
```diff
@@ -0,0 +1,117 @@
...
+            if (begin.equals(line)) {
+                addLine = true;
+            }
+        }
+        Base64 encoder = new Base64(Base64.PEM_CHUNK_SIZE);
+        byte[] decoded = encoder.decode(base64Encoded.toString());
+        return decoded;
+    }
+
+    private static List<String> readLines(String text) throws IOException {
```
DwayneJengSage: Would it be cleaner to replace this with String.split()? This way, we can avoid having to create streams and readers. If we're worried about windows newlines vs unix newlines, we could just use a regex that looks like "\r?\n".

app/org/sagebionetworks/bridge/crypto/PemUtils.java
```diff
@@ -0,0 +1,117 @@
...
+        } catch (InvalidKeySpecException e) {
+            throw new RuntimeException(e);
+        } catch (NoSuchProviderException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private static byte[] decode(List<String> lines, String begin, String end) {
+        checkNotNull(lines, "Lines cannot be null");
+        checkArgument(lines.size() >= 2, "Two few lines.");
```
DwayneJengSage: Nitpick: Typo - "too"

app/org/sagebionetworks/bridge/crypto/PemUtils.java
eric-wu: You are right "\r?\n" is what BufferedReader does and is what I need here.  I can certainly replace it with String.split().

DwayneJengSage:
Looks good to me.

PR #328, opened 2014-11-20
Consent Signature Image (REST API, DynamoDB object model, Send Mail)

DwayneJengSage:
Server-side implementation of Consent Signature Image. Includes updating the object models used by the REST API and by DynamoDB and updating the send mail service to write the signature as an embedded Base64 image.

Tested this with play compile and play testOnly for the tests I changed. Still need to run play test and play run to test the changes end-to-end.


app/controllers/ConsentController.java
```diff
@@ -26,7 +26,11 @@ public void setOptionsService(ParticipantOptionsService optionsService) {
 
     public Result give() throws Exception {
         final UserSession session = getAuthenticatedSession();
+
+        // ConsentSignature validates name and birthdate, but for backwards compatibility, does not validate image.
```
alxdarksage: Some studies may collect a signature, but some may not. It's okay to let the client validate this until we're sure this 100% required for all consents (IMO).

app/org/sagebionetworks/bridge/models/studies/ConsentSignature.java
```diff
@@ -1,65 +1,96 @@
...
+import org.sagebionetworks.bridge.json.JsonUtils;
+import org.sagebionetworks.bridge.models.BridgeEntity;
 
 public class ConsentSignature implements BridgeEntity {
 
     private static final String NAME_FIELD = "name";
     private static final String BIRTHDATE_FIELD = "birthdate";
+    private static final String IMAGE_FIELD = "image";
+
+    private final @Nonnull String name;
```
alxdarksage, line 22: If we have annotations for stuff like this, the validators could validate it. Otherwise, I'm not sure what will prevent an image set without a mime type, but you may have handled this as a validation case.

app/org/sagebionetworks/bridge/models/studies/ConsentSignature.java
```diff
@@ -1,65 +1,96 @@
...
+     *         if the JSON contains invalid fields
+     */
+    public static ConsentSignature fromJson(JsonNode node) throws InvalidEntityException {
         String name = JsonUtils.asText(node, NAME_FIELD);
         String birthdate = JsonUtils.asText(node, BIRTHDATE_FIELD);
-        return new ConsentSignature(name, birthdate);
+        ObjectNode imageNode = JsonUtils.asObjectNode(node, IMAGE_FIELD);
+
+        if (name == null || name.isEmpty()) {
+            throw new InvalidEntityException("name must be specified");
```
alxdarksage: I would rather allow an object like this to safely collect what input has been sent from the client, and then validate the object, returning errors for everything that is wrong in one message. Just for the usability of the API, as people try and integrate with it, it is nice to get everything back that is wrong at once, rather than an IEE for a single problem, then another, etc. I don't know if the pattern was followed here, but the consent service can validate the object before calling any DAOs.

app/org/sagebionetworks/bridge/models/studies/ConsentSignatureImage.java
```diff
@@ -0,0 +1,75 @@
...
+     *         JSON node to parse
+     * @return consent signature image
+     * @throws InvalidEntityException
+     *         if the JSON contains invalid fields
+     */
+    public static ConsentSignatureImage fromJson(JsonNode jsonNode) throws InvalidEntityException {
+        String data = JsonUtils.asText(jsonNode, DATA_FIELD);
+        String mimeType = JsonUtils.asText(jsonNode, MIME_TYPE_FIELD);
+
+        if (data == null || data.isEmpty()) {
```
alxdarksage: Same comment as above _except_ I think this could be validated as part of the signature object when it is based on user input.

app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
```diff
@@ -69,9 +68,6 @@ public User consentToResearch(final User caller, final ConsentSignature consentS
         if (caller.doesConsent()) {
             throw new EntityAlreadyExistsException(consentSignature);
         }
-        if (consentSignature.getBirthdate() == null) {
```
alxdarksage, line 72: There's now enough here that there could be a validator... I see there wasn't one here, as initially we weren't checking for much.

test/org/sagebionetworks/bridge/dynamodb/DynamoUserConsentDaoTest.java
```diff
@@ -31,7 +37,7 @@ public static void initialSetUp() {
 
     @AfterClass
     public static void finalCleanUp() {
-        DynamoTestUtil.clearTable(DynamoUserConsent2.class);
+    //    DynamoTestUtil.clearTable(DynamoUserConsent2.class);
```
alxdarksage: Intended to leave commented out?

app/org/sagebionetworks/bridge/models/studies/ConsentSignature.java
DwayneJengSage, line 22: Right now, the annotations are mainly there for documentation and for IDE hints. I like the idea of annotation-based validation. I created a JIRA issue for this https://sagebionetworks.jira.com/browse/BRIDGE-271

PR #329, opened 2014-11-21
Upgrade PhatomJS to 1.9.8; Upgrade Play to 2.2.6

eric-wu:


bootstrap.sh
```diff
@@ -48,16 +48,16 @@ gem install sass
...
 
 # Redis
 apt-get -q -y install redis-server
 
 # .bash_profile
 su - vagrant -c "echo 'source ~/.profile' > .bash_profile"
-su - vagrant -c "echo 'export PATH=$PATH:~/phantomjs-1.9.7-linux-x86_64/bin:~/play-2.2.5' >> ~/.bash_profile"
-su - vagrant -c "echo 'export SBT_OPTS=\"-Xmx2000M -Xss2M -XX:PermSize=128M -XX:MaxPermSize=256M -XX:+UseConcMarkSweepGC -XX:+CMSClassUnloadingEnabled\"' >> ~/.bash_profile"
+su - vagrant -c "echo 'export PATH=$PATH:~/phantomjs-1.9.8-linux-x86_64/bin:~/play-2.2.6' >> ~/.bash_profile"
+su - vagrant -c "echo 'export SBT_OPTS=\"-Xmx2000M -Xss1M -XX:PermSize=128M -XX:MaxPermSize=256M -XX:+UseConcMarkSweepGC -XX:+CMSClassUnloadingEnabled\"' >> ~/.bash_profile"
```
DwayneJengSage: Do we need to reduce the SBT stack size? Are we at risk of running out of memory?

eric-wu: No. Let me change it back to 2M.

PR #351, opened 2014-12-01
AB testing updated to be stateless and calculated at request time

alxdarksage:
Implementing AB testing use a stateless, at-request-time calculation of the groups. The only thing lost here is the ability to divide into exact groups by percentage, but I don't believe we need that for UI A/B testing. This now completely avoids plan or user-based calculations of all the schedules for a study.


app/org/sagebionetworks/bridge/models/schedules/ABTestScheduleStrategy.java
```diff
@@ -83,52 +79,24 @@ public void addGroup(int percent, Schedule schedule) {
...
             return null;
         }
         // Randomly assign to a group, weighted based on the percentage representation of the group.
         ScheduleGroup group = null;
         int i = 0;
-        int perc = rand.nextInt(100)+1; // 1-100
+        // Stable random percentage generated by feeding stable but unique seed to Random, based 
+        // on this strategy's seed + a stable user seed. Each run this will generate the same number.
+        long seed = plan.getGuid().hashCode() + user.getHealthCode().hashCode();
+        int perc = new Random(seed).nextInt(100) + 1;
```
alxdarksage: SecureRandom, not Random. Also you can set the seed on the instance per call.

On discussing, it's probably the case that % 100 of two GUID hashCodes will create a normally (equally) distributed set of numbers, and no randomization is needed at all. Will test.

PR #352, opened 2014-12-01
New API to submit identifier for a survey response

alxdarksage:


alxdarksage:
- Survey responses really aren't tied to a user, which seems crazy;
- Adding hdd to the GUID field would make user-submitted identifiers much more robust

PR #356, opened 2014-12-03
Modified study service to no longer use cache. Hopefully this will resol...

dr-willikers:
...ve issues with UAT failing.


app/org/sagebionetworks/bridge/services/StudyServiceImpl.java
```diff
@@ -93,30 +83,25 @@ public Tracker getTrackerByIdentifier(String trackerId) {
     @Override
     public Study getStudyByIdentifier(String identifier) {
         checkArgument(isNotBlank(identifier), Validate.CANNOT_BE_BLANK, "identifier");
-        
+
         try {
-            return studyCache.get(identifier);
+            return studyDao.getStudy(identifier);
         } catch (UncheckedExecutionException e) {
```
alxdarksage: I believe this cannot be thrown either, the whole try/catch can be removed.

PR #360, opened 2014-12-04
Survey in use

alxdarksage:
Tries to clean up the SurveyDao/Service naming scheme to be more explicit about what the getter methods do; adding REST APIS to reference the most recent, or most recently published, version of a survey without explicitly knowing the createdOn timestamp.

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
@@ -254,64 +254,6 @@ public class DynamoSurveyDao implements SurveyDao {
         }
         return saveSurvey(copy);
     }
-    // --------------------------------------------------------------------------------------
-
-    @Override
-    public List<Survey> getSurveyAllVersions(String studyKey, String guid) {
-        return new QueryBuilder().setStudy(studyKey).setSurvey(guid).getAll(false);
-    }
-    
-    
-    // ------------------------- HEREIN LIES THE GET DIVIDING LINE -------------------------
-    
-    @Override
-    public List<Survey> getSurveys(String studyKey) {
-        return new QueryBuilder().setStudy(studyKey).getAll(false);
-    }
-    
-    @Override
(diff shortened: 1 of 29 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
```diff
@@ -200,7 +201,7 @@ public void setDynamoDbClient(AmazonDynamoDB client) {
 
     @Override
     public Survey createSurvey(Survey survey) {
-        checkNotNull(survey.getStudyKey(), "Survey study key is null");
+        checkNotNull(survey.getStudyIdentifier(), "Survey study key is null");
```
alxdarksage: study identifier

conf/routes
```diff
@@ -55,11 +56,13 @@ POST   /api/v1/upload                  @controllers.UploadController.upload
 POST   /api/v1/upload/:id/complete     @controllers.UploadController.uploadComplete(id: String)
 
 # Researchers - Surveys
-GET    /researchers/v1/surveys                                 @controllers.SurveyController.getAllSurveysAllVersions
+GET    /researchers/v1/surveys                                 @controllers.SurveyController.getAllSurveysMostRecentVersion
 POST   /researchers/v1/surveys                                 @controllers.SurveyController.createSurvey
-GET    /researchers/v1/surveys/recent                          @controllers.SurveyController.getMostRecentSurveys
-GET    /researchers/v1/surveys/published                       @controllers.SurveyController.getMostRecentlyPublishedSurveys
-GET    /researchers/v1/surveys/:surveyGuid/versions            @controllers.SurveyController.getAllVersionsOfASurvey(surveyGuid: String)
+GET    /researchers/v1/surveys/recent                          @controllers.SurveyController.getAllSurveysMostRecentVersion2
```
alxdarksage, line 61: These two could probably be hidden. I do use them in the SDK to create the schedule plans.

How about we remove some of these from the documentation and don't add them to the iOS SDK, as a way of having them be "experimental" or "unsupported" API calls?

PR #373, opened 2014-12-09
Preventing duplicate accounts in different studies (Stormpath directories)

alxdarksage:


app/org/sagebionetworks/bridge/services/AuthenticationServiceImpl.java
```diff
@@ -150,13 +152,19 @@ public void signOut(String sessionToken) {
...
-        Validate.entityThrowingException(signUpValidator, signUp);
-        if (consentService.isStudyAtEnrollmentLimit(study)) {
-            throw new StudyLimitExceededException(study);
-        }
-        
+        checkNotNull(signUp.getEmail(), "Sign up email cannot be null");
+
+        String lockId = null;
         try {
+            lockId = lockDao.acquireLock(SignUp.class, signUp.getEmail());
```
alxdarksage, line 159: must lower-case the email for the lock

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
```diff
@@ -193,6 +194,21 @@ public void createAdminAndSignInWithoutConsentError() {
         }
     }
 
+    @Test
+    public void cannotCreateTheSameEmailAccountTwice() {
+        TestUser user = helper.createUser(AuthenticationServiceImplTest.class, false, false);
+        try {
+            authService.signUp(user.getSignUp(), user.getStudy(), false);
```
alxdarksage, line 201: Should create a second study, and and try to add to that, because this is the real test that we are checking at application scope for the email address.

PR #375, opened 2014-12-09
Write consent signature encrypted to Stormpath

eric-wu:


app/org/sagebionetworks/bridge/services/AccountEncryptionService.java
```diff
@@ -17,4 +18,19 @@
      * Returns null if the account has no health ID.
      */
     HealthId getHealthCode(Study study, Account account);
+
+    /**
+     * Encrypts and saves the consent signature.
+     */
+    void putConsentSignature(Study study, Account account, ConsentSignature consentSignature);
```
eric-wu, line 25: Like the Map.put(), it's both create and update.

app/org/sagebionetworks/bridge/services/AccountEncryptionService.java
```diff
@@ -17,4 +18,19 @@
...
+
+    /**
+     * Gets the decrypted consent signature.
+     */
+    ConsentSignature getConsentSignature(Study study, Account account);
+
+    /**
+     * Removes consent signature.
+     */
+    ConsentSignature removeConsentSignature(Study study, Account account);
```
eric-wu: Document the nulls.

app/org/sagebionetworks/bridge/services/AccountEncryptionServiceImpl.java
```diff
@@ -2,17 +2,24 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.stormpath.sdk.account.Account;
 import com.stormpath.sdk.directory.CustomData;
 
 public class AccountEncryptionServiceImpl implements AccountEncryptionService {
 
     private static final int CURRENT_VERSION = 2;
+    private final ObjectMapper mapper = new ObjectMapper();
```
eric-wu: static

app/org/sagebionetworks/bridge/services/AccountEncryptionServiceImpl.java
```diff
@@ -50,10 +57,48 @@ public HealthId getHealthCode(Study study, Account account) {
         return (getHealthId(healthIdObj, versionObj));
     }
 
+    @Override
+    public void putConsentSignature(Study study, Account account, ConsentSignature consentSignature) {
+        JsonNode json = mapper.valueToTree(consentSignature);
+        try {
+            String encrypted = healthCodeEncryptor.encrypt(mapper.writeValueAsString(json));
```
eric-wu: Combine the mapper methods into one to write object to JSON string.

app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
```diff
@@ -67,50 +67,58 @@ public void setApplicationEventPublisher(ApplicationEventPublisher publisher) {
...
+
+            if (sendEmail) {
+                sendMailService.sendConsentAgreement(caller, consentSignature, studyConsent);
+            }
+
+            caller.setConsent(true);
+            caller.setHealthCode(healthCode);
+            return caller;
+        } finally {
+            decrementStudyEnrollment(study);
```
eric-wu: Remove try-finally.

1) Move the decrement() method to before write-new
2) try-catch around write-new and log the exceptions

test/org/sagebionetworks/bridge/services/AccountEncryptionServiceImplTest.java
```diff
@@ -0,0 +1,70 @@
...
+import org.sagebionetworks.bridge.models.studies.Study;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+import com.stormpath.sdk.account.Account;
+import com.stormpath.sdk.client.Client;
+
+@ContextConfiguration("classpath:test-context.xml")
+@RunWith(SpringJUnit4ClassRunner.class)
+public class AccountEncryptionServiceImplTest {
```
DwayneJengSage, line 28: Do we want a test case for remove()?

app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
```diff
@@ -67,49 +71,61 @@ public void setApplicationEventPublisher(ApplicationEventPublisher publisher) {
...
         checkNotNull(caller, Validate.CANNOT_BE_NULL, "user");
         checkNotNull(study, Validate.CANNOT_BE_NULL, "study");
-
         final StudyConsent consent = studyConsentDao.getConsent(study.getIdentifier());
         if (consent == null) {
             throw new EntityNotFoundException(StudyConsent.class);
         }
-        ConsentSignature consentSignature = userConsentDao.getConsentSignature(caller.getHealthCode(), consent);
+        Account account = stormpathClient.getResource(caller.getStormpathHref(), Account.class);
+        ConsentSignature consentSignature = accountEncryptionService.getConsentSignature(study, account);
```
DwayneJengSage: Could stormpath client or aes throw? If so, should we propagate the error? Or should we log the exception and fall back to dynamo?

app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
```diff
@@ -67,49 +71,61 @@ public void setApplicationEventPublisher(ApplicationEventPublisher publisher) {
...
+
         incrementStudyEnrollment(study);
-        userConsentDao.giveConsent(healthId.getCode(), studyConsent, consentSignature);
-        // Publish event
+        try {
+            userConsentDao.giveConsent(healthCode, studyConsent, consentSignature);
+        } catch (Throwable e) {
+            decrementStudyEnrollment(study);
+            throw e;
+        }
```
DwayneJengSage, line 128: I feel like it would be cleaner to have increment() after the try-catch instead of increment, try, and then decrement in the catch. I'd run it by Alx in case he has special insights into this.

app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
```diff
@@ -146,18 +166,23 @@ public User withdrawConsent(User caller, Study study) {
...
         if (consent == null) {
             throw new EntityNotFoundException(StudyConsent.class);
         }
-        ConsentSignature consentSignature = userConsentDao.getConsentSignature(caller.getHealthCode(), consent);
+
+        Account account = stormpathClient.getResource(caller.getStormpathHref(), Account.class);
+        ConsentSignature consentSignature = accountEncryptionService.getConsentSignature(study, account);
+        if (consentSignature == null) {
+            consentSignature = userConsentDao.getConsentSignature(caller.getHealthCode(), consent);
+        }
```
DwayneJengSage: Should we call getConsentSignature() so we don't have duplicate logic?

app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
eric-wu: Yes, we should try-catch the exception and log it at INFO.

app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
eric-wu, line 128: Alx and I had a discussion about this.  There can be race conditions.  If we increment later, there is the risk of getting over the limit.  On the other hand if we increment early, we run the risk of not reaching the limit.  Among the two, we have opted for incrementing early and accepted the risk of not reaching the limit.

test/org/sagebionetworks/bridge/services/AccountEncryptionServiceImplTest.java
eric-wu, line 28: Added.

PR #380, opened 2014-12-12
Add study id to the health code table

eric-wu:


app/org/sagebionetworks/bridge/dynamodb/DynamoHealthCodeDao.java
```diff
@@ -14,24 +16,50 @@
 
 public class DynamoHealthCodeDao implements HealthCodeDao {
 
+    private final Logger logger = LoggerFactory.getLogger(DynamoHealthCodeDao.class);
```
DwayneJengSage: Nitpick: Conventially, Loggers are static. (That said, this is a singleton class, so it won't make a difference.)

PR #382, opened 2014-12-15
Fixed test users explosion.

dr-willikers:
All new users created in testing are placed in "test_users" group. Additionally, added a method in the UserAdminService that first adds all users with email addresses starting with "bridge-testing" to the "test_users" group, then deletes all users in that group.


app/controllers/UserManagementController.java
```diff
@@ -41,4 +41,11 @@ public Result deleteUser(String email) throws Exception {
         return okResult("User deleted.");
     }
 
+    public Result deleteAllTestUsers() throws Exception {
```
alxdarksage: I would prefer we only take this to the service layer, and then have a server-side test that does this each run through of the test suite.

app/org/sagebionetworks/bridge/services/UserAdminService.java
```diff
@@ -30,11 +28,18 @@ public UserSession createUser(SignUp signUp, Study userStudy, boolean signUserIn
...
     public void deleteUser(String userEmail) throws BridgeServiceException;
 
-}
\ No newline at end of file
+    /**
+     * Deletes all all users in the "test users" group in Stormpath.
+     *
+     * @throws BridgeServiceException
+     */
+    public void deleteAllTestUsers() throws BridgeServiceException;
```
alxdarksage: deleteAllUsersInGroup() ?

app/org/sagebionetworks/bridge/services/UserAdminServiceImpl.java
```diff
@@ -97,18 +110,52 @@ public void deleteUser(String userEmail) throws BridgeServiceException {
...
+        Group testUsers = null;
+        for (Group group : groups) {
+            if (group.getName().equalsIgnoreCase(BridgeConstants.TEST_USERS_GROUP)) {
+                testUsers = group;
+            }
+        }
+        return testUsers;
+    }
+
+    private void addAllOrphanedTestUsersToGroup(Group testUsers, Directory directory) {
```
alxdarksage: I think you can remove this, as I'll clean up manually once this is deployed.

app/org/sagebionetworks/bridge/services/UserAdminServiceImpl.java
```diff
@@ -97,18 +110,52 @@ public void deleteUser(String userEmail) throws BridgeServiceException {
             }
         }
     }
-    
+
+    @Override
+    public void deleteAllTestUsers() throws BridgeServiceException {
+        for (Study study : studyService.getStudies()) {
+            Directory directory = directoryDao.getDirectoryForStudy(study.getIdentifier());
```
alxdarksage: I think you can get a group and it has getAccounts().

test/org/sagebionetworks/bridge/TestUserAdminHelper.java
```diff
@@ -42,9 +42,19 @@ public TestUser(String username, String email, String password, List<String> rol
             this.username = username;
             this.email = email;
             this.password = password;
-            this.roles = (roleList == null) ? null : roleList.toArray(new String[roleList.size()]);
```
alxdarksage, line 45: Set<String> would be better; then you can test_users without worrying if it's already there or not. And convert to an array later, wherever it was we needed it

test/org/sagebionetworks/bridge/TestUserAdminHelper.java
```diff
@@ -100,35 +110,35 @@ public String makeRandomUserName(Class<?> cls) {
...
         return createUser(cls, true, true, roles);
     }
-    
+
     public TestUser createUser(Class<?> cls, boolean signIn, boolean consent, String...roles) {
         checkNotNull(cls, "Class must not be null");
-        
+
         String name = makeRandomUserName(cls);
         SignUp signUp = new SignUp(name, name + "@sagebridge.org", PASSWORD, roles);
```
alxdarksage, line 118: ... and signUp should be Set<String> roles as well!

app/org/sagebionetworks/bridge/stormpath/StormpathDirectoryDao.java
```diff
@@ -63,35 +63,35 @@ public String createDirectoryForStudy(String identifier) {
             mapping.setListIndex(10); // this is a priority number
             app.createAccountStoreMapping(mapping);
         }
-        
-        Group group = getResearcherGroup(app, groupName);
+
+        Group group = getGroup(app, groupName);
```
alxdarksage, line 67: Do we need to create the test_users group when we create a new directory? I think we do.

PR #384, opened 2014-12-15
Where to lunch

eric-wu:
Seriously this may be used for health-check pinging.

### DIFF
--- a/app/controllers/LunchController.java
+++ b/app/controllers/LunchController.java
@@ -1,0 +1,24 @@
+package controllers;
+
+import java.security.SecureRandom;
+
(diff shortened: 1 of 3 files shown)


app/controllers/LunchController.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.security.SecureRandom;
+
+import play.mvc.Result;
+
+public class LunchController extends BaseController {
+
+    private static final SecureRandom random = new SecureRandom();
+
+    public Result where() throws Exception {
```
alxdarksage, line 11: How about a redirect to the yelp page?

PR #386, opened 2014-12-17
Backfill studies for health codes

eric-wu:


conf/test-context.xml
```diff
@@ -18,4 +18,8 @@
         <property name="studyService" ref="studyService"/>
     </bean>
 
+    <bean id="testBackfillService" class="org.sagebionetworks.bridge.services.backfill.TestBackfillService">
```
eric-wu: Don't need this

test/org/sagebionetworks/bridge/dynamodb/DynamoHealthCodeDaoTest.java
```diff
@@ -37,5 +37,11 @@ public void test() {
         assertFalse(healthCodeDao.setIfNotExist("123", "789"));
         assertEquals("789", healthCodeDao.getStudyIdentifier("123"));
         assertNull(healthCodeDao.getStudyIdentifier("xyz"));
+        healthCodeDao.setStudyId("123", "789");
+        try {
+            healthCodeDao.setStudyId("123", "456");
+        } catch (RuntimeException e) {
+            assertTrue("Expected", true);
+        }
```
eric-wu, line 59: fail() here

test/org/sagebionetworks/bridge/dynamodb/DynamoHealthCodeDaoTest.java
```diff
@@ -37,5 +38,12 @@ public void test() {
         assertFalse(healthCodeDao.setIfNotExist("123", "789"));
         assertEquals("789", healthCodeDao.getStudyIdentifier("123"));
         assertNull(healthCodeDao.getStudyIdentifier("xyz"));
+        healthCodeDao.setStudyId("123", "789");
+        try {
+            healthCodeDao.setStudyId("123", "456");
+            fail();
+        } catch (RuntimeException e) {
+            assertTrue("Expected", true);
+        }
```
DwayneJengSage, line 59: Should we add test cases for:
1. setting study ID on a healthcode that does not have study ID?
2. setting study ID on a non-existent health code?

app/org/sagebionetworks/bridge/services/backfill/AsyncBackfillTemplate.java
```diff
@@ -0,0 +1,108 @@
...
+            callback.start(task);
+            doBackfill(task, callback);
+        } finally {
+            callback.done();
+        }
+    }
+
+    private BackfillTask createBackfillTask(final String user, final String name) {
+        final String taskId = UUID.randomUUID().toString();
+        return new BackfillTask() {
```
DwayneJengSage, line 76: It looks like this is the only place we use BackfillTask. Do we intend to have different implementations of BackfillTask? If not, should we making BackfillTask a concrete class with this implementation, and pass in user, name, and taskId in the constructor?

app/org/sagebionetworks/bridge/services/backfill/StudyIdBackfill.java
```diff
@@ -0,0 +1,118 @@
...
+                            public String getTaskId() {
+                                return task.getId();
+                            }
+                            @Override
+                            public long getTimestamp() {
+                                return DateTime.now(DateTimeZone.UTC).getMillis();
+                            }
+                            @Override
+                            public String getRecord() {
+                                ObjectNode node = MAPPER.createObjectNode();
```
DwayneJengSage: It looks like all the instances of BackfillRecord are almost identical. getTaskId() and getTimestamp() are the same for all of them. getRecord looks like it creates an object node. Would it be better to make a single concrete class of BackfillRecord (without the interface, since it seems like we don't need it) and have task ID, timestamp, and possibly the record as a Map<String, String> passed in the constructor?

app/org/sagebionetworks/bridge/services/backfill/AsyncBackfillTemplate.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private void backfillWithLock(final String user, final String name, final BackfillCallback callback) {
+        final Class<? extends AsyncBackfillTemplate> clazz = getClass();
+        final String identifier = getClass().getSimpleName();
+        String lock = null;
+        try {
+            lock = lockDao.acquireLock(clazz, identifier, getLockExpireInSeconds());
+            backfillTemplate(user, name, callback);
+        } catch (ConcurrentModificationException e) {
+            // TODO: Query DynamoDB and report back progress
```
DwayneJengSage, line 55: We should log something here.

app/controllers/BackfillChunksAdapter.java
```diff
@@ -0,0 +1,47 @@
...
+        checkNotNull(task);
+        chunksOut.write(task.getName()
+                + " started by " + task.getUser()
+                + " on " + new DateTime(task.getTimestamp(), DateTimeZone.UTC).toString("YYYY-MM-dd HH:mm:ss"));
+    }
+
+    @Override
+    public void newRecords(BackfillRecord... records) {
+        checkNotNull(records);
+        checkNotNull(records);
```
DwayneJengSage: It looks like you checkNotNull(records) twice. Did you mean a different precondition check?

test/org/sagebionetworks/bridge/dynamodb/DynamoHealthCodeDaoTest.java
eric-wu, line 59: Both done.

app/org/sagebionetworks/bridge/services/backfill/AsyncBackfillTemplate.java
eric-wu, line 76: Once I add the DynamoDB tables for the backfill, these will be created, saved, and returned by the DAO.  That would be a better place to define the object.  And yes, it will have a constructor that takes user, name, and id.

app/org/sagebionetworks/bridge/services/backfill/StudyIdBackfill.java
```diff
@@ -0,0 +1,118 @@
...
+
+    @Override
+    void doBackfill(final BackfillTask task, final BackfillCallback callback) {
+        List<Study> studies = studyService.getStudies();
+        for (final Study study : studies) {
+            Application application = StormpathFactory.getStormpathApplication(stormpathClient);
+            StormpathAccountIterator iterator = new StormpathAccountIterator(application);
+            while (iterator.hasNext()) {
+                List<Account> accountList = iterator.next();
+                for (final Account account : accountList) {
```
DwayneJengSage: I suspect that this will make multiple calls to Stormpath, and Stormpath calls tend to be slow.. Given this, I think it would be better to have accounts be the other loop and studies be the inner loop, so we make fewer calls to Stormpath. Ideally, we could just call studyService.getStudies() once and just iterate on the same list of studies.

app/org/sagebionetworks/bridge/services/backfill/StudyIdBackfill.java
eric-wu: The same here.  This will be moved into a DynamoDB data object.

app/org/sagebionetworks/bridge/services/backfill/AsyncBackfillTemplate.java
eric-wu, line 55: Done.

app/controllers/BackfillChunksAdapter.java
eric-wu: Indeed redundant. Removed.

app/org/sagebionetworks/bridge/services/backfill/StudyIdBackfill.java
```diff
@@ -0,0 +1,118 @@
...
+                                node.put("studyIdentifier", study.getIdentifier());
+                                node.put("account", account.getEmail());
+                                node.put("operation", e.getMessage());
+                                try {
+                                    return MAPPER.writeValueAsString(node);
+                                } catch (JsonProcessingException e) {
+                                    throw new RuntimeException(e);
+                                }
+                            }
+                        });
```
DwayneJengSage, line 98: The try and the catch are really similar. Maybe consolidate to a helper function that takes the operation as a string.

app/org/sagebionetworks/bridge/services/backfill/StudyIdBackfill.java
eric-wu: Good point!  Moved Stormpath call to the outer loop.

app/org/sagebionetworks/bridge/services/backfill/StudyIdBackfill.java
eric-wu, line 98: Done.  The backfill record will go to DynamoDB too.

PR #388, opened 2014-12-17
Edited ConsentService to not provide as many nullpointerexceptions, edit...

dr-willikers:
...ed UserAdminService to increase chances of success for delete, and changed name of ManualCleanup to ManualCleanupTest so that JUnit actually runs it.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/UserAdminServiceImpl.java
+++ b/app/org/sagebionetworks/bridge/services/UserAdminServiceImpl.java
@@ -118,12 +118,12 @@ public class UserAdminServiceImpl implements UserAdminService {
             Account account = authenticationService.getAccount(userEmail);
             if (account != null) {
                 for (Study study : studyService.getStudies()) {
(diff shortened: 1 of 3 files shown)


app/org/sagebionetworks/bridge/services/UserAdminServiceImpl.java
```diff
@@ -111,23 +111,26 @@ public void deleteAllUsers(Group group) throws BridgeServiceException {
...
             Account account = authenticationService.getAccount(userEmail);
             if (account != null) {
                 for (Study study : studyService.getStudies()) {
+                    System.out.println("STUDY NAME: " + study.getName());
                     User user = authenticationService.getSessionFromAccount(study, account).getUser();
                     deleteUserInStudy(study, account, user);
                 }
                 account.delete();
             }
-            return true;
```
alxdarksage, line 125: Is this a test after the fact to verify that the account was deleted?

PR #393, opened 2014-12-18
Added new method to ConsentService to reduce number of calls to Dynamo when signing in.

dr-willikers:


test/org/sagebionetworks/bridge/dynamodb/DynamoUserConsentDaoTest.java
```diff
@@ -84,6 +84,19 @@ public void canCountStudyParticipants() {
         assertEquals("Correct number of participants", 5, count);
     }
 
+    @Test
```
alxdarksage, line 87: Verify there's a test where you consent, and you're up-to-date, then a new consent is created, and you're consented but not up-to-date, then you consent to that new record, and both are true again. Through the service methods.

PR #395, opened 2014-12-18
Record the exception name in the callback; test the page iterator

eric-wu:


app/org/sagebionetworks/bridge/services/backfill/PageIterator.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Iterates page by page. A page is a list of items.
+ * That is the page is of known size and the items are in order.
+ */
+abstract class PageIterator<T> implements Iterator<List<T>> {
```
DwayneJengSage: Optional but recommended: Consider also implementing Iterable<List<T>> so we can use this in a foreach loop. The iterator() method could just return "this".

test/org/sagebionetworks/bridge/services/backfill/PageIteratorTest.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        @Override
+        Iterator<String> nextPage() {
+            int end = start + pageSize;
+            if (end > words.size()) {
+                end = words.size();
+            }
+            List<String> page = words.subList(start, end);
+            start += end;
```
DwayneJengSage: Should be start += pageSize

test/org/sagebionetworks/bridge/services/backfill/PageIteratorTest.java
```diff
@@ -0,0 +1,112 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.List;
+
+import org.junit.Test;
+
+public class PageIteratorTest {
```
DwayneJengSage, line 13: Recommend adding a test case with 5+ pages.

Recommend adding a test case with page size 3 or more.

PR #402, opened 2014-12-23
add tracker schemas to unpack uploads from iOS apps

DwayneJengSage:
Added tracker schemas for the different data types we're getting back from the iOS apps. We only look at the top-level keys. For any JSON blobs or for survey answers (which can be of any time), we use the empty schema, which allows any type and value.

For the timestamps in ios-survey-tracker, JSON doesn't have a native timestamp type, so I decided to treat them as strings. If we want to do stricter validation, I can add a rudimentary regex to the schema.

Testing:
- created unit tests to test schema validation
- added all 5 trackers to "api" study on my local dev instance. Called /api/v1/trackers and /api/v1/trackers/schema/(tracker) to validate that we can get the tracker schemas back.
- put data for all trackers (except ios-survey-tracker, which will never actually write health data) to verify that the data can be written back to the server.

### DIFF
--- a/conf/application-context.xml
+++ b/conf/application-context.xml
@@ -316,8 +316,13 @@
         <property name="trackers">
             <list>
                 <ref bean="pb-tracker"/>
+                <ref bean="cardio-walk-tracker"/>
+                <ref bean="diabetes-glucose-tracker"/>
+                <ref bean="ios-survey-tracker"/>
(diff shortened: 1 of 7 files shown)


DwayneJengSage:
Schemas may change, we want researchers to be able to create and update schemas. (Example: In dynamo, with list of field names/types, with versioning.)

PR #406, opened 2014-12-27
Remove study ID backfill as it is done; refactor backfill record creation

eric-wu:
The change worth mentioning is the refactoring where I am moving the logic of creating backfill records to a new class BackillRecordFactory.  There record creation is split into 2 types -- one that persists a backfill record in Dynamo; the other does not persist into Dynamo.  The 2nd one can be used for dry runs to see what backfills are needed and to verify the backfill after it's done.

### DIFF
--- a/app/controllers/BackfillChunksAdapter.java
+++ b/app/controllers/BackfillChunksAdapter.java
@@ -19,7 +19,7 @@ import play.mvc.Results.Chunks;
  */
 class BackfillChunksAdapter implements BackfillCallback {
 
-    private final ObjectMapper MAPPER = new ObjectMapper();
+    private static final ObjectMapper MAPPER = new ObjectMapper();
 
     private final Chunks.Out<String> chunksOut;
 
--- a/app/controllers/BackfillChunksAdapter.java
+++ b/app/controllers/BackfillChunksAdapter.java
@@ -8,12 +8,18 @@ import org.sagebionetworks.bridge.models.BackfillRecord;
 import org.sagebionetworks.bridge.models.BackfillTask;
 import org.sagebionetworks.bridge.services.backfill.BackfillCallback;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
 import play.mvc.Results.Chunks;
 
 /**
  * Adapts backfill callback to Play chunked responses.
  */
 class BackfillChunksAdapter implements BackfillCallback {
+
+    private final ObjectMapper MAPPER = new ObjectMapper();
 
     private final Chunks.Out<String> chunksOut;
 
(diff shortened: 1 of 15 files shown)


app/controllers/BackfillChunksAdapter.java
```diff
@@ -15,6 +19,8 @@
  */
 class BackfillChunksAdapter implements BackfillCallback {
 
+    private final ObjectMapper MAPPER = new ObjectMapper();
```
DwayneJengSage: Do we want to use BridgeObjectMapper.get()? This way, we don't have to create an ObjectMapper instance for every class.

Minor: If we still want to use ObjectMapper directly, we should declare it static. That said, since BackfillChunksAdapter is singleton, it's not a drastic difference.

app/org/sagebionetworks/bridge/services/backfill/BackfillRecordFactory.java
```diff
@@ -0,0 +1,91 @@
...
+import org.sagebionetworks.bridge.models.studies.Study;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.stormpath.sdk.account.Account;
+
+public class BackfillRecordFactory {
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
```
DwayneJengSage, line 19: Similarly here?

app/org/sagebionetworks/bridge/services/backfill/StudyIdBackfill.java
```diff
@@ -63,15 +69,19 @@ void doBackfill(final BackfillTask task, final BackfillCallback callback) {
...
-                                boolean set = healthCodeDao.setStudyId(healthCode, study.getIdentifier());
-                                if (set) {
-                                    callback.newRecords(createRecord(task, study, account, "backfilled"));
+                                final String studyId = healthCodeDao.getStudyIdentifier(healthCode);
+                                if (isBlank(studyId)) {
+                                    String msg = "Backfill needed as study ID is blank.";
+                                    callback.newRecords(backfillRecordFactory.createOnly(task, study, account, msg));
+                                } else {
+                                    String msg = "Study ID already exists.";
+                                    callback.newRecords(backfillRecordFactory.createOnly(task, study, account, msg));
```
DwayneJengSage, line 78: Would it be cleaner to just delete this class, since the study ID backfill is done and the code is being deleted at the dao level?

test/org/sagebionetworks/bridge/services/backfill/BackfillRecordFactoryTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.joda.time.DateTimeZone;
+import org.junit.Test;
+import org.sagebionetworks.bridge.models.BackfillRecord;
+import org.sagebionetworks.bridge.models.BackfillTask;
+import org.sagebionetworks.bridge.models.studies.Study;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.stormpath.sdk.account.Account;
+
+public class BackfillRecordFactoryTest {
```
DwayneJengSage, line 22: For completeness, should we also test createAndSave(), if if only to make sure it calls through to a mock dao?

PR #411, opened 2015-01-06
upload schema in DDB

DwayneJengSage:
Created upload schema model and getter/setter API for researchers. Code is complete except implementing global secondary indices for the DDB table.

Code has been tested programmatically via test cases and via the following manual test cases:
- create new schema
- update schema
- create conflict
- update conflict
- get schema between each step

This also includes implementing global secondary indices (hash-only). To see these changes in isolation, see:
- https://github.com/DwayneJengSage/BridgePF/commit/eb71cfa7e2c383ff1603fdbc3fad904dbd7bb8c5
- https://github.com/DwayneJengSage/BridgePF/commit/fb09525499cf5b010b66f2a0fc9d1e7c852f1375


app/controllers/BaseController.java
```diff
@@ -222,5 +223,23 @@ protected JsonNode requestToJSON(Request request) {
             throw new InvalidEntityException("Expected JSON in the request body is missing or malformed");
         }
     }
-    
+
+    protected static <T> T parseJson(Request request, Class<? extends T> clazz) {
```
eric-wu: Is there a reason that this method is static?

app/controllers/BaseController.java
```diff
@@ -222,5 +223,23 @@ protected JsonNode requestToJSON(Request request) {
...
+        try {
+            // Whether asText() or asJson() works depends on the content-type header of the request
+            String jsonText = request.body().asText();
+            if (!Strings.isNullOrEmpty(jsonText)) {
+                return mapper.readValue(jsonText, clazz);
+            }
+
+            JsonNode jsonNode = request.body().asJson();
```
eric-wu, line 250: Need to confirm that we can call request.body() twice.  Behind the scenes, it could read off an input stream.

app/org/sagebionetworks/bridge/services/UploadSchemaService.java
```diff
@@ -0,0 +1,28 @@
+package org.sagebionetworks.bridge.services;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import org.sagebionetworks.bridge.dao.UploadSchemaDao;
+import org.sagebionetworks.bridge.models.studies.Study;
+import org.sagebionetworks.bridge.models.upload.UploadSchema;
+
+@Component
```
eric-wu, line 18: Is this redundant as we are already using XML?

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchema.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    // We don't use the DynamoDBVersionAttribute here because we want to keep multiple versions of the schema so we can
+    // parse older versions of the data. Similarly, we make this a range key so that we can always find the latest
+    // version of the schema.
+    @DynamoDBRangeKey
```
eric-wu, line 105: Do we still want a separate DynamoDBVersionAttribute for optimistic locking?  This additional version attribute does not need to be exposed in the UploadSchema API.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchema.java
```diff
@@ -0,0 +1,118 @@
...
+    public String getSchemaId() {
+        return schemaId;
+    }
+
+    public void setSchemaId(String schemaId) {
+        this.schemaId = schemaId;
+    }
+
+    // TODO: Implement global secondary indices in DynamoInitializer
+    @DynamoDBIndexHashKey(attributeName = "studyId", globalSecondaryIndexName = "studyId-index")
```
eric-wu, line 138: It's unclear to me if we will ever need to query the list of schemas for a particular study.  Another thing is that, if we don't have many studies, the benefit of an index is questionable.  Think about an extreme example of an index on a boolean type, the query using the index does not make much of a difference from a scan.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
```diff
@@ -0,0 +1,88 @@
...
+        int oldRev;
+        if (oldUploadSchema != null) {
+            oldRev = oldUploadSchema.getRevision();
+        } else {
+            oldRev = 0;
+        }
+
+        // Request should match old rev. If it does, auto-increment and write.
+        if (oldRev != uploadSchema.getRevision()) {
+            // TODO throw something
```
eric-wu: If we are implementing our own optimistic locking, we could throw a ConcurrentModificationException.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
```diff
@@ -0,0 +1,88 @@
...
+        DynamoUploadSchema ddbUploadSchema;
+        if (uploadSchema instanceof DynamoUploadSchema) {
+            ddbUploadSchema = (DynamoUploadSchema) uploadSchema;
+        } else {
+            ddbUploadSchema = BridgeObjectMapper.get().convertValue(uploadSchema, DynamoUploadSchema.class);
+        }
+        ddbUploadSchema.setRevision(oldRev + 1);
+
+        try {
+            mapper.save(uploadSchema, DOES_NOT_EXIST_EXPRESSION);
```
eric-wu, line 68: Does this require the hash key does not exist yet? If so, how can we update an existing schema?

app/org/sagebionetworks/bridge/dao/UploadSchemaDao.java
```diff
@@ -0,0 +1,11 @@
+package org.sagebionetworks.bridge.dao;
+
+import org.sagebionetworks.bridge.models.upload.UploadSchema;
+
+public interface UploadSchemaDao {
+    UploadSchema createOrUpdateUploadSchema(String studyId, String schemaId, UploadSchema uploadSchema);
```
eric-wu: When creating a new schema, it looks like the schema id is supplied by the user? As it is part of the hash key, I would suggest we control the creation of the schema id instead.  The user is responsible for supplying a good name.  Otherwise, the id seems duplicate of the schema name.  Alternatively, we can get rid of the id and use (name, revision) to reference a schema.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchema.java
alxdarksage, line 138: If we ever build a UI for researchers, and we plan to, we will need to show
a list of the schemas that have been posted for a study.

Alx

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadFieldDefinition.java
```diff
@@ -0,0 +1,90 @@
...
+import org.sagebionetworks.bridge.models.upload.UploadFieldType;
+import org.sagebionetworks.bridge.validators.Validate;
+
+/**
+ * Dynamo DB implementation of UploadFieldDefinition. While there is nothing specific to Dynamo DB in this class, this
+ * class exists to distinguish itself from potential other implementations.
+ */
+@JsonDeserialize(builder = DynamoUploadFieldDefinition.Builder.class)
+public class DynamoUploadFieldDefinition implements UploadFieldDefinition {
+    private final @Nonnull String name;
```
eric-wu, line 19: Is @Nonnull a new feature in Java 8?  Not sure if we have fully migrated to Java 8 yet.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchema.java
```diff
@@ -0,0 +1,168 @@
...
+
+    /**
+     * This is the DynamoDB key. It is used by the DynamoDB mapper. This should not be used directly. The key format is
+     * "[studyID]$[schemaID]". The schema ID may contain dollar signs. The study ID may not. Since the key is created
+     * from the study ID and schema ID, this will throw an InvalidEntityException if either one is blank.
+     */
+    @DynamoDBHashKey
+    @JsonIgnore
+    public String getKey() throws InvalidEntityException {
+        if (Strings.isNullOrEmpty(studyId)) {
```
eric-wu: If we use Guava's checkArgument, we could combine the check and exception throwing into one line of code.

app/org/sagebionetworks/bridge/models/upload/UploadFieldDefinition.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * The field's type.
+     *
+     * @see org.sagebionetworks.bridge.models.upload.UploadFieldType
+     */
+    @Nonnull UploadFieldType getType();
+
+    /** This class validates the intrinsic properties of the field definition. */
+    static class Validator implements org.springframework.validation.Validator {
```
eric-wu: Can we free this inner class to a package-scope class by itself?

app/org/sagebionetworks/bridge/models/upload/UploadSchema.java
```diff
@@ -0,0 +1,118 @@
...
+     * data. This can be human readable, such as "tapping-task". This cannot be changed across different schema
+     * revisions.
+     */
+    String getSchemaId();
+
+    /** Study ID for the study this schema lives under. This cannot be changed across different schema revisions. */
+    String getStudyId();
+
+    /** This class validates the intrinsic properties of the schema. */
+    public static class Validator implements org.springframework.validation.Validator {
```
eric-wu: Again, I am not sure about the benefits of inner class "UploadSchema.Validaor" over "UploadSchemaValidator".  What's the reason?

app/org/sagebionetworks/bridge/services/UploadSchemaService.java
```diff
@@ -0,0 +1,86 @@
...
+     */
+    public UploadSchema createOrUpdateUploadSchema(Study study, UploadSchema uploadSchema) {
+        // validate schema
+        if (uploadSchema == null) {
+            throw new InvalidEntityException(String.format(Validate.CANNOT_BE_NULL, "upload schema"));
+        }
+        Validate.entityThrowingException(UploadSchema.Validator.INSTANCE, uploadSchema);
+
+        // validate that the schema's study ID matches the server endpoint's study
+        if (!uploadSchema.getStudyId().equals(study.getIdentifier())) {
```
eric-wu: This cross-study check is interesting.  I don't think we do this check anywhere else.  Alternatively the upload schema from the client can skip the study field which is then determined by the bridge endpoint the client is calling.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadFieldDefinition.java
DwayneJengSage, line 19: This actually comes from the Google implementation of JSR-305:
https://code.google.com/p/jsr-305/source/browse/trunk/ri/src/main/java/javax/annotation/Nonnull.java

It has 3 benefits:
1. It provides documentation about the preconditions of a given method or class.
2. Static code analysis tools (such as IntelliJ, FindBugs, and possibly Eclipse) can tell us if our preconditions are violated.
3. There are tools that can validate objects using annotations.

It looks like there's a separate annotation coming in Java 8. If/when we migrate, we can consider migrating all of our annotations to the Java 8 version.

app/org/sagebionetworks/bridge/models/upload/UploadFieldDefinition.java
DwayneJengSage: Discussed this with Eric. I will move this to a separate class in the validators pacakge.

app/org/sagebionetworks/bridge/services/UploadSchemaService.java
DwayneJengSage: Discussed this with Eric. I'll add a @JsonIgnore to the study property so that it's never exposed to the caller, and we'll just use the endpoint-derived study.

PR #421, opened 2015-01-12
You can add a phone number to your user profile

alxdarksage:


app/controllers/UserProfileController.java
```diff
@@ -16,14 +16,15 @@ public void setUserProfileService(UserProfileService userProfileService) {
     }
 
     public Result getUserProfile() throws Exception {
-        UserSession session = getAuthenticatedSession();
```
alxdarksage, line 19: Authenticated only.

app/org/sagebionetworks/bridge/services/UserProfileServiceImpl.java
```diff
@@ -1,27 +1,67 @@
...
+    public UserProfile getProfile(String email) {
+        Account account = authService.getAccount(email);
+
+        UserProfile profile = new UserProfile();
+        profile.setFirstName(removeEmpty(account.getGivenName()));
+        profile.setLastName(removeEmpty(account.getSurname()));
+        profile.setUsername(removeEmpty(account.getUsername()));
+        profile.setEmail(removeEmpty(account.getEmail()));
+        
+        String encryptedPhone = (String)account.getCustomData().get("phone");
```
alxdarksage: Make a constant out of "Phone"

app/org/sagebionetworks/bridge/services/UserProfileServiceImpl.java
```diff
@@ -1,27 +1,67 @@
...
 
     @Override
+    public UserProfile getProfile(String email) {
+        Account account = authService.getAccount(email);
+
+        UserProfile profile = new UserProfile();
+        profile.setFirstName(removeEmpty(account.getGivenName()));
+        profile.setLastName(removeEmpty(account.getSurname()));
+        profile.setUsername(removeEmpty(account.getUsername()));
+        profile.setEmail(removeEmpty(account.getEmail()));
```
DwayneJengSage: Are empty usernames and emails possible? We get the account from Stormpath using email, so I would assume email addresses always exist. Similarly, our signUp API provides a username, so I assume we'll always have a username.

app/org/sagebionetworks/bridge/services/UserProfileServiceImpl.java
```diff
@@ -1,27 +1,67 @@
...
+        
+        String encryptedPhone = (String)account.getCustomData().get("phone");
+        if (encryptedPhone != null) {
+            profile.setPhone(healthCodeEncryptor.decrypt(encryptedPhone));
+        }
```
DwayneJengSage, line 40: If phone number isn't specified, the field will return null. If name isn't specified, the field will return empty string. Should we change phone number to empty string or removeEmpty to null to be consistent?

PR #423, opened 2015-01-14
Survey enhancements

alxdarksage:


app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
```diff
@@ -53,9 +53,11 @@
         private static final String PUBLISHED_PROPERTY = "published";
```
alxdarksage: We should verify, before returning a survey, that it is associated to the study of the hostname. The code does not/cannot enforce that for a key-based retrieval.

app/org/sagebionetworks/bridge/validators/SurveyAnswerValidator.java
```diff
@@ -195,17 +197,52 @@ private void validateType(Errors errors, StringConstraints con, String answer) {
...
-            // TODO: The docs don't say it throws an exception when wrong...
-            Period.parse(answer);
+            seconds = new DateTime(0L).plus(Period.parse(answer)).toDateTime().getMillis()/1000L;
         } catch(Throwable t) {
             rejectField(errors, "constraints", "%s is not a valid ISO 8601 duration string", answer);
+            return;
+        }
+        if (con.getUnit() != null) {
+            double value = convertPeriodToDurationUnits(con, seconds);
+            if (value > 0) {
```
alxdarksage: Should this be removed? the options are: declined, there's no validation, or it's within the range, we don't skip just because the answer is zero.

app/org/sagebionetworks/bridge/validators/SurveyAnswerValidator.java
```diff
@@ -71,55 +73,55 @@ public void validate(Object object, Errors errors) {
...
     
     private boolean hasNoAnswer(SurveyAnswer answer) {
-        if (expectsMultipleValues()) {
-            return BridgeUtils.isEmpty(answer.getAnswers());
-        }
-        return StringUtils.isEmpty(answer.getAnswer());
+        return BridgeUtils.isEmpty(answer.getAnswers());
     }
 
     private void validateType(Errors errors, TimeConstraints constraints, String answer) {
```
alxdarksage: DurationConstraints --> IntegerConstraints, unless it's actually in use, in which case functionally it's the same but with time units only.

PR #424, opened 2015-01-15
Redo Upload DynamoDB table, add global secondary indices

DwayneJengSage:
Includes the following changes:
- Added a DynamoUpload2 table, with secondary indices and a few changes
- Added global secondary indices on upload date (needed for downloader performance), healthcode (so users can filter on their own files very quickly), and filename (which could be useful, unless there are a lot of filenames named "encrypted.zip")
- Combine uploadId and objectId. This eliminates the possibility of an S3 name collision. (Which was extremely unlikely, but possible before. Now, since it's the same as the dynamo key, it should be impossible.) This also allows us to, given an S3 file, get the DynamoDB metadata without needing a separate index.
- Re-worked the Upload interface to just include the bare minimum fields and not the whole slew of internal fields
- Re-worked UploadDao to get the Upload object instead of making individual calls for objectId and isComplete.

Currently, all new uploads will be written to Upload2. However, there is a small window where upload IDs with the old table are still being uploaded to, so uploadComplete will first check Upload2, then fall back to Upload. We can either backfill the old uploads, or we can drop them in favor of the new uploads table.

Testing done:
- manually tested uploads
- unit tests


app/org/sagebionetworks/bridge/services/UploadServiceImpl.java
```diff
@@ -82,7 +85,17 @@ public UploadSession createUpload(User user, UploadRequest uploadRequest) {
...
     public void uploadComplete(final String uploadId) {
-        final String objectId = uploadDao.getObjectId(uploadId);
+        if (Strings.isNullOrEmpty(uploadId)) {
+            throw new BadRequestException(String.format(Validate.CANNOT_BE_BLANK, "uploadId"));
+        }
+
+        // can't re-complete uploads that are already complete
+        Upload upload = uploadDao.getUpload(uploadId);
+        if (upload.isComplete()) {
+            throw new BadRequestException(String.format("Upload ID %s is already complete", uploadId));
```
eric-wu: Not sure about the value of an exception here.  This boolean flag may be set by a worker, for example.  This worker is an internal backend worker, not the worker for unpacking and exporting the uploads by the way.  If the client for some reason never called uploadComplete() but did finish the upload with S3.  The worker may pick up the discrepancy and update the upload table accordingly.

app/org/sagebionetworks/bridge/dynamodb/DynamoUpload2.java
```diff
@@ -0,0 +1,155 @@
...
+        return contentType;
+    }
+
+    /** @see #getContentType */
+    public void setContentType(String contentType) {
+        this.contentType = contentType;
+    }
+
+    /** Name of the file to upload. */
+    @DynamoDBIndexHashKey(attributeName = "filename", globalSecondaryIndexName = "filename-index")
```
eric-wu: Maintaining an index can be expensive.  If not desperately needed, I would suggest not add too many indices on tables directly supporting transactions.  On the other hand, once the data is exported, tables there are to support queries.  Indices should be set up there, downstream of the data pipeline.

app/org/sagebionetworks/bridge/dynamodb/DynamoUpload2.java
```diff
@@ -0,0 +1,155 @@
...
+        return filename;
+    }
+
+    /** @see #getFilename */
+    public void setFilename(String filename) {
+        this.filename = filename;
+    }
+
+    /** Health code of the user from which this upload originates from. */
+    @DynamoDBIndexHashKey(attributeName = "healthCode", globalSecondaryIndexName = "healthCode-index")
```
eric-wu, line 97: The same comment as the one on "filename-index".

PR #426, opened 2015-01-15
Survey units

alxdarksage:


app/org/sagebionetworks/bridge/validators/SurveyAnswerValidator.java
```diff
@@ -176,45 +171,62 @@ private void validateType(Errors errors, DecimalConstraints con, String answer)
...
+        if (con.getMaxValue() != null && numberAsDouble > con.getMaxValue()) {
+            if (con.getUnit() != null) {
+                rejectField(errors, "constraints", "%s is higher than the maximum value of %s %s", number, con.getMaxValue(), con.getUnit().name().toLowerCase());
+            } else {
+                rejectField(errors, "constraints", "%s is higher than the maximum value of %s", number, con.getMaxValue());
+            }
+        }
+        if (con.getStep() != null) {
+            String result = new BigDecimal(number).remainder(new BigDecimal(con.getStep()), MATH_CONTEXT).toString();
+            if (!result.startsWith("0.0") && !result.equals("0")) { // A simple approach to rounding that works
```
alxdarksage: Could do straight math on doubles and check that it's less than 0.1.  0.0000001 would be an issue, however.

test/org/sagebionetworks/bridge/validators/SurveyAnswerValidatorTest.java
```diff
@@ -217,13 +234,42 @@ public void validateIntegerMinValue() {
...
+        IntegerConstraints constraints = new IntegerConstraints();
+        constraints.setMaxValue(20d);
+        constraints.setStep(5d);
+        validator = new SurveyAnswerValidator(createQuestion(constraints));
+
+        SurveyAnswer answer = createAnswer("15");
+        Validate.entityThrowingException(validator, answer);
+    }
+    @Test
+    public void validateIntegerCanHaveUnits() {
```
alxdarksage: This is rather silly

app/org/sagebionetworks/bridge/validators/SurveyAnswerValidator.java
```diff
@@ -157,18 +163,7 @@ private void validateType(Errors errors, TimeBasedConstraints con, String answer
 
     private void validateType(Errors errors, DecimalConstraints con, String answer) {
```
DwayneJengSage: validateType(Errors, DecimalConstraints, String) and validateType(Errors, IntegerConstraints, String) differ only by the error message (which we could probably consolidate). Should we delete these and just use validateType(Errors, NumericalConstraints, String) and let Java's dynamic type system take care of this for us? (And move the try-catch.)

test/org/sagebionetworks/bridge/validators/SurveyAnswerValidatorTest.java
```diff
@@ -125,10 +125,27 @@ public void validateDecimalMinValue() {
     @Test(expected = InvalidEntityException.class)
     public void validateDecimalMaxValue() {
         DecimalConstraints constraints = new DecimalConstraints();
+        constraints.setMaxValue(10d); 
+        validator = new SurveyAnswerValidator(createQuestion(constraints));
+        Validate.entityThrowingException(validator, createAnswer("15.0"));
+    }
+    @Test(expected = InvalidEntityException.class)
+    public void validateInvalidDecimalStep() {
```
DwayneJengSage, line 135: Do we have unit tests for malformatted integers and decimals?

app/org/sagebionetworks/bridge/validators/SurveyAnswerValidator.java
```diff
@@ -176,45 +171,62 @@ private void validateType(Errors errors, DecimalConstraints con, String answer)
...
-        if (StringUtils.isNotBlank(con.getPattern()) && answer != null && !answer.matches(con.getPattern())) {
-            rejectField(errors, "constraints", "%s does not match the regular expression /%s/", answer, con.getPattern());
-        }
-    }
-
     private void validateType(Errors errors, DurationConstraints con, String answer) {
         if (con.getUnit() == null) {
             rejectField(errors, "constraints", "%s does not have a specified unit", answer);
+        } else if (!Unit.DURATION_UNITS.contains(con.getUnit())) {
+            rejectField(errors, "constraints", "%s is not a time unit", answer);
```
DwayneJengSage: Do we have unit tests for missing units or non-duration units?

test/org/sagebionetworks/bridge/validators/SurveyAnswerValidatorTest.java
DwayneJengSage, line 135: Do we have unit tests that have constraints with both min and max?

app/org/sagebionetworks/bridge/validators/SurveyAnswerValidator.java
alxdarksage: Yes, most of what you're looking for is in SurveyAnswerValidatorTest... but I'll double-check for each one.

test/org/sagebionetworks/bridge/validators/SurveyAnswerValidatorTest.java
alxdarksage, line 135: An interesting case for that would be also allowing multiple values, and multiple values are out of range... I'll add that.

Actually scratch that, if you include multiple values it must always be a MultiValueConstraint where you supply a list of the enumerated values. We don't validate past validating that the value supplied is one of the enumerated values.

PR #437, opened 2015-01-21
Remove JavaScript and other static content

eric-wu:
Do NOT merge yet.  There are paths in both packages that need to fixed.


app/interceptors/LoggingInterceptor.java
```diff
@@ -16,7 +16,7 @@ public Object invoke(MethodInvocation method) throws Throwable {
         if (logger.isInfoEnabled()) {
             Http.Request request = Http.Context.current().request();
             String[] values = request.headers().get("User-Agent");
-            String userAgent = (values.length > 0) ? values[0] : "<NONE>";
```
alxdarksage, line 19: Don't log UA if there really is none.

PR #438, opened 2015-01-21
wire up UploadArchiveService to CMS encryptors per study

DwayneJengSage:
This change adds study to the methods of the UploadArchiveService, in preparation for using it for the upload validation chain. This also includes adding a Guava cache for the CMS encryptors per study, as well as an S3 Helper to simplify loading files from S3.

This also includes Spring configs for an async worker thread pool, included mainly because it was difficult to separate it out.

Testing included:
- writing and running unit tests
- play run, to ensure Spring configs wired up correctly

Next steps:
- create upload validation service to call UploadArchiveService, refactor UploadArchiveService to parse non-JSON files


app/org/sagebionetworks/bridge/crypto/CmsEncryptorCacheLoader.java
```diff
@@ -0,0 +1,51 @@
...
+import org.sagebionetworks.bridge.s3.S3Helper;
+
+/**
+ * This is the cache loader that supports loading CMS encryptors on demand, keyed by the study ID. If the study's
+ * encryptor is already in the cache, this returns that encryptor. If it isn't, this study will pull the PEM fils for
+ * the cert and private key from the configured S3 bucket and construct an encryptor using those encryption materials.
+ */
+@Component
+public class CmsEncryptorCacheLoader extends CacheLoader<String, CmsEncryptor> {
+    public static final String CERT_BUCKET = BridgeConfigFactory.getConfig().getProperty("upload.cms.cert.bucket");
```
eric-wu: Why is this public? Outside this class, it is only referenced by the unit test. Could we make it at least package scope? The same comment for the PRIV_KEY_BUCKET field below.

app/org/sagebionetworks/bridge/crypto/CmsEncryptorCacheLoader.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /** S3 helper, configured by Spring. */
+    @Autowired
+    public void setS3Helper(S3Helper s3Helper) {
+        this.s3Helper = s3Helper;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public CmsEncryptor load(@Nonnull String study) throws CertificateEncodingException, IOException {
```
eric-wu: Just to be super clear and consistent, perhaps rename the parameter 'study' to 'studyIdentifier'?

app/org/sagebionetworks/bridge/s3/S3Helper.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.stereotype.Component;
+
+/** Helper class that simplifies reading S3 files. */
+@Component
+public class S3Helper {
+    private AmazonS3Client s3Client;
+
+    /** S3 Client. This is configured by Spring. */
+    @Autowired
+    public void setS3Client(AmazonS3Client s3Client) {
```
eric-wu, line 25: We have multiple beans of AmazonS3Client.  It's the "s3CmsClient" that should be injected here.  The client has only permissions to the bucket holding the key pairs and it uses different IAM credentials for different stacks. However, @Autowired AFAIK is injection by type.  It is not clear which one is injected.  I'd suggest using @Resource to inject here.

app/org/sagebionetworks/bridge/services/UploadArchiveService.java
```diff
@@ -5,73 +5,72 @@
...
-            throw new RuntimeException(e);
-        }
+    /** Loading cache for CMS encryptor, keyed by study ID. This is configured by Spring. */
+    @Autowired
+    public void setCmsEncryptorCache(LoadingCache<String, CmsEncryptor> cmsEncryptorCache) {
+        this.cmsEncryptorCache = cmsEncryptorCache;
     }
 
-    public byte[] zipAndEncrypt(List<ArchiveEntry> entries) {
+    public byte[] zipAndEncrypt(Study study, List<ArchiveEntry> entries) {
```
eric-wu, line 46: Minor comment: checkNotNull(study)

PR #441, opened 2015-01-22
add fields to Upload to support upload validation; add read bytes to S3 Helper

DwayneJengSage:
Part 2 of Upload Validation

Most of the files in this change are to add fields to Upload models (and the consequent changes to the UploadService, UploadDao, and dynamo marshallers to support this change). This also includes changes to S3 Helper to support reading either as bytes or as a String.

Testing done:
- added and ran unit tests for all files involved
- test upload manually to verify it still works

Next steps:
- hook this up with the UploadService and the UploadArchiveService to kick off upload validation


app/org/sagebionetworks/bridge/models/upload/Upload.java
```diff
@@ -1,18 +1,41 @@
...
+     * Calendar date the file was uploaded (specifically, the uploadComplete() call. Date is determined using Pacific
+     * local time.
+     */
+    LocalDate getUploadDate();
+
+    /** Upload ID. This is the key in the Dynamo DB table that uniquely identifies this upload. */
+    String getUploadId();
+
+    /** List of validation messages, generally contains error messages. */
+    List<String> getValidationMessageList();
```
eric-wu, line 50: Can you explain why we are expecting more than one message for a single upload?  An example would help.

app/org/sagebionetworks/bridge/models/upload/Upload.java
```diff
@@ -1,18 +1,41 @@
...
-    /** True if the upload is complete. False otherwise. */
-    boolean isComplete();
+    /** Represents upload status, such as requested, validation in progress, validation failed, or succeeded. */
+    UploadStatus getStatus();
+
+    /**
+     * Calendar date the file was uploaded (specifically, the uploadComplete() call. Date is determined using Pacific
+     * local time.
+     */
+    LocalDate getUploadDate();
```
eric-wu, line 40: Is there a reason that we don't use UTC?

PR #448, opened 2015-01-24
refactor UploadDao to use Upload instead of Upload ID; don't clean buckets in UploadCertServiceTest

DwayneJengSage:
Staging a few more changes so that we don't have a massive CR for all of the upload validation stuff.

Changes:
- refactored UploadDao and UploadService to take and return Upload instead of Upload ID - This removes most of the redudant calls to getUpload, and it allows the controller to call getUpload once and pass it into both the UploadService and the UploadValidationService
- merged UploadServiceImpl with UploadService - There was no value to having a separate interface for that.
- remove the cleanBucket() calls from UploadCertServiceTest - These wipe the bucket on a regular basis, which makes it really difficult to test uploads on local boxes.

Testing done:
- added and ran unit tests
- manually tested uploads

Next steps:
- upload validation service and async handlers


app/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -1,22 +1,114 @@
...
 import org.sagebionetworks.bridge.models.User;
+import org.sagebionetworks.bridge.models.upload.Upload;
 import org.sagebionetworks.bridge.models.upload.UploadRequest;
 import org.sagebionetworks.bridge.models.upload.UploadSession;
+import org.sagebionetworks.bridge.validators.Validate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.validation.Validator;
 
-public interface UploadService {
```
alxdarksage, line 7: add JIRA issue for merging Serivce interfaces and ServiceImpls together, for consistency

test/org/sagebionetworks/bridge/services/UploadCertificateServiceImplTest.java
```diff
@@ -52,8 +47,6 @@
     public void before() {
         assertNotNull(uploadCertificateService);
         assertNotNull(s3Client);
-        cleanBucket(CERT_BUCKET);
-        cleanBucket(PRIV_BUCKET);
     }
 
     @After
```
alxdarksage, line 52: add separate After methods to make sure S3 objects get cleaned

app/org/sagebionetworks/bridge/services/UploadService.java
DwayneJengSage, line 7: https://sagebionetworks.jira.com/browse/BRIDGE-330

test/org/sagebionetworks/bridge/services/UploadServiceMockTest.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Test
+    public void getUpload() {
+        // This is a simple call through to the DAO. Test the data flow.
+
+        // mock upload dao
+        UploadDao mockDao = mock(UploadDao.class);
+        Upload mockUpload = new DynamoUpload2();
+        when(mockDao.getUpload("test-upload-id")).thenReturn(mockUpload);
```
alxdarksage, line 32: is the upload always complete when this call works, or does the upload represent the upload when uploading as well? Is that something that would throw an exception if it were still incomplete?

PR #452, opened 2015-01-28
Upload Validation - Service and Handler Chain

DwayneJengSage:
This is the prototype for the Upload Validation Service and Handler Chain. I still need to add unit tests, comments, and a little refactoring, and there's a merge conflict I still need to resolve. However, I'm making this available for an initial high-level design review.

The overall architecture is UploadController calls UploadService.getUpload(), then passes it to UploadService.uploadComplete() and UploadValidationService.validateUpload(). The UploadValidationService kicks off a separate thread that does the actual validation asynchronously. (We could have used Spring context.getBean() instead of creating UploadValidationTaskFactory, but for testing and code factoring, it felt cleaner to not have a hard dependency on Spring.)

The UploadValidationTask uses handlers, so we can easily add handlers in the future without bloating Validation Task. (Future improvement, non-short-circuiting handlers - handlers that can fail but still allow the next steps of handlers to happen. This is when having a message list is more useful than just having a single message.)

Also, UploadCertServiceTest likes to wipe our cert and priv-key buckets in the local domain. If that happens to you, just upload test/resources/cms/rsacert.pem and rsaprivkey.pem to their respective buckets and rename them to api.pem. (There's a code change to make UploadCertServiceTest not do that.)

What the handler chain currently does:
- download from S3
- decrypt and unzip
- writes validation status and messages back to DynamoDB

What we still need to do:
- add API for get validation status and messages
- separate decrypt from unzip and from JSON parsing
- validate against schemas
- write intermediate artifacts
- wipe the files from S3 if validation fails

Testing done:
- manually tested it with a good file (test/resources/cms/data/archive) and a bad file (regular old text file)
- TODO add unit tests

### DIFF
--- a/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
+++ b/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
@@ -32,13 +32,13 @@ import org.sagebionetworks.bridge.upload.UploadValidationHandler;
 @ComponentScan(basePackages = "org.sagebionetworks.bridge")
 @Configuration
 public class BridgeSpringConfig {
-    @Bean(name = "AsyncExecutorService")
+    @Bean(name = "asyncExecutorService")
     @Resource(name = "numAsyncWorkerThreads")
     public ExecutorService asyncExecutorService(Integer numAsyncWorkerThreads) {
         return Executors.newFixedThreadPool(numAsyncWorkerThreads);
     }
 
-    @Bean(name = "CmsEncryptorCache")
+    @Bean(name = "cmsEncryptorCache")
     @Autowired
     public LoadingCache<String, CmsEncryptor> cmsEncryptorCache(CmsEncryptorCacheLoader cacheLoader) {
         return CacheBuilder.newBuilder().build(cacheLoader);
@@ -60,7 +60,7 @@ public class BridgeSpringConfig {
         return s3Helper;
     }
 
-    @Bean(name = "UploadDdbMapper")
+    @Bean(name = "uploadDdbMapper")
     @Autowired
     public DynamoDBMapper uploadDdbMapper(AmazonDynamoDB client) {
         DynamoDBMapperConfig mapperConfig = new DynamoDBMapperConfig.Builder().withSaveBehavior(
@@ -71,7 +71,7 @@ public class BridgeSpringConfig {
     }
 
     // TODO: Remove this when the migration is done
-    @Bean(name = "UploadDdbMapperOld")
+    @Bean(name = "uploadDdbMapperOld")
     @Autowired
     public DynamoDBMapper uploadDdbMapperOld(AmazonDynamoDB client) {
         DynamoDBMapperConfig mapperConfig = new DynamoDBMapperConfig.Builder().withSaveBehavior(
@@ -81,7 +81,7 @@ public class BridgeSpringConfig {
         return new DynamoDBMapper(client, mapperConfig);
     }
 
-    @Bean(name = "UploadValidationHandlerList")
+    @Bean(name = "uploadValidationHandlerList")
     @Autowired
     public List<UploadValidationHandler> uploadValidationHandlerList(S3DownloadHandler s3DownloadHandler,
             DecryptAndUnzipHandler decryptAndUnzipHandler) {
@@ -91,7 +91,7 @@ public class BridgeSpringConfig {
         return ImmutableList.of(s3DownloadHandler, decryptAndUnzipHandler);
     }
 
-    @Bean(name = "UploadSchemaDdbMapper")
+    @Bean(name = "uploadSchemaDdbMapper")
     @Autowired
     public DynamoDBMapper uploadSchemaDdbMapper(AmazonDynamoDB client) {
         DynamoDBMapperConfig mapperConfig = new DynamoDBMapperConfig.Builder()
--- a/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
+++ b/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
@@ -2,6 +2,7 @@ package org.sagebionetworks.bridge.config;
 
 import javax.annotation.Resource;
 
+import java.util.List;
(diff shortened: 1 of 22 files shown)


app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
```diff
@@ -0,0 +1,59 @@
...
+    public void setUploadDao(UploadDao uploadDao) {
+        this.uploadDao = uploadDao;
+    }
+
+    @Override
+    public void run() {
+        for (UploadValidationHandler oneHandler : handlerList) {
+            try {
+                oneHandler.handle(context);
+            } catch (UploadValidationException ex) {
```
alxdarksage: Do you want this to catch any unintended exceptions as well? If not, how will we know if these background tasks are failing? Just from logs?

app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
```diff
@@ -0,0 +1,59 @@
...
+                // unwrap to find cause, if any
+                Throwable inner = ex.getCause() != null ? ex.getCause() : ex;
```
alxdarksage: Throwables.getRootCause(ex); // guava libraries

app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
```diff
@@ -0,0 +1,59 @@
...
+
+                // log stuff, set the success flag, and break out of the loop
+                String handlerName = oneHandler.getClass().getName();
+                context.setSuccess(false);
+                context.addMessage(String.format("Error running upload validation handler %s: %s", handlerName,
+                        inner.getMessage()));
+                LOG.warn(String.format("Error running upload validation handler %s for study %s, upload %s",
```
alxdarksage: We haven't followed this style for loggers so far, they've been named "loggers"

app/org/sagebionetworks/bridge/upload/UploadValidationTaskFactory.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import org.sagebionetworks.bridge.dao.UploadDao;
+import org.sagebionetworks.bridge.models.studies.Study;
+import org.sagebionetworks.bridge.models.upload.Upload;
+
+@Component
+public class UploadValidationTaskFactory {
```
alxdarksage, line 20: RE: your comment: we are welded to Spring at this point. If it's less code to use context.getBean(), please do it. That's what Spring does after all, it makes objects.

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
```diff
@@ -68,6 +81,16 @@ public DynamoDBMapper uploadDdbMapperOld(AmazonDynamoDB client) {
         return new DynamoDBMapper(client, mapperConfig);
     }
 
+    @Bean(name = "UploadValidationHandlerList")
```
alxdarksage: By convention so far, our bean names have all started lower-case.

app/org/sagebionetworks/bridge/upload/UploadValidationException.java
```diff
@@ -0,0 +1,19 @@
+package org.sagebionetworks.bridge.upload;
+
+@SuppressWarnings("serial")
+public class UploadValidationException extends Exception {
```
eric-wu, line 8: extends RuntimeException?

app/org/sagebionetworks/bridge/services/UploadValidationService.java
```diff
@@ -0,0 +1,32 @@
...
+    public void setAsyncExecutorService(ExecutorService asyncExecutorService) {
+        this.asyncExecutorService = asyncExecutorService;
+    }
+
+    @Autowired
+    public void setTaskFactory(UploadValidationTaskFactory taskFactory) {
+        this.taskFactory = taskFactory;
+    }
+
+    public void validateUpload(Study study, Upload upload) {
```
eric-wu: Do we need to lock on the upload object to avoid race conditions?

app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
DwayneJengSage: Added RuntimeException to the multi-catch.

app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
DwayneJengSage: Talked offline with Alx. Throwables.getRootCause() gets the innermost exception, while I only want to remove the outermost. This is because intermediate exceptions contain useful information and stack traces. After some quick discussion, we decided it would be cleaner to remove the unwrapping code altogether.

app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
DwayneJengSage: Changed this to "logger".

app/org/sagebionetworks/bridge/upload/UploadValidationTaskFactory.java
DwayneJengSage, line 20: Discussed this offline with Alx. Having context.getBean() in my code means my unit tests have to be JUnit aware, which adds a bunch of complications to unit tests. The net complexity of using a factory object instead of calling context.getBean() is much lower, so I'm going to stick with the Factory class.

DwayneJengSage, line 20: Correction: "JUnit aware" should be "Spring aware".

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
DwayneJengSage: Updated this and all other beans in BridgeSpringConfig.java.

PR #454, opened 2015-01-29
Participants

alxdarksage:


app/controllers/BaseController.java
```diff
@@ -126,6 +126,16 @@ protected UserSession getAuthenticatedResearchOrAdminSession(Study study) {
         }
         throw new UnauthorizedException();
     }
+    
+    protected UserSession getAuthenticatedResearchSession(Study study) {
```
DwayneJengSage: Naming: Should it be ResearchSession or ResearcherSession? Also note that above, there are two identical methods named ResearcherOrAdminSession and ResearchOrAdminSession...

app/controllers/StudyController.java
```diff
@@ -28,6 +37,20 @@ public Result getStudyForResearcher() throws Exception {
...
+    public Result sendStudyParticipantsRoster() throws Exception {
+        Study study = studyService.getStudyByHostname(getHostname());
+        // Researchers only, administrators cannot get this list so easily
+        getAuthenticatedResearchSession(study);
+        
+        // One thing we can do here is verify that an email address has been set for the email
+        // that is about to be sent.
+        if (StringUtils.isBlank(study.getConsentNotificationEmail())) {
+            throw new BridgeServiceException("No consent notification contact email exists for this study. You must update the study with this contact before retrieving a roster of study participants.");
+        }
```
DwayneJengSage: Should this check be in the controller or in the service? I feel it would be better to have the controllers be minimal and have all the validation in the service.

app/org/sagebionetworks/bridge/services/SendMailViaAmazonService.java
```diff
@@ -87,29 +101,54 @@ public void sendConsentAgreement(User user, ConsentSignature consentSignature, S
...
-                sendEmailTo(study.getName(), email, bodyPart, sigPart);
+                sendEmailTo(subject, email, bodyPart, sigPart);
             }
         } catch(IOException | MessagingException e) {
             throw new BridgeServiceException(e);
         }
     }
+    
+    @Override
+    public void sendStudyParticipantsRoster(Study study, List<StudyParticipant> participants) {
```
DwayneJengSage, line 123: We should create a unit test for this, with mocks so we can verify the emails sent are reasonable.

app/org/sagebionetworks/bridge/services/SendMailViaAmazonService.java
```diff
@@ -162,4 +201,53 @@ private String createSignedDocument(ConsentSignature consent, StudyConsent study
...
+                sb.append(participant.getFirstName()).append(" ").append(participant.getLastName());
+            } else if (participant.getFirstName() != null) {
+                sb.append(participant.getFirstName());
+            } else if (participant.getLastName() != null) {
+                sb.append(participant.getLastName());
+            }
+            sb.append(")"+NEWLINE);
+            if (participant.getPhone() != null) {
+                sb.append("Phone: ").append(participant.getPhone()).append(NEWLINE);    
+            }
```
DwayneJengSage, line 171: Should this include email address?

app/org/sagebionetworks/bridge/services/SendMailViaAmazonService.java
```diff
@@ -162,4 +201,53 @@ private String createSignedDocument(ConsentSignature consent, StudyConsent study
         html = html.replace("@@signing.date@@", signingDate);
         return html;
     }
+    
+    private String createInlineParticipantRoster(List<StudyParticipant> participants) {
```
DwayneJengSage: It looks like there's some non-trivial logic here. Should we expose this to unit tests so we can get good branch coverage on this?

app/org/sagebionetworks/bridge/services/SendMailViaAmazonService.java
```diff
@@ -162,4 +201,53 @@ private String createSignedDocument(ConsentSignature consent, StudyConsent study
...
+        return sb.toString();
+    }
+    
+    private void append(StringBuilder sb, String value) {
+        if (value != null) {
+            sb.append(escapeCsv(value)).append(COMMA_DELIMITER);
+        } else {
+            sb.append("").append(COMMA_DELIMITER);
+        }
+    }
```
DwayneJengSage, line 256: From my understanding, there's no universal standard for CSV escaping. Should we consider an alternate format, like TSV (tab-separated values)? TSVs aren't as popular, but they're just as widely supported (by both Excel and by Synapse), easy to parse, and don't have escaping issues the CSVs do.

app/org/sagebionetworks/bridge/services/UserProfileServiceImpl.java
```diff
@@ -58,6 +84,67 @@ public User updateProfile(User user, UserProfile profile) {
...
     }
     
+    @Override
+    public void sendStudyParticipantRoster(Study study) {
+        checkNotNull(study, Validate.CANNOT_BE_NULL, "study");
+
+        executorService.submit(new ParticipantRosterGenerator(study));
+    }
+    
+    private class ParticipantRosterGenerator implements Runnable {
```
DwayneJengSage: We should expose this to unit tests so that we can meaningfully test it.

app/org/sagebionetworks/bridge/services/UserProfileServiceImpl.java
```diff
@@ -58,6 +84,67 @@ public User updateProfile(User user, UserProfile profile) {
...
+        
+        @Override
+        public void run() {
+            try {
+                List<StudyParticipant> participants = Lists.newArrayList();
+
+                // We must iterate over every account in the application, not just the study directory, in case 
+                // a user signed up with a different study. There is no way to search/query for specific custom 
+                // data, we must iterate over all accounts. Will this exceed 30 seconds eventually? It could.
+                // One solution would be to do this in a worker thread and email the results to the researcher.
```
DwayneJengSage: Nitpick: Reword this comment to something like "Our solution is to..." to make it clear that this is, in fact, what we're doing, and not some kind of TODO in disguise.

app/org/sagebionetworks/bridge/services/UserProfileServiceImpl.java
```diff
@@ -58,6 +84,67 @@ public User updateProfile(User user, UserProfile profile) {
...
+                String consentKey = study.getIdentifier() + BridgeConstants.CUSTOM_DATA_CONSENT_SIGNATURE_SUFFIX; 
+                Application app = StormpathFactory.getStormpathApplication(StormpathFactory.getStormpathClient());
```
DwayneJengSage: If we want to test this meaningfully, we're going to want to mock out the stormpath stuff. This means we'd have to make StormpathFactory an instantiable (and therefore mockable) factory instead of a static factory.

app/org/sagebionetworks/bridge/services/UserProfileServiceImpl.java
```diff
@@ -58,6 +84,67 @@ public User updateProfile(User user, UserProfile profile) {
...
+                StormpathAccountIterator iterator = new StormpathAccountIterator(app);
+                for (List<Account> page : iterator) {
+                    for (Account account : page) {
+                        if (account.getCustomData().containsKey(consentKey)) {
+                            UserProfile profile = profileFromAccount(account);
+                            StudyParticipant p = new StudyParticipant();
+                            p.setFirstName(profile.getFirstName());
+                            p.setLastName(profile.getLastName());
+                            p.setEmail(profile.getEmail());
+                            p.setPhone(profile.getPhone());
```
DwayneJengSage: Is it worthwhile encapsulating this in a StudyParticipant constructor?

app/org/sagebionetworks/bridge/models/studies/StudyParticipant.java
```diff
@@ -0,0 +1,81 @@
+package org.sagebionetworks.bridge.models.studies;
+
+public final class StudyParticipant {
```
DwayneJengSage, line 3: Does this class need to be mutable?

app/org/sagebionetworks/bridge/services/UserProfileServiceImpl.java
```diff
@@ -58,6 +84,67 @@ public User updateProfile(User user, UserProfile profile) {
...
+                            participants.add(p);
+                        }
+                    }
+                }
+                Collections.sort(participants, STUDY_PARTICIPANT_COMPARATOR);
+                sendMailService.sendStudyParticipantsRoster(study, participants);
```
DwayneJengSage: In our tests, we should have our mock iterator return the participants out of order, so we can exercise our comparator. Also, we should mock out the sendMailService so we can validate that we're properly filtering and sorting our participant list.

test/org/sagebionetworks/bridge/services/UserProfileServiceImplTest.java
```diff
@@ -60,4 +60,10 @@ public void canUpdateUserProfile() {
         assertEquals(testUser.getUsername(), userProfile.getUsername());
         assertEquals("123-456-7890", userProfile.getPhone());
     }
+    
+    @Test
+    public void canRetrieveStudyParticipants() {
+        // All we an really do here is verify no error is thrown.
+        profileService.sendStudyParticipantRoster(testUser.getStudy());
+    }
```
DwayneJengSage, line 89: Will this send a real email? How can we ensure that all our test studies are configured properly (with an email address set to "bridge-testing@sagebase.org", or similar) to ensure we don't get hammered by AWS's spam block?

app/controllers/BaseController.java
alxdarksage: Deleted duplicate, renamed for consistency, resorted the methods

app/controllers/StudyController.java
alxdarksage: Yes... moved

app/org/sagebionetworks/bridge/models/studies/StudyParticipant.java
alxdarksage, line 3: Subclassable? No, and final-ing the class makes the equals/hashCode methods work correctly without changing the default implementation provided by Eclipse. Fields aren't final simply so I can use setters rather than collecting all the data for a large set of constructor arguments. Okay for instances to be mutable.

app/org/sagebionetworks/bridge/services/SendMailViaAmazonService.java
alxdarksage, line 171: It does (always, see line 210).

alxdarksage, line 171: Adding tests for this and for CSV output.

app/org/sagebionetworks/bridge/services/SendMailViaAmazonService.java
alxdarksage, line 256: I tested this and the escaping library didn't seem to do enough to protect the cdv format. Switching to tabs.

app/org/sagebionetworks/bridge/services/UserProfileServiceImpl.java
alxdarksage: I am going to put Stormpath behind a facade that we could then easily mock. I can work around that for the moment by mocking storm path.

test/org/sagebionetworks/bridge/services/UserProfileServiceImplTest.java
alxdarksage, line 89: Mock this out, should not be sending email

alxdarksage, line 89: Now prevents this and the emailer service is tested separately.

app/org/sagebionetworks/bridge/services/SendMailViaAmazonService.java
```diff
@@ -162,4 +247,11 @@ private String createSignedDocument(ConsentSignature consent, StudyConsent study
         html = html.replace("@@signing.date@@", signingDate);
         return html;
     }
+    
+    private void append(StringBuilder sb, String value, boolean withComma) {
+        sb.append( (value != null) ? escapeCsv(value) : "" );
```
DwayneJengSage: If we're using tabs now, instead of escapeCsv(), we should probably use String.replace() to replace tabs with spaces.

alxdarksage: Yes

test/org/sagebionetworks/bridge/services/ParticipantRosterGeneratorTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.sagebionetworks.bridge.models.studies.StudyParticipant;
+
+import com.google.common.collect.Lists;
+import com.stormpath.sdk.account.Account;
+import com.stormpath.sdk.account.AccountCriteria;
+import com.stormpath.sdk.account.AccountList;
+import com.stormpath.sdk.application.Application;
+import com.stormpath.sdk.directory.CustomData;
+
+public class ParticipantRosterGeneratorTest {
```
DwayneJengSage, line 28: How about a test where some accounts get filtered out?

PR #456, opened 2015-01-30
get upload validation status API

DwayneJengSage:
This is a relatively small piece in the next steps for upload validation. This is an API in the upload controller that allows clients to get the upload validation status. (This will allow them to identify failed uploads and retry. Or at the very least, it can signal to the client that something is wrong with their upload file.)

Testing done:
- unit tests
- manually tested by calling get validation status via HTTP interface, once after the upload is requested, once after the upload validation fails

Next steps:
- separate decrypt from unzip from JSON parsing


conf/routes
```diff
@@ -55,6 +55,7 @@ POST   /api/v1/consent/dataSharing/resume    @controllers.ConsentController.resu
 # API - Upload
 POST   /api/v1/upload                  @controllers.UploadController.upload
 POST   /api/v1/upload/:id/complete     @controllers.UploadController.uploadComplete(id: String)
+GET    /api/v1/upload/:id/validationStatus  @controllers.UploadController.getValidationStatus(id: String)
```
alxdarksage: /api/v1/upload/:id/status? Is there any external notion of validation vs any other problem in the status of the upload that a user would be aware of?

test/org/sagebionetworks/bridge/models/upload/UploadValidationStatusTest.java
```diff
@@ -0,0 +1,184 @@
...
+    @Test
+    public void validatorDoesntSupportWrongClass() {
+        assertFalse(UploadValidationStatusValidator.INSTANCE.supports(String.class));
+    }
+
+    // branch coverage
+    // we call the validator directly, since Validate.validateThrowingException filters out nulls and wrong types
+    @Test
+    public void validateNull() {
+        MapBindingResult errors = new MapBindingResult(new HashMap(), "UploadValidationStatus");
```
alxdarksage: In Eclipse, simply using new HashMap<>() removed the warnings this generates.

PR #467, opened 2015-02-05
add timing to UploadValidationTask; separate decrypt, unzip, and parse JSON; add UploadArchiveUtil

DwayneJengSage:
The next step in Upload Validation. This change includes 3 parts:
1. Add simple timing to UploadValidationTask, which measures the time elasped for each handler and writes it to the log. (Future improvement is to write it to somewhere where we can make a dashboard out of it.)
2. Separate the Decrypt, Unzip, and Parse JSON steps, and make Parse JSON more robust (since not all data is JSON data).
3. Add UploadArchiveUtil, which allows us to use play as a command-line tool for encrypting and decrypting files.

Testing done:
- unit tests
- manually tested upload validation, with a file that includes both JSON and non-JSON, and with a file that includes two files with the same name
- manually tested UploadArchiveUtil, by encrypting a file, then decrypting to get the same file back

Next steps:
- parsing metadata from info.json and schema validation (iOS-specific)


app/org/sagebionetworks/bridge/services/UploadArchiveService.java
```diff
@@ -1,29 +1,29 @@
...
 import org.bouncycastle.cms.CMSException;
+
 import org.sagebionetworks.bridge.crypto.CmsEncryptor;
+import org.sagebionetworks.bridge.exceptions.BadRequestException;
 import org.sagebionetworks.bridge.exceptions.BridgeServiceException;
-import org.sagebionetworks.bridge.models.studies.Study;
-import org.sagebionetworks.bridge.models.upload.ArchiveEntry;
+import org.sagebionetworks.bridge.validators.Validate;
 
-import com.fasterxml.jackson.databind.JsonNode;
```
eric-wu, line 25: I like that we are getting rid of JSON here.

app/org/sagebionetworks/bridge/upload/DecryptHandler.java
```diff
@@ -25,6 +25,7 @@ public void setUploadArchiveService(UploadArchiveService uploadArchiveService) {
     /** {@inheritDoc} */
     @Override
     public void handle(@Nonnull UploadValidationContext context) {
-        uploadArchiveService.decryptAndUnzip(context.getStudy(), context.getData());
+        byte[] decryptedData = uploadArchiveService.decrypt(context.getStudy().getIdentifier(), context.getData());
+        context.setData(decryptedData);
```
eric-wu: This would wipe out the original data.  Alternatively can the decrypted data be on its own field?  When we call context.getDecryptedData(), we will know we are reading decrypted data.

DwayneJengSage: Done

PR #476, opened 2015-02-11
Caching

alxdarksage:
- Caches at the view layer (JSON, which is returned directly on a cached request)
- Doesn't handle a potential rush of requests to replace the cache when it is expired... if we really need this, it can be done with a blocking queue to reduce the requests to one per machine, worst case
- Caching surveys (individual GETs) and UserProfile. Will look for any other read requests that should be cached given current usage.


app/controllers/SurveyController.java
```diff
@@ -164,7 +201,8 @@ public Result versionSurvey(String surveyGuid, String createdOnString) throws Ex
         
         Survey survey = surveyService.getSurvey(keys);
         verifySurveyIsInStudy(session, study, survey);
-        
+
+        // This creates a new version of the survey, and so, no cache clearing needs to occur
```
DwayneJengSage: Won't this impact the most recent and published versions of the survey? (In the sense that they will now be pointing to different versions?)

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
```diff
@@ -42,7 +42,7 @@
 import com.google.common.collect.Maps;
 
 public class DynamoSurveyDao implements SurveyDao {
```
DwayneJengSage, line 44: There are still uncached code paths, and it only takes 2 r/s to potentially hit DDB throttling. Should we add a caching layer to the survey elements? Or is this unlikely enough that we don't need to defend against it?

app/controllers/UserProfileController.java
```diff
@@ -1,25 +1,39 @@
...
+    }
 
     public Result getUserProfile() throws Exception {
-        UserSession session = getAuthenticatedSession();
-
-        UserProfile profile = userProfileService.getProfile(session.getUser().getEmail());
-        return okResult(profile);
+        final UserSession session = getAuthenticatedSession();
+
+        String json = viewCache.getView(UserProfile.class, session.getUser().getEmail(), new Supplier<UserProfile>() {
```
DwayneJengSage: If we use email as the cache key, could this impact our ability to re-use email addresses across multiple studies?

app/org/sagebionetworks/bridge/cache/CacheProvider.java
```diff
@@ -89,7 +93,42 @@ public Study getStudy(String identifier) {
...
+            promptToStartRedisIfLocalEnv(e);
+            throw new BridgeServiceException(e);
+        }
+    }
+    
+    public String getString(String cacheKey) {
+        try {
+            String ser = stringOps.get(cacheKey);
+            if (ser != null) {
+                stringOps.expire(cacheKey, BridgeConstants.BRIDGE_SESSION_EXPIRE_IN_SECONDS);
```
DwayneJengSage: This extends the cache TTL every time a string is accessed. This is probably not desirable for the general case.

For cases where we do want to extend the cache TTL, this should be configurable, rather than hardcoded at 24 hours. (Also, the cache TTL shouldn't use the same constant as session TTL.)

app/org/sagebionetworks/bridge/cache/CacheProvider.java
```diff
@@ -16,13 +16,17 @@
  */
 public class CacheProvider {
```
DwayneJengSage, line 17: We should add unit tests for getString, setString, and removeString.

app/org/sagebionetworks/bridge/cache/ViewCache.java
```diff
@@ -0,0 +1,59 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.google.common.base.Supplier;
+
+public class ViewCache {
+    
+    // Could add a BlockingQueue and Executors.newSingleThreadExecutor() to prevent a run of 
+    // requests when the view is expired. But I don't think this is likely to be a very big
+    // problem.
+
+    private static Logger logger = LoggerFactory.getLogger(ViewCache.class);
```
DwayneJengSage: Nitpick: final

app/org/sagebionetworks/bridge/cache/ViewCache.java
```diff
@@ -0,0 +1,59 @@
...
+    private <T> String cacheView(String cacheKey, Supplier<T> supplier) throws JsonProcessingException {
+        logger.debug("Caching JSON for " +cacheKey+"'");
+        T object = supplier.get();
+        String value = BridgeObjectMapper.get().writeValueAsString(object);
+        cache.setString(cacheKey, value, 18000); // 5 hours
+        return value;
+    }
+    
+    private String getCacheKey(Class<?> clazz, String id) {
+        return RedisKey.VIEW.getRedisKey(id + ":" + clazz.getSimpleName());
```
DwayneJengSage: Should this be getName() instead of getSimpleName()? In the off-chance we end up caching 2 different types with the same class name?

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
alxdarksage: Survey elements are always retrieved as part of a survey. If we cache a survey, we cache the elements. The only place we query for them is in the query for a survey.

app/controllers/UserProfileController.java
alxdarksage: Yes. There will have to be separate profiles for each account. We'll need to add the study to this cache key.

app/org/sagebionetworks/bridge/cache/CacheProvider.java
alxdarksage: I think this is okay if we're manually expiring the cache (it could never expire). I will change the expire constant if we keep it.

app/org/sagebionetworks/bridge/cache/ViewCache.java
alxdarksage: I don't know if the length of the Redis key matters in any way, but yes.

test/org/sagebionetworks/bridge/cache/ViewCacheTest.java
```diff
@@ -0,0 +1,144 @@
...
+        cache.setCacheProvider(provider);
+        
+        // It doesn't get wrapped or transformed or anything
+        try {
+            cache.getView(Study.class, study.getIdentifier(), new Supplier<Study>() {
+                @Override public Study get() {
+                    throw new BridgeServiceException("There has been a problem retrieving the study");
+                }
+            });
+        } catch(BridgeServiceException e) {
```
DwayneJengSage, line 72: Add a fail() so that if it doesn't throw, this test will catch that.

test/org/sagebionetworks/bridge/cache/ViewCacheTest.java
```diff
@@ -0,0 +1,144 @@
...
+        
+        CacheProvider provider = mock(CacheProvider.class);
+        when(provider.getString(cacheKey)).thenReturn(originalStudyJson);
+        
+        ViewCache cache = new ViewCache();
+        cache.setCacheProvider(provider);
+        
+        String json = cache.getView(Study.class, study.getIdentifier(), new Supplier<Study>() {
+            @Override public Study get() {
+                Study study = new DynamoStudy();
```
DwayneJengSage: Add a fail() here to make sure get() never gets called.

test/org/sagebionetworks/bridge/cache/ViewCacheTest.java
```diff
@@ -0,0 +1,144 @@
...
+            }
+            public void removeString(String cacheKey) {
+                map.remove(cacheKey);
+            }
+        };   
+    }
+    
+    private String getCacheKey(Class<?> clazz, String id) {
+        return RedisKey.VIEW.getRedisKey(id + ":" + clazz.getSimpleName());
+    }
```
DwayneJengSage: Instead of duplicating this, would it be better to expose ViewCache.getCacheKey(), either at the protected or package-scoped level?

test/org/sagebionetworks/bridge/cache/ViewCacheTest.java
```diff
@@ -0,0 +1,144 @@
...
+        when(provider.getString(cacheKey)).thenReturn(null);
+        
+        ViewCache cache = new ViewCache();
+        cache.setCacheProvider(provider);
+        
+        // It doesn't get wrapped or transformed or anything
+        try {
+            cache.getView(Study.class, study.getIdentifier(), new Supplier<Study>() {
+                @Override public Study get() {
+                    throw new BridgeServiceException("There has been a problem retrieving the study");
```
DwayneJengSage, line 68: Do we care about the JsonProcessingException case?

test/org/sagebionetworks/bridge/cache/ViewCacheTest.java
alxdarksage: Yes. I changed ViewCache so it has a method, getCacheKey() that returns a ViewCacheKey, and that is used in the other signatures. Now all code uses that method to construct the key.

test/org/sagebionetworks/bridge/cache/ViewCacheTest.java
alxdarksage, line 68: Insofar as it shouldn't happen.. I think that would fail either way and we want that, no?

app/org/sagebionetworks/bridge/cache/CacheProvider.java
alxdarksage, line 17: Added

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
alxdarksage, line 44: After discussing, I think we are ok with the fact this can cache the same survey 3 times before all paths are cached. In practice 99% of requests will go through one path and the others will not put too much additional load on the system.

app/org/sagebionetworks/bridge/cache/ViewCache.java
```diff
@@ -0,0 +1,82 @@
...
+    
+    /**
+     * Get the view cached with the given identifier, or if that view has not been cached, 
+     * call the supplier, convert the response to JSON and cache it, then return that view.
+     * @param clazz
+     * @param id
+     * @param supplier
+     * @return
+     */
+    public <T> String getView(ViewCacheKey<T> key, Supplier<T> supplier) {
```
DwayneJengSage, line 44: Minor: params in the comments don't match the actual params.

app/org/sagebionetworks/bridge/cache/ViewCache.java
```diff
@@ -0,0 +1,82 @@
...
+            throw new BridgeServiceException(e);
+        }
+    }
+
+    /**
+     * Remove the view with the given identifier.
+     * @param clazz
+     * @param id
+     */
+    public <T> void removeView(ViewCacheKey<T> key) {
```
DwayneJengSage, line 62: Minor: same params issue here

app/org/sagebionetworks/bridge/cache/ViewCache.java
```diff
@@ -0,0 +1,82 @@
...
+        private final String key;
+        public ViewCacheKey(String key) {
+            this.key = key;
+        }
+        String getKey() {
+            return key;
+        }
+    };
+    
+    private static Logger logger = LoggerFactory.getLogger(ViewCache.class);
```
DwayneJengSage: nitpick: final

alxdarksage: Yes sorry I forgot that one. Marked final.

PR #477, opened 2015-02-11
BRIDGE-343 - adding email to consents, removing birthdate

alxdarksage:


app/org/sagebionetworks/bridge/services/SendMailViaAmazonService.java
```diff
@@ -87,7 +87,7 @@ public void sendConsentAgreement(User user, ConsentSignature consentSignature, S
         try {
```
alxdarksage: So, do we want the birth date sent as part of the participant report, so the consent admin can prove everyone declared themselves of age.

PR #480, opened 2015-02-13
Scope and associate user's study to their session

alxdarksage:
An important step towards study-scoped authentication:
- Fixes security hole where you could sign in under one study, then act under another one by using a different study identifier in the header/host. Study information now comes from a user session. Once authenticated, that's the study you're operating under.
- Stop using study when you're only looking for the identifier or researcher role, but use StudyIdentifier to provide type safety. No trip to Redis or DynamoDB.
- Only look for study in the header, host name or configuration file when doing an action like signing in, where we don't have an authenticated session.
- v2 authentication will explicitly ask for the study you are signing in to, and go to that study's directory to authenticate you.


app/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDao.java
```diff
@@ -53,13 +53,13 @@ public void setDynamoDbClient(AmazonDynamoDB client) {
     }
     
     @Override
-    public SchedulePlan getSchedulePlan(Study study, String guid) {
-        checkNotNull(study, "Study is null");
-        checkArgument(StringUtils.isNotBlank(study.getIdentifier()), "Study key is null");
+    public SchedulePlan getSchedulePlan(StudyIdentifier studyIdentifier, String guid) {
+        checkNotNull(studyIdentifier, "studyIdentifier is null");
+        checkArgument(StringUtils.isNotBlank(studyIdentifier.getIdentifier()), "Study key is null");
```
alxdarksage: this is now impossible so these can be removed.

test/controllers/AuthenticationControllerTest.java
```diff
@@ -90,4 +106,35 @@ public void testCode() throws Exception {
         });
     }
     
+    @Test
+    public void onceAuthenticatedUserCannotSwitchStudies() {
```
alxdarksage, line 115: This could be an SDK test.

app/controllers/BaseController.java
```diff
@@ -64,17 +68,25 @@ public void setBridgeConfig(BridgeConfig bridgeConfig) {
     }
     
     /**
+     * Returns a session. Will not throw exception if user is not authorized or has not consented to research.
+     * @return session if it exists, or null otherwise.
+     */
+    protected UserSession getSessionIfItExists() {
+        String sessionToken = getSessionToken();
+        return authenticationService.getSession(sessionToken);
```
DwayneJengSage, line 79: What happens if you call getSession with a null or empty session token? Should have have a check here? Or possibly in authService?

test/controllers/AuthenticationControllerTest.java
```diff
@@ -45,19 +57,49 @@
...
+    
     @Before
     public void before() {
         testUser = helper.createUser(AuthenticationControllerTest.class);
+        
+        String id = RandomStringUtils.randomAlphabetic(7).toLowerCase();
+        secondStudy = new DynamoStudy();
+        secondStudy.setIdentifier(id);
+        secondStudy.setName("Second Test Study");
+        studyService.createStudy(secondStudy);
```
DwayneJengSage: @Before is run with every test. Not every test needs the second study. You should either change this to @BeforeClass or move it into the individual test method.

PR #481, opened 2015-02-13
health data v2

DwayneJengSage:
New version of health data. Key changes:
- Wiped out all the old health data APIs. Since no one is using health data v1, and since we aren't supporting health data v1, this change will make it so health data v1 API calls will return 400 Bad Request.
- Rewrote the health data controllers, services, and daos. Two APIs:
  *\* createRecord - used by worker apps to create health data records from unpacked uploads
  *\* getRecordsByUploadDate - used by worker apps for exporting health data
- A smattering of DDB marshallers and Jackson serializers for Joda classes.
- A DynamoIndexHelper class, which allows querying against Global Secondary Indices.

Testing done:
- unit tests
- manually tested health data by creating 2 records, then calling getByUploadDate

Next steps:
- add validation in createRecord() to cross-check health codes and schema IDs
- remove the deprecated Tracker APIs
- implement worker accounts, so that worker accounts don't need to be admins
- make upload validation call createRecord


app/org/sagebionetworks/bridge/dynamodb/DateTimeMarshaller.java
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * DynamoDB marshaller for Joda DateTime (because the built-in ones don't work very well). This simply marshalls
+ * to and from the yyyy-MM-dd'T'HH:mm:ss.SSSZZ format.
+ */
+public class DateTimeMarshaller implements DynamoDBMarshaller<DateTime> {
+    /** {@inheritDoc} */
+    @Override
+    public String marshall(DateTime dateTime) {
+        return dateTime.toString(ISODateTimeFormat.dateTime());
```
alxdarksage: We have this: DateUtils.getISODateTime(DateTime date)

app/org/sagebionetworks/bridge/dynamodb/DateTimeMarshaller.java
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public DateTime unmarshall(Class<DateTime> clazz, String data) {
+        return DateTime.parse(data, ISODateTimeFormat.dateTime());
```
alxdarksage: But not this, could you add this to DateUtils? These conversions come up all the time.

app/org/sagebionetworks/bridge/dao/HealthDataDao.java
```diff
@@ -1,24 +1,36 @@
...
+     * DAO method used by worker apps to creating a health data record and persisting it, generally from unpacking
+     * uploads.
+     *
+     * @param record
+     *         health data record prototype, from which the record should be created from, must be non-null
+     * @return the unique ID of the created record
+     */
+    String createRecord(@Nonnull HealthDataRecord record);
 
-    public List<HealthDataRecord> appendHealthData(HealthDataKey key, List<HealthDataRecord> records);
```
alxdarksage, line 10: It makes me a little nervous to delete all this, because I believe post-Apple, we'll probably return at least partly to the original idea of Bridge. But, we can create this stuff again.

app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
```diff
@@ -1,185 +1,154 @@
...
-    private static final String END_DATE_FIELD = "endDate";
-    private static final String VERSION_FIELD = "version";
-    private static final String DATA_FIELD = "data";
-    
-    private String key;
-    private String guid;
-    private long startDate;
-    private long endDate;
     private JsonNode data;
+    private String healthCode;
```
alxdarksage, line 26: Can you still retrieve one type of data record for one individual?

app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
```diff
@@ -1,185 +1,154 @@
...
+    private String id;
+    private DateTime measuredTime;
```
alxdarksage: What is measured time? The time when the data was recorded? Sage conventionally would call that createdOn, as opposed to uploadedOn. Elsewhere they use long for these timestamps, in Synapse, and we did so as well in Bridge, although I like the use of DateTime better.

app/org/sagebionetworks/bridge/dynamodb/DynamoIndexHelper.java
```diff
@@ -0,0 +1,97 @@
...
+                }
+
+                recordList.add((T) oneResult);
+            }
+        }
+
+        return recordList;
+    }
+
+    /**
```
alxdarksage, line 89: Internal comment /\* */ or // ... ?

app/org/sagebionetworks/bridge/json/DateTimeToStringSerializer.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Custom serializer for Joda DateTime, because the one in jackson-datatype-joda serializes to a long. This one
+ * serializes to a string in yyyy-MM-dd'T'HH:mm:ss.SSSZZ format.
+ */
+public class DateTimeToStringSerializer extends JsonSerializer<DateTime> {
+    /** {@inheritDoc} */
+    @Override
+    public void serialize(DateTime dateTime, JsonGenerator jgen, SerializerProvider provider) throws IOException {
+        jgen.writeString(dateTime.toString(ISODateTimeFormat.dateTime()));
```
alxdarksage: DateUtils.convertToISODateTime(long l)

app/org/sagebionetworks/bridge/json/JodaDateTimeDeserializer.java
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Custom Joda DateTime deserializer, because the one in jackson-datatype-joda doesn't work. This one deserializes from
+ * a string in yyyy-MM-dd'T'HH:mm:ss.SSSZZ format.
+ */
+public class JodaDateTimeDeserializer extends JsonDeserializer<DateTime> {
+    /** {@inheritDoc} */
+    @Override
+    public DateTime deserialize(JsonParser jp, DeserializationContext ctxt) throws IOException {
+        String dateTimeStr = jp.getText();
+        return ISODateTimeFormat.dateTimeParser().parseDateTime(dateTimeStr);
```
alxdarksage: Again this conversion is missing from DateUtils, could you move it over to there?

app/org/sagebionetworks/bridge/json/LocalDateToStringSerializer.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Custom serializer for Joda LocalDate, because the one in jackson-datatype-joda serializes to a weird format. This
+ * one serializes to a string in a simple YYYY-MM-DD format.
+ */
+public class LocalDateToStringSerializer extends JsonSerializer<LocalDate> {
+    /** {@inheritDoc} */
+    @Override
+    public void serialize(LocalDate date, JsonGenerator jgen, SerializerProvider provider) throws IOException {
+        jgen.writeString(date.toString(ISODateTimeFormat.date()));
```
alxdarksage: DateUtils, same comment.

app/org/sagebionetworks/bridge/services/UserAdminServiceImpl.java
```diff
@@ -154,16 +146,4 @@ private boolean deleteUserInStudy(Study study, Account account, User user) throw
             return false;
         }
     }
-
-    private void removeAllHealthDataRecords(Study study, User user) throws BridgeServiceException {
```
alxdarksage, line 158: We have the requirement to delete everything of a user when we delete a user. I mean, a regulatory requirement to do this. We have to have that functionality somewhere... do we?

conf/routes
```diff
@@ -111,5 +102,9 @@ POST   /admin/v1/backfill/:name/start    @controllers.BackfillController.start(n
 POST   /admin/v1/users          @controllers.UserManagementController.createUser
 DELETE /admin/v1/users          @controllers.UserManagementController.deleteUser(email: java.lang.String ?= null)
 
+# Worker - Health Data
+POST /worker/v2/healthdata @controllers.HealthDataController.createRecord
```
alxdarksage, line 106: We should be able to add a worker role now and enforce that in the controller, if you don't want to use admin.

app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
alxdarksage: I'm also curious how you query in a date range against this value stored as a timestamp string... I didn't know DynamoDB could do that.

The other thing I wonder about is local vs. UTC time. Is this stored in UTC time? Let's say the user collects data in Los Angeles and New York, will a time-based query correctly account for time zone changes?

conf/routes
```diff
@@ -16,15 +16,6 @@ POST   /api/v1/auth/signUp                  @controllers.AuthenticationControlle
 POST   /api/v1/auth/verifyEmail             @controllers.AuthenticationController.verifyEmail
 POST   /api/v1/auth/resendEmailVerification @controllers.AuthenticationController.resendEmailVerification
 
-# API - Individual Health Data Record
```
alxdarksage, line 19: There are probably integration tests to remove with the removal of this functionality.

app/org/sagebionetworks/bridge/services/HealthDataService.java
```diff
@@ -2,24 +2,74 @@
...
+     * Returns a list of all health data records with the given upload date. THis method is generally called by
+     * worker apps as part of data export.
+     *
+     * @param uploadDate
+     *         an upload date in YYYY-MM-DD format, must be non-null, non-empty, and represent a valid date
+     * @return list of health data records
+     */
+    public List<HealthDataRecord> getRecordsForUploadDate(String uploadDate) {
+        // validate upload date
+        if (Strings.isNullOrEmpty(uploadDate)) {
```
alxdarksage: FWIW, this method allows empty strings like "   ", whereas the Apache version isNotBlank() does not.

app/org/sagebionetworks/bridge/dynamodb/DateTimeMarshaller.java
DwayneJengSage: Adding relevant utils to DateUtils and changing my code to use them.

app/org/sagebionetworks/bridge/dao/HealthDataDao.java
DwayneJengSage, line 10: Discussed this offlne with Alx. We're still planning on supporting these features in the future. We're just not implementing these yet, since no one needs them yet. However, Health Data v2 is built in a way that we could add these easily.

app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
DwayneJengSage, line 26: Yes, we have a secondary index on health code, and we could filter by schema. (Whether we want a specialized secondary index on both health code and schema can be easily answered in the future.)

app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
DwayneJengSage: Renamed this to createdOn.

Re: long millis vs DateTime - After discussion with the team, we decided it would be simpler for our object model to use long millis, but the JSON serialization should use ISO timestamps.

Re: Time zone. Standard epoch time is always in UTC, and DateTime is backed by epoch time, so regardless of time zone, it always refers to the same instant in time. (Time zone is, in fact, only used for formatting ISO 8601 data and to provide a reference point for the hour.) Ultimately, the the timestamp comes from the client, so it's up to the client to provide a non-ambiguous timestamp.

app/org/sagebionetworks/bridge/dynamodb/DynamoIndexHelper.java
DwayneJengSage, line 89: Discussed this offline with Alx. I felt this was important enough to include in the javadoc.

app/org/sagebionetworks/bridge/services/UserAdminServiceImpl.java
DwayneJengSage, line 158: Discussed this with Alx. I had broken up the changes in order to keep the code review small. This will be implemented in the next code change. Tracked by https://sagebionetworks.jira.com/browse/BRIDGE-363

conf/routes
DwayneJengSage, line 106: Similarly, this was pushed off to the next code change to keep the code review small. This is tracked in https://sagebionetworks.jira.com/browse/BRIDGE-364

conf/routes
DwayneJengSage, line 19: Fixed.

app/org/sagebionetworks/bridge/services/HealthDataService.java
DwayneJengSage: I like that better. Fixed here and in HeatlhDataRecordBuilder.

app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
alxdarksage: Yes with longs we're fine, they're all UTC, they sort, etc.

app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
```diff
@@ -2,184 +2,151 @@
...
-    private String guid;
-    private long startDate;
-    private long endDate;
+    private Long createdOn;
     private JsonNode data;
+    private String healthCode;
+    private String id;
+    private JsonNode metadata;
+    private String schemaId;
+    private LocalDate uploadDate;
```
alxdarksage, line 30: What's the thinking behind having uploadDate a LocalDate? It would seem that we know the time in an absolute sense, wouldn't it be a Long?

app/org/sagebionetworks/bridge/json/DateUtils.java
```diff
@@ -3,21 +3,37 @@
...
+    /** Returns current calendar date as a Joda LocalDate object, using UTC as the reference time zone. */
+    public static LocalDate getCurrentCalendarDateInUtc() {
+        return LocalDate.now(DateTimeZone.UTC);
+    }
+
+    /**
+     * Returns current calendar date as a Joda LocalDate object, using Pacific local time as the reference time zone.
+     */
+    public static LocalDate getCurrentCalendarDateInLocalTime() {
+        return LocalDate.now(BridgeConstants.LOCAL_TIME_ZONE);
```
alxdarksage, line 35: I'm a little confused by this because once you set a time zone, it's no longer a local date, it's a date date. Or is that not true because there's no time component?

PR #484, opened 2015-02-16
Annotate tables with specific throughputs

eric-wu:
1) Remove the interface DynamoTable.
2) Add annotation DynamoThroughput. Use this annotation to specify throughputs that not the default values.
3) Remove the throughput validation on global indices.

TODO: Update throughputs via the initializer.

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoInitializer.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoInitializer.java
@@ -383,16 +383,12 @@ public class DynamoInitializer {
     }
 
     /**
-     * Compares hash key, range key, secondary indices of the two tables. Throws an exception
+     * Compares hash key, range key of the two tables. Throws an exception
      * if there is difference.
      */
     static void compareSchema(TableDescription table1, TableDescription table2) {
         if (table1.getTableName().equals(table2.getTableName())) {
             compareKeySchema(table1, table2);
-            compareSecondaryIndices(table1.getTableName(), table1.getGlobalSecondaryIndexes(),
-                    table2.getGlobalSecondaryIndexes(), true);
-            compareSecondaryIndices(table1.getTableName(), table1.getLocalSecondaryIndexes(),
-                    table2.getLocalSecondaryIndexes(), false);
         }
     }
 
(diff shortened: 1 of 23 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoInitializer.java
```diff
@@ -39,22 +48,14 @@
...
 public class DynamoInitializer {
 
     private static Logger logger = LoggerFactory.getLogger(DynamoInitializer.class);
 
-    // DynamoDB Free tier
-    private static final Long READ_CAPACITY = Long.valueOf(25);
-    private static final Long WRITE_CAPACITY = Long.valueOf(25);
+    // Default capacities are the DynamoDB free tier
+    static final long READ_CAPACITY = 25;
+    static final long WRITE_CAPACITY = 25;
```
DwayneJengSage: Can we rename these to DEFAULT_READ_CAPACITY and DEFAULT_WRITE_CAPACITY so it's clear that's what they are?

Also, can we make them public for code cleanliness?

app/org/sagebionetworks/bridge/dynamodb/DynamoInitializer.java
```diff
@@ -79,59 +80,49 @@ public static void init(String dynamoPackage) {
...
-     * For phasing out obsolete schemas.
+     * Actions performed before init(), e.g. for phasing out obsolete schemas.
      */
     static void beforeInit() {
     }
 
-    static void deleteTable(String table) {
-        table = getTableName(table);
+    static void deleteTable(Class<?> table) {
+        final String tableName = TableNameOverrideFactory.getTableNameOverride(table).getTableName();
```
DwayneJengSage: TableNameOverrideFactory.getTableNameOverride(table).getTableName() seems to come up a lot. Consider making a DynamoUtils class with this as a helper method. (At the very least, it might be easier to mock and test this way.)

app/org/sagebionetworks/bridge/dynamodb/DynamoThroughput.java
```diff
@@ -0,0 +1,13 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface DynamoThroughput {
+    long writeCapacity() default DynamoInitializer.WRITE_CAPACITY;
+    long readCapacity() default DynamoInitializer.READ_CAPACITY;
+}
```
DwayneJengSage, line 13: There's already defaulting logic in DynamoInitializer. I feel like having defaults here would be redundant. Consider making read and write capacity required. Also, consider shortening them to write() and read() so we can write the annotation shorter:
@DynamoThroughput(read = 50, write = 50)

app/org/sagebionetworks/bridge/dynamodb/DynamoInitializer.java
```diff
@@ -490,18 +488,6 @@ static void compareSecondaryIndices(String tableName, List<?> indices1, List<?>
                             " is changing the projection.");
                 }
 
-                if (!globalIndex1.getProvisionedThroughput().getReadCapacityUnits().equals(
-                        globalIndex2.getProvisionedThroughput().getReadCapacityUnits())) {
-                    throw new BridgeInitializationException("Table " + tableName + " global index " + indexName +
-                            " is changing the provisioned read capacity.");
-                }
```
DwayneJengSage, line 497: Let's take this a step further and wipe out compareSecondaryIndices() altogether. Otherwise, this will prevent us from creating, deleting, or modifying indices in the future.

app/org/sagebionetworks/bridge/dynamodb/DynamoInitializer.java
```diff
@@ -315,7 +311,9 @@ static void initTables(final List<TableDescription> tables) {
                 logger.info("Creating table " + table.getTableName());
                 DYNAMO.createTable(createTableRequest);
             } else {
-                compareSchema(table, existingTables.get(table.getTableName()));
+                final TableDescription existingTable = existingTables.get(table.getTableName());
+                compareSchema(table, existingTable);
+                updateThroughput(table,existingTable);
```
DwayneJengSage: I'm not sure we should do this. This means that if someone changes the throughput in Dynamo, when DynamoInitializer comes back, it'll revert those changes. It would be much safer to use DynamoThroughput only for table creation, and if we want to sync up our DDB tables with our code, there should be an explicit process to do that.

alxdarksage: I agree with these remarks. We don't want to limit our ability to scale horizontally by requiring we change and deploy code to do it.

app/org/sagebionetworks/bridge/dynamodb/DynamoUpload.java
```diff
@@ -17,8 +17,9 @@
 import org.joda.time.DateTimeZone;
 import org.joda.time.LocalDate;
 
+@DynamoThroughput(writeCapacity=50, readCapacity=50)
 @DynamoDBTable(tableName = "Upload")
-public class DynamoUpload implements Upload, DynamoTable {
+public class DynamoUpload implements Upload {
```
DwayneJengSage, line 22: This is safe, but probably not necessary, since no one uses DynamoUpload anymore. (I'll need to delete it in a future code change.) Everything goes through DynamoUpload2 now.

DwayneJengSage:
I think the overall approach is okay, but I think we need to address the following:
- If we change the throughput in the DDB dashboard, the DynamoInitializer shouldn't change them back.
- We should delete compareSecondaryIndices(), as this will prevent us from updating our indices outside of code.
- Don't forget to add/update unit tests!

app/org/sagebionetworks/bridge/dynamodb/DynamoInitializer.java
eric-wu, line 497: This prevention might be necessary.  Without it we are running the risk of wiping out data by accident.  The schema check in general enforces data migration for schema changes.

alxdarksage: If we need to meet about this, let's meet about this. Whatever error you are seeking to prevent here, you are also preventing us from reacting in real time to scaling challenges.

eric-wu, line 497: Just to be clear.  Throughput validate is being removed in this pull request.

But adding, deleting, renaming indices are operations with no small impact  and, in my opinion, should always go through code deploy and data migrations.

DwayneJengSage:
Let's have a 30 minute meeting and talk about validation on global secondary indices. Other than that, the change looks good.

PR #487, opened 2015-02-17
Access-style log of UA and other; removing thread from loggers

alxdarksage:
On my dev machine the logging for a request now includes this:

2015-02-17 15:55:51,769 INFO  interceptors.LoggingInterceptor - 0:0:0:0:0:0:0:1 "GET /api/v1/schedules HTTP/1.1" 401 Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_5) AppleWebKit/600.3.18 (KHTML, like Gecko) Version/7.1.3 Safari/537.85.12

Which tells us what, where, and what the result was. I did take out the thread because I believe the request ID is a more accurate way to tie separate logging statements together, and the lines are already noisy.

### DIFF
--- a/app/interceptors/LoggingInterceptor.java
+++ b/app/interceptors/LoggingInterceptor.java
@@ -19,18 +19,25 @@ public class LoggingInterceptor implements MethodInterceptor {
             // Common Log Format for access logging, minus the first part which 
             // is already part of our standard logging output.
             Http.Request request = Http.Context.current().request();
-            String address = request.remoteAddress();
+            
             String path = request.path();
             String verb = request.method();
             String version = request.version();
-            String[] values = request.headers().get("User-Agent");
-            String userAgent = (values != null && values.length > 0) ? values[0] : "";
+            String address = header(request, "X-Forwarded-For", request.remoteAddress());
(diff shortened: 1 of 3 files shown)


conf/application-logger.xml
```diff
@@ -2,7 +2,7 @@
 <configuration>
   <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
     <encoder>
-      <pattern>%d{ISO8601} %-5p [%t] %logger - %message%n%xException%n%mdc</pattern>
```
alxdarksage, line 5: Add threads back

app/interceptors/LoggingInterceptor.java
```diff
@@ -13,13 +14,23 @@
 
     @Override
     public Object invoke(MethodInvocation method) throws Throwable {
+        Result result = (Result)method.proceed();
         if (logger.isInfoEnabled()) {
+            // Common Log Format for access logging, minus the first part which 
+            // is already part of our standard logging output.
             Http.Request request = Http.Context.current().request();
+            String address = request.remoteAddress();
```
alxdarksage: X-Forwarded-For (instead of remoteAddress)
X-Request-Id (add)

PR #490, opened 2015-02-19
add DeleteRecordsForHealthCode to Health Data API

DwayneJengSage:
Add DeleteRecordsForHealthCode to Health Data API and hook it up to UserManagementService.deleteUser(), to delete health data records when a user is deleted.

This change also includes dummying out the tracker APIs (by removing them from the routing tables, so Play returns 404 when they are called). I did not remove the tracker code, since there's a lot of code, and I want to keep the code review small. (Removing the tracker code can be done at a later date.)

Testing done:
- unit tests
- added 4 health data records for 2 different health codes, then called delete for health code and verified the right records are deleted
- created a user through the admin APIs, added health data records, then deleted the user and verified the user's records are deleted


app/controllers/HealthDataController.java
```diff
@@ -34,6 +34,22 @@ public Result createRecord() {
...
 
     /**
+     * Play controller for DELETE /admin/v2/healthdata/byHealthCode/:healthCode. This API is used by admins to delete
+     * health data records for a given health code (user in study).
+     *
+     * @param healthCode
+     *         health code to delete records for
+     * @return Play result indicating number of records deleted
+     */
+    public Result deleteRecordsForHealthCode(String healthCode) {
```
alxdarksage: Nothing in the API should ever expose the health code such that it could be used to delete records through the API, and the only place we want to do this is when we're deleting a user, so I think we should remove the routes and controller code. Note the user has the alternative to turn sharing of these records off through the api, and that preference is written to the ParticipantOptions table. When exporting, any health codes mapped to a preference not to share, should have their data records excluded from the export. But that's all that is exposed.

PR #492, opened 2015-02-20
Accounts

alxdarksage:
With this PR we use the new Account object and AccountDao that completely remove direct reference to Stormpath outside of these classes. A _lot_ of code has been touched in this refactor, including significant logic around onboarding.

Also changed the signatures of the ConsentService to match every other part of the application where study is passed as the first argument.

AccountEncryptionService was subsumed into the account functionality and/or the HealthCodeService, where appropriate.

Health codes are now only checked for on authentication, and created and saved at that time if necessary (previously had been checking on sign up, sign in and consent). This is the riskiest of the changes made here.


test/org/sagebionetworks/bridge/services/ConsentServiceImplTest.java
```diff
@@ -96,89 +89,81 @@ public void after() {
...
-        study.setIdentifier("api");
-        study.setName("Test Study");
-        study.setMinAgeOfConsent(18);
-
         LocalDate now = LocalDate.now();
         int year = now.getYear();
         int month = now.getMonthOfYear();
         int day = now.getDayOfMonth();
 
+        // The API study is set for 17 years... just go with that as it exists, it's easiest
```
DwayneJengSage, line 142: [MINOR] It looks like from the code change in ConsentBootstrapper and from the DDB tables, it's actually 18 years. Is this comment accurate?

conf/application-context.xml
```diff
@@ -368,41 +358,37 @@
...
+        <property name="accountDao" ref="accountDao" />
         <property name="consentService" ref="consentService" />
         <property name="optionsService" ref="optionsService" />
+        <property name="healthCodeService" ref="healthCodeService" />
         <property name="signInValidator">
             <bean class="org.sagebionetworks.bridge.validators.SignInValidator"/>
         </property>
+        <property name="signUpValidator">
+            <bean class="org.sagebionetworks.bridge.validators.SignUpValidator"/>
+        </property>
```
DwayneJengSage: [MINOR] Is there any value in making this a bean? It's just a bunch of simple null checks and empty string checks, so we're unlikely to ever need a separate implementation or a mock. Why not just create a static instance object and call SignUpValidator.INSTANCE?

app/org/sagebionetworks/bridge/services/HealthCodeServiceImpl.java
```diff
@@ -43,15 +44,43 @@ public String getCode() {
...
+        return new HealthId() {
+            @Override
+            public String getId() {
+                return healthId;
+            }
+            @Override
+            public String getCode() {
+                return healthCode;
+            }
+        };
```
DwayneJengSage: [POLISH] Instead of an anonymous local class, it would be much cleaner to make this a full-fledged class (similar to the StudyIdentifier class).

app/org/sagebionetworks/bridge/services/AuthenticationServiceImpl.java
```diff
@@ -353,11 +191,12 @@ public UserSession getSessionFromAccount(Study study, Account account) {
...
         final String healthCode = getHealthCode(study, account);
-        user.setHealthCode(healthCode);
-
-        user.setSignedMostRecentConsent(consentService.hasUserSignedMostRecentConsent(user, study));
-        user.setConsent(consentService.hasUserConsentedToResearch(user, study));
-        user.setDataSharing(optionsService.getBooleanOption(healthCode, Option.DATA_SHARING));
+        if (healthCode != null) {
+            user.setHealthCode(healthCode);
+            user.setDataSharing(optionsService.getBooleanOption(healthCode, Option.DATA_SHARING));
+        }
```
DwayneJengSage: [POLISH] Under what conditions would this would healthCode be null? getHealthCode() creates the HealthId if it's not present, and HealthCodeService.createMapping() ensures both are created. Plus, the semantics of getSession() is that it will create the health code if not present, and this code looks confusingly like the healthcode might not have been created at this point.

app/org/sagebionetworks/bridge/services/HealthCodeServiceImpl.java
```diff
@@ -43,15 +44,43 @@ public String getCode() {
...
 
     @Override
-    public String getHealthCode(String id) {
-        return healthIdDao.getCode(id);
+    public HealthId getMapping(String healthId) {
+        checkNotNull(healthId);
+        final String healthCode = healthIdDao.getCode(healthId);
+        if (healthCode == null) {
+            return null;
+        }
```
DwayneJengSage: [POLISH] Would it be worthwhile creating a mock test to sanity test that even if healthIdDao returns a null healthcode (which could also happen if we call with an invalid health ID), that we return null (instead of a mapping with health ID and no code)?

app/org/sagebionetworks/bridge/services/HealthCodeServiceImpl.java
```diff
@@ -43,15 +44,43 @@ public String getCode() {
...
+        
+        final String healthId = account.getHealthId();
+        if (healthId == null) {
+            return null;
+        }
+        final String healthCode = healthIdDao.getCode(healthId);
+        if (healthCode == null) {
+            return null;
+        }
+        return getHealthIdObject(healthId, healthCode);
```
DwayneJengSage: [POLISH] We could call getMapping(String healthId) instead of duplicating code.

app/org/sagebionetworks/bridge/services/HealthCodeServiceImpl.java
```diff
@@ -43,15 +44,43 @@ public String getCode() {
...
     @Override
-    public String getStudyIdentifier(String healthCode) {
-        return healthCodeDao.getStudyIdentifier(healthCode);
+    public HealthId getMapping(Account account) {
+        checkNotNull(account);
+        
+        final String healthId = account.getHealthId();
+        if (healthId == null) {
+            return null;
+        }
```
DwayneJengSage, line 41: [POLISH] Is it possible to have an account with no health ID? Should we have a unit test to test this scenario?

app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
```diff
@@ -80,6 +81,13 @@ public void setStudyService(StudyService studyService) {
         this.studyService = studyService;
     }
     
+    public StormpathAccountDao() {
+        BridgeConfig config = BridgeConfigFactory.getConfig();
+        String url = config.getStormpathApplicationHref().trim();
+        String[] parts = url.split("/");
+        applicationId = parts[parts.length-1];
+    }
```
DwayneJengSage: [POLISH] It would be much cleaner and much more testable to have the StormpathAccountDao take in the applicationId (either in the constructor or as a setter) and have the Spring bean configure it. This way, we don't have to hit production configs to run our tests.

conf/application-context.xml
alxdarksage: Sometimes validators need to inject a service. I just removed a case of this that went to Stormpath to check for email duplicates. If we replace Stormpath, we'd need to do something like this again. I can convert these to @Autowired so it's not an annoyance.

app/org/sagebionetworks/bridge/services/AuthenticationServiceImpl.java
alxdarksage: You're right it should be unnecessary. It was there to start with a different form, but I'll remove it and rerun tests.

app/org/sagebionetworks/bridge/services/HealthCodeServiceImpl.java
alxdarksage: It's redundant, I'll remove it in favor of getMapping(healthId) in all cases.

app/org/sagebionetworks/bridge/services/HealthCodeServiceImpl.java
alxdarksage, line 41: Yes, between sign up and sign in. Added a test

app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
alxdarksage: It's possible to remove this entirely, it turns out. Removed in favor of using application.getHref() as the base for making a REST call with no string parsing.

app/org/sagebionetworks/bridge/services/HealthCodeServiceImpl.java
alxdarksage: Yes, added test.

test/org/sagebionetworks/bridge/services/HealthCodeServiceImplTest.java
```diff
@@ -41,19 +45,60 @@ public void after() {
...
+        assertNull(id2);
+        
+    }
+    
+    /*
+   @Override
+    public HealthId getMapping(String healthId) {
+        checkNotNull(healthId);
+        final String healthCode = healthIdDao.getCode(healthId);
+        if (healthCode == null) {
```
DwayneJengSage: [MINOR] Should probably remove this commented out implementation of getMapping, especially since it's already diverged from the actual implementation.

PR #507, opened 2015-02-26
TranscribeConsentHandler and UploadArtifactsHandler

DwayneJengSage:
TranscribeConsentHandler and UploadArtifactsHandler.

This change beefs up some of the health data records stuff, adds health data attachments, transcribes consent information into the health data record (so the exporter can download it all without having to make a bunch of queries to another table), and uploads everything to the right place.

This change also includes a few last minute changes to the data format being parsed by IosSchemaValidationHandler.

Lastly, this change includes autowiring of a few classes. This was because Spring was behaving weirdly on me, and the only way I could get it working again was by adding the Autowired annotation to a few places.

Testing done:
- basic unit tests
- manual tests
  *\* give consent
  *\* create health data
  *\* update health data
  *\* get by ID
  *\* get by upload date
  *\* create attachment
  *\* update attachment
  *\* upload:
  **\* Parkinsons voice test
  **\* BCS Mood Journal
  **\* Asthma survey
  **\* BCS survey
  **\* Parkinsons survey


app/org/sagebionetworks/bridge/dao/HealthDataDao.java
```diff
@@ -29,6 +29,15 @@
     int deleteRecordsForHealthCode(@Nonnull String healthCode);
 
     /**
+     * DAO method used by worker apps to fetch a health data record by the record ID.
+     *
+     * @param id
+     *         record ID
+     * @return health data record
+     */
+    HealthDataRecord getRecordById(@Nonnull String id);
```
eric-wu, line 38: Trivial.  If there are not other getRecord() method, I feel "ById" is kind of redundant.

app/org/sagebionetworks/bridge/models/healthdata/HealthDataAttachment.java
```diff
@@ -0,0 +1,30 @@
...
+@JsonDeserialize(as = DynamoHealthDataAttachment.class)
+public interface HealthDataAttachment {
+    /**
+     * Attachment ID. This is used as both the unique attachment ID and the S3 key in the attachments bucket to
+     * guarantee filename uniqueness in S3.
+     */
+    String getId();
+
+    /** Record ID of the health data record this file is attached to. */
+    String getRecordId();
```
eric-wu, line 22: Are we expecting one record many attachments?  What can be a valid case that we want to store multiple files for one health record.  I am wondering if we could simplify it to a 1:1 relationship.  Then we don't need the index on record id.  I am usually not a big fan of mixing up writes with reads especially for tables receiving batch updates.

app/org/sagebionetworks/bridge/models/healthdata/HealthDataUserConsent.java
```diff
@@ -0,0 +1,85 @@
...
+
+import org.sagebionetworks.bridge.json.DateTimeToStringSerializer;
+import org.sagebionetworks.bridge.json.JodaDateTimeDeserializer;
+
+/**
+ * User consent metadata stored with the health data record. We store this in the Health Data Record table, so we can
+ * avoid doing a bunch of extra calls to other tables during export.
+ */
+@JsonDeserialize(builder = HealthDataUserConsent.Builder.class)
+public class HealthDataUserConsent {
```
eric-wu: I understand the benefits for performance.  I'm not clear about the cost of the maintaining a copy of user consent. For example, if the user turned off sharing, shouldn't this copy be searched and modified accordingly?  Another example is about maintaining the code.  Alx just changed sharing from a boolean to an enum.  This class is an additional place that needs to be changed accordingly.

alxdarksage: ... and Mike just proposed something different in regard to exporting than we discussed yesterday, where it seemed like the settings would be "frozen" at that moment of export for that piece of data. Maybe we should meet with him to clarify this.

app/org/sagebionetworks/bridge/dao/HealthDataDao.java
DwayneJengSage, line 38: There's getRecordsForUploadDate(), so I wanted to contrast with getRecordById(). Plus, the HTTP APIs are /healthdata/byId/:id and /healthdata/byUploadDate/:uploadDate, so the DAO API matches the HTTP API.

app/org/sagebionetworks/bridge/models/healthdata/HealthDataAttachment.java
DwayneJengSage, line 22: The Cardio 6-Minute Walk Test has multiple JSON blobs that we'd want to save as separate attachments. It's not really meaningful to try to combine them into a single attachment. Plus, in the long term, if clients are uploading attachments directly instead of in upload bundles, it makes more sense to store them as separate attachments.

app/org/sagebionetworks/bridge/dao/HealthDataDao.java
alxdarksage, line 38: I don't love these byWhatever path segments. Should those be moved to a query parameter? /healthdata?id=foo or /healthdata?date=bar?

app/org/sagebionetworks/bridge/models/healthdata/HealthDataAttachment.java
eric-wu, line 22: I wonder, if we define one attachment as its own record, would this destroy anything?

app/org/sagebionetworks/bridge/models/healthdata/HealthDataUserConsent.java
DwayneJengSage: Based on the email thread, I _think_ the consensus we reached is that we do want to tag each health data record with sharing prefs at the time of record creation. If the user flips their sharing prefs, we don't retroactively change the sharing prefs on the record. (The two big reasons are (a) it's more user-friendly and more intuitive to _not_ retroactively export data when the user flips from not sharing to sharing and (b) if for any reason we need to re-export, we need a historical record of when the user was and wasn't sharing.)

Re: code maintenance - I agree 100%. Given more time, I would have preferred to use the same class shared across both APIs. Unfortunately, this is a non-trivial refactor that I'm afraid we don't have time for right now.

alxdarksage: Agree to all of this, the only thing I would add is that for the purpose of export, I don't think you need to know anything about the consent except the user's sharing preference. And this record doesn't record the newly added sharing enumeration which specifies no sharing, share to researchers, or share to all researchers (see SharingScope). That's really all you need to know at the time of export, granted this enumeration may change making it more difficult.

I didn't add that preference to the consent record, I added it to the ParticipantOptions table so it would be easier to acquire while working on records (without having to go to Stormpath).

app/org/sagebionetworks/bridge/models/healthdata/HealthDataAttachment.java
DwayneJengSage, line 22: Not sure what you mean here.

Do you mean split the record so that each record has only one attachment? A multi-attachment record represents multiple recordings of a single event. This is probably relevant to the researchers. Breaking the record up into multiple records means we lose that information.

Alternatively, do you mean store attachments in the records table alongside records instead of creating a new table? This doesn't solve the problem of multiple writes, and the health data records table has too heavy of a schema for attachments.

app/org/sagebionetworks/bridge/models/healthdata/HealthDataUserConsent.java
DwayneJengSage: For the purposes of export functionality, that's true. For the purposes of actual research data, Andrew asked me for the consent date (strictly speaking, the join date, which is the same thing). And based on our conversation yesterday, there might be an interesting correlation between data sharing prefs and data. So at this point, I figured I might as well include all consent information.

Re: sharing scope - Yes, I'll need to integrate that. I started the TranscribeConsentHandler before the sharing scope stuff was finalized, and I didn't want to block on that before getting eyeballs on this code review. I'll definitely be sure to pull it in.

alxdarksage: Sorry this is a terrible time to be having this discussion as everything is rushed. I think that request is ridiculous... I don't know what he's going to do with the join date vis-a-vis some research data. What could that possibly signify that would be meaningful? If we had known they wanted this kind of arbitrary relationship to be maintained, we could have created an auditable history of consent records.

app/org/sagebionetworks/bridge/dao/HealthDataDao.java
DwayneJengSage, line 38: Play doesn't route on query params, so I have both now routing to getRecord and I'm using a bunch of if statements to manually route on the query params.

app/org/sagebionetworks/bridge/models/healthdata/HealthDataUserConsent.java
DwayneJengSage: SharingScope stuff is now integrated.

DwayneJengSage: After a brief discussion, we decided that it's not appropriate to put consentedOn in the user data. So I'm ripping all that out and replacing it with only the Sharing Scope (defaults to "no sharing" if not present, of course).

PR #509, opened 2015-02-26
Change bad requests to quiet exceptions

alxdarksage:
Clients sending bad requests should not send ERROR notifications, these are now logged like other quiet exceptions;

The quiet exception logging now includes the full stack trace. I think if we need that, we'll be glad we have it.


app/interceptors/ExceptionInterceptor.java
```diff
@@ -62,16 +64,7 @@ public Object invoke(MethodInvocation method) throws Throwable {
             }
             
             if (!config.isLocal() && quietExceptions.contains(throwable.getClass())) {
-                // quiet exceptions, just a one liner at the debug level. May suppress eventually.
-                if (throwable instanceof InvalidEntityException) {
-                    // For the near future, log this in the event it represents an error
-                    // on our side.
-                    Http.Request request = Http.Context.current().request();
-                    String bodyContent = request.body().asText();
-                    logger.debug(throwable.getMessage() + ": " + bodyContent);
```
DwayneJengSage, line 71: I remember this being a point of contention back in November, so you'll prolly want to sync up with Eric to see if he's okay with this, or if the stack trace gives enough information.

On the other hand, this might be sensitive health data, so we prolly don't want to put it in the logs anyway.

PR #513, opened 2015-03-02
Autowiring

alxdarksage:


conf/application-context.xml
```diff
@@ -2,169 +2,34 @@
...
         http://www.springframework.org/schema/beans/spring-beans-3.0.xsd
         http://www.springframework.org/schema/context 
         http://www.springframework.org/schema/context/spring-context-3.0.xsd
         http://www.springframework.org/schema/mvc
-        http://www.springframework.org/schema/mvc/spring-mvc.xsd">
-
-    <context:annotation-config />
+        http://www.springframework.org/schema/mvc/spring-mvc.xsd
+        http://www.springframework.org/schema/util 
+        http://www.springframework.org/schema/util/spring-util-2.5.xsd">
```
DwayneJengSage: We should use spring-util-3.0. Really weird things happen if we have mismatched spring versions.

app/org/sagebionetworks/bridge/services/SendMailViaAmazonService.java
```diff
@@ -67,18 +71,18 @@
     private static final String NEWLINE = "\n";
     private static final Region region = Region.getRegion(Regions.US_EAST_1);
 
-    private String supportEmail;
+    private String supportEmail = "support@sagebridge.org";
```
DwayneJengSage: We should move this value to Spring, so we can still change it in config instead of code if we need it.
<bean id="numAsyncWorkerThreads" class="java.lang.String">
  <constructor-arg value="support@sagebridge.org" />
</bean>

Also, the previous value was "Bridge (Sage Bionetworks) &lt;support@sagebridge.org&gt;". Was this change intentional?

app/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -41,19 +47,24 @@
...
         this.uploadCredentailsService = uploadCredentialsService;
     }
+    @Autowired
     public void setS3UploadClient(AmazonS3 s3UploadClient) {
         this.s3UploadClient = s3UploadClient;
     }
+    @Autowired
     public void setS3Client(AmazonS3 s3Client) {
         this.s3Client = s3Client;
     }
```
DwayneJengSage, line 62: We have multiple S3 clients (and in this case, 2 S3 clients in the same class). You'll probably want to use @Resource(name = "s3Client") and @Resource(name = "s3UploadClient") to disambiguate.

app/org/sagebionetworks/bridge/services/SendMailViaAmazonService.java
alxdarksage: Spring won't mix Java-based configuration and XML-based configuration, so
moving that value out as a constant to the application-context.xml file
doesn't work. So this ended up in the code for the moment.

app/org/sagebionetworks/bridge/services/UserProfileServiceImpl.java
```diff
@@ -22,14 +25,17 @@
...
 
+    @Autowired
     public void setAccountDao(AccountDao accountDao) {
         this.accountDao = accountDao;
     }
     
+    @Autowired
     public void setExecutorService(ExecutorService executorService) {
         this.executorService = executorService;
     }
```
DwayneJengSage, line 38: We don't currently have more than one ExecutorService, but we might. Consider changing this to
@Resource(name = "asyncExecutorService")

alxdarksage, line 38: This raises the more general point: @Resource(name="foo") is always less
ambiguous than @Autowired, should we switch to naming all components and
using annotations that take component names instead of @Autowired?

app/org/sagebionetworks/bridge/services/SendMailViaAmazonService.java
DwayneJengSage: This should work. The string is just a bean. It looks like GitHub stripped out my XML example, but see numAsyncWorkerThreads in application-context.xml for an example.

app/org/sagebionetworks/bridge/services/UserProfileServiceImpl.java
DwayneJengSage, line 38: I think it depends. For singleton classes that we define (services, controllers, DAOs), @Autowired is almost never ambiguous. For non-singleton classes (S3 clients), @Autowired is almost always ambiguous.

The ExecutorService is one of those things where right now we only have one, but I feel like it's singleton by coincidence, not by design. There's nothing stopping us from having different executors for different worker pools, so I think we should disambiguate on this.

One advantage of using @Autowired for singleton classes is if we change the implementation (fictitious example: DynamoFooDao to StormpathFooDao), @Autowired will pick that up automatically. That said, you can achieve the same effect by having a bean name "fooDao" and referencing using @Resource. (You'd just need to type a bit more.)

app/org/sagebionetworks/bridge/services/SendMailViaAmazonService.java
alxdarksage: Okay, I see what you're saying, I can make that an entirely separate bean.

conf/application-context.xml
alxdarksage: The documentation suggests removing the versions (it looks for the version shipped with the jars I believe). All versions removed.

PR #514, opened 2015-03-03
Request reset password call requires a study to scope account being reset

alxdarksage:


alxdarksage:
Ack... I did not mean to merge that!

PR #522, opened 2015-03-05
Stormpath migration

alxdarksage:
DON'T MERGE YET.
- code to switch to the enterprise version of storm path
- two consents updated.


eric-wu:
Looks fine to me.  Just keep in mind that, if we don't the follow the "standard" migration protocol and shadow the old with the new, there will be no easy rollback.  I guess it's still acceptable before launch.

alxdarksage:
I  can copy the existing HREFs from the old system into that class so I can
flip the logic and have it save back to the old URLs on a redeployment.

Alx

PR #548, opened 2015-03-10
Updated consent form with new approval date

alxdarksage:


eric-wu:
Do we want release all in develop to production?

alxdarksage:
I wouldn't right now, Stormpath is having issues and I don't think we
should make it worse by deploying to production at the same time.

Alx

PR #553, opened 2015-03-10
Better Stormpath error-handling, including intelligible password message...

alxdarksage:
...s for the user and a 400 response.

It turns out that getDeveloperMessage() provides a little bit cleaner message from storm path on what has gone wrong when we call their API (there are some things, like rules for passwords, where the code doesn't tell us exactly what has happened, so the best thing to do is to send the message along to the client).

### DIFF
--- a/app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
+++ b/app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
@@ -312,7 +312,7 @@ public class StormpathAccountDao implements AccountDao {
     }
     
     private void rethrowResourceException(ResourceException e, Account account) {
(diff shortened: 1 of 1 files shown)


app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
```diff
@@ -314,21 +312,21 @@ public void deleteAccount(Study study, String email) {
     }
     
     private void rethrowResourceException(ResourceException e, Account account) {
+        logger.info(String.format("Stormpath error: %s, exception: %s, mapped to BadRequestException", e.getCode(), e.getMessage()));
```
eric-wu: The exception can be mapped to other exceptions, not just BadRequestException, right?

PR #554, opened 2015-03-11
logging updates

DwayneJengSage:
Changes include:
- ExceptionInterceptor quietExceptions uses ImmutableSet instead of a static instantiator. This is cleaner.
- Created NotFoundException to represent 404s not tied to BridgeEntities and added it to quietExceptions.
- Removed Throwables.getRootCause() from ExceptionInterceptor. The exception we're interested in is always the outermost exception, not the innermost exception.
- Updated DynamoUploadDao to use NotFoundException. This catches the /api/v1/upload/(null)/complete issue.
- Fixed warning format for calling upload complete for uploads that are already complete.
- Added try-catch to catch S3 exceptions in upload complete and wrap them in a NotFoundException.
- Removed generally non-useful warnings from IosSchemaValidationHandler. (The exception logged by UploadValidationTask is still useful, though.)

Testing done:
- /api/v1/upload/(null)/complete
- upload complete without uploading file to S3
- upload bad file, ensure validation messages aren't logged as warnings
- upload complete again, and verify warning format

### DIFF
--- a/app/interceptors/ExceptionInterceptor.java
+++ b/app/interceptors/ExceptionInterceptor.java
@@ -2,6 +2,7 @@ package interceptors;
 
 import java.util.Set;
 
+import com.google.common.collect.ImmutableSet;
 import models.ExceptionMessage;
 
 import org.aopalliance.intercept.MethodInterceptor;
@@ -16,6 +17,7 @@ import org.sagebionetworks.bridge.exceptions.EntityAlreadyExistsException;
 import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
 import org.sagebionetworks.bridge.exceptions.InvalidEntityException;
 import org.sagebionetworks.bridge.exceptions.NotAuthenticatedException;
+import org.sagebionetworks.bridge.exceptions.NotFoundException;
 import org.sagebionetworks.bridge.exceptions.UnauthorizedException;
 import org.sagebionetworks.bridge.models.UserSessionInfo;
 import org.slf4j.Logger;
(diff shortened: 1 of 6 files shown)


app/interceptors/ExceptionInterceptor.java
```diff
@@ -57,8 +53,6 @@ public Object invoke(MethodInvocation method) throws Throwable {
         try {
             return method.proceed();
         } catch(Throwable throwable) {
-            throwable = Throwables.getRootCause(throwable);
```
eric-wu, line 60: The current behavior looks at the root cause to decide whether to log it or not, which I think is wrong.  So this change makes total sense by fixing the bug.  In the meantime, logger.debug(message, throwable) downstream logs the entire stack trace to the root cause, which is the desired behavior.  When we look at the logs for errors we do want to know the root cause.  Overall this is a nice fix.

app/interceptors/ExceptionInterceptor.java
```diff
@@ -28,23 +30,17 @@
...
-        quietExceptions.add(BadRequestException.class);
-        quietExceptions.add(InvalidEntityException.class);
-        quietExceptions.add(EntityAlreadyExistsException.class);
-        quietExceptions.add(EntityNotFoundException.class);
-        quietExceptions.add(UnauthorizedException.class);
-        quietExceptions.add(NotAuthenticatedException.class);
-    }
-    
+
+    private static Set<Class<? extends BridgeServiceException>> quietExceptions = ImmutableSet.of(
```
eric-wu, line 39: Not related to this change.  But can we create a marker Interface QuietException?  If I write a new exception not to be logged, I just implement QuietException.  I don't need to update anything in ExceptionInterceptor.

DwayneJengSage, line 39: Created JIRA issue https://sagebionetworks.jira.com/browse/BRIDGE-426

PR #558, opened 2015-03-11
first refactor to submit study as part of JSON payload

alxdarksage:


app/controllers/AuthenticationController.java
```diff
@@ -104,13 +118,28 @@ public Result resetPassword() throws Exception {
      * @param email
      * @return
      */
-    private Study getStudyOrThrowException(Email email) {
-        if (email.getStudyIdentifier() != null) {
-            return studyService.getStudy(email.getStudyIdentifier());
+    @SuppressWarnings("deprecation")
+    private Study getStudyOrThrowException(JsonNode node) {
```
DwayneJengSage, line 121: This and getStudyIdentifierOrThrowException() have some common code. Would it be better to refactor this to a getStudyIdString() with the fall back logic and have both getStudy() and getStudyId() call it?

app/org/sagebionetworks/bridge/dynamodb/DynamoParticipantOptionsDao.java
```diff
@@ -118,8 +116,6 @@ public OptionLookup getOptionForAllStudyParticipants(StudyIdentifier studyIdenti
         scan.addFilterCondition("studyKey", condition);
         
         List<DynamoParticipantOptions> mappings = mapper.scan(DynamoParticipantOptions.class, scan);
-        List<FailedBatch> failures = mapper.batchDelete(mappings);
-        BridgeUtils.ifFailuresThrowException(failures);
```
DwayneJengSage: Should we add a mock unit test to verify the mapper isn't being used for anything other than scan?

alxdarksage: We should do something because this was a scary mistake had we used it on production. A mock that verifies no other interactions would be good. Should do it elsewhere too.

alxdarksage: BTW this should already have been removed in a separate PR I had Eric review and submit immediately, because it was destructive.

PR #561, opened 2015-03-13
Remove the questions property from surveys

alxdarksage:
It's deprecated and nearly identical to the elements property. Since Ymedia is currently using no surveys from the server, this can finally be removed without breaking the applications (they should have already moved to the elements property, but we couldn't completely confirm this).

### DIFF
--- a/test/org/sagebionetworks/bridge/dynamodb/DynamoSurveyTest.java
+++ b/test/org/sagebionetworks/bridge/dynamodb/DynamoSurveyTest.java
@@ -1,6 +1,7 @@
 package org.sagebionetworks.bridge.dynamodb;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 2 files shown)


test/org/sagebionetworks/bridge/dynamodb/DynamoSurveyTest.java
```diff
@@ -121,4 +124,15 @@ public void serializationTestWithInfoScreen() throws Exception {
...
     }
+    
+    // This was a legacy property that we want removed. It cuts the size of surveys by half 
+    // (before compression)
+    @Test
+    public void surveyJsonDoesNotIncludeQuestionsProperty() throws Exception {
+        ObjectMapper mapper = BridgeObjectMapper.get();
+        String json = mapper.writeValueAsString(survey);
+        
+        System.out.println(json);
```
alxdarksage, line 135: Remove

PR #573, opened 2015-03-17
Changing CORS to aid development of web-based clients

alxdarksage:
Switching to use wildcard with host

The headers header doesn't accept a wildcard, SFAIK, so added the specific header that comes through from client that's not whitelisted.


app/controllers/ApplicationController.java
```diff
@@ -28,9 +28,9 @@ public Result resetPassword() {
     }
 
     public Result preflight(String all) {
-        response().setHeader(ACCESS_CONTROL_ALLOW_ORIGIN, "https://" + ASSETS_HOST);
+        response().setHeader(ACCESS_CONTROL_ALLOW_ORIGIN, "*");
         response().setHeader(ACCESS_CONTROL_ALLOW_METHODS, "HEAD, GET, OPTIONS, POST, PUT, DELETE");
-        response().setHeader(ACCESS_CONTROL_ALLOW_HEADERS, "*");
+        response().setHeader(ACCESS_CONTROL_ALLOW_HEADERS, "Content-Type");
```
eric-wu: Why do we want to limit to only "Content-Type"?

PR #577, opened 2015-03-20
Log metrics

eric-wu:
Log structured metrics that can also serve as audit logs, access logs.  Tracing is achieved via request ID, session ID, and user ID.  Here is an example,

```
2015-03-20 06:37:31,254 INFO  [play-akka.actor.default-dispatcher-5] interceptors.MetricsInterceptor - 
{
    "version":1,
    "start":"2015-03-20T06:37:31.245Z",
    "request_id":"2",
    "method":"GET",
    "uri":"/api/v1/auth/signOut",
    "protocol":"HTTP/1.1",
    "remote_address":"127.0.0.1",
    "user_agent":"NING/1.0",
    "session_id":"f3164e72-b022-42cb-a2bb-e3e7af01c91e",
    "user_id":"4AKWCDd4U1Zbvb3lhzYikL",
    "study":"api",
    "status":200,
    "end":"2015-03-20T06:37:31.254Z"
}
```


app/org/sagebionetworks/bridge/BridgeConstants.java
```diff
@@ -5,15 +5,21 @@
...
-    
+
     public static final String BRIDGE_HOST_HEADER = "Bridge-Host";
-    
+
+    public static final String X_REQUEST_ID_HEADER = "X-Request-Id";
+
+    public static final String X_FORWARDED_FOR_HEADER = "X-Forwarded-For";
+
+    public static final String USER_AGENT_HEADER = "User-Agent";
```
eric-wu: Remove. Use the Apache standard header.

app/controllers/BaseController.java
```diff
@@ -295,7 +299,17 @@ protected JsonNode requestToJSON(Request request) {
...
         throw new InvalidEntityException("Expected JSON in the request body is missing");
     }
+
+    Metrics getMetrics() {
+        final String requestId = RequestUtils.getRequestId(request());
+        final String cacheKey = Metrics.getCacheKey(requestId);
+        Metrics metrics = (Metrics)Cache.get(cacheKey);
+        if (metrics == null) {
+            metrics = new Metrics(requestId);
+            Cache.set(cacheKey, metrics, METRICS_EXPIRE_SECONDS);
```
alxdarksage: Wouldn't it be cleaner to do all the metrics work in the interceptor? It could access the user's session using the session as set in the returned header for user-specific stuff like study or userId. It would be nice if this kind of logging were completely orthogonal to the application logic.

I think this would also eliminate the need to cache the metrics object (though that seems to just be in memory, it's a little like a thread local variable, which we were avoiding).

app/interceptors/MetricsInterceptor.java
```diff
@@ -0,0 +1,50 @@
...
+        final Metrics metrics = initMetrics();
+        Cache.set(metrics.getCacheKey(), metrics, METRICS_EXPIRE_SECONDS);
+        try {
+            final Result result = (Result)method.proceed();
+            metrics.setStatus(result.toScala().header().status());
+            return result;
+        } finally {
+            Cache.remove(metrics.getCacheKey());
+            metrics.end();
+            logger.info(metrics.toJsonString());
```
alxdarksage, line 36: Is there any other structured format that would be easier to search? I think JSON is okay but it might be messy to read in the logs and it's a little tedious to search for stuffy like "requestId":"foo" if requestId=foo would work. We just need it to be parseable by an external log processor.

app/models/Metrics.java
```diff
@@ -0,0 +1,122 @@
...
+        json.put("version", VERSION);
+        start();
+        setRequestId(requestId);
+    }
+
+    public String getCacheKey() {
+        return Metrics.getCacheKey(json.get("request_id").asText());
+    }
+
+    public String toJsonString() {
```
alxdarksage, line 47: json.toString() works and I don't think it throws an exception.

app/models/Metrics.java
```diff
@@ -0,0 +1,122 @@
...
+            return MAPPER.writeValueAsString(json);
+        } catch (JsonProcessingException e) {
+            final String msg = "Failed to write metrics.";
+            logger.error(msg, e);
+            return msg;
+        }
+    }
+
+    public void start() {
+        json.put("start", DateTime.now(DateTimeZone.UTC).toString(ISODateTimeFormat.dateTime()));
```
alxdarksage: DateUtils.getCurrentISODateTime()

app/models/Metrics.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    public void end() {
+        json.put("end", DateTime.now(DateTimeZone.UTC).toString(ISODateTimeFormat.dateTime()));
```
alxdarksage: DateUtils.getCurrentISODateTime()

app/org/sagebionetworks/bridge/models/UserSession.java
```diff
@@ -20,6 +21,12 @@ public String getSessionToken() {
     public void setSessionToken(String sessionToken) {
         this.sessionToken = sessionToken;
     }
+    public String getSessionId() {
```
alxdarksage: This is exactly the same as sessionToken, is it not? Both are session-scoped GUIDs. Why two?

app/org/sagebionetworks/bridge/services/AuthenticationServiceImpl.java
```diff
@@ -198,6 +198,7 @@ private UserSession getSessionFromAccount(Study study, Account account) {
         session.setAuthenticated(true);
         session.setEnvironment(config.getEnvironment().name().toLowerCase());
         session.setSessionToken(BridgeUtils.generateGuid());
+        session.setSessionId(BridgeUtils.generateGuid());
         session.setStudyIdentifier(study.getStudyIdentifier());
```
alxdarksage, line 203: See comment above. Is this for anonymity? If so could we call it internalSessionToken or something to queue later developers that it's not to be exposed to the response? Actually I'm not sure if that's the use case, that this couldn't be done as part of the metrics, and not be in the session object at all (i.e. a metricsGuid ... and I would still consider metricsId to be an integer key, not a GUID).

app/controllers/BaseController.java
```diff
@@ -67,31 +69,37 @@ public void setAuthenticationService(AuthenticationService authenticationService
...
     }
-    
+
     /**
      * Returns a session. Will not throw exception if user is not authorized or has not consented to research.
      * @return session if it exists, or null otherwise.
      */
-    protected UserSession getSessionIfItExists() {
+    UserSession getSessionIfItExists() {
         String sessionToken = getSessionToken();
```
alxdarksage, line 83: Just from looking at the code I wasn't sure why you were increasing the visibility of these to package scope. Needed?

app/models/RequestUtils.java
```diff
@@ -0,0 +1,20 @@
...
+    public static String getRequestId(final Request request) {
+        return header(request, X_REQUEST_ID_HEADER,
+                Integer.toHexString(System.identityHashCode(request)));
+    }
+
+    public static String header(final Request request, final String name, final String defaultVal) {
+        final String[] values = request.headers().get(name);
+        return (values != null && values.length > 0) ? values[0] : defaultVal;
+    }
+}
```
alxdarksage, line 20: If this work were all done in the interceptor, there would be no need for these separate static utils.

test/models/RequestUtilsTest.java
```diff
@@ -0,0 +1,37 @@
...
+        when(mockRequest.headers()).thenReturn(headerMap);
+        assertEquals("123", RequestUtils.getRequestId(mockRequest));
+        assertEquals("ifeng 6", RequestUtils.header(mockRequest, "User-Agent", null));
+        headerMap.remove("X-Request-Id");
+        assertFalse("123".equals(RequestUtils.getRequestId(mockRequest)));
+        assertNotNull(RequestUtils.getRequestId(mockRequest));
+        headerMap.remove("User-Agent");
+        assertNull(RequestUtils.header(mockRequest, "User-Agent", null));
+    }
+}
```
alxdarksage, line 37: Any way to test at the request level that the right stuff is output? It seems like the integration of these parts isn't verified... I don't know if we can do this without starting Play's integration system. For some things, like verifying that a cookie is set in the response, this seems to be necessary.

app/controllers/BaseController.java
```diff
@@ -295,7 +299,17 @@ protected JsonNode requestToJSON(Request request) {
...
         }
-
         throw new InvalidEntityException("Expected JSON in the request body is missing");
     }
+
+    Metrics getMetrics() {
+        final String requestId = RequestUtils.getRequestId(request());
+        final String cacheKey = Metrics.getCacheKey(requestId);
+        Metrics metrics = (Metrics)Cache.get(cacheKey);
+        if (metrics == null) {
```
eric-wu: Remove this block

app/org/sagebionetworks/bridge/models/UserSession.java
eric-wu: Rename to InternalSessionId or MetricsToken.

app/controllers/BaseController.java
eric-wu, line 83: We are decreasing the visibility.

PR #586, opened 2015-03-23
Refactoring the email sender into some components

alxdarksage:
The SendEmailService implementation was factored out so the service just sends email, and takes a parameter object, MimeTypeEmail, that has all the content for the email.

There's a new package with two MimeTypeEmailProvider implementations that create the consent email, and the roster email, respectively. That package also includes a builder for the email object.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/email/ConsentEmailProvider.java
+++ b/app/org/sagebionetworks/bridge/services/email/ConsentEmailProvider.java
@@ -58,14 +58,14 @@ public class ConsentEmailProvider implements MimeTypeEmailProvider {
     }
 
     @Override
-    public MimeTypeEmail getEmail(String defaultFromEmail) throws MessagingException {
+    public MimeTypeEmail getEmail(String defaultSender) throws MessagingException {
         MimeTypeEmailBuilder builder = new MimeTypeEmailBuilder();
         
         String subject = String.format(CONSENT_EMAIL_SUBJECT, study.getName());
         builder.withSubject(subject);
         
         final String sendFromEmail = isNotBlank(study.getSupportEmail()) ?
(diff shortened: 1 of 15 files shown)


app/org/sagebionetworks/bridge/services/email/ConsentEmailProvider.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+        builder.withMessageParts(sigPart);
+
+        return builder.build();
+    }
+
+    private String createSignedDocument() {
+        final String filePath = studyConsent.getPath();
+        final FileSystemResource resource = new FileSystemResource(filePath);
+        try (InputStreamReader isr = new InputStreamReader(resource.getInputStream(), StandardCharsets.UTF_8);) {
```
eric-wu: This is not your change.  But I think it is safer to declare the input stream of resource.getInputStream() separately so that it is guaranteed its close() method will be called at the end of the try block.

test/org/sagebionetworks/bridge/services/email/ConsentEmailProviderTest.java
```diff
@@ -0,0 +1,5 @@
+package org.sagebionetworks.bridge.services.email;
+
+public class ConsentEmailProviderTest {
```
eric-wu: Empty?

app/org/sagebionetworks/bridge/services/email/ConsentEmailProvider.java
alxdarksage: Like this?
Like this?

try (InputStream is = resource.getInputStream();
      InputStreamReader isr = new InputStreamReader(is, StandardCharsets.UTF_8);) {
            
} catch(IOException e) {
}

PR #588, opened 2015-03-24
Adding sharing option to output, removed inline text

alxdarksage:
The inline text was going to get very long, so I removed it. There is still the TSV file of course.

Sharing added. However to get this information you have to make N requests to DynamoDB, one for every user. This isn't good (although I think it's the only case where this happens). We should discuss alternatives, though I think all of them require altering the ParticipantOptions table.


app/org/sagebionetworks/bridge/services/email/ParticipantRosterProvider.java
```diff
@@ -65,29 +63,11 @@ String createInlineParticipantRoster() {
...
             sb.append("There are no users enrolled in this study.");
         } else if (participants.size() == 1) {
-            sb.append("There is 1 user enrolled in this study:");
+            sb.append("There is 1 user enrolled in this study. Please see the attached TSV file.");
         } else {
-            sb.append("There are "+participants.size()+" users enrolled in this study:");
+            sb.append("There are "+participants.size()+" users enrolled in this study. Please see the attached TSV file.");
         }
         sb.append(NEWLINE);
-        for (int i=0; i < participants.size(); i++) {
```
eric-wu: I see the for loop is being removed.  Does it mean we are not sending detailed information about individual participants?

PR #591, opened 2015-03-26
External alternative ID

alxdarksage:


app/controllers/UserProfileController.java
```diff
@@ -19,15 +24,21 @@
 @Controller("userProfileController")
 public class UserProfileController extends BaseController {
     
-    private UserProfileService userProfileService;
+    protected UserProfileService userProfileService;
```
DwayneJengSage: Instead of making these protected, it would be much cleaner to create a mock userProfileService and call the setter (which you already do), and access the the mock through the variables defined in the test.

app/controllers/UserProfileController.java
```diff
@@ -59,4 +70,20 @@ public Result updateUserProfile() throws Exception {
         
         return okResult("Profile updated.");
     }
+    
+    public Result createExternalIdentifier() throws Exception {
+        UserSession session = getAuthenticatedSession();
+
+        ExternalIdentifier externalId = ExternalIdentifier.fromJson(requestToJSON(request()));
```
DwayneJengSage: You can use BaseController.parseJson(request, clazz). parseJson wraps both the requestToJSON() call and the Jackson call, so you can leverage Jackson directly instead of building a custom fromJson() method.

app/controllers/UserProfileController.java
```diff
@@ -59,4 +70,20 @@ public Result updateUserProfile() throws Exception {
...
+        
+        // TODO: An annotation-based validator would make these trivial validations 
+        // easier to factor out.
+        if (isBlank(externalId.getIdentifier())) {
```
DwayneJengSage, line 81: Can we push this check to the Service? I know it's possible to test this through the Controller (because that's what you did), but I feel it's cleaner, more testable, and more consistent to have validation in the Service.

app/org/sagebionetworks/bridge/StudyConsentBootstrapper.java
```diff
@@ -37,6 +37,7 @@ public StudyConsentBootstrapper(StudyService studyService, StudyConsentDao study
             study.setMinAgeOfConsent(18);
             study.setResearcherRole("api_researcher");
             study.setConsentNotificationEmail("bridge-testing+consent@sagebridge.org");
+            study.setSupportEmail("bridge-testing+support@sagebase.org");
```
DwayneJengSage: Has this email been verified? I'm not seeing it in the SES console, nor did I see any verification emails in bridge-testing@

app/controllers/UserProfileController.java
alxdarksage, line 81: I agree with this but there's no business-specific service involved here. It's calling directly to the options service which throws an IllegalArgumentException if the identifier isn't submitted, not the invalid entity exception you should get if your JSON is wrong.

app/org/sagebionetworks/bridge/StudyConsentBootstrapper.java
alxdarksage: It is referenced in one of Eric's tests, but hadn't been set in my local environment, so I added this to get his test to pass. I wonder if bridge-testing@sagebase.org is sufficient to make these bridge-testing emails work?

DwayneJengSage: It looks like the emails are still coming from support@sagebridge.org. So either Eric's tests are using a cached version of the study (either cached in Redis or DDB still has the old email), or Eric's tests don't go through SES at all.

test/org/sagebionetworks/bridge/services/SendEmailIntegTest.java
```diff
@@ -38,7 +38,7 @@ public void test() {
         final ConsentSignature signature = ConsentSignature.create("Eggplant McTester", "1970-05-01", IMG, "image/png");
         final User user = new User();
         final Study study = studyService.getStudy(TestConstants.TEST_STUDY_IDENTIFIER);
-        user.setEmail(study.getSupportEmail());
+        user.setEmail("eggplant.mctester@tester.com");
```
DwayneJengSage: This won't work either. It looks like this integration test sends email for real, and eggplant.mctester@tester.com doesn't exist. (As evidenced by the bounce notification I just received.)

You should change this to bridge-testing@sagebase.org (with or without a plus), which was created specifically for this purpose.

PR #600, opened 2015-03-27
Cleanup

alxdarksage:
- remove hostname from study
- move phone to be an extended attribute
- removed some fromJson() methods that were unnecessary
- some test refactoring for maintainability

### DIFF
--- a/app/org/sagebionetworks/bridge/StudyConsentBootstrapper.java
+++ b/app/org/sagebionetworks/bridge/StudyConsentBootstrapper.java
@@ -33,7 +33,6 @@ public class StudyConsentBootstrapper {
             Study study = new DynamoStudy();
             study.setName("Test Study");
             study.setIdentifier("api");
-            study.setHostname("api-local.sagebridge.org");
             study.setMinAgeOfConsent(18);
             study.setResearcherRole("api_researcher");
             study.setConsentNotificationEmail("bridge-testing+consent@sagebridge.org");
--- a/app/org/sagebionetworks/bridge/StudyConsentBootstrapper.java
+++ b/app/org/sagebionetworks/bridge/StudyConsentBootstrapper.java
@@ -22,7 +22,8 @@ public class StudyConsentBootstrapper {
         try {
             Study study = studyService.getStudy("api");
             try {
-                if (study.getUserProfileAttributes().isEmpty()) {
+                if (study.getUserProfileAttributes().size() < 2) {
+                    study.getUserProfileAttributes().add("phone");
                     study.getUserProfileAttributes().add("can_be_recontacted");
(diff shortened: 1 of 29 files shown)


app/org/sagebionetworks/bridge/StudyConsentBootstrapper.java
```diff
@@ -22,7 +22,8 @@ public StudyConsentBootstrapper(StudyService studyService, StudyConsentDao study
         try {
             Study study = studyService.getStudy("api");
             try {
-                if (study.getUserProfileAttributes().isEmpty()) {
+                if (study.getUserProfileAttributes().size() < 2) {
```
eric-wu: Could it be more robust if we check against the expected number of attributes. If not the same as expected, we clear and add all of them.

app/org/sagebionetworks/bridge/models/Email.java
```diff
@@ -1,31 +1,25 @@
...
-    
-    public static final Email fromJson(JsonNode node) {
-        String email = JsonUtils.asText(node, EMAIL_FIELD);
-        String study = JsonUtils.asText(node, STUDY_FIELD);
-        StudyIdentifier studyIdentifier = (study == null) ? null : new StudyIdentifierImpl(study);
-        return new Email(studyIdentifier, email);
-    }
 
+    public Email(StudyIdentifier studyId, String email) {
+        this.studyIdentifier = studyId;
```
eric-wu, line 20: These two fields being null can be confusing. Do we have valid cases where they will be null? Otherwise, we should check null and throw exceptions.

app/org/sagebionetworks/bridge/models/EmailVerification.java
```diff
@@ -1,24 +1,15 @@
...
 
 public class EmailVerification {
 
-    private static final String SPTOKEN_FIELD = "sptoken";
-
     private final String sptoken;
 
-    public EmailVerification(String sptoken) {
+    public EmailVerification(@JsonProperty("sptoken") String sptoken) {
         this.sptoken = sptoken;
```
eric-wu, line 10: The same checkNull() on sptoken.

app/org/sagebionetworks/bridge/models/Email.java
alxdarksage, line 20: Email has an EmailValidator while EmailVerification does not. So I should add that. For these trivial classes it would be nicest to have a validator that used annotations and that was reusable for all these classes.

app/org/sagebionetworks/bridge/models/EmailVerification.java
alxdarksage, line 10: I can add these as well.

PR #610, opened 2015-04-07
Scheduling

alxdarksage:
- Update the schedule object to contain new fields: interval and delay.
- Add a task scheduler that can schedule either using cron expressions or interval/delay


app/org/sagebionetworks/bridge/json/JsonUtils.java
```diff
@@ -96,13 +98,27 @@ public static Long asMillisDurationLong(JsonNode parent, String property) {
         }
         return null;
     }
-
+    
+    public static Period asPeriod(JsonNode parent, String property) {
```
DwayneJengSage, line 95: Can we get some basic unit tests on this?

app/org/sagebionetworks/bridge/json/JsonUtils.java
```diff
@@ -96,13 +98,27 @@ public static Long asMillisDurationLong(JsonNode parent, String property) {
...
+        return null;
+    }
+    
     public static long asMillisSinceEpoch(JsonNode parent, String property) {
         if (parent != null && parent.hasNonNull(property)) {
             return DateUtils.convertToMillisFromEpoch(parent.get(property).asText());
         }
         return 0L;
     }
+
```
DwayneJengSage: And this?

app/org/sagebionetworks/bridge/json/JsonUtils.java
```diff
@@ -96,13 +98,27 @@ public static Long asMillisDurationLong(JsonNode parent, String property) {
...
+    }
+    
     public static long asMillisSinceEpoch(JsonNode parent, String property) {
         if (parent != null && parent.hasNonNull(property)) {
             return DateUtils.convertToMillisFromEpoch(parent.get(property).asText());
         }
         return 0L;
     }
+    
+    public static DateTime asDateTime(JsonNode parent, String property) {
```
DwayneJengSage, line 109: And this?

app/org/sagebionetworks/bridge/json/ScheduleDeserializer.java
```diff
@@ -20,7 +23,7 @@
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.JsonDeserializer;
 import com.fasterxml.jackson.databind.JsonNode;
-import com.google.common.collect.Lists;
+import com.fasterxml.jackson.databind.node.ArrayNode;
 
 public class ScheduleDeserializer extends JsonDeserializer<Schedule> {
```
DwayneJengSage: Can we leverage Jackson directly instead of writing a custom serializer and deserializer?

test/org/sagebionetworks/bridge/models/schedules/ScheduleTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class ScheduleTest {
+
+    @Test
+    public void equalsHashCode() {
+        EqualsVerifier.forClass(Schedule.class).suppress(Warning.NONFINAL_FIELDS).allFieldsShouldBeUsed().verify();
+    }
+    
+    @Test
+    public void canRountripSerialize() throws Exception {
```
DwayneJengSage, line 24: This tests verifies that we can round trip back to a POJO, but it doesn't validate JSON. Can we add a check to validate the JSON?

alxdarksage, line 24: It's not possible for the JSON to be correct if the objects are equal after deserialization. You mean to lock down the structure of the JSON?

app/org/sagebionetworks/bridge/json/ScheduleDeserializer.java
alxdarksage: I don't think so because of the times array, but I can look for a way to annotate it.

DwayneJengSage: I think you should be able to use a serializer for just the local time list without having to create a serializer for the whole class.

test/org/sagebionetworks/bridge/models/schedules/ScheduleTest.java
DwayneJengSage, line 24: Given that the JSON is part of our external interface, I feel we should define the JSON structure and test against it.

app/org/sagebionetworks/bridge/json/ScheduleDeserializer.java
DwayneJengSage: So I know Jackson can handle lists of things without a problem:
https://github.com/Sage-Bionetworks/BridgePF/blob/develop/app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchema.java#L41

So at most we need custom serializers for LocalTime, which should be easy to hook up in our custom module:
https://github.com/Sage-Bionetworks/BridgePF/blob/develop/app/org/sagebionetworks/bridge/json/BridgeObjectMapper.java#L57

app/org/sagebionetworks/bridge/json/ScheduleDeserializer.java
```diff
@@ -33,19 +36,23 @@ public Schedule deserialize(JsonParser parser, DeserializationContext context) t
...
         if (node.has(ACTIVITIES_PROPERTY)) {
             schedule.setActivities(JsonUtils.asEntityList(node.get(ACTIVITIES_PROPERTY), Activity.class));
-        } else if (node.has(ACTIVITY_TYPE_PROPERTY) && node.has(ACTIVITY_REF_PROPERTY)) {
-            String label = JsonUtils.asText(node, LABEL_PROPERTY);
-            String ref = JsonUtils.asText(node, ACTIVITY_REF_PROPERTY);
-            Activity activity = new Activity(label, ref);
-            schedule.setActivities(Lists.newArrayList(activity));
+        }
+        if (node.has(TIMES_PROPERTY)) {
+            ArrayNode array = (ArrayNode)node.get(TIMES_PROPERTY);
```
DwayneJengSage: You'll want to use node.hasNonNull() instead of node.has(). Also, this code will throw a ClassCastException if times isn't an array. (This may or may not be handled by Jackson.) Also, you don't actually need to cast it to an ArrayNode, since JsonNode has a get() method. If this node is null, or if any of its children are null, this will throw a NullPointerException. Lastly, if any of the children are a malformatted LocalTime, this will throw some kind of RuntimeException.

All of this can be avoided by using Jackson directly instead of a custom serializer.

app/org/sagebionetworks/bridge/json/ScheduleSerializer.java
```diff
@@ -31,14 +34,23 @@ public void serialize(Schedule schedule, JsonGenerator gen, SerializerProvider p
...
         }
-        if (schedule.getEndsOn() != null && schedule.getEndsOn() != 0L) {
-            writeString(gen, ENDS_ON_PROPERTY, DateUtils.convertToISODateTime(schedule.getEndsOn()));    
+        if (schedule.getEndsOn() != null) {
+            writeString(gen, ENDS_ON_PROPERTY, DateUtils.getISODateTime(schedule.getEndsOn()));    
         }
-        if (schedule.getExpires() != null && schedule.getExpires() != 0L) {
-            writeString(gen, EXPIRES_PROPERTY, DateUtils.convertToDuration(schedule.getExpires()));    
+        if (schedule.getExpires() != null) {
+            writeString(gen, EXPIRES_PROPERTY, schedule.getExpires().toString());
```
DwayneJengSage: Do we want to enforce a specific format? Or is the default formatter sufficient? Similarly for delay, interval, and times.

app/org/sagebionetworks/bridge/models/schedules/TaskScheduler.java
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * Get all the tasks for the schedule that start before up to the "until" date and time passed to the method. Tasks
+     * will be returned from the event that triggers the schedule (by default, the enrollment date of the user). If the
+     * events map contains a later event for the last time the schedule plan generated tasks, schedules will be
+     * generated after that date and time.
+     * 
+     * More information about schedules themselves can be found at:
+     * https://sagebionetworks.jira.com/wiki/display/BRIDGE/Schedule
+     * 
+     * @param events
```
DwayneJengSage: What is "events"?

app/org/sagebionetworks/bridge/models/schedules/TaskScheduler.java
```diff
@@ -0,0 +1,26 @@
...
+     * @param startAt
+     * @param endsAt
```
DwayneJengSage: Fix comment.

app/org/sagebionetworks/bridge/models/schedules/TaskScheduler.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+import org.joda.time.DateTime;
+
+public interface TaskScheduler {
+
+    /**
+     * Get all the tasks for the schedule that start before up to the "until" date and time passed to the method. Tasks
```
DwayneJengSage: Does this return _all_ tasks before the "until" date? Or all tasks between now and the "until" date?

app/org/sagebionetworks/bridge/models/schedules/SchedulerFactory.java
```diff
@@ -0,0 +1,18 @@
+package org.sagebionetworks.bridge.models.schedules;
+
+import org.sagebionetworks.bridge.validators.ScheduleValidator;
+import org.sagebionetworks.bridge.validators.Validate;
+
+public class SchedulerFactory {
+
+    public static TaskScheduler getScheduler(String schedulePlanGuid, Schedule schedule) {
+        // TODO: This validation will be moved to create/update, not the time of use.
+        Validate.entityThrowingException(new ScheduleValidator(), schedule);
```
DwayneJengSage: Can we make a singleton ScheduleValidator (preferably a public static final in ScheduleValidator) instead of creating a new one every time?

app/org/sagebionetworks/bridge/models/schedules/SchedulerFactory.java
```diff
@@ -0,0 +1,18 @@
+package org.sagebionetworks.bridge.models.schedules;
+
+import org.sagebionetworks.bridge.validators.ScheduleValidator;
+import org.sagebionetworks.bridge.validators.Validate;
+
+public class SchedulerFactory {
```
DwayneJengSage, line 3: How do we plan on wiring this up to the rest of Bridge?

app/org/sagebionetworks/bridge/json/ScheduleSerializer.java
alxdarksage: We're following the ISO8601 format for durations, there aren't any alternative formats I'm aware of (there aren't short versions of it as there are with dates, for example).

app/org/sagebionetworks/bridge/models/schedules/TaskScheduler.java
```diff
@@ -0,0 +1,26 @@
+package org.sagebionetworks.bridge.models.schedules;
+
+import java.util.List;
+import java.util.Map;
+
+import org.joda.time.DateTime;
+
+public interface TaskScheduler {
```
DwayneJengSage: Is there value in having this be a separate class from BaseTaskScheduler? Why not consolidate the two into a single abstract class?

app/org/sagebionetworks/bridge/models/schedules/TaskScheduler.java
alxdarksage: All tasks from the last recorded created task (the event "startsOn" or "scheduledOn" in the event map) to the "until" timestamp.

alxdarksage: Note: this isn't done yet. Right now it returns all tasks from the event to the until timestamp.

app/org/sagebionetworks/bridge/models/schedules/SchedulerFactory.java
alxdarksage: As noted, this is only temporary: this is referenced elsewhere during schedule plan CRUD and will be removed here.

app/org/sagebionetworks/bridge/models/schedules/SchedulerFactory.java
alxdarksage, line 3: It'll be called at request time when a study's schedule plans are retrieved, the schedule for the user is retrieved, and then the scheduler is called to retrieve the tasks. These will be compared against tasks already created for the user, and saved if they are new (extending forwarding time from the last request), before the list of tasks are returned to the user.

app/org/sagebionetworks/bridge/json/ScheduleSerializer.java
DwayneJengSage: What about LocalTime?

app/org/sagebionetworks/bridge/models/schedules/TaskScheduler.java
DwayneJengSage: Can you update the comments to clarify this and to clarify what the events map is? What is a triggering event and how do I specify it when I call the API?

app/org/sagebionetworks/bridge/models/schedules/SchedulerFactory.java
DwayneJengSage, line 3: More specifically, which controllers/services and which APIs/methods will this be hooked up to?

app/org/sagebionetworks/bridge/json/ScheduleSerializer.java
alxdarksage: No, it has to be a duration. It's always stated as such or convertible to such.

DwayneJengSage: No, I mean for times.

app/org/sagebionetworks/bridge/models/schedules/BaseTaskScheduler.java
```diff
@@ -0,0 +1,82 @@
...
+    protected final Schedule schedule;
+    
+    BaseTaskScheduler(String schedulePlanGuid, Schedule schedule) {
+        this.schedulePlanGuid = schedulePlanGuid;
+        this.schedule = schedule;
+    }
+    
+    protected DateTime getStartTimeBasedOnEvent(Schedule schedule, Map<String, DateTime> events) {
+        // If no event is specified, it's enrollment by default.
+        String eventId = schedule.getEventId();
```
DwayneJengSage: It's not clear that eventId actually means starting event ID. Can we rename this to make it more clear?

app/org/sagebionetworks/bridge/models/schedules/TaskScheduler.java
alxdarksage: I will, but as it isn't done yet, it's more like writing a design document in the javadocs. I've done that as well as I'm able in the wiki, and will need to return to the javadocs when the tasks portion of this work is done.

app/org/sagebionetworks/bridge/models/schedules/SchedulerFactory.java
alxdarksage, line 3: That's not a discussion for a code review... none of the code exists yet, please see the design document I sent out over the last week for an overview of the tasks API.

app/org/sagebionetworks/bridge/json/ScheduleSerializer.java
alxdarksage: It's whatever LocalTime can parse, which seems to be all formats for a time value without reference to a date or time zone, as per ISO 8601.

app/org/sagebionetworks/bridge/models/schedules/BaseTaskScheduler.java
alxdarksage: This is not the only event or possibly even the most important event that can be scheduled against. Scheduling against the last scheduled task is an internal event that we will care about, but researchers will probably schedule against the completion of specific tasks, so event IDs are going to be a combination of the event type, and an identifier for an activity, or a special event like enrollment (as if the first task a user completes is enrolling themselves).

Although this doesn't exist yet, it may take the form of things like
scheduledOn:survey:guid --> timestamp
finishedOn:task:tapTest --> timestamp

See task events in this document: https://sagebionetworks.jira.com/wiki/display/BRIDGE/Scheduling+and+Task+Design

app/org/sagebionetworks/bridge/models/schedules/SchedulerFactory.java
DwayneJengSage, line 3: Respectfully disagree. Looking at the big picture is an important part of the code review process.

Also, unless I'm looking at the wrong page, I'm not seeing the design details in https://sagebionetworks.jira.com/wiki/display/BRIDGE/Schedule

app/org/sagebionetworks/bridge/json/ScheduleSerializer.java
DwayneJengSage: No, I mean, for serializing LocalTimes. LocalTime.toString() uses a default format, which may or may not be the format we want to use. I recommend explicitly setting a format, even if it's the same as the default format.

alxdarksage: SFAICT, all of the toString() methods in the Joda Time library produce standard ISO 8601 formatted output, e.g. http://joda-time.sourceforge.net/apidocs/org/joda/time/LocalTime.html#toString() where it says as much, or http://joda-time.sourceforge.net/apidocs/org/joda/time/base/AbstractInstant.html#toString(), where it says the same for DateTime.

app/org/sagebionetworks/bridge/json/JsonUtils.java
alxdarksage, line 109: Added tests for this and most other static methods in this utility class.

app/org/sagebionetworks/bridge/json/ScheduleDeserializer.java
alxdarksage: It looks from the SDK like we can get rid of the schedule serializer/deserializer, although there we include the jackson-datatype-joda package. Testing this now.

app/org/sagebionetworks/bridge/json/ScheduleSerializer.java
DwayneJengSage: From the Javadocs, they return HH:mm:ss.SSS and yyyy-MM-ddTHH:mm:ss.SSSZZ respectfully. These are ISO8601 strings, but not all ISO8601 strings are in this format. If this is the format we want, then I'm fine with that. I'd still strongly prefer to explicitly declare a format rather than relying on the default, though, but I'll leave it up to you.

test/org/sagebionetworks/bridge/models/schedules/ScheduleTest.java
alxdarksage, line 24: Added assertion

app/org/sagebionetworks/bridge/json/BridgeObjectMapper.java
```diff
@@ -54,12 +55,24 @@ public void setupModule(SetupContext context) {
...
-        serializers.addSerializer(Schedule.class, new ScheduleSerializer());
-        this.registerModule(serializers);
-        
+        this.registerModule(new JodaModule());
         this.registerModule(new LowercaseEnumModule());
+        this.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
+        /* This works, but we don't want to do it. While it's nice to have shorter dates and times in some classes, like 
+         * the Schedule (where seconds/milliseconds aren't needed), we do not know if this creates challenges for some 
+         * language libraries when parsing ISO 8601 strings, and there are DateTimes in the system that we do want to 
+         * server as timestamps to the millisecond. If used, we'd need to scope to specific properties of specific objects.
```
DwayneJengSage: Let's remove this comment and the commented out code. We already have examples of custom Joda Jackson serializers in our JSON package, and if we really want to serialize LocalTimes in just HH:mm format, we can write a custom serializer for that and use the JsonSerialize annotation.

test/org/sagebionetworks/bridge/models/schedules/IntervalTaskSchedulerTest.java
```diff
@@ -0,0 +1,462 @@
...
+        schedule.setExpires("PT24H");
+        
+        tasks = SchedulerFactory.getScheduler("", schedule).getTasks(events, ENROLLMENT.plusMonths(1));
+        assertDates(tasks, "2015-03-30 08:00");
+    }
+    
+    @Test
+    public void scheduleCreatesTasksOnDayAfterEventTime() {
+        //org.junit.Assert.fail("Not implemented");
+    }
```
DwayneJengSage: I think this is implemented now?

test/org/sagebionetworks/bridge/models/schedules/TaskSchedulerTest.java
```diff
@@ -0,0 +1,168 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import com.newrelic.agent.deps.com.google.common.collect.Maps;
+
+/**
+ * These tests cover other aspects of the scheduler besides the accuracy of its scheduling.
+ *
+ */
+public class TaskSchedulerTest {
```
DwayneJengSage, line 23: Other unit tests I feel we should add:
- No event found - Events is an empty map. No enrollment or anything. From reading the code, it looks like the scheduler would return an empty list, but I didn't see a unit test for this.
- A test encapsulating something like "1 hour after breakfast".

app/org/sagebionetworks/bridge/models/schedules/Schedule.java
```diff
@@ -42,6 +53,25 @@ public void addActivity(Activity activity) {
...
+            addTime(time);
+        }
+    }
+    public void addTime(LocalTime time) {
+        checkNotNull(time);
+        this.times.add(time);
+    }
+    public void addTime(String time) {
+        addTime(LocalTime.parse(time));
+    }
```
DwayneJengSage, line 74: Some basic unit tests for setTimes(String...) and the addTime() methods?

app/org/sagebionetworks/bridge/models/schedules/Schedule.java
```diff
@@ -60,24 +90,62 @@ public String getCronTrigger() {
...
     }
-    public void setStartsOn(Long startsOn) {
+    @JsonProperty("startsOn")
+    public void setStartsOn(DateTime startsOn) {
         this.startsOn = startsOn;
     }
-    public Long getEndsOn() {
+    public void setStartsOn(String startsOn) {
+        setStartsOn(DateTime.parse(startsOn));
+    }
```
DwayneJengSage, line 102: Should we have unit tests for the parsing setters for: setStartsOn(String), setEndsOn(String), setExpires(String), setDelay(String), setInterval(String)?

app/org/sagebionetworks/bridge/models/schedules/SchedulerFactory.java
```diff
@@ -0,0 +1,18 @@
...
+
+    public static TaskScheduler getScheduler(String schedulePlanGuid, Schedule schedule) {
+        // TODO: This validation will be moved to create/update, not the time of use.
+        Validate.entityThrowingException(new ScheduleValidator(), schedule);
+        
+        if (schedule.getCronTrigger() != null) {
+            return new CronTaskScheduler(schedulePlanGuid, schedule);
+        }
+        return new IntervalTaskScheduler(schedulePlanGuid, schedule);
```
DwayneJengSage, line 9: Would it be cleaner for TaskScheduler to be an immutable singleton that takes in the guid and schedule as an input, rather than creating a new one for each schedule? At the very least, this saves you the trouble of creating a scheduler cache.

app/org/sagebionetworks/bridge/validators/ActivityValidator.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public void validate(Object obj, Errors errors) {
+        Activity activity = (Activity)obj;
+        if (activity.getActivityType() == null) {
+            errors.rejectValue(ACTIVITY_TYPE_PROPERTY, CANNOT_BE_NULL);
+        }
+        if (isBlank(activity.getLabel())) {
+            errors.rejectValue("label", CANNOT_BE_BLANK);
+        }
```
DwayneJengSage, line 34: It looks like label is a new check to the validation. Can we add this to the corresponding unit test in ScheduleValidatorTest?

DwayneJengSage, line 34: Is it also worth testing with empty strings in addition to nulls?

test/org/sagebionetworks/bridge/validators/ScheduleValidatorTest.java
```diff
@@ -102,4 +102,110 @@ public void activityCorrectlyParsesPublishedSurveyPath() {
         assertEquals("2015-01-27T17:46:31.237Z", ref.getCreatedOn());
     }
     
+    @Test
+    public void recurringSchedulesMustHaveCronTriggerOrInterval() {
+        Schedule schedule = new Schedule();
+        schedule.setScheduleType(ScheduleType.RECURRING);
+        try {
+            Validate.entityThrowingException(validator, schedule);
+        } catch(InvalidEntityException e) {
```
DwayneJengSage, line 132: Add a fail() after the validate to make sure it actually throws. Same for all the rest of the tests.

test/org/sagebionetworks/bridge/validators/ScheduleValidatorTest.java
```diff
@@ -102,4 +102,110 @@ public void activityCorrectlyParsesPublishedSurveyPath() {
...
+            assertEquals("Schedule recurring schedules should have either a cron expression, or an interval", e.getErrors().get("Schedule").get(0));
+        }
+        
+        schedule.setInterval("P1D");
+        try {
+            Validate.entityThrowingException(validator, schedule);
+        } catch(InvalidEntityException e) {
```
DwayneJengSage, line 140: Should this throw or not? If it's supposed to throw, add fail() as mentioned above. If it's not supposed to throw, don't include the try catch.

app/org/sagebionetworks/bridge/validators/ScheduleValidator.java
```diff
@@ -16,86 +21,133 @@
...
+     * @return
+     */
+    private boolean intervalTooShort(Period interval) {
+        if (interval == null) {
+            return false;
+        }
+        DateTime now = DateTime.now();
+        DateTime dur = now.plus(interval);
+        DateTime hours24 = now.plusHours(24);
+        return dur.isBefore(hours24);
```
DwayneJengSage: This will fail when Daylight Savings starts (the one day of the year that's only 23 hours long). The safest thing to do would be to check that the period has at least 1 in getYears() or getMonths() or getWeeks() or getDays() (and optionally has none of getHours(), getMinutes(), getSeconds(), and getsMillis(), unless we decide that a period of 1 day and 30 minutes makes sense.)

DwayneJengSage: Alternatively, now.plusDays(1) should do the trick.

test/org/sagebionetworks/bridge/validators/ScheduleValidatorTest.java
```diff
@@ -102,4 +102,110 @@ public void activityCorrectlyParsesPublishedSurveyPath() {
...
+            assertEquals("interval must be at least one day", e.getErrors().get("interval").get(0));
+        }
+    }
+    
+    @Test
+    public void rejectsDelayOfLessThanADay() {
+        
+    }
```
DwayneJengSage: Empty unit test, which is also redundant with delaysWithTimesOfDayMustBeMoreThanOneDayLong()

test/org/sagebionetworks/bridge/validators/ScheduleValidatorTest.java
```diff
@@ -25,11 +26,12 @@ public void before() {
     }
     
     @Test
-    public void mustHaveAtLeastOneActivity() {
+    public void mustHaveAtLeastOneActivityAndAScheduleType() {
```
DwayneJengSage, line 30: Add unit tests for:
- schedule with both cron and interval
- expires too short

test/org/sagebionetworks/bridge/models/schedules/CronTaskSchedulerTest.java
```diff
@@ -0,0 +1,120 @@
...
+ * Tests for the cron scheduler. Most details of the schedule object have been tested in 
+ * the IntervalTaskSchedulerTask class, here we're testing some of the specifics of 
+ * the cron schedules.  
+ */
+public class CronTaskSchedulerTest {
+    
+    private Map<String, DateTime> events;
+    private List<Task> tasks;
+    
+    private DateTime ENROLLMENT = DateTime.parse("2015-03-23T10:00:00Z");
```
DwayneJengSage, line 28: Can we change all our tests to run in a non-UTC time zone (like -0700) just to make sure we aren't accidentally forcing UTC anywhere?

test/org/sagebionetworks/bridge/models/schedules/TaskSchedulerTest.java
alxdarksage, line 23: Adding the first test. The second test is called IntervalTaskSchedulerTest.onceEventScheduleWorks()

app/org/sagebionetworks/bridge/models/schedules/Schedule.java
alxdarksage, line 102: Adding these, along with the times methods. The tests use these setters and test the resulting values, but we can make a test that just does this.

app/org/sagebionetworks/bridge/models/schedules/SchedulerFactory.java
alxdarksage, line 9: I like, a lot, in the schedulers, not having to pass the schedule from method to method. I would prefer to leave it that way... in truth, all the values could be passed in the constructor and getTasks() could take no arguments. But the version that allows some arguments was easiest to re-execute in tests. These scheduler objects won't be cached or re-used (creating an object like this is not expensive).

test/org/sagebionetworks/bridge/models/schedules/TaskSchedulerTest.java
```diff
@@ -0,0 +1,168 @@
...
+        events.put("task:task1", TASK1_EVENT);
+        
+        tasks = SchedulerFactory.getScheduler("", schedule2).getTasks(events, NOW.plusMonths(2));
+        assertEquals(TASK1_EVENT, tasks.get(0).getScheduledOn());
+    }
+    
+    @Test
+    public void taskSchedulerWorksInDifferentTimezone() {
+        
+    }
```
DwayneJengSage, line 124: Blank test?

app/org/sagebionetworks/bridge/validators/ActivityValidator.java
alxdarksage, line 34: This was just moving activity validation out of the schedule validator, because it was getting complicated. Going to create an ActivityValidatorTest and add/move tests there.

app/org/sagebionetworks/bridge/validators/ScheduleValidator.java
alxdarksage: Yes this is good, will change

test/org/sagebionetworks/bridge/models/schedules/CronTaskSchedulerTest.java
alxdarksage, line 28: Well, I have one test that specifically tests a different time zone, and it works, it doesn't seem necessary to switch all of them to another time zone... I can add a test that mixes timezones. That should also work.

alxdarksage, line 28: Actually it looks like that test got deleted, I'll add it back + mixed time zones.

test/org/sagebionetworks/bridge/validators/ScheduleValidatorTest.java
alxdarksage, line 30: Added

test/org/sagebionetworks/bridge/models/schedules/IntervalTaskSchedulerTest.java
```diff
@@ -0,0 +1,457 @@
...
+import org.joda.time.DateTime;
+import org.junit.Before;
+import org.junit.Test;
+
+import com.google.common.collect.Maps;
+
+public class IntervalTaskSchedulerTest {
+
+    private Map<String, DateTime> events;
+    private List<Task> tasks;
```
DwayneJengSage, line 21: Super minor: Is there any reason why this has to be a class variable instead of a local variable? It's not being shared across multiple tests, nor used on Before or After. It's prolly not going to affect anything, but it does seem unclean.

alxdarksage, line 21: Tasks? To spare me from typing List&lt;Task&gt; tasks several dozen times. I can null it in a @Before method but the class is instantiated each time a test is run.

PR #618, opened 2015-04-17
add delete Upload Schema APIs

DwayneJengSage:
Add the ability to delete Upload Schemas using researcher APIs. These APIs are useful for 2 reasons:
- gives researchers the ability to delete old schemas, so they don't clutter up Synapse (or Dynamo or any future researcher UI)
- allows integration tests to create schemas, then clean up after itself

Testing done:
- unit tests for DAO, service, and controllers
- manually tested by
  *\* creating schema "manual-test-schema" with revisions 1, 2, and 3
  *\* delete by ID and rev manual-test-schema v3 and verify that v1 and v2 were unaffected
  *\* delete by ID manual-test-schema and verify that all manual-test-schemas were deleted

### DIFF
--- a/test/org/sagebionetworks/bridge/TestUtils.java
+++ b/test/org/sagebionetworks/bridge/TestUtils.java
@@ -1,6 +1,12 @@
 package org.sagebionetworks.bridge;
 
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.commons.lang3.RandomStringUtils;
+import play.mvc.Http;
 
 public class TestUtils {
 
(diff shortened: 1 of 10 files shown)


test/org/sagebionetworks/bridge/TestUtils.java
```diff
@@ -17,7 +23,22 @@ public void run() {
             }
         }
     }
-    
+
+    public static Http.Context mockPlayContextWithJson(String json) throws Exception {
```
eric-wu, line 27: Nice!

DwayneJengSage, line 27: Credit goes to Alx. He wrote the original method. I just ported it to the shared TestUtils class.

eric-wu, line 27: I wrote similar mock code. It is nice you move it to TestUtils.

PR #622, opened 2015-04-25
iOS Upload Validation Handler v2

DwayneJengSage:
This change sets up an iOS Upload Validation Handler v2 and has it running in shadow mode behind iOS Handler v1. This includes the following components:
- added schema type to the upload schema - This allows schemas to identify as iOS surveys or non-surveys, so the iOS handler can properly parse the data. This has the side effect of tying the schema to a specific data format. This may or may not be ideal, but there's currently no way around it as the iOS uploads don't self identify in any other way.
- added TestingHandler and ContextValidator to the upload validator chain - The TestingHandler can be used to run a "prod" and "test" handlers (in this case the v1 and v2 handlers) side by side, using the Context Validator to compare them, and ensure the test handler does not impact the prod data stream. This allows us to test new iOS Handler on a real data stream before enabling it in prod.
- minor patch to iOS Handler v1 to allow floats and ints to accept any numeric type (but will flatten to the expected type when exporting to Synapse) - This is done because Research Kit data doesn't differentiate between ints and floats
- iOS Handler v2 implementation and the corresponding context validator
- added shallow copy to the Upload Validation Context - This allows the test handler to use a copy of the context instead of the real context

Testing done:
- a bunch of unit tests
- pointed the integration tests to my local box and ensured Upload Tests and Upload Schema Tests still work
- verified in my local box using debug statements that the TestingHandler was executing both the v1 and v2 handlers

Next steps:
- We'll need to add the schema type field to all production schemas, or else the iOS Handler v2 will spit out a bunch of warnings.
- We'll also need to figure out a way to stitch together old tables and new tables, although I suspect the number of impacted tables will be very small, if any.


app/org/sagebionetworks/bridge/upload/IosSchemaValidationHandler2.java
```diff
@@ -0,0 +1,463 @@
...
+        removeTimestampsFromFilenames(unzippedDataMap);
+
+        // extract other data from info.json
+        JsonNode itemNode = infoJson.get(KEY_ITEM);
+        if (itemNode == null) {
+            // fall back to "identifier"
+            itemNode = infoJson.get(KEY_IDENTIFIER);
+        }
+        if (itemNode == null || itemNode.isNull()) {
+            throw new UploadValidationException("info.json is missing \"item\" field");
```
alxdarksage, line 148: UploadValidationException... is this returned to the caller of the rest service? Should it be a BridgeServiceException with a 400 response, or is it caught and translated?

app/org/sagebionetworks/bridge/upload/IosSchemaValidationHandler2.java
```diff
@@ -0,0 +1,463 @@
...
+                fileMap.put(newFilename, fileData);
+            }
+        }
+    }
+
+    private static Map<String, JsonNode> convertSurveyToJsonData(UploadValidationContext context, String uploadId,
+            Map<String, JsonNode> jsonDataMap) {
+        // Currently, the 3rd party iOS apps don't tag surveys or questions with guids. (In fact, some of the
+        // surveys aren't even in the Surveys table yet.) So we have to store them in Health Data Records instead of
+        // Survey Responses.
```
alxdarksage, line 289: Boy this is going to  messy when we switch over...

test/org/sagebionetworks/bridge/upload/IosSchemaValidationHandler2Test.java
```diff
@@ -0,0 +1,418 @@
...
+                        .withType(UploadFieldType.ATTACHMENT_BLOB).build(),
+                new DynamoUploadFieldDefinition.Builder().withName("attachment.json")
+                        .withType(UploadFieldType.ATTACHMENT_JSON_BLOB).build(),
+                new DynamoUploadFieldDefinition.Builder().withName("inline.json")
+                        .withType(UploadFieldType.INLINE_JSON_BLOB).build(),
+                new DynamoUploadFieldDefinition.Builder().withName("field.json.attachment")
+                        .withType(UploadFieldType.ATTACHMENT_JSON_BLOB).build(),
+                new DynamoUploadFieldDefinition.Builder().withName("field.json.string")
+                        .withType(UploadFieldType.STRING).build()));
+
```
alxdarksage, line 114: I'm harping on this but:
mixedSchema.addFieldDefinition("nonJsonFile.txt", ATTACHMENT_BLOB);
mixedSchema.addFieldDefinition("attachment.json", ATTACHMENT_JSON);
mixedSchema.addFieldDefinition("inline.json", INLINE_JSON_BLOB);
mixedSchema.addFieldDefinition("field.json.attachment", ATTACHMENT_JSON_BLOB);
mixedSchema.addFieldDefinition("field.json.string", STRING);

Seems like a reasonable shortcut for these, assuming fieldDefinitions has a mutable list initialized. Could be made immutable when getFieldDefinitions() is called.

app/org/sagebionetworks/bridge/upload/IosSchemaValidationHandler2.java
DwayneJengSage, line 148: This is done in an asynch worker. The upload validation task has code which catches these, writes these to the log, and updates the upload validation status accordingly:
https://github.com/Sage-Bionetworks/BridgePF/blob/develop/app/org/sagebionetworks/bridge/upload/UploadValidationTask.java#L70

PR #625, opened 2015-04-28
Tasks

alxdarksage:


app/org/sagebionetworks/bridge/dynamodb/DynamoTaskDao.java
```diff
@@ -68,15 +72,27 @@ public void setSchedulePlanService(SchedulePlanService schedulePlanService) {
...
     @Override
     public List<Task> getTasks(User user, DateTime endsOn) {
+        StudyIdentifier studyId = new StudyIdentifierImpl(user.getStudyKey());
+        
         Map<String,List<Task>> scheduledTasks = scheduleTasksForPlans(user, endsOn);
         List<Task> tasksToSave = Lists.newArrayList();
         for (String runKey : scheduledTasks.keySet()) {
             if (taskHasNotBeenSaved(user.getHealthCode(), runKey)) {
-                tasksToSave.addAll(scheduledTasks.get(runKey));
+                for (Task task : scheduledTasks.get(runKey)) {
```
alxdarksage: dao shouldn't rely on services. You'd need to pull this up to the service level, do the iteration there, then maybe call back to the dao to save a list of tasks that are verified not in the db.

Lower down, also referring to a service to get schedules.

PR #630, opened 2015-04-30
Changing how survey responses are deleted; redoing hash/range/indices of...

alxdarksage:
- changing how survey responses are deleted (deleted with user delete);
- redoing hash/range/indices of the table. Tables have to be deleted and redone
- collapsing identifier and guid, leaving identifier as the external property seen through the API. I am now checking the identifier in advance of using it to verify it is truly unique for that user.

This creates a SurveyResponse2 table, after rollout, we can delete SurveyResponse.

### DIFF
--- a/app/controllers/SurveyController.java
+++ b/app/controllers/SurveyController.java
@@ -171,7 +171,7 @@ public class SurveyController extends BaseController {
         Survey survey = surveyService.getSurvey(keys);
         verifySurveyIsInStudy(session, studyId, survey);
         
-        surveyService.deleteSurvey(session.getUser().getHealthCode(), studyId, survey);
+        surveyService.deleteSurvey(studyId, survey);
         expireCache(surveyGuid, createdOnString);
         
         return okResult("Survey deleted.");
--- a/app/controllers/SurveyController.java
+++ b/app/controllers/SurveyController.java
@@ -171,7 +171,7 @@ public class SurveyController extends BaseController {
         Survey survey = surveyService.getSurvey(keys);
         verifySurveyIsInStudy(session, studyId, survey);
         
-        surveyService.deleteSurvey(studyId, survey);
+        surveyService.deleteSurvey(session.getUser().getHealthCode(), studyId, survey);
         expireCache(surveyGuid, createdOnString);
         
         return okResult("Survey deleted.");
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponse.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponse.java
@@ -3,7 +3,6 @@ package org.sagebionetworks.bridge.dynamodb;
 import java.util.List;
 import java.util.Objects;
 
-import org.apache.commons.lang3.StringUtils;
 import org.sagebionetworks.bridge.json.BridgeObjectMapper;
 import org.sagebionetworks.bridge.json.DateTimeJsonDeserializer;
 import org.sagebionetworks.bridge.json.DateTimeJsonSerializer;
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponse.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponse.java
@@ -6,7 +6,9 @@ import java.util.Objects;
 import org.sagebionetworks.bridge.json.BridgeObjectMapper;
 import org.sagebionetworks.bridge.json.DateTimeJsonDeserializer;
 import org.sagebionetworks.bridge.json.DateTimeJsonSerializer;
+import org.sagebionetworks.bridge.json.DateUtils;
 import org.sagebionetworks.bridge.json.JsonUtils;
+import org.sagebionetworks.bridge.models.GuidCreatedOnVersionHolder;
 import org.sagebionetworks.bridge.models.surveys.SurveyAnswer;
 import org.sagebionetworks.bridge.models.surveys.SurveyResponse;
 
@@ -71,6 +73,9 @@ public final class DynamoSurveyResponse implements SurveyResponse {
     }
     public void setSurveyKey(String surveyKey) {
         this.surveyKey = surveyKey;
+    }
+    public void setSurveyKey(GuidCreatedOnVersionHolder keys) {
+        this.surveyKey = String.format("%s:%s", keys.getGuid(), DateUtils.convertToISODateTime(keys.getCreatedOn()));
     }
     @DynamoDBIgnore
     @JsonIgnore
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponse.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponse.java
@@ -27,7 +27,7 @@ import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.collect.Lists;
 
-@DynamoDBTable(tableName = "SurveyResponse")
+@DynamoDBTable(tableName = "SurveyResponse2")
 public class DynamoSurveyResponse implements SurveyResponse {
 
     private static final String ANSWERS_PROPERTY = "answers";
@@ -35,10 +35,13 @@ public class DynamoSurveyResponse implements SurveyResponse {
     private String healthCode;
     private String identifier;
     
-    private String surveyKey; // stored in dynamo, local index range key
-    private String surveyGuid; // stored in dynamo
-    private long surveyCreatedOn; // stored in dynamo
-    private Survey survey; // constructed and returned to the consumer
+    // These three are stored in Dynamo.
+    private String surveyKey;
+    private String surveyGuid;
+    private long surveyCreatedOn;
+    
+    // This retrieved and returned as part of the JSON to the consumer
+    private Survey survey;
     
     private Long startedOn;
     private Long completedOn;
@@ -67,7 +70,6 @@ public class DynamoSurveyResponse implements SurveyResponse {
         this.identifier = identifier;
     }
     
-    // Local index: surveyKey
     @JsonIgnore
     @DynamoDBAttribute
     @DynamoDBIndexRangeKey(localSecondaryIndexName = "healthCode-surveyKey-index")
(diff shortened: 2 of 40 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponse.java
```diff
@@ -13,54 +15,72 @@
...
     
+    @JsonIgnore
+    @DynamoDBAttribute
+    @DynamoDBIndexRangeKey(localSecondaryIndexName = "healthCode-surveyKey-index")
+    public String getSurveyKey() {
+        return surveyKey;
+    }
+    public void setSurveyKey(String surveyKey) {
+        this.surveyKey = surveyKey;
+    }
```
DwayneJengSage, line 75: It's possible to call these APIs in a way to have surveyKey not consistent with surveyGuid and surveyCreatedOn. Would it be cleaner to have getSurveyKey always derive the result from surveyGuid and surveyCreatedOn, and to have setSurveyKey always parse that string and set surveyGuid and surveyCreatedOn accordingly?

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponse.java
```diff
@@ -77,24 +97,11 @@ public long getSurveyCreatedOn() {
...
-        return healthCode + ":" + identifier;
-    }
-    public void setGuid(String guid) {
-        // Ignore. Always calculated as above
+    @DynamoDBIgnore
+    public void setSurveyKey(GuidCreatedOnVersionHolder keys) {
+        this.surveyGuid = keys.getGuid();
+        this.surveyCreatedOn = keys.getCreatedOn();
+        this.surveyKey = String.format("%s:%s", keys.getGuid(), Long.toString(keys.getCreatedOn()));
     }
```
DwayneJengSage, line 100: Should we have a unit test to enforce that setSurveyKey() modifies the surveyGuid and surveyCreatedOn? (For this method and the setSurveyKey(String), if we decide to change it to work like this one.

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponse.java
```diff
@@ -202,49 +202,19 @@ public int hashCode() {
...
-                return false;
-        } else if (!version.equals(other.version))
-            return false;
-        return true;
+        return (Objects.equals(healthCode, other.healthCode) && Objects.equals(identifier, other.identifier) && 
+            Objects.equals(surveyKey, other.surveyKey) && Objects.equals(surveyGuid, other.surveyGuid) && 
+            Objects.equals(surveyCreatedOn, other.surveyCreatedOn) && Objects.equals(survey, other.survey) && 
+            Objects.equals(startedOn, other.startedOn) && Objects.equals(completedOn, other.completedOn) && 
+            Objects.equals(answers, other.answers));
     }
```
DwayneJengSage, line 189: hashCode and equals are both missing version. Do we care about that field? Should we add an EqualsVerifier test for these?

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponseDao.java
```diff
@@ -59,25 +49,33 @@ public void setSurveyDao(DynamoSurveyDao surveyDao) {
...
+        this.lockDao = lockDao;
     }
-
+    
     @Override
     public SurveyResponse createSurveyResponse(GuidCreatedOnVersionHolder keys, String healthCode,
             List<SurveyAnswer> answers, String identifier) {
+        
+        String key = RedisKey.SURVEY_RESPONSE_LOCK.getRedisKey(identifier);
+        String lock = null;
```
DwayneJengSage: DDB optimistic locking will take care of this for us, so we don't need to use our own locking mechanism.

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponseDao.java
```diff
@@ -59,25 +49,33 @@ public void setSurveyDao(DynamoSurveyDao surveyDao) {
...
         try {
+            lock = lockDao.acquireLock(SurveyResponse.class, key);
             SurveyResponse response = getSurveyResponseInternal(healthCode, identifier);
             if (response == null) {
-                // This is the response we want. This is good, carry on.
                 return createSurveyResponseInternal(keys, healthCode, answers, identifier);
             }
```
DwayneJengSage: We actually don't even need get-before-put semantics, since DDB optimistic locking also takes care of this for us.

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponseDao.java
```diff
@@ -105,34 +103,41 @@ public SurveyResponse appendSurveyAnswers(SurveyResponse response, List<SurveyAn
...
+    public void deleteSurveyResponses(String healthCode) {
+        DynamoSurveyResponse hashKey = new DynamoSurveyResponse();
+        hashKey.setHealthCode(healthCode);
+        
+        DynamoDBQueryExpression<DynamoSurveyResponse> query = new DynamoDBQueryExpression<DynamoSurveyResponse>();
+        query.setHashKeyValues(hashKey);
+        PaginatedQueryList<DynamoSurveyResponse> results = mapper.query(DynamoSurveyResponse.class, query);
+        
+        List<DynamoSurveyResponse> responsesToDelete = Lists.newArrayList();
+        responsesToDelete.addAll(results);
```
DwayneJengSage, line 94: A PaginatedQueryList implements List, so it should be safe to pass this directly into batchDelete.

app/controllers/SurveyController.java
```diff
@@ -171,7 +171,7 @@ public Result deleteSurvey(String surveyGuid, String createdOnString) throws Exc
         Survey survey = surveyService.getSurvey(keys);
         verifySurveyIsInStudy(session, studyId, survey);
         
-        surveyService.deleteSurvey(studyId, survey);
+        surveyService.deleteSurvey(session.getUser().getHealthCode(), studyId, survey);
```
DwayneJengSage: This is the researcher's healthcode. This is later used in surveyDao to check whether the survey already has responses, but since it's keyed off the healthcode of the participant, this check will always wrongfully say the survey has no responses.

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponse.java
alxdarksage, line 75: We need surveyKey in the table for look-ups, but use survey guid/createdOn to retrieve a survey, so these could be derived from the surveyKey. Will change this.

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponse.java
alxdarksage, line 100: Survey really shouldn't be in this object, it's only attached to create a compound survey/response object when generating JSON. I've moved this to the service and created an object, SurveyResponseWithSurvey, to represent that compound object. Once that's moved, there's only surveyKey and there's no ambiguity about how it is set (it's set or it's not).

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponse.java
alxdarksage, line 189: Adding.

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponseDao.java
alxdarksage: removing (yay)

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponseDao.java
alxdarksage: Yeah okay, it took me awhile to work that out. You're right, removing.

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponseDao.java
alxdarksage, line 94: What you say makes sense, which is why I've tried it multiple times, and it won't delete. You have to copy to a list before the delete will work.

app/controllers/SurveyController.java
alxdarksage: Modified local index to global index on surveyKey, changed all delete apis so they only take a survey.

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponse.java
```diff
@@ -1,100 +1,101 @@
...
     @Override
-    public void setHealthCode(String healthCode) {
-        this.healthCode = healthCode;
+    public String getSurveyGuid() {
+        try {
+            return (surveyKey != null) ? surveyKey.split(":")[0] : null;    
+        } catch(Exception e) {
+            return null;
+        }
     }
```
DwayneJengSage, line 89: Is it worth having a unit test to make sure getSurveyGuid() and getSurveyCreatedOn() correctly parse their fields out of surveyKey?

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyResponse.java
```diff
@@ -1,100 +1,101 @@
...
+        this.surveyKey = surveyKey;
     }
-    @Override
-    @JsonIgnore
-    @DynamoDBAttribute
-    public String getHealthCode() {
-        return healthCode;
+    public void setSurveyKey(GuidCreatedOnVersionHolder keys) {
+        this.surveyKey = String.format("%s:%s", keys.getGuid(), DateUtils.convertToISODateTime(keys.getCreatedOn()));
     }
```
DwayneJengSage, line 78: Similarly, a unit test to extract the survey key from the version holder?

app/org/sagebionetworks/bridge/models/surveys/SurveyResponseView.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+    public Long getStartedOn() {
+        return response.getStartedOn();
+    }
+    public Long getCompletedOn() {
+        return response.getCompletedOn();
+    }
+    public List<SurveyAnswer> getAnswers() {
+        return response.getAnswers();
+    }
```
DwayneJengSage, line 55: Is it worth having all these convenience methods? They add a tiny amount of convenience, but they create double the work if you need to update SurveyResponse.

PR #643, opened 2015-05-08
Upload Validation v2 - write null for absent fields to maintain compatibility with v1

DwayneJengSage:
v1 would write null for absent fields. v2 would just drop the absent fields altogether. This is causing a bunch of noise in in the v1-v2 validation, so I'm going to patch v2 to make validation easier.

Testing done:
- manually tested with an example file from production
- added to unit tests


test/org/sagebionetworks/bridge/upload/IosSchemaValidationHandler2Test.java
```diff
@@ -380,8 +391,9 @@ public void mixedData() throws Exception {
         assertEquals(1, recordBuilder.getSchemaRevision());
 
         JsonNode dataNode = recordBuilder.getData();
-        assertEquals(2, dataNode.size());
+        assertEquals(3, dataNode.size());
         assertEquals("This is a string", dataNode.get("field.json.string").textValue());
+        assertTrue(dataNode.get("optional").isNull());
```
alxdarksage, line 396: Is there a path that tests when it's required?

Other than that, I can merge when you're ready.

PR #649, opened 2015-05-11
Generate task event for enrollment when user consents to research

alxdarksage:


test/org/sagebionetworks/bridge/services/ConsentServiceImplTest.java
```diff
@@ -70,7 +78,10 @@ public void before() {
         
         studyConsent = studyConsentDao.addConsent(study.getStudyIdentifier(), "/path/to", study.getMinAgeOfConsent());
         studyConsentDao.setActive(studyConsent, true);
-
+        
+        taskEventService = mock(TaskEventService.class);
+        consentService.setTaskEventService(taskEventService);
```
DwayneJengSage: This test mixes mocks with real consent services. I feel like it would be cleaner to create an entirely new test that just uses mocks.

PR #651, opened 2015-05-11
Support fallback events in Schedule.eventId as comma-delimited list

alxdarksage:


app/org/sagebionetworks/bridge/models/schedules/TaskScheduler.java
```diff
@@ -26,11 +27,12 @@
     
     protected DateTime getScheduledTimeBasedOnEvent(Schedule schedule, Map<String, DateTime> events) {
         // If no event is specified, it's enrollment by default.
-        String eventId = schedule.getEventId();
-        if (eventId == null) {
-            eventId = "enrollment";
+        String eventIdsString = schedule.getEventId();
```
alxdarksage, line 30: So if schedule is delayed 7 days, it's delay 7 days after enrollment, delay 7 days after survey finished event... not quite what we want.

app/org/sagebionetworks/bridge/models/schedules/TaskScheduler.java
```diff
@@ -26,11 +27,12 @@
...
+        if (eventIdsString == null) {
+            eventIdsString = "enrollment";
```
alxdarksage, line 32: TaskEventType.ENROLLMENT.name().toLowerCase()

app/org/sagebionetworks/bridge/models/schedules/TaskScheduler.java
```diff
@@ -42,6 +44,21 @@ protected DateTime getScheduledTimeBasedOnEvent(Schedule schedule, Map<String, D
...
+        } else if (!eventIdsString.contains(",")) {
+            return events.get(eventIdsString);
+        }
+        List<String> eventIds = Arrays.asList(eventIdsString.split("\\s*,\\s*"));
+        for (String thisEventId : eventIds) {
+            if (events.get(thisEventId) != null) {
+                return events.get(thisEventId);
+            }
+        }
+        return null;
```
alxdarksage, line 59: This is an error in the schedule itself. What can we do here? Probably fall back to enrollment.

PR #654, opened 2015-05-12
Added backfill for enrollment events

alxdarksage:


app/org/sagebionetworks/bridge/services/backfill/EnrollmentEventBackfill.java
```diff
@@ -0,0 +1,77 @@
...
+    void doBackfill(BackfillTask task, BackfillCallback callback) {
+        for (Iterator<Account> i = accountDao.getAllAccounts(); i.hasNext();) {
+            Account account = i.next();
+            Study study = studyService.getStudy(account.getStudyIdentifier());
+            HealthId mapping = healthCodeService.getMapping(account.getHealthId());
+            if (mapping != null) {
+                String healthCode = mapping.getCode();
+                
+                UserConsent consent = userConsentDao.getUserConsent(healthCode, study.getStudyIdentifier());
+                if (consent != null) {
```
eric-wu, line 69: Not critical but I recommend to call backfillFactory.createOnly() when consent is null and when health code mapping is null. The method createOnly() reports back to the caller and helps monitor the progress.

PR #657, opened 2015-05-13
Throw exceptions on zip bomb

eric-wu:


app/org/sagebionetworks/bridge/util/DuplicateZipEntryException.java
```diff
@@ -0,0 +1,13 @@
+package org.sagebionetworks.bridge.util;
```
alxdarksage, line 1: These will still be returned as 500s, but are they subclasses of BridgeServiceException?

(I didn't do a full review, I think Dwayne wrote this code so he should look at your refactoring, but this caught my eye at a consistency level.)

app/org/sagebionetworks/bridge/services/UploadArchiveService.java
```diff
@@ -175,24 +168,14 @@ public void setCmsEncryptorCache(LoadingCache<String, CmsEncryptor> cmsEncryptor
...
-            }
-            return dataMap;
+        try {
+            return zipper.unzip(bytes);
         } catch (IOException e) {
             throw new BridgeServiceException(e);
+        } catch (ZipOverflowException e) {
+            throw new BridgeServiceException(e);
+        } catch (DuplicateZipEntryException e) {
+            throw new BridgeServiceException(e);
```
eric-wu: Throw BadRequestException.

app/org/sagebionetworks/bridge/services/UploadArchiveService.java
```diff
@@ -175,24 +168,14 @@ public void setCmsEncryptorCache(LoadingCache<String, CmsEncryptor> cmsEncryptor
...
-                dataMap.put(entryName, content);
-                zipEntry = zis.getNextEntry();
-            }
-            return dataMap;
+        try {
+            return zipper.unzip(bytes);
         } catch (IOException e) {
             throw new BridgeServiceException(e);
+        } catch (ZipOverflowException e) {
+            throw new BridgeServiceException(e);
```
eric-wu: Should be a BadRequestException.

app/org/sagebionetworks/bridge/util/DuplicateZipEntryException.java
```diff
@@ -0,0 +1,13 @@
...
+
+@SuppressWarnings("serial")
+public class DuplicateZipEntryException extends RuntimeException {
+
+    public DuplicateZipEntryException() {
```
eric-wu: Not needed.

app/org/sagebionetworks/bridge/services/UploadArchiveService.java
```diff
@@ -1,38 +1,40 @@
...
+
 /**
  * Archives messages and then encrypts the archive using CMS.
  * Conversely also decrypts the archive and unpacks it.
  */
 @Component
 public class UploadArchiveService {
+
+    /** Max number of uncompressed bytes per zip entry. */
+    private final Zipper zipper = new Zipper(
```
DwayneJengSage: This probably doesn't matter since UploadArchiveService is a singleton, but can we also make this static?

app/org/sagebionetworks/bridge/util/DuplicateZipEntryException.java
DwayneJengSage, line 1: This should be fine, since we never return these exceptions directly to the user. (We could enforce catching them by making these checked exceptions.)

PR #667, opened 2015-05-14
Add a whitelist to protect production studies

eric-wu:


app/org/sagebionetworks/bridge/config/BridgeConfig.java
```diff
@@ -159,6 +162,11 @@ public int getPropertyAsInt(String name) {
         return Integer.parseInt(properties.getProperty(name));
     }
 
+    public List<String> getPropertyAsList(String name) {
+        final String prop = properties.getProperty(name);
+        return Lists.newArrayList(prop.split(";"));
```
alxdarksage: This is nitpicky, but I've been using commas to separate lists like this, and in case someone puts spaces in there, e.g. "api; asthma; breastcancer", something like split("\\s_[,;]\\s_") would be more robust. I also need this in a couple of places, so it could go to BridgeUtils.

PR #671, opened 2015-05-14
improve upload validation logging, add bulk download util

DwayneJengSage:
Two changes:
1. Improved upload validation logging, so that the exception class, error message, and all the relevant context is on a single line in the logs. This makes searching and filtering the logs much easier.
2. Added the bulk download util. This is the util I use for downloading raw upload files to diagnose issues with the data stream. Having it in the main branch means I don't have to switch workspaces or branches every time I want to use the tool.

Testing done:
- manually tested bulk download util
- ran unit tests for UploadValidationTask
- ran UploadTest integration test against my local server

### DIFF
--- a/app/org/sagebionetworks/bridge/util/BulkDownloadUtil.java
+++ b/app/org/sagebionetworks/bridge/util/BulkDownloadUtil.java
@@ -15,7 +15,7 @@ import com.google.common.base.Charsets;
 import com.google.common.base.Joiner;
 import com.google.common.io.Files;
 import org.joda.time.format.ISODateTimeFormat;
-import org.springframework.context.ApplicationContext;
+import org.springframework.context.support.AbstractApplicationContext;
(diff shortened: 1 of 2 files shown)


app/org/sagebionetworks/bridge/util/BulkDownloadUtil.java
```diff
@@ -0,0 +1,249 @@
...
+        File tmpDir = new File("tmp");
+        if (!tmpDir.exists()) {
+            tmpDir.mkdir();
+        }
+
+        System.out.println(String.format("Downloading %s files for S3 keys (%s)", s3KeyArr.length,
+                Joiner.on(", ").join(s3KeyArr)));
+
+        // spring beans
+        ApplicationContext springCtx = new ClassPathXmlApplicationContext("application-context.xml");
```
eric-wu: You might want to call registerShutdownHook() to gracefully shut down the container and properly close the resources.

PR #677, opened 2015-05-20
Add the ability to do Redis transactions

eric-wu:
Bundle up multiple Redis calls and execute them together.

This is particularly useful for writing related keys together as a unit.  For example, we keep track of user sessions using a map of session token -> session object.  To remove the session for a specific user, we want to also maintain a reverse lookup of user id --> session token.  Then the two keys are better written together as a transactional unit:

session token --> session object
user id --> session token

It greatly reduces the chance where a session is written in the cache for a user but the write of id --> session fails on a separate call.  It also guarantees that the expiration of the two keys will happen at about the same time.


app/controllers/BaseController.java
```diff
@@ -49,19 +49,15 @@
...
-    protected CacheProvider cacheProvider;
-    protected BridgeConfig bridgeConfig;
+    private BridgeConfig bridgeConfig;
+    private CacheProvider cacheProvider;
 
-    @Autowired
-    public void setStudyService(StudyService studyService) {
-        this.studyService = studyService;
-    }
+    StudyService studyService;
```
alxdarksage, line 55: Why not private or protected? This is "package private", so another class in the package could see these (maybe I've forgotten my Java). I don't think we want that.

DwayneJengSage, line 55: I agree with Alx here. We shouldn't make this package-scoped. The tests can use the setters if they need access to these, and if subclasses need access, we should make them protected.

app/controllers/EmailController.java
```diff
@@ -19,13 +20,17 @@
 @Controller("emailController")
 public class EmailController extends BaseController {
 
+    private BridgeConfig bridgeConfig;
```
DwayneJengSage: This is potentially dangerous. This bridgeConfig is actually different from BaseController.bridgeConfig, but we override the setter, so weird things could happen. If the subclass needs access to this, I strongly recommend using a getBridgeConfig() instead.

app/controllers/BaseController.java
eric-wu, line 55: "protected" has more visibility than package scope.  Members in the same package can access protected fields even if there is not inheritance.  I'd like to reduce it to private scope but that involves too many places to change.  In general, I'm very much against any field that is not private.  But let's reduce it to package scope first.

app/org/sagebionetworks/bridge/redis/JedisTransaction.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public String discard() {
+        try (Jedis jedis = this.jedis) {
+            return transaction.discard();
+        }
+    }
+
+    @Override
+    public void finalize() {
```
DwayneJengSage, line 45: finalize() is called by the garbage collector, and iirc it's generally considered an unreliable form of cleanup. I think it would be cleaner to have JedisTransaction implement Closeable and have a close method instead of a finalize() method.

Yes, this would require callers to call close (or at the very least, try-with-resources), but I think explicitly enforcing close would be cleaner than exec() and discard() implicitly calling close.

app/controllers/EmailController.java
eric-wu: Yes, I agree.  It's really ugly when injection and inheritance get mixed up.  It's actually the intention of this refactoring to move away from inheritance.  Maybe I should be more aggressive and split out part of the BaseController that uses other services into a new service?  As far as I can tell, it's mainly about managing session.  Maybe call it SessionService.

app/controllers/BaseController.java
alxdarksage, line 55: You blew my mind with this one. I looked online and you're right: https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html The only difference is visibility by subclasses. Crazy.

app/org/sagebionetworks/bridge/redis/JedisTransaction.java
```diff
@@ -0,0 +1,52 @@
...
+import redis.clients.jedis.Jedis;
+import redis.clients.jedis.Transaction;
+import redis.clients.jedis.exceptions.JedisException;
+
+/**
+ * Provides abstraction over Jedis's transactions.
+ */
+final class JedisTransaction {
+
+    private final Logger logger = LoggerFactory.getLogger(JedisTransaction.class);
```
DwayneJengSage, line 17: Nitpick: static

test/org/sagebionetworks/bridge/redis/JedisTransactionTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.junit.runner.RunWith;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+@ContextConfiguration("classpath:test-context.xml")
+@RunWith(SpringJUnit4ClassRunner.class)
+public class JedisTransactionTest {
+
+    private final String key1 = JedisTransactionTest.class.getSimpleName() + "Key1";
+    private final String key2 = JedisTransactionTest.class.getSimpleName() + "Key2";
```
DwayneJengSage, line 22: Do we have shared Redis endpoints? Or does each stack (including each dev's local box) have a separate Redis endpoint? If any of them are shared, we might want to stick some random seed in the key so the tests don't interfere with each other.

app/controllers/EmailController.java
DwayneJengSage: I'm actually more concerned that, as written, EmailController.bridgeConfig will have the correct value, but the parent BaseController.bridgeConfig will be null.

This is unlikely to cause a crash, since BaseController.bridgeConfig is only used in getStudyIdentifier(), which is never called by EmailController. But if in the future, someone adds code in BaseController that uses bridgeConfig, it may cause an obscure crash.

I strongly recommend changing it to getBridgeConfig(). It's an easy fix to do now, and it'll save us some hassle in the future.

eric-wu: I can make that quick adjustment as you are suggesting.  Still the fundamental problem is dependency injection in an abstract class, which is a big no-no.  An abstract bean (<bean abstract="true" />) is probably the original intention which is to share common properties.  But a bean marked as "abstract" is totally different from an abstract class.

test/org/sagebionetworks/bridge/redis/JedisTransactionTest.java
eric-wu, line 22: Not shared. Each has its own.

PR #678, opened 2015-05-22
remove user session by user Id

eric-wu:
Please do not merge yet.  It compiles fine and all the tests pass.  But I am still adding more tests especially around CacheProvider.  The core idea is to use Redis transactions to keep track of two mappings, sessionToken --> session and userId --> sessionToken.  Doing so allows us to remove the session by userId.  Please review and let me know your feedback.

### DIFF
--- a/app/org/sagebionetworks/bridge/cache/CacheProvider.java
+++ b/app/org/sagebionetworks/bridge/cache/CacheProvider.java
@@ -73,12 +73,13 @@ public class CacheProvider {
     }
 
     public UserSession getUserSession(final String sessionToken) {
-        final String sessionKey = RedisKey.SESSION.getRedisKey(sessionToken);
-        final String ser = jedisOps.get(sessionKey);
-        if (ser == null) {
-            return null;
-        }
+        checkNotNull(sessionToken);
         try {
+            final String sessionKey = RedisKey.SESSION.getRedisKey(sessionToken);
+            final String ser = jedisOps.get(sessionKey);
+            if (ser == null) {
+                return null;
+            }
             final UserSession session = bridgeObjectMapper.readValue(ser, UserSession.class);
             final String userKey = RedisKey.USER_SESSION.getRedisKey(session.getUser().getId());
             jedisOps.getTransaction(sessionKey)
@@ -93,8 +94,15 @@ public class CacheProvider {
     }
 
     public UserSession getUserSessionByUserId(final String userId) {
-        final String userKey = RedisKey.USER_SESSION.getRedisKey(userId);
-        final String sessionToken = jedisOps.get(userKey);
+        checkNotNull(userId);
+        String sessionToken = null;
+        try {
+            final String userKey = RedisKey.USER_SESSION.getRedisKey(userId);
+            sessionToken = jedisOps.get(userKey);
+        } catch(Throwable e) {
+            promptToStartRedisIfLocal(e);
+            throw new BridgeServiceException(e);
+        }
         if (sessionToken == null) {
             return null;
         }
--- a/app/org/sagebionetworks/bridge/cache/CacheProvider.java
+++ b/app/org/sagebionetworks/bridge/cache/CacheProvider.java
@@ -17,7 +17,6 @@ import org.sagebionetworks.bridge.redis.RedisKey;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 /**
@@ -66,7 +65,7 @@ public class CacheProvider {
                     throw new BridgeServiceException("Session storage error.");
                 }
             }
-        } catch (JsonProcessingException e) {
+        } catch (Throwable e) {
             promptToStartRedisIfLocal(e);
             throw new BridgeServiceException(e);
         }
--- a/app/org/sagebionetworks/bridge/cache/CacheProvider.java
+++ b/app/org/sagebionetworks/bridge/cache/CacheProvider.java
@@ -9,6 +9,7 @@ import org.sagebionetworks.bridge.config.BridgeConfigFactory;
 import org.sagebionetworks.bridge.dynamodb.DynamoStudy;
 import org.sagebionetworks.bridge.exceptions.BridgeServiceException;
 import org.sagebionetworks.bridge.json.BridgeObjectMapper;
+import org.sagebionetworks.bridge.models.accounts.User;
 import org.sagebionetworks.bridge.models.accounts.UserSession;
 import org.sagebionetworks.bridge.models.studies.Study;
 import org.sagebionetworks.bridge.redis.JedisOps;
@@ -17,8 +18,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
-
-import redis.clients.jedis.exceptions.JedisException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 
 /**
  * A wrapper around whatever cache provider we ultimately decide to go with (probably Redis). 
@@ -27,7 +27,13 @@ import redis.clients.jedis.exceptions.JedisException;
 @Component
 public class CacheProvider {
 
+    private ObjectMapper bridgeObjectMapper;
     private JedisOps jedisOps;
+
+    @Autowired
+    public void setBridgeObjectMapper(BridgeObjectMapper bridgeObjectMapper) {
+        this.bridgeObjectMapper = bridgeObjectMapper;
+    }
 
     @Autowired
     public void setJedisOps(JedisOps jedisOps) {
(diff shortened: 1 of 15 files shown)


app/org/sagebionetworks/bridge/cache/CacheProvider.java
```diff
@@ -26,41 +34,72 @@ public void setJedisOps(JedisOps jedisOps) {
...
-            String ser = BridgeObjectMapper.get().writeValueAsString(session);
-            String redisKey = RedisKey.SESSION.getRedisKey(key);
-            String result = jedisOps.setex(redisKey, BridgeConstants.BRIDGE_SESSION_EXPIRE_IN_SECONDS, ser);
+            final String userId = session.getUser().getId();
+            final String userKey = RedisKey.USER_SESSION.getRedisKey(userId);
+            final String sessionToken = session.getSessionToken();
+            final String sessionKey = RedisKey.SESSION.getRedisKey(sessionToken);
+            final String ser = BridgeObjectMapper.get().writeValueAsString(session);
+            results = jedisOps.getTransaction()
+                    .setex(userKey, BridgeConstants.BRIDGE_SESSION_EXPIRE_IN_SECONDS, sessionToken)
```
alxdarksage, line 57: Users can create more than one session, it looks like your implementation will need to check by user ID on sign in to see if the session already exists, and if so, return that instead of creating a new one... ?

DwayneJengSage, line 57: It looks like we already do this:
https://github.com/Sage-Bionetworks/BridgePF/blob/develop/app/controllers/AuthenticationController.java#L29

Eric, can you explicitly test this? Ideally in an integration test, but a manual test will suffice for now.

app/org/sagebionetworks/bridge/cache/CacheProvider.java
```diff
@@ -26,41 +34,72 @@ public void setJedisOps(JedisOps jedisOps) {
...
+            throw new BridgeServiceException(e);
+        }
+    }
+
+    public void removeSessionByUserId(final String userId) {
+        try {
+            final String userKey = RedisKey.USER_SESSION.getRedisKey(userId);
+            final String sessionToken = jedisOps.get(userKey);
+            final String sessionKey = RedisKey.SESSION.getRedisKey(sessionToken);
+            jedisOps.getTransaction().del(sessionKey).del(userKey).exec();
```
DwayneJengSage: What if the user is not logged in? I believe this will actually crash with a NullPointerException.

app/org/sagebionetworks/bridge/cache/CacheProvider.java
```diff
@@ -129,14 +168,16 @@ public void removeString(String cacheKey) {
...
+            promptToStartRedisIfLocal(e);
             throw new BridgeServiceException(e);
         }
     }
 
-    private void promptToStartRedisIfLocalEnv(Throwable e) {
-        if (BridgeConfigFactory.getConfig().isLocal()) {
-            throw new BridgeServiceException("Cannot find cache service, have you started a Redis server? (original message: "+e.getMessage()+")");
+    private void promptToStartRedisIfLocal(Throwable e) {
+        if (e.getClass() == JedisException.class && BridgeConfigFactory.getConfig().isLocal()) {
```
DwayneJengSage: Should this be e instanceof JedisException? This way, subclasses of JedisException are caught by this if.

DwayneJengSage:
Test cases we'll want to test (ideally in an integration test):
- signing in
- signing in when already signed in (returns same session)
- signing out
- signing out when already signed out (shouldn't crash)
- deleting a signed in user (user's session token no longer valid)
- deleting a signed out user (shouldn't crash)

app/org/sagebionetworks/bridge/cache/CacheProvider.java
eric-wu: The code being replaced catches Throwable.  I'm limiting it to JedisException.  Still not ideal but should be less noisy than the original code.

app/org/sagebionetworks/bridge/cache/CacheProvider.java
eric-wu, line 57: Indeed the code that does all this is in `AuthenticationController`:

```
    public Result signIn() throws Exception {
        UserSession session = getSessionIfItExists();
        if (session != null) {
            setSessionToken(session.getSessionToken());
            return okResult(new UserSessionInfo(session));
        }
        try {
            SignIn signIn = parseJson(request(), SignIn.class);
            Study study = getStudyOrThrowException(json);
            session = authenticationService.signIn(study, signIn);
            setSessionToken(session.getSessionToken());
            Result result = okResult(new UserSessionInfo(session));
            return result;
        } catch(ConsentRequiredException e) {
            setSessionToken(e.getUserSession().getSessionToken());
            throw e;
        }
    }
```

I believe we're already testing them.

What's new in my change is that session.getUser() is null or user.getId() is null.  They should be exceptions. What I can do is checking on them and throw exceptions.

alxdarksage, line 57: This returns the session by GUID, unless you've changed it to look up the session by the user's ID. As a consequence, I can sign in multiple times without re-submittin the GUID I got from my other sessions. Then when deleted/signed out, only one of those sessions will be destroyed (the one whose GUID was submitted for the deletion or signing out). If getSessionIfItExists() is first looking up the user's ID from the GUID, then this should no longer happen, there can only be one session for a given user (user ID).

app/org/sagebionetworks/bridge/cache/CacheProvider.java
eric-wu: Actually it also processes JsonProcessingException. I'm removing the check on JedisException.

app/org/sagebionetworks/bridge/cache/CacheProvider.java
eric-wu: Good catch. I am adding a check on null.

app/org/sagebionetworks/bridge/cache/CacheProvider.java
eric-wu, line 57: I see.  Browsing through the code, it looks like the best place to check on this is AuthenticationServiceImpl.getSessionFromAccount() when the user have signed in with Stormpath but haven't created a session yet.

eric-wu:
Updated:

1) More mocked tests on CacheProvider. It now covers all the methods about sessions.
2) Reuse the existing session if it exists.
3) More smoked tests on the services.

Test cases:

1) Signing in -- This is already covered extensively by AuthenticationServiceImplTest
2) Signing in when already signed in -- Covered by AuthenticationServiceImplTest.signInWhenSignedIn(). I added a check to make sure the same session is reused.
3) Signing out -- Added a test AuthenticationServiceImplTest.testSignOut()
4) Signing out when already signed out -- Added a test AuthenticationServiceImplTest.testSignOutWhenSignedOut()
5) Deleting a signed in user -- Covered extensively by StormPathUserAdminServiceTest
6) Deleting a signed out user -- Added a test StormPathUserAdminServiceTest.testDeleteUserWhenSignedOut()

Future work:

1) [BRIDGE-545 Refactor SessionService](https://sagebionetworks.jira.com/browse/BRIDGE-545)
2) [BRIDGE-546 Lock user on signing in](https://sagebionetworks.jira.com/browse/BRIDGE-546)

app/org/sagebionetworks/bridge/cache/CacheProvider.java
```diff
@@ -19,62 +27,115 @@
...
+                throw new BridgeServiceException("Session storage error.");
             }
-        } catch (Throwable e) {
-            promptToStartRedisIfLocalEnv(e);
+            for (Object result : results) {
+                if (!"OK".equals(result)) {
+                    throw new BridgeServiceException("Session storage error.");
+                }
+            }
+        } catch (JsonProcessingException e) {
```
DwayneJengSage: What about JedisException? Do we need to catch that?

app/org/sagebionetworks/bridge/cache/CacheProvider.java
```diff
@@ -19,62 +27,115 @@
...
+            promptToStartRedisIfLocal(e);
             throw new BridgeServiceException(e);
         }
     }
 
-    public UserSession getUserSession(String key) {
+    public UserSession getUserSession(final String sessionToken) {
+        final String sessionKey = RedisKey.SESSION.getRedisKey(sessionToken);
+        final String ser = jedisOps.get(sessionKey);
```
DwayneJengSage: What happens if Jedis throws? Should we leave this inside the try block?

app/org/sagebionetworks/bridge/cache/CacheProvider.java
```diff
@@ -19,62 +27,115 @@
...
+            promptToStartRedisIfLocal(e);
             throw new BridgeServiceException(e);
         }
-        return null;
     }
 
-    public void removeSession(String key) {
+    public UserSession getUserSessionByUserId(final String userId) {
+        final String userKey = RedisKey.USER_SESSION.getRedisKey(userId);
+        final String sessionToken = jedisOps.get(userKey);
```
DwayneJengSage: Do we need a try-catch around this?

app/org/sagebionetworks/bridge/cache/CacheProvider.java
eric-wu: Jedis exceptions will be wrapped into BridgeServiceException by downstream service/controller code.  Behind services and controllers we can just pass whatever exception it is.  Actually does it make sense to remove all the try-catch-rethrow in CacheProvider? What do you think, Alx?

eric-wu: Actually let me restore the try-catch-rethrow blocks.  Don't see we are catching and wrapping exceptions into BridgeServiceException in the downstream code.

test/org/sagebionetworks/bridge/services/StormPathUserAdminServiceTest.java
```diff
@@ -131,6 +131,13 @@ public void cannotCreateUserWithSameUsernameOrEmail() {
...
         }
-        
+    }
+
+    @Test
+    public void testDeleteUserWhenSignedOut() {
+        UserSession session = userAdminService.createUser(signUp, study, true, true);
+        authService.signOut(session);
+        // Shouldn't crash
+        userAdminService.deleteUser(study, session.getUser().getEmail());
```
DwayneJengSage, line 142: Can we validate the user session (or lack thereof) to make sure deleting the user deletes the session?

DwayneJengSage:
Sorry, missed one more. We'll need the same session check in StormPathUserAdminServiceTest.deletedUserHasBeenDeleted()

PR #684, opened 2015-05-27
add upload schema revision to surveys, refactor JSON serialization

DwayneJengSage:
This code change is needed to integrate upload schemas with surveys, with the ultimate goal of being able to support server-side surveys in our upload and data export pipeline. See https://sagebionetworks.jira.com/wiki/display/BRIDGE/Upload+Survey+Integration+Design for more design details.

Testing done:
- unit tests
- integration tests


app/controllers/SurveyController.java
```diff
@@ -190,7 +190,7 @@ public Result createSurvey() throws Exception {
         UserSession session = getAuthenticatedResearcherSession();
         StudyIdentifier studyId = session.getStudyIdentifier();
         
-        Survey survey = DynamoSurvey.fromJson(requestToJSON(request()));
+        Survey survey = parseJson(request(), Survey.class);
```
alxdarksage: Deep in the internals of DynamoSurvey.fromJson(), it uses SurveyElementFactory.fromJson(), and this creates the correct SurveyElement subtype based on the type attribute of the JSON in the SurveyElement record. Without this, you won't have type-safe survey screens vs. questions, and these have different data. I think we need to maintain this factory method.

app/controllers/SurveyController.java
```diff
@@ -190,7 +190,7 @@ public Result createSurvey() throws Exception {
...
         survey.setStudyIdentifier(studyId.getIdentifier());
```
alxdarksage, line 194: (Never mind earlier comment. It looks like you're preserving that factory method for survey elements.)

test/org/sagebionetworks/bridge/models/surveys/TestSurvey.java
```diff
@@ -167,6 +167,7 @@ public TestSurvey(boolean makeNew) {
         setCreatedOn(DateUtils.getCurrentMillisFromEpoch());
         setVersion(2L);
         setPublished(true);
+        setSchemaRevision(42);
```
alxdarksage, line 170: Is schemaRevision editable through the API? It looks like it could be changed and submitted through the UI. Do you want to prevent this?

PR #685, opened 2015-05-28
Upgrade dependencies & use Stormpath client to send verification email

eric-wu:
Mostly minor version upgrades.  Except that the HTTP client has moved to a different Apache group and received a major upgrade.  The only place we are using a HTTP client is  StormpathAccountDao where verification email is sent.  I think I have found the Stormpath client API for doing that and we don't need to code directly against the REST  URL.


app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
```diff
@@ -130,63 +119,16 @@ public Account verifyEmail(StudyIdentifier study, EmailVerification verification
...
-            }
-            String message = node.get("message").asText();
-            if (message.contains("does not match a known resource")) {
-                throw new EntityNotFoundException(Account.class);
-            }
-            throw new ServiceUnavailableException(message);
-        }
+        VerificationEmailRequestBuilder requestBuilder = Applications.verificationEmailBuilder();
+        VerificationEmailRequest request = requestBuilder
+                .setAccountStore(application.getDefaultAccountStore())
```
alxdarksage: It needs to use the directory (account store) for the study.... we can get this, the study has the HREF of that directory.

app/interceptors/ExceptionInterceptor.java
```diff
@@ -72,7 +71,7 @@ private int getStatusCode(final Throwable throwable) {
     private String getMessage(final Throwable throwable, final int status) {
         String message = throwable.getMessage();
         if (StringUtils.isBlank(message)) {
-            message = HttpStatus.getStatusText(status);
+            message = Integer.toString(status);
```
alxdarksage, line 74: No way to get something like "Not Found", only "404" here?

eric-wu, line 74: As far as I know, the exceptions need to store StatusLine (or the reason phrase (StatusLine.getReasonPhrase()) plus status code).  The change is probably not trivial.

alxdarksage, line 74: org.eclipse.jetty.http.HttpStatus.getMessage(int statusCode) works, there's also one in Spring I believe.

eric-wu, line 74: Actually I'm able to remove the dependency to 'httpcore' altogether.  See the updated pull request.  If we use org.eclipse.jetty.http.HttpStatus.getMessage(int statusCode) or the Spring one, we will have to add one more package to the code base.  I am not sure it's worth it.

alxdarksage, line 74: I think those are transitive dependencies that are already there, and don't have to be added (just check Eclipse). But I would not add them just for this if they aren't.

eric-wu, line 74: It's a new dependency to Jetty (owned by Eclipse.org).  The dependency to apache httpcore is transitive.  I didn't check.  It's likely introduced by the AWS client or the Stormpath client.  HttpStatus is still used in our code base.  The version is 4.3.6 not the latest.  But I expect HttpStatus shouldn't change often.

eric-wu, line 74: To be clear, if we use Apache httpcore's HttpStatus, it is transitive, we already have it.  Jetty is a new dependency.  I suspect the reason you see it in Eclipse is because it is used by some Eclipse plugins.

PR #697, opened 2015-06-04
refactor jedis.close into a util class; set testOnBorrow and testOnReturn to true

DwayneJengSage:
These changes work around a Jedis bug that throws exceptions on closing a connection and turns on the connection validation flags in the Jedis config. These are needed to solve the Jedis errors we're seeing in production.

### DIFF
--- a/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
+++ b/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
@@ -199,6 +199,9 @@ public class BridgeSpringConfig {
         // configure Jedis pool
         JedisPoolConfig poolConfig = new JedisPoolConfig();
         poolConfig.setMaxTotal(config.getPropertyAsInt("redis.max.total"));
+        poolConfig.setTestOnBorrow(true);
+        poolConfig.setTestOnReturn(true);
+
         String host = config.getProperty("redis.host");
         int port = config.getPropertyAsInt("redis.port");
         int timeout = config.getPropertyAsInt("redis.timeout");
--- a/app/org/sagebionetworks/bridge/redis/JedisTransaction.java
+++ b/app/org/sagebionetworks/bridge/redis/JedisTransaction.java
@@ -3,10 +3,7 @@ package org.sagebionetworks.bridge.redis;
 import java.util.List;
 
 import redis.clients.jedis.Jedis;
-import redis.clients.jedis.JedisPool;
 import redis.clients.jedis.Transaction;
-import redis.clients.jedis.exceptions.JedisException;
-import redis.clients.util.Pool;
 
 /**
  * Provides abstraction over Jedis's transactions.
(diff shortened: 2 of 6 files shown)


app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
```diff
@@ -199,6 +199,9 @@ public JedisPool jedisPool(BridgeConfig config) {
         // configure Jedis pool
         JedisPoolConfig poolConfig = new JedisPoolConfig();
         poolConfig.setMaxTotal(config.getPropertyAsInt("redis.max.total"));
+        poolConfig.setTestOnBorrow(true);
```
eric-wu, line 202: Not sure of the value of this. The doc says, "Using the provided factory's PoolableObjectFactory.validateObject(T) method".  But what exactly does it do?  What's the performance overhead?  And if the connection is invalid, does it imply a problem somewhere else?

app/org/sagebionetworks/bridge/redis/JedisTransaction.java
```diff
@@ -37,44 +34,32 @@ public JedisTransaction del(final String key) {
     }
 
     public List<Object> exec() {
-        try (Jedis jedis = this.jedis) {
+        try {
             return transaction.exec();
+        } finally {
+            close();
```
eric-wu, line 40: Isn't `close()` in finally block the same as `try (Jedis jedis = this.jedis)`?  It's just syntax sugar, right?

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
DwayneJengSage, line 202: As far as I can tell, it checks that the pool thread has an active connection with the Redis server when it first borrows a connection from the pool and again when it returns the connection to the pool. It's not clear to me what happens if the test fails (although in the worst case, we'd be fast failing here instead of failing later in the Redis call).

Presumably, there is a small overhead, since we'll need to ping the Redis server to get the connection.

After researching possible Redis connection errors, I found this thread https://github.com/xetorthio/jedis/issues/91 which recommends setting both testOnBorrow and testOnReturn. The thread is about 4 years old, so it might be outdated, but there isn't any newer documentation.

I think it's worth trying, since there are still a lot of JedisConnectionExceptions in our logs.

app/org/sagebionetworks/bridge/redis/JedisTransaction.java
DwayneJengSage, line 40: No, the try-with-resources in the old code would call jedis.close(). We need to call this.close(), or we end up bypassing the "could not return resource" check.

Force-King:
I also had this problem, setting TestOnBorrow to true still had an error :

```
redis.clients.jedis.exceptions.JedisException: Could not get a resource from the pool
	at redis.clients.util.Pool.getResource(Pool.java:51)
	at redis.clients.jedis.JedisPool.getResource(JedisPool.java:226)
	at io.codis.jodis.RoundRobinJedisPool.getResource(RoundRobinJedisPool.java:218)
	at com.br.hermes.center.service.impl.IndexServiceV3Impl.initLimitProductMap(IndexServiceV3Impl.java:2780)
	at com.br.hermes.center.quartz.SendMailQuartz.refreshFlowControllConfig(SendMailQuartz.java:608)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.springframework.scheduling.support.ScheduledMethodRunnable.run(ScheduledMethodRunnable.java:65)
	at org.springframework.scheduling.support.DelegatingErrorHandlingRunnable.run(DelegatingErrorHandlingRunnable.java:54)
	at org.springframework.scheduling.concurrent.ReschedulingRunnable.run(ReschedulingRunnable.java:81)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:266)
	at java.util.concurrent.FutureTask.run(FutureTask.java)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.util.NoSuchElementException: Unable to validate object
	at org.apache.commons.pool2.impl.GenericObjectPool.borrowObject(GenericObjectPool.java:506)
	at org.apache.commons.pool2.impl.GenericObjectPool.borrowObject(GenericObjectPool.java:363)
	at redis.clients.util.Pool.getResource(Pool.java:49)
	... 19 more
15:40:01.399 INFO  
```

PR #704, opened 2015-06-08
a few more minor fixes

DwayneJengSage:
- fixed NullPointerException for signIn - If the lock fails, the finally block will try to release the lock for a null lock ID, which throws an NPE. This check prevents that NPE, and the original exception will be propagated.
- UploadValidationBackfill end date of June 5 - The fix was confirmed on June 5. Set the end date to that date instead of "yesterday".
- fixed ResourceException for invalid emails for resendEmailVerification - This is a common pattern in all of the other StormpathAccountDao calls, but we missed it for this one.
- UploadValidationTask - We catch RuntimeExceptions, but not Throwables. Since Throwables (specifically Errors) represent things like OutOfMemoryError, we want to log these as errors so that our logger picks them up. (Long-term, we'll need a way to integrate the exception interceptor with helper thread pools.)

Testing done:
- unit tests for all changed files
- loaded the local server and signed in/out

### DIFF
--- a/app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
+++ b/app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
@@ -131,7 +131,12 @@ public class StormpathAccountDao implements AccountDao {
                 .setAccountStore(directory)
                 .setLogin(email.getEmail())
                 .build();
-        application.sendVerificationEmail(request);
+
+        try {
+            application.sendVerificationEmail(request);
+        } catch (ResourceException e) {
+            rethrowResourceException(e, null);
+        }
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
```diff
@@ -131,7 +131,12 @@ public void resendEmailVerificationToken(StudyIdentifier studyIdentifier, Email
                 .setAccountStore(directory)
                 .setLogin(email.getEmail())
                 .build();
-        application.sendVerificationEmail(request);
+
+        try {
+            application.sendVerificationEmail(request);
+        } catch (ResourceException e) {
+            rethrowResourceException(e, null);
```
alxdarksage, line 138: Awesome, this closes a bug I filled on this.

PR #735, opened 2015-06-16
add record ID to upload validation logging

DwayneJengSage:
When an upload successfully validates, we want the record IDs in the log in addition to the upload ID so we can track its progress through the data pipeline.

Testing done:
- unit tests for changed files
- upload integration tests


app/org/sagebionetworks/bridge/upload/UploadArtifactsHandler.java
```diff
@@ -46,6 +46,7 @@ public void handle(@Nonnull UploadValidationContext context) {
         HealthDataRecordBuilder recordBuilder = context.getHealthDataRecordBuilder();
         HealthDataRecord record = recordBuilder.build();
         String recordId = healthDataService.createOrUpdateRecord(record);
+        context.setRecordId(recordId);
```
alxdarksage, line 49: This should be a recordGuid. In particular, you can migrate this record between data stores without an issue, it's not an auto-generated integer (or what we'd probably call an id). See the table of four different kinds of identifiers we have going on, on the code conventions page.

PR #739, opened 2015-06-19
More refactoring after the Play upgrade

eric-wu:
After this all the Java code is name-spaced under "org.sagebionetworks.bridge" and the code that depend on Play are in the  "org.sagebionetworks.bridge.play.*" packages.

### DIFF
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -24,6 +24,9 @@ play.filters {
     # The X-Content-Type-Options header. If null, the header is not set.
     contentTypeOptions = "nosniff"
 
+    # The X-Permitted-Cross-Domain-Policies header. If null, the header is not set.
+    permittedCrossDomainPolicies = "master-only"
+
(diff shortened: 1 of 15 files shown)


conf/application.conf
```diff
@@ -24,6 +24,9 @@ play.filters {
     # The X-Content-Type-Options header. If null, the header is not set.
     contentTypeOptions = "nosniff"
 
+    # The X-Permitted-Cross-Domain-Policies header. If null, the header is not set.
+    permittedCrossDomainPolicies = "master-only"
```
eric-wu, line 28: The code deployed to staging has the `permittedCrossDomainPolicies` entry removed.  But it is still supplied by the server with the default value.  (Run `curl -i https://webservices-staging.sagebridge.org/` and look at the results.)  I though it would be more explicit if we specify it in `application.conf`.

PR #742, opened 2015-06-20
Lock sign-in on both user name and study ID

eric-wu:
We're seeing 0.5% of the sign-in requests get back ConcurrentModificationException "Lock already set".   We don't see such a high ratio in other places where the distributed lock is used.  I am **_guessing**_ we only lock on user during sign-in while the actual sign-in involves both the user and the study.  I suspect there are users using the same user name for more than one study.  This change is to help us understand why the race conditions occur.  Or, if the race conditions do happen naturally, try to queue up the sign-ins in a very limited manner -- retry at most 3 times with delay of 1 second in between.

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
@@ -4,6 +4,7 @@ import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static org.sagebionetworks.bridge.BridgeConstants.STUDY_PROPERTY;
 
 import org.sagebionetworks.bridge.BridgeConstants;
+import org.sagebionetworks.bridge.exceptions.ConcurrentModificationException;
 import org.sagebionetworks.bridge.exceptions.ConsentRequiredException;
 import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
 import org.sagebionetworks.bridge.json.JsonUtils;
@@ -19,6 +20,7 @@ import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
 import org.sagebionetworks.bridge.models.studies.StudyIdentifierImpl;
 import org.springframework.stereotype.Controller;
 
+import play.Logger;
 import play.mvc.Result;
 
 import com.fasterxml.jackson.databind.JsonNode;
@@ -27,24 +29,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 public class AuthenticationController extends BaseController {
 
     public Result signIn() throws Exception {
-        UserSession session = getSessionIfItExists();
-        if (session != null) {
-            setSessionToken(session.getSessionToken());
-            return okResult(new UserSessionInfo(session));
-        }
-        try {
-            JsonNode json = requestToJSON(request());
-            SignIn signIn = parseJson(request(), SignIn.class);
-            Study study = getStudyOrThrowException(json);
-            
-            session = authenticationService.signIn(study, signIn);
-            setSessionToken(session.getSessionToken());
-            Result result = okResult(new UserSessionInfo(session));
-            return result;
-        } catch(ConsentRequiredException e) {
-            setSessionToken(e.getUserSession().getSessionToken());
-            throw e;
-        }
+        return signInWithRetry(3);
     }
 
     public Result signOut() throws Exception {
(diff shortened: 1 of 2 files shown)


app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
```diff
@@ -102,7 +87,40 @@ public Result resetPassword() throws Exception {
...
+        } catch(ConsentRequiredException e) {
+            setSessionToken(e.getUserSession().getSessionToken());
+            throw e;
+        } catch(ConcurrentModificationException e) {
+            if (retryCounter > 0) {
+                Logger.info("User " + signIn.getUsername() +
+                        " is having a race condition with signing in for study " + study.getIdentifier() + "." +
+                        " Will retry after 1 second.");
+                // controller.signIn() 95% is < 1000 ms 
+                Thread.sleep(1000);
```
DwayneJengSage: 1000ms sleep time seems a little high to me. This means if the initial sign-in fails due to the lock issue, users could be waiting a long time for their sign-in to complete.

The median and avg for signIn is < 500ms, which means on average, the lock would be released after 250ms. So I'd prefer either 250ms sleep time with more retries or 1000ms timeout with only 1 retry.

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
```diff
@@ -102,7 +87,40 @@ public Result resetPassword() throws Exception {
         authenticationService.resetPassword(passwordReset);
         return okResult("Password has been changed.");
     }
-    
+
+    private Result signInWithRetry(final int retryCounter) throws Exception {
```
DwayneJengSage, line 97: Minor: Add comment that retryCounter is the number of retries, excluding the initial try. For example, if retryCounter=3, there will be 4 tries total.

app/org/sagebionetworks/bridge/services/AuthenticationServiceImpl.java
```diff
@@ -112,28 +113,27 @@ public UserSession getSession(String sessionToken) {
...
         checkNotNull(study, "Study cannot be null");
         checkNotNull(signIn, "Sign in cannot be null");
-
         Validate.entityThrowingException(signInValidator, signIn);
+
         String lockId = null;
         try {
-            lockId = lockDao.acquireLock(SignIn.class, signIn.getUsername());
+            lockId = lockDao.acquireLock(SignIn.class,
+                    study.getIdentifier() + RedisKey.SEPARATOR + signIn.getUsername());
```
DwayneJengSage: The default lock timeout is 3 minutes. It's possible that another process acquired but failed to release the lock, which would cause the lock to sit around for a very long time. I would recommend setting the explicitly setting lock timeout timeout to something more reasonable, something in the 1-5 sec range.

I'd also recommend making the same changes for signUp.

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
```diff
@@ -102,7 +87,40 @@ public Result resetPassword() throws Exception {
...
+        if (session != null) {
+            setSessionToken(session.getSessionToken());
+            return okResult(new UserSessionInfo(session));
+        }
+
+        final JsonNode json = requestToJSON(request());
+        final SignIn signIn = parseJson(request(), SignIn.class);
+        final Study study = getStudyOrThrowException(json);
+        try {
+            Logger.info("User " + signIn.getUsername() + " signing in for study " + study.getIdentifier() + ".");
```
DwayneJengSage: Do we need to log this for _every_ sign in? Or just for the ones that throw?

Also, I'm concerned that the user name could be identifying info. Are we able to log these in the logs? Would it be safer to log a hash of the username, so we can see if/when they repeat without having to expose the actual user name?

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
```diff
@@ -102,7 +87,40 @@ public Result resetPassword() throws Exception {
...
+                        " Will retry after 1 second.");
+                // controller.signIn() 95% is < 1000 ms 
+                Thread.sleep(1000);
+                return signInWithRetry(retryCounter - 1);
+            }
+            throw e;
+        }
+
+        setSessionToken(session.getSessionToken());
+        return okResult(new UserSessionInfo(session));
```
DwayneJengSage, line 127: Optional: I feel like this would be cleaner in the try block, so that every execution path has its own block.

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
```diff
@@ -102,7 +87,40 @@ public Result resetPassword() throws Exception {
...
+        final Study study = getStudyOrThrowException(json);
+        try {
+            Logger.info("User " + signIn.getUsername() + " signing in for study " + study.getIdentifier() + ".");
+            session = authenticationService.signIn(study, signIn);
+        } catch(ConsentRequiredException e) {
+            setSessionToken(e.getUserSession().getSessionToken());
+            throw e;
+        } catch(ConcurrentModificationException e) {
+            if (retryCounter > 0) {
+                Logger.info("User " + signIn.getUsername() +
```
DwayneJengSage: Same concern with username in logs.

PR #747, opened 2015-06-23
create upload schema from survey

DwayneJengSage:
Publishing a survey now makes a corresponding upload schema for that survey. This allows surveys and upload schemas to be created dynamically.

Also refactored the survey DAO and survey service tests to not wipe the survey and survey elements table before each unit test, and for the survey tests to clean after themselves.

Lastly, moved the DynamoInitializer calls to BeforeClass instead of Before, which reduces the time to run these tests from 20 min to 8 min.

Testing done:
- ran unit tests
- manually tested creating and publishing a survey, then verifying the resulting schema
- TODO: integration tests


app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
```diff
@@ -304,6 +328,7 @@ public Survey closeSurvey(GuidCreatedOnVersionHolder keys) {
         Survey existing = getSurvey(keys);
         existing.setPublished(false);
         existing.setModifiedOn(DateUtils.getCurrentMillisFromEpoch());
+        existing.setSchemaRevision(null);
```
alxdarksage: A survey that is unpublished may still be referenced in a cache implemented by a client. Is it okay if that survey has no schema associated with it when the user submits answers to that old and potentially unpublished version of the survey? We still accept those answers.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
```diff
@@ -82,6 +86,101 @@ public void setStudyIdIndex(DynamoIndexHelper studyIdIndex) {
...
+            String name = oneQuestion.getIdentifier();
+            UploadFieldType type = getFieldTypeFromConstraints(oneQuestion.getConstraints());
+            UploadFieldDefinition oneFieldDef = new DynamoUploadFieldDefinition.Builder().withName(name)
+                    .withType(type).build();
+            fieldDefList.add(oneFieldDef);
+        }
+
+        // Get the current rev.
+        String studyId = studyIdentifier.getIdentifier();
+        String schemaId = survey.getIdentifier();
```
alxdarksage, line 103: surveyId?

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
```diff
@@ -82,6 +86,101 @@ public void setStudyIdIndex(DynamoIndexHelper studyIdIndex) {
...
+                            // be an attachment.
+                            return UploadFieldType.ATTACHMENT_BLOB;
+                        } else {
+                            // Short string. Can be a string.
+                            return UploadFieldType.STRING;
+                        }
+                    } else {
+                        // Constraints aren't StringConstraints, so we can't determine max length (if any). To be
+                        // safe, make this an attachment.
+                        return UploadFieldType.ATTACHMENT_BLOB;
```
alxdarksage: I believe this is very likely to be short if it's not multi-value, it will be a single boolean, int, decimal, or time/duration/datetime string. I think all would fit in under 100 characters.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
```diff
@@ -82,6 +86,101 @@ public void setStudyIdIndex(DynamoIndexHelper studyIdIndex) {
...
+                // Multiple choice string answers frequently become longer than the 100-char limit. This will have
+                // to be a JSON attachment.
+                return UploadFieldType.ATTACHMENT_JSON_BLOB;
+            } else {
+                // Multiple choice non-string answers (frequently ints) tend to be very short, so this can fit in
+                // an inline_json_blob.
+                return UploadFieldType.INLINE_JSON_BLOB;
+            }
+        } else {
+            switch (constraints.getDataType()) {
```
alxdarksage, line 141: DataType/constraints type should always match, though there's no harm in having fallbacks.

test/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDaoTest.java
```diff
@@ -75,7 +120,9 @@ public void createPreventsEmptyStudyKey() {
 
     @Test(expected = BridgeServiceException.class)
     public void createPreventsRecreatingASurvey() {
-        surveyDao.createSurvey(testSurvey);
+        Survey survey = surveyDao.createSurvey(testSurvey);
+        surveysToDelete.add(new GuidCreatedOnVersionHolderImpl(survey));
```
alxdarksage, line 104: should be able to add survey, it's also a GuidCreatedOnVersionHolder.

test/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDaoTest.java
```diff
@@ -107,8 +181,10 @@ public void crudSurvey() {
     @Test
     public void canUpdateASurveyVersion() {
         Survey survey = surveyDao.createSurvey(testSurvey);
+        surveysToDelete.add(new GuidCreatedOnVersionHolderImpl(survey));
```
alxdarksage, line 165: ditto for all cases

test/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDaoTest.java
```diff
@@ -206,34 +295,81 @@ public void versioningASurveyCopiesTheQuestions() {
...
+        assertEquals(UploadFieldType.TIMESTAMP, fieldDefList.get(2).getType());
+
+        assertEquals("deleuterium_dosage", fieldDefList.get(3).getName());
+        assertEquals(UploadFieldType.FLOAT, fieldDefList.get(3).getType());
+
+        assertEquals("bp_x_day", fieldDefList.get(4).getName());
+        assertEquals(UploadFieldType.INT, fieldDefList.get(4).getType());
+
+        assertEquals("time_for_appt", fieldDefList.get(5).getName());
+        assertEquals(UploadFieldType.STRING, fieldDefList.get(5).getType());
```
alxdarksage, line 313: Parenthetically, I would call the UploadFieldDefinition name an identifier. Sometimes, these things also have "labels" which can be changed for UI presentations without changing the identifier. So name ends up being a little ambiguous.

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
DwayneJengSage: Alx updated the survey API so there's no longer a "close", so this is moot. Once published, surveys can only be versioned (which creates a new survey version entirely) or deleted (which marks the survey version as deleted, but still exists in the database so users can submit answers to it).

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
DwayneJengSage, line 103: Discussed this with Alx over IM. The idea is that the schema ID should be something that researchers can set, and it should match the survey, so survey ID (which is different from survey GUID) is what we want here.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
DwayneJengSage, line 141: Discussed this with Alx over IM. MultiValueConstraints are used for multiple choice questions and can have any data type.

test/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDaoTest.java
DwayneJengSage, line 104: Discussed this with Alx over IM. The problem is that Survey objects are mutable, so if we save a Survey object and it gets changed (which happens in some of the tests), when we try to delete the survey, we can't find it. We use GuildCreatedOnVersionHolderImpl because it's immutable.

test/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDaoTest.java
DwayneJengSage, line 313: Agreed, although renaming this is outside the scope of this changelist.

app/org/sagebionetworks/bridge/dao/SurveyDao.java
```diff
@@ -48,7 +48,16 @@
...
-    
+
+    /**
+     * Admin API to remove the survey from the backing store. This exists to clean up surveys from tests. This will
+     * remove the survey regardless of publish status, whether it has responses. This will delete all survey elements
+     * as well.
+     *
+     * @param keys survey keys (guid, created-on timestamp)
+     */
+    public void deleteSurveyPermanently(GuidCreatedOnVersionHolder keys);
```
alxdarksage, line 59: Thanks, this was bothering me. The tables are filling up with tons and tons of test data

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
```diff
@@ -359,6 +391,17 @@ public Survey getSurvey(GuidCreatedOnVersionHolder keys) {
...
+        // end up with end up with orphaned survey elements.
+        try {
+            surveyMapper.save(survey);
+        } catch(ConditionalCheckFailedException throwable) {
+            throw new ConcurrentModificationException(survey);
+        } catch(Throwable t) {
+            throw new BridgeServiceException(t);
+        }
+
+        // Delete old survey elements (if any) and create new, updated survey elements.
```
alxdarksage: What if you save the survey and then there's an error saving the elements, you'll have a corrupted survey then. Better to have orphaned survey elements I think.

test/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDaoTest.java
alxdarksage, line 313: Just filed a JIRA for this

PR #767, opened 2015-07-10
upload validation backfill to take a list of record IDs from S3

DwayneJengSage:
This is to support https://sagebionetworks.jira.com/browse/BRIDGE-621. The idea is we have some records that were recorded under an old schema. We want to re-validate these and have them appear in the new schema. We already have a list of affected record IDs, saved to a file in S3.

The strategy we take is we take those record IDs, look up their upload IDs, and then kick off upload validation for those uploads again. This is done in a generic way that all we need is a list of record IDs.

In the future, we may want to genericize this into two separate generic backfills (with shared code refactored into a shared utility or parent class), one that takes upload IDs and one that takes record IDs.

Also changes upload validation to write today's date into the record instead of copying from the upload. This way, each object's upload date reflects when it was created.

Testing done:
- manually tested on local dev box
- unit tests for IosSchemaValidationHandler
- upload integration tests

### DIFF
--- a/app/org/sagebionetworks/bridge/services/backfill/UploadValidationBackfill.java
+++ b/app/org/sagebionetworks/bridge/services/backfill/UploadValidationBackfill.java
@@ -106,10 +106,8 @@ public class UploadValidationBackfill extends AsyncBackfillTemplate {
                     HealthDataRecord record = healthDataDao.getRecordById(oneRecordId);
                     String uploadId = record.getUploadId();
 
-                    // Call uploadComplete() to reset the upload status and uploadDate. This will make the upload
-                    // eligible for validation again.
+                    // Get upload.
                     Upload oneUpload = uploadDao.getUpload(uploadId);
-                    uploadDao.uploadComplete(oneUpload);
 
                     // Get study ID from health code. Upload validation needs this.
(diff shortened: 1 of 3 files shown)


app/org/sagebionetworks/bridge/services/backfill/UploadValidationBackfill.java
```diff
@@ -65,48 +81,56 @@ public final void setUploadValidationService(UploadValidationService uploadValid
...
-                recordMessage(task, callback, errMsg);
+                try {
+                    // get upload ID for record ID
+                    HealthDataRecord record = healthDataDao.getRecordById(oneRecordId);
+                    String uploadId = record.getUploadId();
+
+                    // Call uploadComplete() to reset the upload status and uploadDate. This will make the upload
+                    // eligible for validation again.
+                    Upload oneUpload = uploadDao.getUpload(uploadId);
+                    uploadDao.uploadComplete(oneUpload);
```
eric-wu: Would resetting the upload date be Ok? I understand it is needed for automatic redriving. But if we reset it,  if we ever want to find the original upload date, it would be in the S3 object's metadata?

PR #771, opened 2015-07-15
Cardio migration

eric-wu:
Please don't merge.  This is migration script.

Please take a look at UploadDumper.  It has a download() method to download and decrypt staging data (still running after a day).  Then after switching the config to production, the upload() method is called to encrypt and upload data to production.  This write operation is done on production stack.  Please help look for potential risks that may damage the production data.  The uploads are the very original data we collects.  Just want to be safe before I run it.


app/org/sagebionetworks/bridge/migration/HealthDataDumper.java
```diff
@@ -0,0 +1,95 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+class HealthDataDumper {
+
+    public static void main(String[] args) throws FileNotFoundException, IOException {
+        final AbstractApplicationContext appContext = Utils.loadAppContext();
+        HealthDataDumper dataDumper = new HealthDataDumper(Environment.UAT, "heroku", appContext);
+        DateTime start = new DateTime(2015, 7, 7, 0, 0);
+        DateTime end = new DateTime(2015, 7, 10, 0, 0);
```
DwayneJengSage, line 30: I assume these are test dates, because the problem started happening in early June.

app/org/sagebionetworks/bridge/migration/Utils.java
```diff
@@ -0,0 +1,49 @@
...
+        checkNotNull(user);
+        final BridgeConfig config = BridgeConfigFactory.getConfig();
+        if (!config.getEnvironment().equals(env)) {
+            throw new RuntimeException("Wrong environment. Expected environment: " + env.name()
+                    + ". Actual environment: " + config.getEnvironment() + ".");
+        }
+        if (!config.getUser().equals(user)) {
+            throw new RuntimeException("Wrong user. Expected user: " + user
+                    + ". Actual user: " + config.getUser() + ".");
+        }
```
DwayneJengSage, line 39: Does this check mean we have to run this on Heroku in UAT and in Prod? Or is there a way to change your local box to run in staging as heroku?

DwayneJengSage:
UploadDumper looks fine. Should I look at the rest of the stuff?

app/org/sagebionetworks/bridge/migration/HealthDataDumper.java
eric-wu, line 30: This is my 2nd batch.  The first batch started from May 1.

app/org/sagebionetworks/bridge/migration/Utils.java
eric-wu, line 39: Run locally.  The user Heroku is only needed by the DynamoDB table names.

DwayneJengSage:
One potential thought is that guid collision is incredibly unlikely, but is theoretically possible. Is it worth putting a check to see if the guid exists before uploading to S3? Or is this so unlikely that it's not worth protecting against?

eric-wu:
Thanks.  Probably not worth it as I am not sure if the code will be re-used.

eric-wu:
The check against collisions is a good point.  The algorithm is impossible to have collisions.  But the use of the algorithm (e.g. how we run code on actual data like we are doing right now with backfills) is a different story.  I'll add a check.  I am thinking if the check finds a collision we need to decrypt both and compare the contents.

PR #826, opened 2015-08-12
backfill by either upload ID or by record ID

DwayneJengSage:
As it turns out, we do need both. So I refactored this one backfill to handle either upload ID or record ID.

Testing done: Manually tested both record ID and upload ID in my local dev server.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/backfill/UploadValidationByRecordIdBackfill.java
+++ b/app/org/sagebionetworks/bridge/services/backfill/UploadValidationByRecordIdBackfill.java
@@ -1,0 +1,90 @@
+package org.sagebionetworks.bridge.services.backfill;
+
+import com.amazonaws.services.s3.model.S3Object;
+import com.google.common.base.Charsets;
+import com.google.common.io.CharStreams;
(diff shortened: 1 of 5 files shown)


app/org/sagebionetworks/bridge/services/backfill/UploadValidationByRecordIdBackfill.java
```diff
@@ -0,0 +1,90 @@
...
+    @Autowired
+    public void setHealthDataDao(HealthDataDao healthDataDao) {
+        this.healthDataDao = healthDataDao;
+    }
+
+    /** @{inheritDoc} */
+    @Override
+    protected List<String> getUploadIdList(BackfillTask task, BackfillCallback callback) throws IOException {
+        // get file of list of record IDs
+        S3Object recordIdFile = getS3Client().getObject(RECORD_ID_BUCKET, RECORD_ID_FILENAME);
```
eric-wu: L51-56 duplicate the sibling class UploadValidationByUploadIdBackfill L35-39.  How about moving the code to the parent class to a new method that takes a bucket name and file name and returns a list of ids?  The two subclass can both use the same method.

PR #833, opened 2015-08-18
User Data Download API

DwayneJengSage:
API which asynchronously requests user data. This writes a message to SQS, which is being polled by the Bridge User Data Download Service.

Testing done:
- unit tests
- manually tested by uploading data, then calling user data download API

### DIFF
--- a/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
+++ b/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
@@ -4,8 +4,33 @@ import java.util.List;
 import java.util.MissingResourceException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-
 import javax.annotation.Resource;
+
+import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB;
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClient;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper;
+import com.amazonaws.services.s3.AmazonS3Client;
+import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClient;
+import com.amazonaws.services.simpleemail.AmazonSimpleEmailServiceClient;
+import com.amazonaws.services.sqs.AmazonSQSClient;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.LoadingCache;
+import com.google.common.collect.ImmutableList;
+import com.stormpath.sdk.api.ApiKey;
+import com.stormpath.sdk.api.ApiKeys;
+import com.stormpath.sdk.application.Application;
+import com.stormpath.sdk.client.Client;
+import com.stormpath.sdk.client.ClientBuilder;
+import com.stormpath.sdk.client.Clients;
+import com.stormpath.sdk.impl.client.DefaultClientBuilder;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+import redis.clients.jedis.Jedis;
+import redis.clients.jedis.JedisPool;
+import redis.clients.jedis.JedisPoolConfig;
 
 import org.sagebionetworks.bridge.crypto.AesGcmEncryptor;
 import org.sagebionetworks.bridge.crypto.CmsEncryptor;
@@ -32,32 +57,6 @@ import org.sagebionetworks.bridge.upload.TranscribeConsentHandler;
 import org.sagebionetworks.bridge.upload.UnzipHandler;
 import org.sagebionetworks.bridge.upload.UploadArtifactsHandler;
 import org.sagebionetworks.bridge.upload.UploadValidationHandler;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.context.annotation.Bean;
-import org.springframework.context.annotation.ComponentScan;
-import org.springframework.context.annotation.Configuration;
-
-import redis.clients.jedis.Jedis;
-import redis.clients.jedis.JedisPool;
-import redis.clients.jedis.JedisPoolConfig;
-
-import com.amazonaws.auth.BasicAWSCredentials;
-import com.amazonaws.services.dynamodbv2.AmazonDynamoDB;
-import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClient;
-import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper;
-import com.amazonaws.services.s3.AmazonS3Client;
-import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClient;
-import com.amazonaws.services.simpleemail.AmazonSimpleEmailServiceClient;
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.LoadingCache;
-import com.google.common.collect.ImmutableList;
-import com.stormpath.sdk.api.ApiKey;
-import com.stormpath.sdk.api.ApiKeys;
-import com.stormpath.sdk.application.Application;
-import com.stormpath.sdk.client.Client;
-import com.stormpath.sdk.client.ClientBuilder;
-import com.stormpath.sdk.client.Clients;
-import com.stormpath.sdk.impl.client.DefaultClientBuilder;
 
 @ComponentScan({"org.sagebionetworks.bridge"})
 @Configuration
@@ -209,6 +208,12 @@ public class BridgeSpringConfig {
     @Resource(name="awsCredentials")
     public AmazonSimpleEmailServiceClient sesClient(BasicAWSCredentials awsCredentials) {
         return new AmazonSimpleEmailServiceClient(awsCredentials);
+    }
+
+    @Bean(name = "sqsClient")
+    @Resource(name = "awsCredentials")
+    public AmazonSQSClient sqsClient(BasicAWSCredentials awsCredentials) {
+        return new AmazonSQSClient(awsCredentials);
     }
 
     @Bean(name = "asyncExecutorService")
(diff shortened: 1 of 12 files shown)


app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
```diff
@@ -211,6 +210,12 @@ public AmazonSimpleEmailServiceClient sesClient(BasicAWSCredentials awsCredentia
         return new AmazonSimpleEmailServiceClient(awsCredentials);
     }
 
+    @Bean(name = "sqsClient")
+    @Resource(name = "awsCredentials")
+    public AmazonSQSClient sqsClient(BasicAWSCredentials awsCredentials) {
+        return new AmazonSQSClient(awsCredentials);
```
alxdarksage, line 216: I'm not sure what credentials we're using here, but I didn't see SQS permissions on any of the app accounts.

app/org/sagebionetworks/bridge/models/DateRange.java
```diff
@@ -0,0 +1,57 @@
...
+     * DateRange constructor. Also validates that startDate and endDate are specified and that startDate isn't after
+     * endDate.
+     *
+     * @param startDate
+     *         start date, must be non-null
+     * @param endDate
+     *         end date, must be non-null
+     */
+    public DateRange(
+            @Nonnull @JsonProperty("startDate") @JsonDeserialize(using = LocalDateDeserializer.class) LocalDate startDate,
```
alxdarksage: Are the deserializers necessary, they shouldn't be if you use BridgeObjectMapper has the JodaModule registered with it. Do we use it to deserialize JSON in the controllers?

app/org/sagebionetworks/bridge/models/DateRange.java
```diff
@@ -0,0 +1,57 @@
...
+    public @Nonnull LocalDate getStartDate() {
+        return startDate;
+    }
+
+    /** End date of the date range, always non-null. */
+    @JsonSerialize(using = LocalDateToStringSerializer.class)
+    public @Nonnull LocalDate getEndDate() {
+        return endDate;
+    }
+}
```
alxdarksage, line 57: toString() method? Useful for debugging.

app/org/sagebionetworks/bridge/play/controllers/UserDataDownloadController.java
```diff
@@ -0,0 +1,40 @@
...
+
+/** Play controller for User Data Download requests. */
+@Controller
+public class UserDataDownloadController extends BaseController {
+    private UserDataDownloadService userDataDownloadService;
+
+    /** Service handler for User Data Download requests. */
+    @Autowired
+    public void setUserDataDownloadService(
+            UserDataDownloadService userDataDownloadService) {
```
alxdarksage: I think we go to 120 characters per line when formatting

app/org/sagebionetworks/bridge/play/controllers/UserDataDownloadController.java
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Play handler for requesting user data. User must be authenticated and consented. (Otherwise, they couldn't have
+     * any data to download to begin with.)
+     */
+    public Result requestUserData() throws JsonProcessingException {
+        UserSession session = getAuthenticatedAndConsentedSession();
+
+        StudyIdentifier studyIdentifier = session.getStudyIdentifier();
+        User user = session.getUser();
+        DateRange dateRange = parseJson(request(), DateRange.class);
```
alxdarksage, line 34: Yes, we use BridgeObjectMapper, so I am surprised it doesn't use the right deserializers without the annotations.

app/org/sagebionetworks/bridge/play/controllers/UserDataDownloadController.java
```diff
@@ -0,0 +1,40 @@
...
+
+        userDataDownloadService.requestUserData(studyIdentifier, user, dateRange);
+        return okResult("Request submitted.");
```
alxdarksage, line 37: This could/should return 202, "Accepted". The other method that should really do this is the one to send the participant roster via email.

app/org/sagebionetworks/bridge/services/UserDataDownloadViaSqsService.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void requestUserData(@Nonnull StudyIdentifier studyIdentifier, @Nonnull User user,
+            @Nonnull DateRange dateRange) throws JsonProcessingException {
+        String studyId = studyIdentifier.getIdentifier();
+        String username = user.getUsername();
+        String startDateStr = dateRange.getStartDate().toString(ISODateTimeFormat.date());
+        String endDateStr = dateRange.getEndDate().toString(ISODateTimeFormat.date());
```
alxdarksage: It's explicit to use this formatting, but it seems error-prone and the toString() method is already documented to "Output the date time in ISO8601 format (yyyy-MM-dd)." Non-compact ISO 8601 is the standard toString() formatting of everything in Joda Time, and they are not going to change that.

conf/routes
```diff
@@ -34,7 +34,10 @@ POST   /v3/users/self/externalId   	    @org.sagebionetworks.bridge.play.control
 GET    /v3/users/self/unsubscribeEmail  @org.sagebionetworks.bridge.play.controllers.EmailController.unsubscribeFromEmail
 POST   /v3/users/self/unsubscribeEmail  @org.sagebionetworks.bridge.play.controllers.EmailController.unsubscribeFromEmail
 POST   /v3/users/self/dataSharing       @org.sagebionetworks.bridge.play.controllers.ConsentController.changeSharingScope
-          
+
+# User data
+POST   /v3/userdata/request  @org.sagebionetworks.bridge.play.controllers.UserDataDownloadController.requestUserData
```
alxdarksage: Because the result is a zip file of many different types, I'm wondering about treating it as a manipulation of the user:

/v3/users/self/dataExport

/v3/userdata _could_ expand with reports and the like, that would be an argument to keep it separate. At the moment though it's just something you can do to your own account, you can export the data.

app/org/sagebionetworks/bridge/services/UserDataDownloadViaSqsService.java
```diff
@@ -0,0 +1,78 @@
...
+import org.sagebionetworks.bridge.models.DateRange;
+import org.sagebionetworks.bridge.models.accounts.User;
+import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
+
+/**
+ * Implementation of {@link UserDataDownloadService} that connects to the Bridge User Data Download Service via SQS.
+ */
+@Component
+public class UserDataDownloadViaSqsService implements UserDataDownloadService {
+    private static final Logger logger = LoggerFactory.getLogger(SendMailViaAmazonService.class);
```
eric-wu: typo: SendMailViaAmazonService.class

app/org/sagebionetworks/bridge/models/DateRange.java
```diff
@@ -0,0 +1,57 @@
...
+            throw new IllegalArgumentException("startDate must be specified");
+        }
+
+        if (endDate == null) {
+            throw new IllegalArgumentException("endDate must be specified");
+        }
+
+        if (startDate.isAfter(endDate)) {
+            throw new IllegalArgumentException("startDate can't be after endDate");
+        }
```
alxdarksage, line 36: These go straight back to the client, so I think these should be EntityInvalidExceptions. This is a very simple object so I guess you don't need a validator, but to if you use the InvalidEntityException(entity, message, map) constructor, it's the same difference.

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
DwayneJengSage, line 216: Good catch. I've added the permissions to our app IAM accounts

app/org/sagebionetworks/bridge/models/DateRange.java
DwayneJengSage: I didn't realize we had updated BridgeObjectMapper to make these serializers and deserializers unnecessary. Removed.

app/org/sagebionetworks/bridge/models/DateRange.java
DwayneJengSage, line 57: Added.

app/org/sagebionetworks/bridge/play/controllers/UserDataDownloadController.java
DwayneJengSage: Not sure what happened here. This is fixed.

app/org/sagebionetworks/bridge/play/controllers/UserDataDownloadController.java
DwayneJengSage, line 34: Fixed. See above.

app/org/sagebionetworks/bridge/play/controllers/UserDataDownloadController.java
DwayneJengSage, line 37: I agree, but there's no Play method to return 202 accepted, and rigging one up seems superfluous.

app/org/sagebionetworks/bridge/services/UserDataDownloadViaSqsService.java
DwayneJengSage: Removed here and in other places.

conf/routes
DwayneJengSage: After discussing this with Alx over IM, we aligned on /v3/users/self/emailData

app/org/sagebionetworks/bridge/services/UserDataDownloadViaSqsService.java
DwayneJengSage: Good catch. Fixed.

app/org/sagebionetworks/bridge/models/DateRange.java
DwayneJengSage, line 36: I think using a validator would be overkill for this class, since it's a super simple class that never needs to become more complicated.

Changed the exceptions to InvalidEntityException.

PR #835, opened 2015-08-27
Upload Validation Strict Validation

DwayneJengSage:
Reviving and refactoring the Strict Validation code to fit in with our upload validation handler chain. Added unit tests.

If a field has the wrong type or if a required field is missing, this will log a warning. In the future, or for new studies, we could configure this to fail upload validation.

Testing done:
- relevant unit tests
- manually tested upload validation
- ran integration tests to make sure existing Upload behavior wasn't broken

### DIFF
--- a/app/org/sagebionetworks/bridge/upload/StrictValidationHandler.java
+++ b/app/org/sagebionetworks/bridge/upload/StrictValidationHandler.java
@@ -1,0 +1,235 @@
+package org.sagebionetworks.bridge.upload;
+
+import javax.annotation.Nonnull;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.base.Joiner;
+import org.joda.time.DateTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import org.sagebionetworks.bridge.json.DateUtils;
+import org.sagebionetworks.bridge.models.healthdata.HealthDataRecordBuilder;
+import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
+import org.sagebionetworks.bridge.models.upload.UploadFieldDefinition;
+import org.sagebionetworks.bridge.models.upload.UploadFieldType;
+import org.sagebionetworks.bridge.models.upload.UploadSchema;
+import org.sagebionetworks.bridge.services.UploadSchemaService;
+
+/**
+ * <p>
+ * Upload validation handler that handles strict schema validation. Specifically, it checks to make sure the record
+ * data contains all the fields marked as "required" are present. This guards against (a) apps sending "useless" data
+ * that's missing key fields and (b) apps changing filenames and fieldnames causing Bridge to no longer see the
+ * incoming data.
+ * </p>
+ * <p>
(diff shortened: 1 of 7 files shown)


app/org/sagebionetworks/bridge/upload/UploadUtil.java
```diff
@@ -0,0 +1,46 @@
...
+            timestampStr = timestampStr.replaceAll("\\s+", "");
+
+            // Log something, so we can keep track of how often this happens.
+            logger.info("Non-standard timestamp in upload data: " + timestampStr);
+        }
+
+        try {
+            return DateUtils.parseISODateTime(timestampStr);
+        } catch (RuntimeException ex) {
+            return null;
```
eric-wu, line 44: Should we log a warning here?

app/org/sagebionetworks/bridge/upload/StrictValidationHandler.java
```diff
@@ -0,0 +1,233 @@
...
+                // For attachment types, since they just get exported as raw files, we only need to check if it's
+                // required and present. Specifically, if it's required and it's not present, then that's an error.
+                if (isRequired && !attachmentFieldNameSet.contains(fieldName)) {
+                    errorList.add("Required attachment field " + fieldName + " missing");
+                }
+            } else {
+                JsonNode fieldValueNode = recordDataNode.get(fieldName);
+
+                if (fieldValueNode != null && !fieldValueNode.isNull()) {
+                    boolean isValid = validateJsonField(fieldType, fieldValueNode);
```
eric-wu, line 162: trivial: Does it justify a boolean variable isValid? It is used only once right after it's initialized.

app/org/sagebionetworks/bridge/upload/StrictValidationHandler.java
```diff
@@ -0,0 +1,233 @@
...
+            case TIMESTAMP:
+                // either it's a string in ISO format, or it's a long in epoch milliseconds
+                if (fieldValueNode.isTextual()) {
+                    DateTime dateTime = UploadUtil.parseIosTimestamp(fieldValueNode.textValue());
+                    return dateTime != null;
+                } else if (fieldValueNode.isIntegralNumber()) {
+                    // any integral value can be converted to an epoch milliseconds, so this is good no
+                    // matter what
+                    return true;
+                } else {
```
eric-wu, line 224: trivial: this last else block is unnecessary.

app/org/sagebionetworks/bridge/upload/UploadUtil.java
DwayneJengSage, line 44: Added a warning to the catch. Also, updated the other log message to a warning.

app/org/sagebionetworks/bridge/upload/StrictValidationHandler.java
DwayneJengSage, line 162: Functionally, it's all the same in byte code.

Stylistically, I prefer that if there's a function call that does a lot of processing, it shouldn't be mixed with other things like if statements. I'd prefer to leave this as is, unless you have strong feelings about this.

app/org/sagebionetworks/bridge/upload/StrictValidationHandler.java
DwayneJengSage, line 224: Unless the code is identical (like in FLOAT vs INT), I prefer not to use fall-through in switch statements.

eric-wu, line 224: I mean you will return false but it does not have to be in a else block.

PR #836, opened 2015-08-27
Upload Validation parse Survey Guid and CreatedOn

DwayneJengSage:
This is needed so that apps can tell the server what version of the survey they are answering.

I also noticed an obscure edge case in DateUtils, where if you pass in an int with 4 digits or less (like 42), it'll parse that int as the year, and you end up with something dumb, like midnight Jan 1 in year 42. So that obscure edge case has been patched.

Testing done:
- relevant unit tests
- manually tested upload with survey Guid and CreatedOn in info.json
- ran UploadTest integration test to verify Upload still works

### DIFF
--- a/app/org/sagebionetworks/bridge/upload/IosSchemaValidationHandler2.java
+++ b/app/org/sagebionetworks/bridge/upload/IosSchemaValidationHandler2.java
@@ -19,22 +19,26 @@ import com.google.common.collect.ImmutableSet;
 import org.apache.commons.lang3.StringUtils;
 import org.joda.time.DateTime;
 import org.joda.time.LocalDate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 import org.sagebionetworks.bridge.BridgeConstants;
 import org.sagebionetworks.bridge.dao.HealthDataDao;
-import org.sagebionetworks.bridge.exceptions.BadRequestException;
-import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
 import org.sagebionetworks.bridge.json.BridgeObjectMapper;
 import org.sagebionetworks.bridge.json.DateUtils;
+import org.sagebionetworks.bridge.json.JsonUtils;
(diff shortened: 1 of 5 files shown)


app/org/sagebionetworks/bridge/upload/IosSchemaValidationHandler2.java
```diff
@@ -165,41 +180,56 @@ public void handle(@Nonnull UploadValidationContext context)
...
+        // Old versions of YML apps sometimes send "identifier" instead of "item". This isn't officially supported,
+        // but we have to do it anyway for backwards compatibility.
+        String identifier = JsonUtils.asText(infoJson, KEY_IDENTIFIER);
+
+        if (StringUtils.isNotBlank(surveyGuid) && StringUtils.isNotBlank(surveyCreatedOn)) {
+            // First try getting the schema by survey.
+            return getUploadSchemaBySurvey(study, surveyGuid, surveyCreatedOn);
+        } else if (StringUtils.isNotBlank(item)) {
+            // Fall back to item field.
+            return getUploadSchemaByItemAndRev(study, item, schemaRev);
```
alxdarksage, line 202: Do you always have schemaRev at this point?

app/org/sagebionetworks/bridge/upload/IosSchemaValidationHandler2.java
```diff
@@ -165,41 +180,56 @@ public void handle(@Nonnull UploadValidationContext context)
...
-            }
-            schemaRev = schemaRevNode.intValue();
+    private UploadSchema getUploadSchemaBySurvey(StudyIdentifier study, String surveyGuid, String surveyCreatedOn)
+            throws UploadValidationException {
+        // surveyCreatedOn is a timestamp. SurveyService takes long epoch millis. Convert.
+        long surveyCreatedOnMillis= DateUtils.convertToMillisFromEpoch(surveyCreatedOn);
+
+        // Get survey. We use the survey identifier as the schema ID and the schema revision. Both of these must be
+        // specified.
+        Survey survey = surveyService.getSurvey(new GuidCreatedOnVersionHolderImpl(surveyGuid, surveyCreatedOnMillis));
```
alxdarksage, line 220: If this doesn't retrieve a survey it throws an EntityNotFoundException, do you want to translate to a UVE?

PR #839, opened 2015-08-31
Use 'bridge-base' for config and crypto

eric-wu:


app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
```diff
@@ -113,8 +117,8 @@ public void run() {
 
     @Bean(name = "healthCodeEncryptor")
     @Resource(name = "bridgeConfig")
-    public AesGcmEncryptor healthCodeEncryptor(BridgeConfig bridgeConfig) {
-        return new AesGcmEncryptor(bridgeConfig.getHealthCodeKey());
+    public BridgeAesGcmEncryptor healthCodeEncryptor(BridgeConfig bridgeConfig) {
```
DwayneJengSage: If we're going to be making use of interfaces, the return type should be BridgeEncryptor, not BridgeAesGcmEncryptor.

app/org/sagebionetworks/bridge/crypto/BridgeAesGcmEncryptor.java
```diff
@@ -0,0 +1,30 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class BridgeAesGcmEncryptor implements BridgeEncryptor {
+
+    private static final Integer VERSION = Integer.valueOf(2);
+
+    private final AesGcmEncryptor encryptor;
+
+    public BridgeAesGcmEncryptor(final AesGcmEncryptor encryptor) {
```
DwayneJengSage: If we're going to be making use of interfaces, this should take an Encryptor, not an AesGcmEncryptor.

app/org/sagebionetworks/bridge/crypto/BridgeAesGcmEncryptor.java
```diff
@@ -0,0 +1,30 @@
+package org.sagebionetworks.bridge.crypto;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class BridgeAesGcmEncryptor implements BridgeEncryptor {
```
DwayneJengSage: Can we genericize BridgeAesGcmEncryptor and just have it take an Encryptor and a version number? This way, if we need to introduce a version 3, assuming version 3 fits into the Encryptor interface, we don't have to update BridgeAesGcmEncryptor, just swap out the internal representation. (And if it doesn't fit in the Encryptor interface, we'd have to make a code change either way.)

This way, we can combine BridgeAesGcmEncryptor with the BridgeEncryptor interface. This reduces the amount of hardcoding, reduces the amount of code duplication, reduces the amount of changes we have to make when we update the encryptor, and reduces the overall complexity of the system, and we'd still be using the Encryptor interface to separate the API from the implementation.

PR #870, opened 2015-10-13
per-study config for Strict Upload Validation

DwayneJengSage:
Add flag to Study for turning Strict Validation on or off (defaults to off).

Testing done:
- unit tests
- manually tested upload with no flag (defaults to false) and with flag set to true
- UploadTest integration tests

### DIFF
--- a/test/org/sagebionetworks/bridge/dynamodb/DynamoStudyTest.java
+++ b/test/org/sagebionetworks/bridge/dynamodb/DynamoStudyTest.java
@@ -53,12 +53,14 @@ public class DynamoStudyTest {
         assertEquals(study.getMaxNumOfParticipants(), node.get("maxNumOfParticipants").asInt());
         assertEquals(study.getStormpathHref(), node.get("stormpathHref").asText());
         assertEquals(study.getPasswordPolicy(), JsonUtils.asEntity(node, "passwordPolicy", PasswordPolicy.class));
-        assertEquals(study.getVerifyEmailTemplate(), JsonUtils.asEntity(node, "verifyEmailTemplate", EmailTemplate.class));
+        assertEquals(study.getVerifyEmailTemplate(), JsonUtils.asEntity(node, "verifyEmailTemplate",
+                EmailTemplate.class));
         assertEquals(study.getResetPasswordTemplate(), JsonUtils.asEntity(node, "resetPasswordTemplate", EmailTemplate.class));
         assertEquals(study.getUserProfileAttributes(), JsonUtils.asStringSet(node, "userProfileAttributes"));
-        assertEquals(study.getConsentHTML(), JsonUtils.asText(node,  "consentHTML"));
-        assertEquals(study.getConsentPDF(), JsonUtils.asText(node,  "consentPDF"));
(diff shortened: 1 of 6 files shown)


test/org/sagebionetworks/bridge/dynamodb/DynamoStudyTest.java
```diff
@@ -53,12 +53,14 @@ public void studyFullySerializesForCaching() throws Exception {
         assertEquals(study.getMaxNumOfParticipants(), node.get("maxNumOfParticipants").asInt());
         assertEquals(study.getStormpathHref(), node.get("stormpathHref").asText());
         assertEquals(study.getPasswordPolicy(), JsonUtils.asEntity(node, "passwordPolicy", PasswordPolicy.class));
-        assertEquals(study.getVerifyEmailTemplate(), JsonUtils.asEntity(node, "verifyEmailTemplate", EmailTemplate.class));
+        assertEquals(study.getVerifyEmailTemplate(), JsonUtils.asEntity(node, "verifyEmailTemplate",
+                EmailTemplate.class));
```
alxdarksage, line 57: I don't know if we ever discussed line width... I'm set to 120 which seems to work even on the laptop, what are you using? What's our standard? Also I'm assuming 4 spaces per tab is our standard. We should probably write down at https://sagebionetworks.jira.com/wiki/display/BRIDGE/Code+Conventions

test/org/sagebionetworks/bridge/dynamodb/DynamoStudyTest.java
```diff
@@ -79,6 +81,9 @@ public void studyFullySerializesForCaching() throws Exception {
         study = BridgeObjectMapper.get().readValue(json, DynamoStudy.class);
         assertNull(study.getStormpathHref());
         assertEquals("Study", node.get("type").asText());
+
+        // Deserialize back to a POJO and verify.
+        DynamoStudy deserStudy = BridgeObjectMapper.get().readValue(json, DynamoStudy.class);
```
alxdarksage: If we're going to have Task.create() we should follow through with things like Study.create() and .readValue(json, Study.class), but that's a separate PR.

alxdarksage:
This all looks good. Just a couple of comments not related to changes for
this PR, so we can merge whenever.

Alx

test/org/sagebionetworks/bridge/dynamodb/DynamoStudyTest.java
DwayneJengSage, line 57: Also 120 lines (which this line is definitely longer than). That said, I have no idea how the auto-format got triggered...

test/org/sagebionetworks/bridge/dynamodb/DynamoStudyTest.java
DwayneJengSage: I figured it was okay for DynamoStudyTest to know about DynamoStudy. That said, for this specific line, we deserializing as a Study instead of a DynamoStudy works just fine, so I went ahead and changed it.

Also, I realized the test as written re-uses variables a lot, which means the last test for deserialization actually doesn't do anything useful. So I slightly refactored the test.

PR #888, opened 2015-10-27
duplicate field names in Schemas

DwayneJengSage:
Researchers can create schemas with duplicate field names. This breaks as you can't have duplicate column names in Synapse (and is also logically bad). Updated the Schema Validator to fail on duplicate field names.

Testing done:
- Schema Validator unit test
- Manual test creating a schema with duplicate field names and verifying that it fails

We'll also need to verify and delete the broken schemas in Staging.

### DIFF
--- a/test/org/sagebionetworks/bridge/validators/UploadSchemaValidatorTest.java
+++ b/test/org/sagebionetworks/bridge/validators/UploadSchemaValidatorTest.java
@@ -352,6 +352,6 @@ public class UploadSchemaValidatorTest {
         } catch (InvalidEntityException ex) {
             thrownEx = ex;
         }
-        assertTrue(thrownEx.getMessage().contains("foo-field"));
+        assertTrue(thrownEx.getMessage().contains("cannot use foo-field (used by another field)"));
     }
 }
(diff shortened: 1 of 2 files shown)


test/org/sagebionetworks/bridge/validators/UploadSchemaValidatorTest.java
```diff
@@ -334,5 +326,32 @@ public void requiresFieldDefinitionHasType() throws Exception {
...
+
+        // validate
+        Exception thrownEx = null;
+        try {
+            Validate.entityThrowingException(UploadSchemaValidator.INSTANCE, schema);
+            fail("expected exception");
+        } catch (InvalidEntityException ex) {
+            thrownEx = ex;
+        }
+        assertTrue(thrownEx.getMessage().contains("foo-field"));
```
alxdarksage: Can you verify the message is correct and is understandable, e.g. UploadSchema is invalid: name 'foo-field' is used for more than one field definition?

alxdarksage: Also, probably need change to documentation

DwayneJengSage: As per IM conversion, changed error message to "cannot use fieldName (used by another field)" and changed unit tests to check for that string.

Updated https://sagebionetworks.jira.com/wiki/display/BRIDGE/UploadSchema and https://sagebionetworks.jira.com/wiki/display/BRIDGE/Bridge+Upload+Data+Format#BridgeUploadDataFormat-Schemas to say "A schema cannot contain fields with the same name."

PR #894, opened 2015-11-03
Delete activities on update

alxdarksage:
- schedule activities from a schedule plan are now deleted when that schedule plan is updated or deleted
- added global secondary index on the schedulePlanGuid (now an SA field) to retrieve them for deletion
- modified JSON serialization so the SecondaryIndexHelper would work as-is (uses Jackson so @JsonIgnore needed to be removed from key fields), using a writer to output in API without healthCode, etc. being exposed
- removed "ref" from Activity (deprecated and not used anywhere).
- tests revealed that schedule activity was not serializing DateTimes using the submitted time zone, we need to use our custom serializers/deserializers to ensure this happens (so added);


app/org/sagebionetworks/bridge/dynamodb/DynamoScheduledActivity.java
```diff
@@ -88,22 +95,26 @@ public void setTimeZone(DateTimeZone zone) {
...
 
     @Override
     @DynamoDBIgnore
+    @JsonSerialize(using = DateTimeSerializer.class)
     public DateTime getExpiresOn() {
         return getInstant(getLocalExpiresOn());
     }
 
     @Override
+    @JsonDeserialize(using = DateTimeDeserializer.class)
```
DwayneJengSage: We currently don't have unit tests to deserialize scheduledOn or expiresOn. If we need this, we should test this. If we don't need this, we should remove JsonDeserialize and replace it with JsonIgnore.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -252,6 +254,36 @@ public void changePublishedAndAbsoluteSurveyActivity() {
...
     }
     
+    @Test
+    public void deleteScheduledActivitiesForUser() {
+        service.deleteActivitiesForUser("AAA");
+        
+        ArgumentCaptor<String> argument = ArgumentCaptor.forClass(String.class);
+        verify(activityDao).deleteActivitiesForUser(argument.capture());
+        
+        assertEquals("AAA", argument.getValue());
```
DwayneJengSage: Another place where you don't need an ArgumentCaptor

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -252,6 +254,36 @@ public void changePublishedAndAbsoluteSurveyActivity() {
...
+    }
+    
+    @Test
+    public void deleteScheduledActivitiesForSchedulePlan() {
+        service.deleteActivitiesForSchedulePlan("BBB");
+        
+        ArgumentCaptor<String> argument = ArgumentCaptor.forClass(String.class);
+        verify(activityDao).deleteActivitiesForSchedulePlan(argument.capture());
+        
+        assertEquals("BBB", argument.getValue());
```
DwayneJengSage: Another place where you don't need an ArgumentCaptor

app/org/sagebionetworks/bridge/dynamodb/DynamoScheduledActivity.java
alxdarksage: Okay, the deserialize values are not needed, these are never deserialized. They are used when calculating dates during scheduling, and the local value is stored separately from the time zone (which is set separately). Removed the annotations.

PR #895, opened 2015-11-03
Defects

alxdarksage:
- Update user's session to be non-consented when consent is withdrawn;
- Do not recreate an enrollment event when a user re-enrolls in a study. This causes one-time activities scheduled at the time of enrollment to be duplicated. We don't want these to happen again when re-enrolling.

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
@@ -79,6 +79,7 @@ public class ConsentController extends BaseController {
         final long withdrewOn = DateTime.now().getMillis();
         
         consentService.withdrawConsent(study, session.getUser(), withdrawal, withdrewOn);
+        updateSessionUser(session, session.getUser());
         
         return okResult("User has been withdrawn from the study.");
     }
--- a/app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
+++ b/app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
@@ -208,8 +208,7 @@ public class ConsentServiceImpl implements ConsentService {
         String externalId = optionsService.getOption(user.getHealthCode(), ParticipantOption.EXTERNAL_IDENTIFIER);
         MimeTypeEmailProvider consentEmail = new WithdrawConsentEmailProvider(study, externalId, user, withdrawal, withdrewOn);
         sendMailService.sendEmail(consentEmail);
-        
-        System.out.println("YES DID GET HERE");
+
         user.setConsent(false);
     }
     
--- a/app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
+++ b/app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
@@ -203,11 +203,13 @@ public class ConsentServiceImpl implements ConsentService {
         decrementStudyEnrollment(study);
 
         optionsService.setOption(study, user.getHealthCode(), SharingScope.NO_SHARING);
+        user.setSharingScope(SharingScope.NO_SHARING);
         
         String externalId = optionsService.getOption(user.getHealthCode(), ParticipantOption.EXTERNAL_IDENTIFIER);
         MimeTypeEmailProvider consentEmail = new WithdrawConsentEmailProvider(study, externalId, user, withdrawal, withdrewOn);
         sendMailService.sendEmail(consentEmail);
         
+        System.out.println("YES DID GET HERE");
         user.setConsent(false);
     }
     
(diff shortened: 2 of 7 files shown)


app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
```diff
@@ -79,6 +79,7 @@ public Result withdrawConsent() throws Exception {
         final long withdrewOn = DateTime.now().getMillis();
         
         consentService.withdrawConsent(study, session.getUser(), withdrawal, withdrewOn);
+        updateSessionUser(session, session.getUser());
```
DwayneJengSage, line 82: Minor: As far as I can tell, consentService.withdrawConsent() sets user.consent to false, but doesn't update the sharing scope on the user session. While this has no functional impact, it does mean that the user sessions returned by our HTTP interface will incorrectly state the user's sharing scope.

alxdarksage, line 82: I'll add that to make sure the session reflects reality.

test/org/sagebionetworks/bridge/play/controllers/ConsentControllerMockedTest.java
```diff
@@ -163,8 +167,14 @@ public void canWithdrawConsent() throws Exception {
...
         assertEquals("User has been withdrawn from the study.", node.get("message").asText());
         
+        // Should call the service and withdraw
         verify(consentService).withdrawConsent(any(Study.class), any(User.class), captor.capture(), any(Long.class));
         assertEquals("Because, reasons.", captor.getValue().getReason());
+        
+        // Should update the user's session to be non-consented
+        ArgumentCaptor<UserSession> sessionCaptor = ArgumentCaptor.forClass(UserSession.class);
+        verify(cacheProvider).setUserSession(sessionCaptor.capture());
+        assertFalse(sessionCaptor.getValue().getUser().doesConsent());
```
DwayneJengSage: This doesn't actually test that the user session was updated. The user session is a mock object with no override on doesConsent(), so it always returns false regardless.

A better thing to do is to create a stub User object that we can update and then either (a) move the user session management to ConsentController.withdrawConsent() so you can test it here or (b) move it to ConsentServiceImpl and move the test there as well.

alxdarksage: Moving this to the service test where the service is not mocked but the user is, and verifying the correct methods are called on the user object.

test/org/sagebionetworks/bridge/play/controllers/ConsentControllerMockedTest.java
```diff
@@ -163,6 +166,7 @@ public void canWithdrawConsent() throws Exception {
         JsonNode node = BridgeObjectMapper.get().readTree(response);
         assertEquals("User has been withdrawn from the study.", node.get("message").asText());
         
+        // Should call the service and withdraw
         verify(consentService).withdrawConsent(any(Study.class), any(User.class), captor.capture(), any(Long.class));
         assertEquals("Because, reasons.", captor.getValue().getReason());
     }
```
DwayneJengSage, line 173: Clarification: I think we should still test that we call cacheProvider.setUserSession(), otherwise we have no automated way to test that.

app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
```diff
@@ -203,11 +203,13 @@ public void withdrawConsent(Study study, User user, Withdrawal withdrawal, long
         decrementStudyEnrollment(study);
 
         optionsService.setOption(study, user.getHealthCode(), SharingScope.NO_SHARING);
+        user.setSharingScope(SharingScope.NO_SHARING);
         
         String externalId = optionsService.getOption(user.getHealthCode(), ParticipantOption.EXTERNAL_IDENTIFIER);
         MimeTypeEmailProvider consentEmail = new WithdrawConsentEmailProvider(study, externalId, user, withdrawal, withdrewOn);
         sendMailService.sendEmail(consentEmail);
         
+        System.out.println("YES DID GET HERE");
```
DwayneJengSage: Minor: remove println()

test/org/sagebionetworks/bridge/play/controllers/ConsentControllerMockedTest.java
```diff
@@ -163,8 +168,11 @@ public void canWithdrawConsent() throws Exception {
         JsonNode node = BridgeObjectMapper.get().readTree(response);
         assertEquals("User has been withdrawn from the study.", node.get("message").asText());
         
+        // Should call the service and withdraw
         verify(consentService).withdrawConsent(any(Study.class), any(User.class), captor.capture(), any(Long.class));
         assertEquals("Because, reasons.", captor.getValue().getReason());
+        
+        verify(cacheProvider).setUserSession(session);
```
DwayneJengSage, line 172: Not sure this does what you think it does. You have two cacheProviders, one at the class level, one local to the method. This is a little confusing.

alxdarksage, line 172: Removed method-specific one.

PR #896, opened 2015-11-04
Study enumerations

alxdarksage:
- adds study identifier enumeration to study;
- adds data group enumeration to study;
- validate that studies only reference declared task identifiers;
- cleans up static imports and redundant declarations of StudyIdentifiers.


app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
```diff
@@ -187,6 +189,30 @@ public void setUserProfileAttributes(Set<String> profileAttributes) {
     }
 
     /** {@inheritDoc} */
+    @DynamoDBMarshalling(marshallerClass = StringSetMarshaller.class)
```
DwayneJengSage: Do we really need the StringSetMarshaller? I know for sure that DDB natively supports StringSets. I've seen code examples of DDB mappers using StringSets without needing any custom marshallers.

app/org/sagebionetworks/bridge/models/schedules/ABTestScheduleStrategy.java
```diff
@@ -125,7 +126,7 @@ public void validate(Errors errors) {
                 return;
             } else {
                 errors.pushNestedPath("schedule");
-                new ScheduleValidator().validate(group.getSchedule(), errors);
+                new ScheduleValidator(taskIdentifiers).validate(group.getSchedule(), errors);
                 errors.popNestedPath();
             }
         }
```
DwayneJengSage: This loop is entirely redundant with the previous loop.

DwayneJengSage:
Tangentially related to this code review: SimpleScheduleStrategy.getAllPossibleSchedules() will throw a NullPointerException if schedule is null. Not sure if that's ever relevant, but since SimpleScheduleStrategy isn't self-validating, it's possible.

[ALX: added NPE check]

app/org/sagebionetworks/bridge/models/schedules/Activity.java
```diff
@@ -145,9 +141,10 @@ public Activity build() {
             if (guid == null) {
                 guid = BridgeUtils.generateGuid();
             }
-            Activity activity = new Activity(label, labelDetail, guid, task, survey, response);
-            Validate.entityThrowingException(VALIDATOR, activity);
-            return activity;
+            // Activity is validated as part of a schedule, and requires task identifiers 
+            // for the study being validated. We can't validate again, here, during construction 
+            // without adding taskIdentifiers to the builder, which is strange.
+            return new Activity(label, labelDetail, guid, task, survey, response);
```
DwayneJengSage, line 144: I think there's still a lot of value in making Activity self-validating, as far as internal validation goes. This way, we can ensure basic things, like label is non-null, etc.

For validating we have valid task IDs, it would be better to have that as a separate step.

app/org/sagebionetworks/bridge/services/SchedulePlanServiceImpl.java
```diff
@@ -59,23 +55,29 @@ public SchedulePlan getSchedulePlan(StudyIdentifier studyIdentifier, String guid
...
 
     @Override
-    public SchedulePlan createSchedulePlan(SchedulePlan plan) {
+    public SchedulePlan createSchedulePlan(Study study, SchedulePlan plan) {
         checkNotNull(plan);
-        
+
+        // Plan must always be in user's study
+        plan.setStudyKey(study.getIdentifier());
+
```
DwayneJengSage, line 64: Might be worth adding a unit test to where the user tries to create a schedule in the wrong study, and we set the correct study.

From a semantics point of view, would it be better to silently set the correct study or throw a 400 Bad Request?

app/org/sagebionetworks/bridge/services/SchedulePlanServiceImpl.java
```diff
@@ -59,23 +55,29 @@ public SchedulePlan getSchedulePlan(StudyIdentifier studyIdentifier, String guid
...
     
     @Override
-    public SchedulePlan updateSchedulePlan(SchedulePlan plan) {
+    public SchedulePlan updateSchedulePlan(Study study, SchedulePlan plan) {
         checkNotNull(plan);
         
-        Validate.entityThrowingException(validator, plan);
+        // Plan must always be in user's study
+        plan.setStudyKey(study.getIdentifier());
+
```
DwayneJengSage, line 81: Similarly here?

app/org/sagebionetworks/bridge/validators/ActivityValidator.java
```diff
@@ -47,10 +57,12 @@ private void validate(Errors errors, TaskReference ref) {
         errors.pushNestedPath("task");
         if (isBlank(ref.getIdentifier())) {
             errors.rejectValue("identifier", CANNOT_BE_BLANK);
+        } else if (taskIdentifiers == null || !taskIdentifiers.contains(ref.getIdentifier())) {
+            errors.rejectValue("identifier", geTaskIdentifierMessage(ref));
```
DwayneJengSage: Minor: Would it be cleaner to have the null check in the constructor, and then substitute it with an empty set? Functionality should be the same, but I feel like it would be cleaner since it moves the null check closer to the source.

app/org/sagebionetworks/bridge/validators/ActivityValidator.java
```diff
@@ -67,4 +79,13 @@ private void validate(Errors errors, SurveyResponseReference ref) {
         errors.popNestedPath();
     }
     
+    private String geTaskIdentifierMessage(TaskReference ref) {
```
DwayneJengSage: Nitpick: Typo "ge" instead of "get". Unless getThing -> geThing is a thing now.

app/org/sagebionetworks/bridge/validators/StudyValidator.java
```diff
@@ -13,9 +13,14 @@
 import org.springframework.validation.Errors;
 import org.springframework.validation.Validator;
 
+import com.newrelic.agent.deps.com.google.common.base.Joiner;
```
DwayneJengSage: I think there's a merge error here. Pretty sure we removed new relic from our codebase.

test/org/sagebionetworks/bridge/validators/StudyValidatorTest.java
```diff
@@ -210,5 +214,31 @@ public void cannotSetMaxNumOfParticipantsLessThanZero() {
...
+    public void shortListOfDataGroupsOK() {
+        study.setDataGroups(Sets.newHashSet("beta_users", "production_users", "testers", "internal"));
+        Validate.entityThrowingException(StudyValidator.INSTANCE, study);
+    }
+    
+    @Test
+    public void longListOfDataGroupsInvalid() {
+        study.setDataGroups(newLinkedHashSet("Antwerp", "Ghent", "Charleroi", "Liege", "Brussels-City", "Bruges", "Schaerbeek", "Anderlecht", "Namur", "Leuven", "Mons", "Molenbeek-Saint-Jean"));
+        assertCorrectMessage(study, "dataGroups", "dataGroups will not export to Synapse (string is over 100 characters: 'Antwerp, Ghent, Charleroi, Liege, Brussels-City, Bruges, Schaerbeek, Anderlecht, Namur, Leuven, Mons, Molenbeek-Saint-Jean')");
+    }
```
DwayneJengSage, line 227: Might be easier if we move MAX_SYNAPSE_LENGTH to a member method, then mock (spy) it out to a smaller value, so our tests are less cumbersome.

app/org/sagebionetworks/bridge/validators/StudyValidator.java
```diff
@@ -79,6 +84,7 @@ public void validate(Object obj, Errors errors) {
         validateEmails(errors, study.getSupportEmail(), "supportEmail");
         validateEmails(errors, study.getTechnicalEmail(), "technicalEmail");
         validateEmails(errors, study.getConsentNotificationEmail(), "consentNotificationEmail");
+        validateDataGroupNamesAndFitForSynapseExport(errors, study.getDataGroups());
```
DwayneJengSage, line 87: Do we also want to validate taskIdentifiers? Or are those entirely unrestricted?

test/org/sagebionetworks/bridge/services/SchedulePlanServiceTest.java
```diff
@@ -29,40 +33,50 @@
...
             assertEquals(baseCount, plans.size());
             
             // Finally, this is outside the upper-bound, so the plan will not be returned
             clientInfo = ClientInfo.fromUserAgentCache("TestApp/16 TestSDK/40");
-            plans = schedulePlanService.getSchedulePlans(clientInfo, new StudyIdentifierImpl("test-study"));
+            plans = schedulePlanService.getSchedulePlans(clientInfo, study);
             assertEquals(baseCount-1, plans.size());
             
+        } catch(Exception e) {
+            e.printStackTrace();
```
DwayneJengSage: This will actually turn test failures into successes. If the problem is the finally block masking test failures, we could move the finally blocks into an @After method, an JUnit _should_ show the errors for both.

(canRoundTripSchedulePlan() doesn't fit the pattern, but that test really should be moved to SchedulePlanTest.)

DwayneJengSage: Incidentally, it might be worth null-checking savedPlan in the finally blocks. I suspect that's why the finally blocks are throwing.

test/org/sagebionetworks/bridge/validators/StudyValidatorTest.java
```diff
@@ -210,5 +214,31 @@ public void cannotSetMaxNumOfParticipantsLessThanZero() {
...
+        assertCorrectMessage(study, "dataGroups", "dataGroups contains invalid tag 'Liège' (only letters, numbers, underscore and dash allowed)");
+    }
+    
+    private Set<String> newLinkedHashSet(String... list) {
+        Set<String> set = new LinkedHashSet<>();
+        for (String element : list) {
+            set.add(element);
+        }
+        return set;
+    }
```
DwayneJengSage: Minor: Might I suggest a TreeSet instead? TreeSets are naturally ordered, and the TreeSet constructor takes a collection, so you won't need this helper method.

app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
alxdarksage: In short, it's needed. Without it and no other modifications, you get an error: 
[error] Test org.sagebionetworks.bridge.dynamodb.DynamoStudyDaoTest.crudOneStudy failed: com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMappingException: Expected SS in value {S: ["phone","can_be_recontacted"],} when invoking public void org.sagebionetworks.bridge.dynamodb.DynamoStudy.setUserProfileAttributes(java.util.Set), took 0.123 sec
[error]     at com.amazonaws.services.dynamodbv2.datamodeling.unmarshallers.SSUnmarshaller.typeCheck(SSUnmarshaller.java:28)

I'm not sure what it wants, it may simply need this to work with sets rather than lists... ?

app/org/sagebionetworks/bridge/models/schedules/ABTestScheduleStrategy.java
alxdarksage: Removed

app/org/sagebionetworks/bridge/services/SchedulePlanServiceImpl.java
alxdarksage, line 64: I have been silently setting healthCode and studyKey to the correct values as these shouldn't be exposed in the API, and even if someone tries to force it with a JSON payload, we should just undo any attempt to write to a different study or a different account. This also makes it easier to copy something between projects later, if we do that.

app/org/sagebionetworks/bridge/validators/ActivityValidator.java
alxdarksage: Changed

app/org/sagebionetworks/bridge/validators/ActivityValidator.java
alxdarksage: Fixed

app/org/sagebionetworks/bridge/validators/StudyValidator.java
alxdarksage: Yeah. Fixed.

test/org/sagebionetworks/bridge/validators/StudyValidatorTest.java
alxdarksage, line 227: This works for me, but if we switch to 1,000 characters, I'm going to create a lot of dummy values rather than hard-code them.

app/org/sagebionetworks/bridge/validators/StudyValidator.java
alxdarksage, line 87: They seem to be unrestricted. An example: 7-APHSpatialSpanMemory-4A04F3D0-AC05-11E4-AB27-0800200C9A66.

test/org/sagebionetworks/bridge/services/SchedulePlanServiceTest.java
alxdarksage: This was left-over debugging code, removed it, the test passes

test/org/sagebionetworks/bridge/validators/StudyValidatorTest.java
alxdarksage: Altered so there's no helper method.

DwayneJengSage:
I just did a test and confirmed that the StringSetMarshaller is _not_ needed. The reason why your previous test failed is because userProfileAttributes is already being serialized as a String, so naturally, we can't deserialize it as a String Set.

We should remove the StringSetMarshaller from taskIdentifiers and dataGroups because:
1. It adds an extra unnecessary step.
2. If we ever need to access Study configuration without using a mapper, using String Set directly saves us the trouble of having to deserialize JSON.

app/org/sagebionetworks/bridge/models/schedules/ABTestScheduleStrategy.java
```diff
@@ -113,22 +114,11 @@ public void validate(Errors errors) {
                 return;
```
DwayneJengSage: I just noticed that this errors.popNestedPath() and return means the validator skips out on the first null schedule instead of validating all the schedules in the list.

DwayneJengSage:
Can we add unit tests for create and update schedule plan to verify that the user specified study ID is ignored and replaced with the correct study ID?

[ALX] I did for the service.

app/org/sagebionetworks/bridge/validators/ActivityValidator.java
```diff
@@ -67,4 +80,13 @@ private void validate(Errors errors, SurveyResponseReference ref) {
         errors.popNestedPath();
     }
     
+    private String getTaskIdentifierMessage(TaskReference ref) {
+        String message = "'" + ref.getIdentifier() + "' is not in enumeration: ";
+        if (taskIdentifiers != null && !taskIdentifiers.isEmpty()) {
```
DwayneJengSage: Minor: We can remove this null check, since it will never be hit.

app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
alxdarksage: Removed for two new sets.

app/org/sagebionetworks/bridge/models/schedules/ABTestScheduleStrategy.java
alxdarksage: Yikes! Fixed this.

app/org/sagebionetworks/bridge/services/SchedulePlanServiceImpl.java
alxdarksage, line 64: Added tests to verify study identifier is set

app/org/sagebionetworks/bridge/services/SchedulePlanServiceImpl.java
alxdarksage, line 81: Test added.

test/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDaoTest.java
```diff
@@ -41,7 +41,7 @@
     
     private StudyIdentifier studyIdentifier;
     
-    private Set<String> plansToDelete;
+    private Set<SchedulePlan> plansToDelete;
```
DwayneJengSage: This is a bad idea. SchedulePlans are mutable, so the HashSet won't work as expected. Furthermore, if we ever introduce a test that changes the schedule plan guid, this test will break in mysterious ways. It's much much much better to use a Set of String of guids.

PR #899, opened 2015-11-06
Support for FPHS' enrollment codes

alxdarksage:
- APIs to validate and register one of these codes
- APIs to add codes and retrieve all codes, along with information as to whether or not each code has been consumed.


conf/routes
```diff
@@ -105,6 +105,13 @@ POST   /v3/backfill/:name/start    @org.sagebionetworks.bridge.play.controllers.
 GET    /v3/cache           @org.sagebionetworks.bridge.play.controllers.CacheAdminController.listItems
 DELETE /v3/cache/:cacheKey @org.sagebionetworks.bridge.play.controllers.CacheAdminController.removeItem(cacheKey: String)
 
+# FOOTBALL PLAYERS HEATH STUDY API --------------------------------------------------------------------------
+
+GET    /fphs/externalId   @org.sagebionetworks.bridge.play.controllers.FPHSController.verifyExternalIdentifier(externalId: String ?= null)
```
DwayneJengSage, line 110: Minor: From a REST perspective, I feel like more of a verb action (verifying an existing externalId) vs a noun action (getting that externalId), so I feel like POST /fphs/externalId/verify?externalId=blah would be better. But I'll leave it to your judgement call.

app/org/sagebionetworks/bridge/play/controllers/FPHSController.java
```diff
@@ -0,0 +1,63 @@
...
+
+import play.mvc.Result;
+
+@Controller("fphsController")
+public class FPHSController extends BaseController {
+    
+    private static final TypeReference<ArrayList<FPHSExternalIdentifier>> externalIdTypeRef = 
+            new TypeReference<ArrayList<FPHSExternalIdentifier>>() {};
+
+    private static final BridgeObjectMapper mapper = BridgeObjectMapper.get();
```
DwayneJengSage: Minor: Conventionally, constants (static finals) should use UPPER_CASE_AND_UNDERSCORES instead of camelCase. So these should be called EXTERNAL_ID_LIST_TYPE_REF and MAPPER.

Also, it should be EXTERNAL_ID_LIST_TYPE_REF instead of just EXTERNAL_ID_TYPE_REF

Also, I don't know if there's a functional difference, but I feel like List would be better than ArrayList, and just let Jackson pick the list implementation.

test/org/sagebionetworks/bridge/play/controllers/FPHSControllerTest.java
```diff
@@ -0,0 +1,215 @@
...
+    public void verifyFailsWhenNull() throws Exception {
+        when(fphsService.verifyExternalIdentifier(any())).thenThrow(new InvalidEntityException("ExternalIdentifier cannot be blank, null or missing."));
+        
+        try {
+            controller.verifyExternalIdentifier(null);
+            fail("Should have thrown exception");
+        } catch(InvalidEntityException e) {
+            assertEquals("ExternalIdentifier cannot be blank, null or missing.", e.getMessage());
+        }
+    }
```
DwayneJengSage, line 129: Minor: From the Controller's perspective, verifyFails() and verifyFailsWhenNull() are identical. Do we need separate Controller tests for these cases?

app/org/sagebionetworks/bridge/play/controllers/FPHSController.java
```diff
@@ -0,0 +1,63 @@
...
+    @Autowired
+    public final void setFPHSService(FPHSService service) {
+        this.fphsService = service; 
+    }
+    
+    public Result verifyExternalIdentifier(String externalId) throws Exception {
+        // public API, no restrictions. externalId can be null so we can create a 400 error in the service.
+        ExternalIdentifier identifier = new ExternalIdentifier(externalId);
+        fphsService.verifyExternalIdentifier(identifier);
+        return okResult("External identifier is valid.");
```
DwayneJengSage, line 38: API mismatch: The Service and Dao return false if the external ID is invalid. The controller seems to expect the service/dao to throw, but it doesn't, so this will always return OK.

test/org/sagebionetworks/bridge/services/FPHSServiceTest.java
```diff
@@ -0,0 +1,118 @@
...
+    public void verifyExternalIdentifierTrue() throws Exception {
+        when(dao.verifyExternalId(externalId)).thenReturn(true);
+        assertTrue(service.verifyExternalIdentifier(externalId));
+    }
+    
+    @Test
+    public void registerExternalIdentifier() throws Exception {
+        service.registerExternalIdentifier(TEST_STUDY, "BBB", externalId);
+        verify(dao).registerExternalId(externalId);
+    }
```
DwayneJengSage, line 74: Test should also verify optionsService call.

test/org/sagebionetworks/bridge/services/FPHSServiceTest.java
```diff
@@ -0,0 +1,118 @@
...
+                    externalId.getIdentifier());
+            verifyNoMoreInteractions(optionsService);
+        }
+    }
+    
+    @Test
+    public void getExternalIdentifiers() throws Exception {
+        service.getExternalIdentifiers();
+        verify(dao).getExternalIds();
+    }
```
DwayneJengSage, line 116: Test should verify getExternalIds() return value.

app/org/sagebionetworks/bridge/models/accounts/FPHSExternalIdentifier.java
```diff
@@ -0,0 +1,23 @@
...
+
+import org.sagebionetworks.bridge.dynamodb.DynamoFPHSExternalIdentifier;
+import org.sagebionetworks.bridge.json.BridgeTypeName;
+import org.sagebionetworks.bridge.models.BridgeEntity;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+@BridgeTypeName("ExternalIdentifier")
+@JsonDeserialize(as=DynamoFPHSExternalIdentifier.class)
+public interface FPHSExternalIdentifier extends BridgeEntity {
```
DwayneJengSage, line 11: Add unit test for serializing/deserializing JSON?

app/org/sagebionetworks/bridge/models/accounts/FPHSExternalIdentifier.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public static FPHSExternalIdentifier create(String externalId) {
+        return new DynamoFPHSExternalIdentifier(externalId);
+    }
+    
+    public String getExternalId();
+    public void setExternalId(String externalId);
+    
+    public boolean getRegistered();
```
DwayneJengSage: Minor: Should this be isRegistered() instead of getRegistered()? (You can use a Jackson annotation to map it to the JSON attribute, if Jackson doesn't pick it up on its own.)

app/org/sagebionetworks/bridge/dao/FPHSExternalIdentifierDao.java
```diff
@@ -0,0 +1,22 @@
...
+    
+    public void registerExternalId(ExternalIdentifier externalId);
+    
+    public void unregisterExternalId(ExternalIdentifier externalId);
+    
+    public List<FPHSExternalIdentifier> getExternalIds();
+    
+    public void addExternalIds(List<FPHSExternalIdentifier> externalIds);
+    
+    public void deleteAll();
```
DwayneJengSage: While this is less likely to come up, I'd strongly prefer a deleteById() call instead of a deleteAll() call. This way, if someone wants to test the external ID stuff locally, they can do that without interfering with the unit tests and vice versa.

test/org/sagebionetworks/bridge/dynamodb/DynamoFPHSExternalIdentifierDaoTest.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @After
+    public void after() {
+        dao.deleteAll();
+    }
+    
+    @Test
+    public void verifyExternalIdDoesNotExist() throws Exception {
+        ExternalIdentifier externalId = new ExternalIdentifier("abc");
```
DwayneJengSage: It might be worthwhile generating a random prefix to the external IDs for this test. This way, if for some reason a previous test run fails to cleanup, or if a dev is locally testing external IDs and just happens to use the same external ID, this won't cause spurious failures.

test/org/sagebionetworks/bridge/dynamodb/DynamoFPHSExternalIdentifierDaoTest.java
```diff
@@ -0,0 +1,133 @@
...
+    
+    @Test
+    public void registerValidCodeOK() throws Exception {
+        DynamoFPHSExternalIdentifier id1 = new DynamoFPHSExternalIdentifier("baz");
+        dao.addExternalIds(Lists.newArrayList(id1));
+        
+        dao.registerExternalId(new ExternalIdentifier("baz"));
+        
+        assertTrue(dao.getExternalIds().get(0).getRegistered());
+    }
```
DwayneJengSage, line 107: If we make getById() available to the unit tests, we could verify for sure that "baz" is registered without depending on the external ID list being empty.

app/org/sagebionetworks/bridge/dynamodb/DynamoFPHSExternalIdentifierDao.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+    
+    @Override
+    public void unregisterExternalId(ExternalIdentifier externalId) {
+        checkNotNull(externalId);
+        
+        DynamoFPHSExternalIdentifier record = getExternalId(externalId, true);
+        record.setRegistered(false);
+        mapper.save(record);
+    }
```
DwayneJengSage, line 59: We should add unit tests for unregister() with all the same cases as register(). Reading through the code, it looks like this wouldn't work, because getExternalId() throws if the ID is already registered.

app/org/sagebionetworks/bridge/dynamodb/DynamoFPHSExternalIdentifierDao.java
```diff
@@ -0,0 +1,112 @@
...
+
+    @Override
+    public List<FPHSExternalIdentifier> getExternalIds() {
+        DynamoDBScanExpression scan = new DynamoDBScanExpression();
+        
+        return mapper.scan(DynamoFPHSExternalIdentifier.class, scan).stream().map(identifier -> {
+            return (FPHSExternalIdentifier)identifier;
+        }).collect(Collectors.toList());
```
DwayneJengSage, line 67: Minor: Streaming APIs are nice and all, but I'm wondering if this would be simpler / more readable if you created a List of FPHSExternalIdentifier and just called List.addAll() on the mapper.scan() results?

test/org/sagebionetworks/bridge/dynamodb/DynamoFPHSExternalIdentifierDaoTest.java
```diff
@@ -0,0 +1,133 @@
...
+    
+    @Test
+    public void getExternalIds() throws Exception {
+        DynamoFPHSExternalIdentifier id1 = new DynamoFPHSExternalIdentifier("bar");
+        DynamoFPHSExternalIdentifier id2 = new DynamoFPHSExternalIdentifier("baz");
+        dao.addExternalIds(Lists.newArrayList(id1, id2));
+        dao.registerExternalId(new ExternalIdentifier(id1.getExternalId()));
+        
+        List<FPHSExternalIdentifier> identifiers = dao.getExternalIds();
+        assertEquals(2, identifiers.size());
```
DwayneJengSage: To make this less dependent on the initial state of the FPHSExternalId table, would it be better to check that identifiers.size() is >= 2, and check that the list contains "bar" (registered) and "baz" (not registered)?

test/org/sagebionetworks/bridge/dynamodb/DynamoFPHSExternalIdentifierDaoTest.java
```diff
@@ -0,0 +1,133 @@
...
+        
+        // Add a third ID, you will have three IDs, none are registered
+        id1 = new DynamoFPHSExternalIdentifier("bar");
+        id2 = new DynamoFPHSExternalIdentifier("baz");
+        DynamoFPHSExternalIdentifier id3 = new DynamoFPHSExternalIdentifier("boo");
+        dao.addExternalIds(Lists.newArrayList(id1, id2, id3));
+        
+        // Two updates by primary key, and an add for 3 records
+        List<FPHSExternalIdentifier> identifiers = dao.getExternalIds();
+        assertEquals(3, identifiers.size());
```
DwayneJengSage: Similarly here.

app/org/sagebionetworks/bridge/dynamodb/DynamoFPHSExternalIdentifierDao.java
```diff
@@ -0,0 +1,112 @@
...
+                BridgeUtils.ifFailuresThrowException(failures);
+            }
+        }
+    }
+    
+    @Override
+    public void deleteAll() {
+        List<DynamoFPHSExternalIdentifier> identifiers = getExternalIds().stream().map(id -> {
+            return new DynamoFPHSExternalIdentifier(id.getExternalId());
+        }).collect(Collectors.toList());
```
DwayneJengSage: As written, this involves redundantly converting a List of DynamoFPHSExternalIdentifier to List of FPHSExternalIdentifier and back. Would it be better to factor out the scan and have deleteAll() use the mapper.scan() result directly?

conf/routes
alxdarksage, line 110: Eric Sieg had the opposite opinion (he wanted GET) and I didn't feel strongly about it, so I switched it from a POST. It's idempotent so it's okay to be a GET.

app/org/sagebionetworks/bridge/play/controllers/FPHSController.java
alxdarksage: Changed to all caps. Changed ArrayList to List (it might be the case that Jackson needs a concrete implementation, we'll see).

test/org/sagebionetworks/bridge/play/controllers/FPHSControllerTest.java
alxdarksage, line 129: I think these are separate. In one case the user has sent an invalid identifier, the controller should pass that along (and not catch the exception that results). If the user does not pass the parameter, the controller gets a null, and I do want to test that the controller does exactly the same thing, it passes it along and returns the exception.

app/org/sagebionetworks/bridge/play/controllers/FPHSController.java
alxdarksage, line 38: Asking Eric what he wants here...

alxdarksage, line 38: Okay, switched to throw exceptions and translate these into 200, 404, or 409 responses as appropriate.

test/org/sagebionetworks/bridge/services/FPHSServiceTest.java
alxdarksage, line 74: Added

test/org/sagebionetworks/bridge/services/FPHSServiceTest.java
alxdarksage, line 116: Added

app/org/sagebionetworks/bridge/models/accounts/FPHSExternalIdentifier.java
alxdarksage, line 11: Added

app/org/sagebionetworks/bridge/models/accounts/FPHSExternalIdentifier.java
alxdarksage: Changed (I don't have strong opinions on this one).

app/org/sagebionetworks/bridge/dao/FPHSExternalIdentifierDao.java
alxdarksage: Changed

test/org/sagebionetworks/bridge/dynamodb/DynamoFPHSExternalIdentifierDaoTest.java
alxdarksage: Added

test/org/sagebionetworks/bridge/dynamodb/DynamoFPHSExternalIdentifierDaoTest.java
alxdarksage, line 107: This works as a test. Do we really need this?

app/org/sagebionetworks/bridge/dynamodb/DynamoFPHSExternalIdentifierDao.java
alxdarksage, line 59: Adding roundtrip test of unregistering.

app/org/sagebionetworks/bridge/dynamodb/DynamoFPHSExternalIdentifierDao.java
alxdarksage, line 67: Personally, I want to use this everywhere I would use a map or forEach in another language like Ruby or JavaScript. It's three lines either way:

```
    List<FPHSExternalIdentifier> results = Lists.newArrayList();
    results.addAll(mapper.scan(DynamoFPHSExternalIdentifier.class, scan));
    return results;
```

test/org/sagebionetworks/bridge/dynamodb/DynamoFPHSExternalIdentifierDaoTest.java
alxdarksage: Changed

test/org/sagebionetworks/bridge/dynamodb/DynamoFPHSExternalIdentifierDaoTest.java
alxdarksage: Again changed (we already pull out specific results by ID to verify they are somewhere in the list)

app/org/sagebionetworks/bridge/dynamodb/DynamoFPHSExternalIdentifierDao.java
alxdarksage: Removed this as requested above.

app/org/sagebionetworks/bridge/dynamodb/DynamoFPHSExternalIdentifierDao.java
```diff
@@ -0,0 +1,109 @@
...
+    @Override
+    public void unregisterExternalId(ExternalIdentifier externalId) {
+        checkNotNull(externalId);
+        
+        DynamoFPHSExternalIdentifier record = getExternalId(externalId, false);
+        if (record != null) {
+            record.setRegistered(false);
+            mapper.save(record);
+        }
+    }
```
DwayneJengSage, line 59: We need unit tests to verify this bug fix. What happens when you unregister an external ID that doesn't exist? What happens when you unregister an external ID that's already registered?

test/org/sagebionetworks/bridge/services/FPHSServiceTest.java
```diff
@@ -0,0 +1,120 @@
...
+    @Test(expected = InvalidEntityException.class)
+    public void validateIdThrowsException() throws Exception {
+        service.verifyExternalIdentifier(new ExternalIdentifier(""));
+    }
+    
+    @Test
+    public void verifyExternalIdentifierSucceeds() throws Exception {
+        service.verifyExternalIdentifier(externalId);
+        verify(dao).verifyExternalId(externalId);
+    }
```
DwayneJengSage, line 55: Is it worth adding a unit test for verify() fails? (An actual verify failure, not an empty string check.)

alxdarksage: The DAO test covers the failure scenario, here I just wanted to verify that we call the DAO as we expect. I can add a failure case if you want. It should just throw the exception right through the service.

alxdarksage, line 55: Added another test of this.

PR #925, opened 2015-11-12
Health code export

alxdarksage:
- study now has a setting as to whether or not health code should be exported as part of the participant roster;
- this setting is false by default and can only be set to true by an administrator
- the roster includes a "health code" column if set to true;
- when a study is created, strict validation is enabled by default (minor fix)

### DIFF
--- a/app/org/sagebionetworks/bridge/services/email/ParticipantRosterProvider.java
+++ b/app/org/sagebionetworks/bridge/services/email/ParticipantRosterProvider.java
@@ -81,6 +81,9 @@ public class ParticipantRosterProvider implements MimeTypeEmailProvider {
         for (String attribute : study.getUserProfileAttributes()) {
             append(sb, StringUtils.capitalize(attribute), true);
         }
+        if (study.isHealthCodeExportEnabled()) {
+            append(sb, "Health Code", true);
+        }
         sb.append(NEWLINE);
         for (int i=0; i < participants.size(); i++) {
             StudyParticipant participant = participants.get(i);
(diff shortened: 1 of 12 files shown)


app/org/sagebionetworks/bridge/services/email/ParticipantRosterProvider.java
```diff
@@ -81,6 +81,9 @@ String createParticipantTSV() {
         for (String attribute : study.getUserProfileAttributes()) {
             append(sb, StringUtils.capitalize(attribute), true);
         }
+        if (study.isHealthCodeExportEnabled()) {
+            append(sb, "Health Code", true);
```
syoung-smallwisdom, line 85: is hardcoding intentional here?

syoung-smallwisdom:
Looks good. 

One question about a "magic string" where I would expect to see a static final keyword. 
Also, didn't see a test for the equality check changed in the DynamoStudy object. Is that already included in different tests?

app/org/sagebionetworks/bridge/services/email/ParticipantRosterProvider.java
alxdarksage, line 85: It is to be consistent, all of these header fields are strings and none are repeated anywhere other than this CSV method, so constants didn't seem necessary.

syoung-smallwisdom, line 85: Only place this is used in a CSV export file. Seems cleaner to use the string inline rather than obfuscate things by pulling it out to a static string at the top.

PR #927, opened 2015-11-13
Email when roster done

alxdarksage:
- Roster is prepared on a different thread, it now emails a systems operations email when that process succeeds or fails so we get visibility into this without having to assign ourselves to the list of consent notification emails (or look through the logs).
- Cleaned up the MimeTypeEmailProvider interface... it's now allowed to leave the sender blank and if the sender is blank, the sendMailService uses a default. No need to pass this value into every provider.
- Made the roster export resilient to accounts that have not consented. At least my account in development supposedly had a signature, but not a healthCode. That won't fail at this point.

### DIFF
--- a/app/org/sagebionetworks/bridge/models/studies/StudyParticipant.java
+++ b/app/org/sagebionetworks/bridge/models/studies/StudyParticipant.java
@@ -45,7 +45,9 @@ public final class StudyParticipant extends HashMap<String,String> {
         return (emptyString == null) ? null : Boolean.valueOf(emptyString);
     }
     public void setNotifyByEmail(Boolean notifyByEmail) {
-        put(UserProfile.NOTIFY_BY_EMAIL_FIELD, notifyByEmail.toString());
+        if (notifyByEmail != null) {
+            put(UserProfile.NOTIFY_BY_EMAIL_FIELD, notifyByEmail.toString());    
+        }
     }
     public String getHealthCode() {
         return getEmpty(UserProfile.HEALTH_CODE_FIELD);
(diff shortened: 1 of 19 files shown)


test/org/sagebionetworks/bridge/services/SendMailViaAmazonServiceConsentTest.java
```diff
@@ -78,25 +77,6 @@ public void setUp() throws Exception {
         studyConsentService = mock(StudyConsentService.class);
         when(studyConsentService.getActiveConsent(any(StudyIdentifier.class))).thenReturn(view);
     }
-    
-    @Test
```
alxdarksage, line 82: This should not happen. Study supportEmail is required and present for all studies, we never want to send these from our default email address.

app/org/sagebionetworks/bridge/models/studies/StudyParticipant.java
```diff
@@ -45,7 +45,9 @@ public Boolean getNotifyByEmail() {
         return (emptyString == null) ? null : Boolean.valueOf(emptyString);
     }
     public void setNotifyByEmail(Boolean notifyByEmail) {
-        put(UserProfile.NOTIFY_BY_EMAIL_FIELD, notifyByEmail.toString());
+        if (notifyByEmail != null) {
+            put(UserProfile.NOTIFY_BY_EMAIL_FIELD, notifyByEmail.toString());    
+        }
```
syoung-smallwisdom, line 50: Why would you get a call to this method with a null Boolean? Additionally, should it PUT "false" if the variable is null rather than do nothing?

syoung-smallwisdom:
looks good

app/org/sagebionetworks/bridge/models/studies/StudyParticipant.java
alxdarksage, line 50: Yes this could be called with null (in an unusual user state, probably only an artifact of testing and not something that happens in production), and we wouldn't consider null the equivalent of being false, so we'd just leave it out of the map entirely.

PR #928, opened 2015-11-16
Data groups for users

alxdarksage:
- Data groups API now exists to edit data groups for a user
- Data groups can be included as part of a sign up
- Data groups are validated against the list of defined data groups for a study.
- Participant options APIs updated to save data groups as a participant option, and to improve type safety
- Changes to support saving data groups as a participant option

[No work yet to export the data groups in the participant roster, the health data records, etc. No work to filter consents or schedule plans using data groups.]


syoung-smallwisdom:
looks good?

app/org/sagebionetworks/bridge/models/accounts/ExternalIdentifier.java
```diff
@@ -15,5 +16,20 @@ public ExternalIdentifier(@JsonProperty("identifier") String identifier) {
     public String getIdentifier() {
         return identifier;
     }
+
+    @Override
+    public int hashCode() {
+        return identifier.hashCode();
```
alxdarksage: Need to use Objects.hash(identifier), which is null-safe.

test/org/sagebionetworks/bridge/BridgeUtilsTest.java
```diff
@@ -61,6 +64,17 @@ public void commaListToSet() {
...
+        set = BridgeUtils.commaListToSet("a,,b");
+        assertEquals(Sets.newHashSet("a","b"), set);
+    }
+    
+    @Test
+    public void setToCommaList() {
+        Set<String> set = Sets.newHashSet("a", null, "", "b");
+        
+        assertEquals("a,b", BridgeUtils.setToCommaList(set));
     }
```
DwayneJengSage, line 82: For completeness, we should also test the null case for setToCommaList()

app/org/sagebionetworks/bridge/dynamodb/OptionLookup.java
```diff
@@ -35,5 +34,10 @@ public String get(String healthCode) {
     public SharingScope getSharingScope(String healthCode) {
         return SharingScope.valueOf(get(healthCode));
     }
+    
+    public Set<String> getDataGroups(String healthCode) {
+        String value = map.get(healthCode);
+        return BridgeUtils.commaListToSet(value);
+    }
```
DwayneJengSage, line 41: Is it worth adding a unit test for getDataGroups(), just to make sure we exercise the code?

app/org/sagebionetworks/bridge/models/accounts/DataGroups.java
```diff
@@ -0,0 +1,45 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public final class DataGroups implements BridgeEntity {
+
+    private final Set<String> dataGroups;
+    
+    @JsonCreator
+    public DataGroups(@JsonProperty("dataGroups") Set<String> dataGroups) {
+        this.dataGroups = (dataGroups == null) ? Collections.emptySet() : dataGroups;
+    }
```
DwayneJengSage, line 19: Do we want to create an immutable copy of dataGroups? This guards against a possible bug where we construct a DataGroups instance, and somewhere else in the code, we modify the underlying set.

test/org/sagebionetworks/bridge/models/accounts/DataGroupsTest.java
```diff
@@ -0,0 +1,43 @@
...
+    public void canRoundTripSerialize() throws Exception {
+        DataGroups groups = new DataGroups(Sets.newHashSet("A","B","C"));
+        
+        String json = BridgeObjectMapper.get().writeValueAsString(groups);
+        assertEquals("{\"dataGroups\":[\"A\",\"B\",\"C\"],\"type\":\"DataGroups\"}", json);
+       
+        DataGroups newGroups = BridgeObjectMapper.get().readValue(json, DataGroups.class);
+        assertEquals(3, newGroups.getDataGroups().size());
+        assertEquals(groups.getDataGroups(), newGroups.getDataGroups());
+    }
```
DwayneJengSage, line 32: This test is brittle. There's no guarantee that the data groups will be serialized in ABC order. It would be better to parse the JsonNode and verify that the dataGroup list contains all the elements that we expect.

Alternatively, if hash performance doesn't matter, we could just use a list.

app/org/sagebionetworks/bridge/models/accounts/ExternalIdentifier.java
```diff
@@ -3,6 +3,7 @@
 import org.sagebionetworks.bridge.models.BridgeEntity;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.stormpath.sdk.lang.Objects;
```
DwayneJengSage: Pretty sure this is the wrong Objects class.

app/org/sagebionetworks/bridge/models/accounts/ExternalIdentifier.java
```diff
@@ -15,5 +16,20 @@ public ExternalIdentifier(@JsonProperty("identifier") String identifier) {
...
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj || obj == null)
+            return false;
+        if (getClass() != obj.getClass())
+            return false;
+        ExternalIdentifier other = (ExternalIdentifier) obj;
+        return Objects.nullSafeEquals(identifier, other.identifier);
+    }
```
DwayneJengSage, line 37: Add EqualsVerifier test?

app/org/sagebionetworks/bridge/models/accounts/SignUp.java
```diff
@@ -1,42 +1,32 @@
...
-    }
-
-    public static final SignUp fromJson(JsonNode node, boolean allowRoles) {
-        String username = JsonUtils.asText(node, USERNAME_FIELD);
-        String email = JsonUtils.asText(node, EMAIL_FIELD);
-        String password = JsonUtils.asText(node, PASSWORD_FIELD);
-        Set<Roles> roles = (allowRoles) ? JsonUtils.asRolesSet(node, ROLES_FIELD) : null;
-        return new SignUp(username, email, password, roles);
+        this.roles = (roles == null) ? Sets.newHashSet() : roles;
+        this.dataGroups = (dataGroups == null) ? Sets.newHashSet() : dataGroups;
```
DwayneJengSage: Similarly, do we want to create immutable copies here?

app/org/sagebionetworks/bridge/models/accounts/User.java
```diff
@@ -123,6 +125,14 @@ public void setStudyKey(String studyKey) {
...
         this.roles = roles;
     }
+    
+    public Set<String> getDataGroups() {
+        return dataGroups;
+    }
+    
+    public void setDataGroups(Set<String> dataGroups) {
+        this.dataGroups = dataGroups;
+    }
```
DwayneJengSage, line 136: Similarly, do we want to create an immutable copy here as well?

app/org/sagebionetworks/bridge/models/accounts/UserSessionInfo.java
```diff
@@ -41,6 +42,7 @@ public UserSessionInfo(UserSession session) {
         this.sharingScope = session.getUser().getSharingScope();
         this.username = session.getUser().getUsername();
         this.roles = session.getUser().getRoles();
+        this.dataGroups = session.getUser().getDataGroups();
```
DwayneJengSage: Minor: If User.getDataGroups() is immutable, it might not be necessary to create an immutable copy here. That said, it would be best practice to do so anyway, in case the semantics of User changes somewhere down the line. Besides, ImmutableSet.copyOf() has optimizations which should make this a relative cheap operation.

app/org/sagebionetworks/bridge/models/accounts/DataGroups.java
alxdarksage, line 19: Added

app/org/sagebionetworks/bridge/dynamodb/OptionLookup.java
alxdarksage, line 41: Added

test/org/sagebionetworks/bridge/BridgeUtilsTest.java
alxdarksage, line 82: Added

test/org/sagebionetworks/bridge/models/accounts/DataGroupsTest.java
alxdarksage, line 32: It's deserialized and equal to the original object which demonstrates the set is correct. I'll parse to JsonNode and test the other thing that matters, the type attribute.

NOTE what I learned yesterday: this is changed in the file list view, even though the old code is visible above. Confusing!

app/org/sagebionetworks/bridge/models/accounts/ExternalIdentifier.java
alxdarksage: Yes. Fixed.

app/org/sagebionetworks/bridge/models/accounts/ExternalIdentifier.java
alxdarksage, line 37: Added

app/org/sagebionetworks/bridge/models/accounts/SignUp.java
alxdarksage: Added. I'm not adding tests because we don't rely on this behavior in these cases, it's more like a runtime check against programming errors.

app/org/sagebionetworks/bridge/models/accounts/User.java
alxdarksage, line 136: Added (also to roles)

app/org/sagebionetworks/bridge/models/accounts/UserSessionInfo.java
alxdarksage: Added immutable copies here during construction. Added helper method to do this in a null safe way: BridgeUtils.nullSafeImmutableSet(Set<T> set). 

All these immutable will require a couple of tests to be rewritten, I'll do that now.

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
```diff
@@ -42,8 +42,10 @@ public Result signOut() throws Exception {
 
     public Result signUp() throws Exception {
         JsonNode json = requestToJSON(request());
-        SignUp signUp = SignUp.fromJson(json, false);
-        signUp.getRoles().clear();
+        SignUp signUp = parseJson(request(), SignUp.class);
+        // this is now the only way to eliminate roles if set by the user...
+        signUp = new SignUp(signUp.getUsername(), signUp.getEmail(), 
+                signUp.getPassword(), null, signUp.getDataGroups());
```
DwayneJengSage, line 48: Do we have a test to make sure user's can't sign up with a role? (This could be a security hole if they can specify "admin", so we should have explicit tests for this.)

test/org/sagebionetworks/bridge/play/controllers/UserProfileControllerTest.java
```diff
@@ -39,28 +53,67 @@ public void canSubmitExternalIdentifier() throws Exception {
...
-        Http.Context.current.set(TestUtils.mockPlayContextWithJson("{\"identifier\":\"\"}"));
+    public void validDataGroupsCanBeAdded() throws Exception {
+        Http.Context.current.set(TestUtils.mockPlayContextWithJson("{\"dataGroups\":[\"group1\"]}"));
         
         UserProfileController controller = controllerForExternalIdTests();
+        
+        Result result = controller.updateDataGroups();
+        
+        ArgumentCaptor<DataGroups> captor = ArgumentCaptor.forClass(DataGroups.class);
+        verify(optionsService).setDataGroups(any(), any(), captor.capture());
```
DwayneJengSage: Minor: We know the study ID and healthcode. Might as well check for them.

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
```diff
@@ -201,6 +225,63 @@ public void testSignOutWhenSignedOut() {
...
     
+    @Test
+    public void signUpWillCreateDataGroups() {
+        authService = spy(authService);
+        optionsService = spy(optionsService);
+        authService.setOptionsService(optionsService);
+        
+        Study tempStudy = TestUtils.getValidStudy(AuthenticationServiceImplTest.class);
+        tempStudy.setDataGroups(Sets.newHashSet("group1","group2","group3"));
+        tempStudy = studyService.createStudy(tempStudy);
```
DwayneJengSage: Is it necessary to create an entire study for this test? This creates a lot of extra overhead for the test.

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
```diff
@@ -57,6 +76,11 @@
     @Before
     public void before() {
         testUser = helper.createUser(AuthenticationServiceImplTest.class);
+        Study study = studyService.getStudy(TestConstants.TEST_STUDY_IDENTIFIER);
+        if (study.getDataGroups().isEmpty()) {
+            study.getDataGroups().add("foo");
+            studyService.updateStudy(study, true);
+        }
```
DwayneJengSage, line 84: It might be cleaner to say !study.getDataGroups().contains("foo"). This will make other tests simpler.

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
```diff
@@ -201,6 +225,63 @@ public void testSignOutWhenSignedOut() {
...
+            accountDao.deleteAccount(tempStudy, email);
+            studyService.deleteStudy(tempStudy.getIdentifier());
+        }
+    }
+    
+    @Test
+    public void userCreatedWithDataGroupsHasThemOnSignIn() throws Exception {
+        // @Before method always ensures there is one data group in the study. Use that for this test.
+        String group = studyService.getStudy(TestConstants.TEST_STUDY_IDENTIFIER).getDataGroups().iterator().next();
+        Set<String> dataGroups = Sets.newHashSet(group);
```
DwayneJengSage: If you change the Before to ensure there's always a group called Foo, you can simplify the test a little.

app/org/sagebionetworks/bridge/services/ParticipantOptionsService.java
```diff
@@ -1,64 +1,90 @@
...
 import org.sagebionetworks.bridge.dynamodb.OptionLookup;
+import org.sagebionetworks.bridge.models.accounts.DataGroups;
+import org.sagebionetworks.bridge.models.accounts.ExternalIdentifier;
 import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
 
+/**
+ * Provides type-safe access to all the options that can be persisted for a health code, as well as validation. 
+ * The values are saved as strings and can be exported as such (in spreadsheets, Synapse, and so forth).
+ */
 public interface ParticipantOptionsService {
```
DwayneJengSage, line 14: I feel like this approach isn't scalable. This approach means that for every option, we'll need to add a new getter/setter pair in ParticipantOptionsService. I feel like it would be more scalable to have get/setOptionAsBoolean(), get/setOptionAsEnum() (which takes the Enum class as a parameter), get/setOptionAsString(), get/setOptionAsStringSet(), etc.

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
alxdarksage, line 48: Added

test/org/sagebionetworks/bridge/play/controllers/UserProfileControllerTest.java
alxdarksage: Switched. Just learned about eq(), which is very handy.

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
alxdarksage: Yes because it's loading the spring context. This is one of those service classes that I will convert to mocks some day, but it already exists in this form.

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
alxdarksage, line 84: Changed

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
alxdarksage: Yes this seems convoluted as in the @Before we add a specific known group. Changed

app/org/sagebionetworks/bridge/services/ParticipantOptionsService.java
alxdarksage, line 14: This is a lot more verbose, but I can change.

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceImplTest.java
```diff
@@ -1,82 +1,149 @@
...
+        verifyNoMoreInteractions(mockDao);
+    }
+
+    @Test
+    public void getAllParticipantOptions() {
+        service.getAllParticipantOptions(HEALTH_CODE);
+        
+        verify(mockDao).getAllParticipantOptions(HEALTH_CODE);
+        verifyNoMoreInteractions(mockDao);
+    }
```
DwayneJengSage, line 167: This test should verify the return value.

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceImplTest.java
```diff
@@ -1,82 +1,149 @@
...
+    
+    @Test
+    public void getOptionForAllStudyParticipants() {
+        service.getOptionForAllStudyParticipants(TEST_STUDY, ParticipantOption.EMAIL_NOTIFICATIONS);
+        
+        verify(mockDao).getOptionForAllStudyParticipants(TEST_STUDY, ParticipantOption.EMAIL_NOTIFICATIONS);
+        verifyNoMoreInteractions(mockDao);
+    }
```
DwayneJengSage, line 179: This test should verify the return value.

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceImplTest.java
```diff
@@ -1,82 +1,149 @@
...
+        verify(mockDao).setOption(TEST_STUDY, HEALTH_CODE, ParticipantOption.EMAIL_NOTIFICATIONS, Boolean.TRUE.toString());
+        verifyNoMoreInteractions(mockDao);
     }
     
     @Test
-    public void canCrudSharingScopeOption() {
-        optionsService.setOption(testUser.getStudyIdentifier(), testUser.getUser().getHealthCode(),
-                SharingScope.ALL_QUALIFIED_RESEARCHERS);
+    public void getBoolean() {
+        when(mockDao.getOption(HEALTH_CODE, ParticipantOption.EMAIL_NOTIFICATIONS)).thenReturn("true");
```
DwayneJengSage, line 63: What about null boolean?

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceImplTest.java
```diff
@@ -1,82 +1,149 @@
...
+        verify(mockDao).setOption(TEST_STUDY, HEALTH_CODE, ParticipantOption.SHARING_SCOPE, "SPONSORS_AND_PARTNERS");
+        verifyNoMoreInteractions(mockDao);
+    }
+    
+    @Test
+    public void getEnum() {
+        when(mockDao.getOption(HEALTH_CODE, ParticipantOption.SHARING_SCOPE)).thenReturn("SPONSORS_AND_PARTNERS");
+        
+        assertEquals(SharingScope.SPONSORS_AND_PARTNERS, service.getEnum(HEALTH_CODE, SHARING_SCOPE, SharingScope.class));
     }
```
DwayneJengSage, line 105: What about null enums?

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceImplTest.java
```diff
@@ -1,82 +1,149 @@
...
     
+    @Test
+    public void getStringSet() {
+        when(mockDao.getOption(HEALTH_CODE, ParticipantOption.DATA_GROUPS)).thenReturn("group1,group2,group3");
+        
+        Set<String> dataGroups = Sets.newHashSet("group1", "group2", "group3");
+        
+        assertEquals(dataGroups, service.getStringSet(HEALTH_CODE, DATA_GROUPS));
+    }
```
DwayneJengSage, line 131: What about null (empty) set?

app/org/sagebionetworks/bridge/services/email/ConsentEmailProvider.java
```diff
@@ -66,11 +67,9 @@ public MimeTypeEmail getMimeTypeEmail() throws MessagingException {
         final String sendFromEmail = String.format("%s <%s>", study.getName(), study.getSupportEmail());
         builder.withSender(sendFromEmail);
 
-        builder.withRecipient(user.getEmail());
-        Set<String> emailAddresses = BridgeUtils.commaListToSet(study.getConsentNotificationEmail());
-        for (String email : emailAddresses) {
-            builder.withRecipient(email);
-        }
+        Set<String> emailAddresses = Sets.newHashSet(BridgeUtils.commaListToSet(study.getConsentNotificationEmail()));
```
DwayneJengSage: Minor: Should add a comment about why we're wrapping a set in a set.

app/org/sagebionetworks/bridge/services/email/MimeTypeEmailBuilder.java
```diff
@@ -37,8 +38,13 @@ MimeTypeEmailBuilder withSender(String senderAddress) {
      * @param recipientAddress
      * @return
      */
```
DwayneJengSage, line 40: Minor: Duplicate comment.

app/org/sagebionetworks/bridge/services/email/MimeTypeEmailBuilder.java
```diff
@@ -37,8 +38,13 @@ MimeTypeEmailBuilder withSender(String senderAddress) {
...
-        this.recipientAddresses.add(recipientAddress);
+    /**
+     * A recipient for this email (may call this more than once and values will be accumulated).
+     * @param recipientAddress
+     * @return
+     */
+    MimeTypeEmailBuilder withRecipients(Collection<String> recipients) {
+        this.recipientAddresses.addAll(recipients);
         return this;
     }
```
DwayneJengSage, line 53: Would it be cleaner to have both withRecipient() and withRecipients()? That way, you don't have to create singleton sets, and you don't have to do weird set manipulation.

test/org/sagebionetworks/bridge/services/email/ParticipantRosterProviderTest.java
```diff
@@ -29,6 +29,20 @@ public void setUp() throws Exception {
...
+        List<StudyParticipant> participants = Lists.newArrayList(participant);
+        
+        study.setConsentNotificationEmail("bridge-testing@sagebase.org,postmaster@sagebase.org");
+        ParticipantRosterProvider provider = new ParticipantRosterProvider(study, participants);
+        
+        MimeTypeEmail email = provider.getMimeTypeEmail();
+        assertEquals(2, email.getRecipientAddresses().size());
+        assertEquals("bridge-testing@sagebase.org", email.getRecipientAddresses().get(0));
+        assertEquals("postmaster@sagebase.org", email.getRecipientAddresses().get(1));
+    }
```
DwayneJengSage, line 45: Since commaListToSet creates a set, the order isn't guaranteed. You're better off either converting this to a set and comparing set equality or changing the util function to commaListToList (or whatever name makes more sense).

test/org/sagebionetworks/bridge/validators/DataGroupsValidatorTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.junit.Before;
+import org.junit.Test;
+
+import org.sagebionetworks.bridge.exceptions.InvalidEntityException;
+import org.sagebionetworks.bridge.models.accounts.DataGroups;
+
+import com.google.common.collect.Sets;
+
+public class DataGroupsValidatorTest {
```
DwayneJengSage, line 18: For completeness, how about the following tests:
- a DataGroups that passes validation
- a study with no data groups, but the DataGroups object has data groups
- an empty DataGroup

test/org/sagebionetworks/bridge/TestUserAdminHelper.java
```diff
@@ -85,40 +89,45 @@ public StudyIdentifier getStudyIdentifier() {
...
+    public TestUser createUser(Class<?> cls, boolean signIn, boolean consent, Set<Roles> roles, Set<String> dataGroups) {
         Study study = studyService.getStudy(TEST_STUDY_IDENTIFIER);
+        return createUser(cls, study, signIn, consent, roles, dataGroups);
+    }
+    
+    public TestUser createUser(Class<?> cls, Study study, boolean signIn, boolean consent, Set<Roles> roles, Set<String> dataGroups) {
+        String name = makeRandomUserName(cls);
+        SignUp signUp = new SignUp(name, name + EMAIL_DOMAIN, PASSWORD, roles, dataGroups);
         return createUser(signUp, study, signIn, consent);
     }
```
DwayneJengSage: Instead of having an every increasing number of "convenience" overrides, would it be cleaner to just use a builder? How much refactoring would this require?

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceImplTest.java
alxdarksage, line 167: Added

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceImplTest.java
alxdarksage, line 179: Added

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceImplTest.java
alxdarksage, line 63: The service just checks for null parameters to the method and then calls to the DAO, and those null checks are trivial and not very interesting. You want those tested as well?

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceImplTest.java
alxdarksage, line 105: This has a some logic around nulls, added

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceImplTest.java
alxdarksage, line 131: Added

test/org/sagebionetworks/bridge/play/controllers/UserProfileControllerTest.java
```diff
@@ -39,28 +54,69 @@ public void canSubmitExternalIdentifier() throws Exception {
...
-        verify(optionsService).setOption(TEST_STUDY, "healthCode", 
-            ParticipantOption.EXTERNAL_IDENTIFIER, "ABC-123-XYZ");
+        verify(optionsService).setString(TEST_STUDY, "healthCode", EXTERNAL_IDENTIFIER, "ABC-123-XYZ");
     }
-    
+
+    @SuppressWarnings({ "rawtypes", "unchecked" })
     @Test
-    public void externalIdentifierVerifiesIdentifierExists() throws Exception {
-        Http.Context.current.set(TestUtils.mockPlayContextWithJson("{\"identifier\":\"\"}"));
```
DwayneJengSage, line 48: externalIdentifierVerifiesIdentifierExists() was deleted. Do we no longer need this test?

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceImplTest.java
DwayneJengSage, line 63: If it's legal input, we should test it.

app/org/sagebionetworks/bridge/services/email/ConsentEmailProvider.java
alxdarksage: Removing this. At some point I think the Set was immutable, but it's created every time you call that comma parsing method, and it's mutable in its final form.

app/org/sagebionetworks/bridge/services/email/MimeTypeEmailBuilder.java
alxdarksage, line 40: Removed

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
```diff
@@ -71,12 +99,12 @@ public void signInNoUsername() throws Exception {
 
     @Test(expected = BridgeServiceException.class)
     public void signInNoPassword() throws Exception {
-        authService.signIn(testUser.getStudy(), new SignIn("foo", null));
+        authService.signIn(testUser.getStudy(), new SignIn(TEST_DATA_GROUP, null));
```
DwayneJengSage: I think you were a little overzealous with your find and replace. SignIn takes a username not a data group. (Here and in signInInvalidCredentials() below.

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
```diff
@@ -103,7 +131,7 @@ public void signInSetsSharingScope() {
 
     @Test
     public void getSessionWithBogusSessionToken() throws Exception {
-        UserSession session = authService.getSession("foo");
+        UserSession session = authService.getSession(TEST_DATA_GROUP);
```
DwayneJengSage: Similarly, this "foo" is a session token.

app/org/sagebionetworks/bridge/services/email/MimeTypeEmailBuilder.java
alxdarksage, line 53: Sure, added

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
```diff
@@ -131,12 +159,12 @@ public void requestPasswordResetFailsOnEmptyString() throws Exception {
     
     @Test(expected = EntityNotFoundException.class)
     public void resetPasswordWithBadTokenFails() throws Exception {
-        authService.resetPassword(new PasswordReset("newpassword", "foo"));
+        authService.resetPassword(new PasswordReset("newpassword", TEST_DATA_GROUP));
```
DwayneJengSage: Similarly, this "foo" is a password reset token.

test/org/sagebionetworks/bridge/services/email/ParticipantRosterProviderTest.java
alxdarksage, line 45: Switched to comparing sets

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
```diff
@@ -201,6 +229,58 @@ public void testSignOutWhenSignedOut() {
...
+        
+        try {
+            SignUp signUp = new SignUp(name, email, "P@ssword1", null, list);
+
+            authService.signUp(study, signUp, true);
+            Account account = accountDao.getAccount(study, email);
+            HealthId healthId = healthCodeService.getMapping(account.getHealthId());
+            
+            verify(authService).signUp(study, signUp, true);
+            verify(optionsService).setStringSet(study, healthId.getCode(), DATA_GROUPS, list);
```
DwayneJengSage: Since optionsService is a real optionsService, I feel like this test would be more complete to verify that optionsService.getStringSet() returns the data groups we expect. This way, we're testing the behavior and not the implementation.

test/org/sagebionetworks/bridge/validators/DataGroupsValidatorTest.java
alxdarksage, line 18: Added

test/org/sagebionetworks/bridge/BridgeUtilsTest.java
```diff
@@ -53,14 +55,43 @@ public void commaListToSet() {
...
+        
+        assertEquals("a,b", BridgeUtils.setToCommaList(set));
+        assertNull(BridgeUtils.setToCommaList(null));
+        assertNull(BridgeUtils.setToCommaList(Sets.newHashSet()));
     }
 
+    @Test(expected = UnsupportedOperationException.class)
+    public void nullsafeImmutableSet() {
+        assertNotNull(BridgeUtils.nullSafeImmutableSet(null));
+        assertNotNull(BridgeUtils.nullSafeImmutableSet(Sets.newHashSet("A")));
```
DwayneJengSage: In addition to testing they're not null, we should check that they contain the values we expect (or are empty, if expected).

test/org/sagebionetworks/bridge/BridgeUtilsTest.java
```diff
@@ -53,14 +55,43 @@ public void commaListToSet() {
...
+        
+        // This should throw an UnsupportedOperationException
+        Set<String> set = BridgeUtils.nullSafeImmutableSet(Sets.newHashSet("A"));
+        set.add("B");
+        
+        // nulls are removed. They have to be to create ImmutableSet
+        assertEquals(Sets.newHashSet("A"), BridgeUtils.nullSafeImmutableSet(Sets.newHashSet(null, "A")));
```
DwayneJengSage, line 97: This assert is dead code, since the previous line will throw. We should separate this out into a separate test.

app/org/sagebionetworks/bridge/dynamodb/OptionLookup.java
```diff
@@ -35,5 +34,11 @@ public String get(String healthCode) {
     public SharingScope getSharingScope(String healthCode) {
         return SharingScope.valueOf(get(healthCode));
     }
+    
+    public Set<String> getDataGroups(String healthCode) {
+        String value = map.get(healthCode);
+        Set<String> set = BridgeUtils.commaListToSet(value);
+        return (set.isEmpty()) ? null : set;
+    }
```
DwayneJengSage, line 41: Do we really want to convert empty sets into nulls? Everywhere else, we use an empty set.

test/org/sagebionetworks/bridge/TestUserAdminHelper.java
alxdarksage: Changing this, need to run all the tests before checking it in.

test/org/sagebionetworks/bridge/play/controllers/UserProfileControllerTest.java
alxdarksage, line 48: This test was about validation in the controller to verify the external ID was not null. I removed this when refactoring the participant options service to take a string generically, then decided that there's no reason to prevent clients from setting null or an empty string if they want to. If the option has a default, setting it to null restores the default.

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
alxdarksage: changed

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
alxdarksage: changed

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
alxdarksage: changed

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
alxdarksage: changed

test/org/sagebionetworks/bridge/BridgeUtilsTest.java
alxdarksage: changed

test/org/sagebionetworks/bridge/BridgeUtilsTest.java
alxdarksage, line 97: made into new test

app/org/sagebionetworks/bridge/dynamodb/OptionLookup.java
alxdarksage, line 41: changed to return empty set here.

PR #929, opened 2015-11-17
Sending 202 for async calls

alxdarksage:
Return 202 (Accepted) for calls that return a 2xx response code before the final success or failure of the operation is known (in our case, the calls that send emails to downloads or with zip files attached, which may fail after the call). It means "The request has been accepted for processing, but the processing has not been completed." 

I know of these two calls, the send reset password request looks similar, but we know it is sent because it's sent during the course of the request, and presumably we would return a 500 if we failed to send the email.


syoung-smallwisdom:
looks good

PR #932, opened 2015-11-18
Fix escaping issue in HTML verifyEmail.html

alxdarksage:
Play does some kind of strange ' escaping that breaks. Escape in the controller so Play doesn't mess with this.


EricSiegNW:
r=eric.sieg

PR #935, opened 2015-11-19
BRIDGE-655 Add logic for setting a kill switch for unsupported client app versions

syoung-smallwisdom:


app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
```diff
@@ -299,6 +303,30 @@ public boolean isHealthCodeExportEnabled() {
...
+    /** {@inheritDoc} */
+    @DynamoDBMarshalling(marshallerClass = StringIntegerMapMarshaller.class)
+    @Override
+    public Map<String,Integer> getMinSupportedAppVersions() {
+        return minSupportedAppVersions;
+    }
+
+    @Override
+    public void setMinSupportedAppVersions(Map<String,Integer> map) {
+        this.minSupportedAppVersions = (map == null) ? new HashMap<>() : map;
```
alxdarksage, line 315: It's set in constructor, so this.minSupportedAppVersions.addAll(map) would work here. Even if there were entries, they'd be updated.

app/org/sagebionetworks/bridge/dynamodb/StringIntegerMapMarshaller.java
```diff
@@ -0,0 +1,12 @@
...
+import java.util.Map;
+
+import com.amazonaws.services.dynamodbv2.datamodeling.JsonMarshaller;
+
+/**
+ * Marshalls Map<String, Integer> values JSON when persisting using DynamoDB.
+ */
+public class StringIntegerMapMarshaller extends JsonMarshaller<Map<String,Integer>> {
+
+}
```
alxdarksage: What was your experience in needing this, because there's no implementation here. What did DynamoDB do without this?

app/org/sagebionetworks/bridge/models/ClientInfo.java
```diff
@@ -134,6 +152,13 @@ public Integer getSdkVersion() {
         return sdkVersion;
     }
     
+    public boolean isSupportedVersion(Integer minSupportedVersion) {
+    	if (appVersion != null && minSupportedVersion != null && appVersion < minSupportedVersion) {
+    		return false;
+    	}
+    	return true;
+    }
+
```
alxdarksage, line 159: return !(appVersion != null && minSupportedVersion != null && appVersion < minSupportedVersion);

app/org/sagebionetworks/bridge/models/ClientInfo.java
```diff
@@ -78,6 +79,16 @@ public ClientInfo load(String userAgent) throws Exception {
...
+    
+    /**
+     * The osName that maps to an iOS client app.
+     */
+    public static final String IOS_OS_NAME = "iPhone OS";
+    
+    /**
+     * The osName that maps to an Android client app.
+     */
+    public static final String ANDROID_OS_NAME = "Android";
```
alxdarksage: I don't see where these are used or how you're addressing the space vs "/" issue... ?

app/org/sagebionetworks/bridge/models/ClientInfo.java
```diff
@@ -252,17 +284,52 @@ static ClientInfo parseUserAgentString(String ua) {
...
+        				withOsVersion(osParts[1].trim());
+        	}
+        	else {
+        		int idx = osInfo.lastIndexOf(" ");
+        		if (idx > 0) {
+            		builder = builder.
+            				withOsName(osInfo.substring(0,idx).trim()).
+            				withOsVersion(osInfo.substring(idx).trim());
+        		}
+        	}
```
alxdarksage: Please replace tabs with four spaces.

app/org/sagebionetworks/bridge/models/studies/Study.java
```diff
@@ -195,4 +196,19 @@
...
     public void setHealthCodeExportEnabled(boolean enabled);
+    
+    /**
+     * Minimum supported app version number. If set, user app clients pointing to an older version will 
+     * fail with an httpResponse status code of 410.
+     */
+    public Integer getMinSupportedAppVersion(String osName);
+    
+    /** @see #getMinSupportedVersion(); */
+    public void setMinSupportedAppVersion(String osName, Integer version);
```
alxdarksage: I would remove this setter, it's jut more common in Java to use getMinSupportedAppVersions().put(osName, version);

If you use it a lot, rename to addMinSupportedAppVersion(osName, version);

In reality this will be set with the map setter both by DynamoDB and the researcher UI, all as a group.

syoung-smallwisdom: Really? seems verbose, but ok.

app/org/sagebionetworks/bridge/models/ClientInfo.java
syoung-smallwisdom, line 159: yeah, I was trying to follow the style of the other method. Figured that the explicit "this is false and this is true" was preferred.

app/org/sagebionetworks/bridge/dynamodb/StringIntegerMapMarshaller.java
syoung-smallwisdom: Followed the previous example of StringSetMarshaller

app/org/sagebionetworks/bridge/play/controllers/BaseController.java
```diff
@@ -160,6 +165,22 @@ private String getSessionToken() {
...
     }
+    
+    void verifySupportedVersionOrThrowException(Study study, UserSession session) throws UnsupportedVersionException {
+
+    	// Exit early if Admin. These accounts should not throw the exception b/c of version support
+    	// b/c might want to look at older data. The supported version is intended for client apps that
+    	// gather data rather than apps used to model it. syoung 11/19/2015
+    	if (session.isAdminRole()) {
+    		return;  
+    	}
```
alxdarksage: I don't think we want to do this. These are orthogonal considerations and when we've made exceptions in the past for admin accounts, they've been used as participant accounts by developers and weird errors result.

app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
syoung-smallwisdom, line 315: Is default constructor _always_ called? (not the case with iOS) Deserializer will use a different constructor from the base.

app/org/sagebionetworks/bridge/play/controllers/BaseController.java
```diff
@@ -160,6 +165,22 @@ private String getSessionToken() {
...
+    	
+        ClientInfo clientInfo = getClientInfoFromUserAgentHeader();
+        String osName = clientInfo.getOsName();
+        if (osName != null && !clientInfo.isSupportedVersion(study.getMinSupportedAppVersion(osName))) {
```
alxdarksage: null check unneeded. study.getMinSupportedAppVersion() will return null, null is checked for in isSupportedVersion. Which is good because if the osName is unknown, that method will return null (no map entry) as well.

app/org/sagebionetworks/bridge/services/StudyServiceImpl.java
```diff
@@ -190,6 +190,7 @@ public Study updateStudy(Study study, boolean isAdminUpdate) {
         if (!isAdminUpdate) {
             study.setMaxNumOfParticipants(originalStudy.getMaxNumOfParticipants());
             study.setHealthCodeExportEnabled(originalStudy.isHealthCodeExportEnabled());
+            // TODO: syoung 11/16/2015 Should the minSupportedVersion require admin privileges? I'm thinking No.
```
alxdarksage: No, can remove comment.

test/org/sagebionetworks/bridge/dynamodb/DynamoStudyTest.java
```diff
@@ -37,34 +37,40 @@ public void equalsHashCode() {
...
+        assertEqualsAndNotNull(study.getConsentPDF(), JsonUtils.asText(node, "consentPDF"));
+        assertEqualsAndNotNull((Long)study.getVersion(), (Long)node.get("version").asLong());
         assertTrue(node.get("strictUploadValidationEnabled").asBoolean());
         assertTrue(node.get("healthCodeExportEnabled").asBoolean());
-        assertEquals("Study", node.get("type").asText());
+        assertEqualsAndNotNull("Study", node.get("type").asText());
+        
+        JsonNode supportedVersionsNode = JsonUtils.asJsonNode(node, "minSupportedAppVersions");
+        assertNotNull(supportedVersionsNode);
+        assertEqualsAndNotNull(study.getMinSupportedAppVersion("iPhone OS"), (Integer)supportedVersionsNode.get("iPhone OS").asInt());
```
alxdarksage: node.get("iPhone OS").asInteger() (or .asInt() I can't remember) -- no cast needed

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
```diff
@@ -100,6 +102,7 @@ private Result signInWithRetry(final int retryCounter) throws Exception {
         final Study study = getStudyOrThrowException(json);
         try {
             session = authenticationService.signIn(study, signIn);
+            verifySupportedVersionOrThrowException(study, session);
```
alxdarksage: maybe  verifyMinSupportedAppVersionOrThrowException?

alxdarksage: I think we should check this after line 95 as well. I don't know if this can happen, but if an application can be updated without losing its session, it would get back the session from Redis and we'd still want to block it at that time.

app/org/sagebionetworks/bridge/dynamodb/StringIntegerMapMarshaller.java
alxdarksage: You'll need to test this against DynamoDB. That marshaller is there because DDB has a weird thing where it will throw an error if an empty Set is set for that attribute. I don't know what it does with an empty Map... don't know if it can even serialize that by default (if not we'd need to serialize it as JSON). The DynamoStudyDaoTest runs against DDB so it can be verified to work there with an empty map, and with a map that has a min supported app version mapping.

app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
alxdarksage, line 315: Yes, if there's no other constructor to call, Java calls the default constructor (always). If there's no default constructor... it adds one that does nothing in the byte code.

app/org/sagebionetworks/bridge/play/controllers/BaseController.java
syoung-smallwisdom: you can ask for a null key in a HashMap? That throws an exception in iOS.

test/org/sagebionetworks/bridge/dynamodb/DynamoStudyTest.java
syoung-smallwisdom: asInt() returns int, and not Integer. Needed to cast or it gave me an error.

app/org/sagebionetworks/bridge/play/controllers/BaseController.java
alxdarksage: Not a null key, but if the key exists and the value doesn't, that's fine. Sorry, you might need the check that the key exists.

app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
syoung-smallwisdom, line 315: addAll is different from set and doesn't meet the contract of the interface.

alxdarksage, line 315: If you don't want to protect against an NPE (and that's not something we normally prevent), just set the map. It's just a little odd to create a new map in this situation. But because of how DDB works, we need to have this setter.

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
syoung-smallwisdom: Would like to keep the naming w/o the "min" since the exception it throws is "unsupported".

app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
syoung-smallwisdom, line 315: Yeah, I get that. I was following the other examples of setters for things that shouldn't be set to null.

app/org/sagebionetworks/bridge/models/accounts/UserSession.java
```diff
@@ -52,4 +53,7 @@ public User getUser() {
     public void setUser(User user) {
         this.user = user;
     }
+    public boolean isAdminRole() {
+    	return this.getUser().isInRole(Roles.ADMINISTRATIVE_ROLES);
+    }
```
alxdarksage: I don't think this is called anywhere now and can be removed... ?

syoung-smallwisdom:
closing until tests can be run. will reopen

PR #936, opened 2015-11-20
Criteria schedule strategies

alxdarksage:
New schedule strategy that selects a schedule based on client info like app version or the data groups of the user. Can be used to server different schedules/activities to different users. I feel it may eventually replace filtering by app version at the schedule plan level, which is getting messy in the UI.

Also implements another method for finding and suppressing persisted tasks that no longer match user-supplied criteria like app version. No longer copying this information into the task, instead, comparing persisted tasks to the tasks retrieved from the scheduler, and if a persisted task isn't in that list, it isn't returned.

DON'T MERGE. This needs to be substantially revised after the introduction of criteria-related code that will be similar for the selection of both schedules and consents.

### DIFF
--- a/app/org/sagebionetworks/bridge/models/schedules/ScheduledActivityStatus.java
+++ b/app/org/sagebionetworks/bridge/models/schedules/ScheduledActivityStatus.java
@@ -41,4 +41,11 @@ public enum ScheduledActivityStatus {
     public static final Set<ScheduledActivityStatus> DELETABLE_STATUSES = EnumSet.of(ScheduledActivityStatus.SCHEDULED,
             ScheduledActivityStatus.AVAILABLE, ScheduledActivityStatus.EXPIRED);
     
+    /**
+     * This activity has been found in database, it should be scheduled, but whether we return it or not ultimately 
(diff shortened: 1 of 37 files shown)


app/org/sagebionetworks/bridge/models/schedules/ScheduledActivityStatus.java
```diff
@@ -41,4 +41,11 @@
     public static final Set<ScheduledActivityStatus> DELETABLE_STATUSES = EnumSet.of(ScheduledActivityStatus.SCHEDULED,
             ScheduledActivityStatus.AVAILABLE, ScheduledActivityStatus.EXPIRED);
     
+    /**
+     * This activity has been found in database, it should be scheduled, but whether we return it or not ultimately 
+     * depends on whether it should be visible. All the visible statuses are enumerated here.
+     */
+    public static final Set<ScheduledActivityStatus> VISIBLE_STATUSES = EnumSet.of(ScheduledActivityStatus.SCHEDULED,
```
alxdarksage, line 48: Do we still use this?

PR #937, opened 2015-11-25
Subpopulations

alxdarksage:
- basic crud support from the REST API to DynamoDB for subpopulations. They are not connected to anything else right now, this seems like enough to review before proceeding (a good place to break up what will be a large change to the system).


app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
```diff
@@ -2,6 +2,7 @@
 
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
```
alxdarksage, line 5: No tests for this class because this is being added in an other PR, and just stubbed out here so I can work on this independently.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulation.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+    @Override
+    public void setStudyIdentifier(String studyIdentifier) {
+        this.studyIdentifier = studyIdentifier;
+    }
+    @Override
+    @DynamoDBRangeKey
+    public String getGuid() {
+        return guid;
+    }
```
DwayneJengSage, line 59: A few key design considerations:
- Is there any particular reason why study is the hash key and the guid is the range key? The guid uniquely identifies the subpop and hash lookups are much more efficient than range lookups. If the goal is to namespace guids within studies, we could always use a compound hash key. If we need to look up all subpops in a study, we could use a secondary index.
- Guids are researcher unfriendly. What if a researcher wants to look up a subpop by name? What if the researcher unintentionally creates two subpops with the same name? Would it be better to use name as the key instead of the guid (and restrict the name to URL-friendly JSON-friendly chars)?

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulation.java
```diff
@@ -0,0 +1,169 @@
...
+    @DynamoDBMarshalling(marshallerClass = StringSetMarshaller.class)
+    @DynamoDBAttribute
+    @Override
+    public Set<String> getAllOfGroups() {
+        return allOfGroups;
+    }
+    public void setAllOfGroups(Set<String> dataGroups) {
+        this.allOfGroups = (dataGroups == null) ? Sets.newHashSet() : dataGroups;
+    }
```
DwayneJengSage, line 120: Should the StringSet setters also be part of the interface? All the other setters are. Otherwise, it seems odd that the only way to set StringSets are parsing them from JSON.

Also, should we make a copy of the data group? Otherwise, weird things might happen if someone changes the original string set out from under us.

app/org/sagebionetworks/bridge/dao/SubpopulationDao.java
```diff
@@ -0,0 +1,76 @@
...
+     * Get all subpopulations defined for this study. It is possible to create a first default
+     * subpopulation if none exists
+     * @param studyId
+     * @param allExistingRecordsOnly
+     *      if true, this call will not create a default subpopulation if none exist, and it will return all 
+     *      records including deleted subpopulations. If false, the opposite: it will create a default 
+     *      subpopulation and it will hide deleted subpopulations.
+     * @return
+     */
+    public List<Subpopulation> getSubpopulations(StudyIdentifier studyId, boolean allExistingRecordsOnly);
```
DwayneJengSage: I think it's really odd that a getter would have a side effect. Is there any compelling reason the getter needs to create a default subpop? Otherwise, I feel it would be much cleaner, simpler, and less error prone to not have this behavior.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -0,0 +1,147 @@
...
+@Component
+public class DynamoSubpopulationDao implements SubpopulationDao {
+    
+    private DynamoDBMapper mapper;
+
+    /** DynamoDB mapper for the HealthDataRecord table. This is configured by Spring. */
+    @Resource(name = "subpopulationDdbMapper")
+    public void setMapper(DynamoDBMapper mapper) {
+        this.mapper = mapper;
+    }
```
DwayneJengSage, line 37: Minor: final

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -0,0 +1,147 @@
...
+    
+    @Override
+    public Subpopulation createSubpopulation(Subpopulation subpop) {
+        checkNotNull(subpop);
+        checkNotNull(subpop.getStudyIdentifier());
+
+        subpop.setVersion(null);
+        subpop.setDeleted(false);
```
DwayneJengSage: Would it be better to throw if version is non-null and deleted is true? This would catch bugs in client-code rather than silently changing what they submitted.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -0,0 +1,147 @@
...
+    public Subpopulation updateSubpopulation(Subpopulation subpop) {
+        checkNotNull(subpop);
+        checkNotNull(subpop.getStudyIdentifier());
+        checkNotNull(subpop.getVersion());
+        checkNotNull(subpop.getGuid());
+        
+        // This will throw a not found exception if this is not an update. 
+        // It also prevents deletion through this call.
+        Subpopulation existing = getSubpopulation(new StudyIdentifierImpl(subpop.getStudyIdentifier()), subpop.getGuid());
+        subpop.setDeleted(existing.isDeleted());
```
DwayneJengSage: Do we want to allow updating a deleted subpop? Do we want to allow undeleting a subpop by updating it? If the answer to both questions is no, I feel like it would be cleaner and more performant to simply check that deleted isn't false and throwing if it is.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -0,0 +1,147 @@
...
+        
+        DynamoDBQueryExpression<DynamoSubpopulation> query = 
+                new DynamoDBQueryExpression<DynamoSubpopulation>().withHashKeyValues(hashKey);
+        
+        // Get all the records because we only create a default if there are no physical records, 
+        // regardless of the deletion status.
+        List<DynamoSubpopulation> subpops = mapper.query(DynamoSubpopulation.class, query);
+        if (!allExistingRecordsOnly && subpops.isEmpty()) {
+            Subpopulation subpop = createDefaultSubpopulation(studyId);
+            return Lists.newArrayList(subpop);
```
DwayneJengSage: Does this have to be mutable? If not, we should always use ImmutableList.

app/org/sagebionetworks/bridge/dao/SubpopulationDao.java
DwayneJengSage: Also, the param name is very confusing to me. Intuitively, allExistingRecordsOnly=true would mean that we hide deleted records, since they don't "exist", which is the opposite of the documentation and code. I would suggest renaming this to "includeDeleted".

If we absolutely need to create the default subpop, we should make that a separate flag "createDefaultOnEmpty", just to make it more intuitive.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -0,0 +1,147 @@
...
+        return subpops.stream()
+            .filter(subpop -> allExistingRecordsOnly || !subpop.isDeleted())
+            .sorted(SPECIFICITY_SORTER).collect(Collectors.toList());
+    }
+    
+    @Override
+    public Subpopulation createDefaultSubpopulation(StudyIdentifier studyId) {
+        DynamoSubpopulation subpop = new DynamoSubpopulation();
+        subpop.setStudyIdentifier(studyId.getIdentifier());
+        subpop.setGuid(studyId.getIdentifier());
```
DwayneJengSage, line 104: We set the guid here with the study ID instead. Is this a bug?

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -0,0 +1,147 @@
...
+            boolean matches = CriteriaUtils.matchCriteria(context, subpop);
+            if (matches) {
+                found.add(subpop);
+            }
+        }
+        if (found.isEmpty()) {
+            throw new EntityNotFoundException(Subpopulation.class);
+        }
+        // Return the most specific match. Eventually we may return them all
+        return found.stream().sorted(SPECIFICITY_SORTER).collect(Collectors.toList()).get(0);
```
DwayneJengSage: Would it be better to use Collections.sort() instead? https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#sort-java.util.List-java.util.Comparator-

I find this much more readable than using a stream, and I'm guessing it involves fewer intermediate objects as well.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -0,0 +1,147 @@
...
+        // regardless of the deletion status.
+        List<DynamoSubpopulation> subpops = mapper.query(DynamoSubpopulation.class, query);
+        if (!allExistingRecordsOnly && subpops.isEmpty()) {
+            Subpopulation subpop = createDefaultSubpopulation(studyId);
+            return Lists.newArrayList(subpop);
+        }
+        // Now filter out deleted subpopulations, if requested
+        return subpops.stream()
+            .filter(subpop -> allExistingRecordsOnly || !subpop.isDeleted())
+            .sorted(SPECIFICITY_SORTER).collect(Collectors.toList());
```
DwayneJengSage: Minor optimization: having the if (allExistingRecordsOnly) outside so we don't have to filter if we don't need to.

It might also be more readable to use Collections.sort() as a separate sorting step.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -0,0 +1,147 @@
...
+
+    @Override
+    public void deleteAllSubpopulations(StudyIdentifier studyId) {
+        List<Subpopulation> subpops = getSubpopulations(studyId, true);
+        
+        if (!subpops.isEmpty()) {
+            List<FailedBatch> failures = mapper.batchDelete(subpops);
+            BridgeUtils.ifFailuresThrowException(failures);
+        }
+    }
```
DwayneJengSage, line 155: As written, this has the unusual side effect that, if you call it on a study with no subpops, it will create a default subpop, then immediately delete it. That's one case for the getter not having side effects, or at the very least having a separate "createOnEmpty" flag.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -0,0 +1,147 @@
...
+    @Override
+    public Subpopulation createSubpopulation(Subpopulation subpop) {
+        checkNotNull(subpop);
+        checkNotNull(subpop.getStudyIdentifier());
+
+        subpop.setVersion(null);
+        subpop.setDeleted(false);
+        subpop.setGuid(BridgeUtils.generateGuid());
+        mapper.save(subpop);
+        return subpop;
```
DwayneJengSage, line 54: Minor: As written, this will return the subpop that was submitted to DDB, but not the subpop that was saved. (Specifically, the version attribute will be bumped, and anything with DynamoDBIgnore will be removed.)

This might not be a big deal, but if a client creates a subpop, edits the returned JSON, and then immediately re-submits it, they'll get a ConcurrentModificationException because of the version number mismatch. The way we handle it in Health Data Records is to just return the guid.

This also affects update() and createDefault()

test/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDaoTest.java
```diff
@@ -0,0 +1,194 @@
...
+    public void crudSubpopulationOK() {
+        DynamoSubpopulation subpop = new DynamoSubpopulation();
+        subpop.setStudyIdentifier(studyId.getIdentifier());
+        subpop.setName("Name");
+        subpop.setDescription("Description");
+        subpop.setMinAppVersion(2);
+        subpop.setMaxAppVersion(10);
+        
+        // CREATE
+        Subpopulation savedSubpop = dao.createSubpopulation(subpop);
```
DwayneJengSage, line 74: Do we want to set a version number and a deleted=true to make sure the DAO will fix or reject those?

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
DwayneJengSage, line 54: Scratch that. Empirical evidence suggests I'm wrong about this. It looks like mapper.save() does have the side effect of updating the POJO in addition to saving it to DDB.

test/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDaoTest.java
```diff
@@ -0,0 +1,194 @@
...
+        assertNotNull(savedSubpop.getVersion());
+        
+        // READ
+        Subpopulation retrievedSubpop = dao.getSubpopulation(studyId, savedSubpop.getGuid());
+        assertEquals(savedSubpop, retrievedSubpop);
+        
+        // UPDATE
+        retrievedSubpop.setName("Name 2");
+        retrievedSubpop.setDescription("Description 2");
+        Subpopulation finalSubpop = dao.updateSubpopulation(retrievedSubpop);
```
DwayneJengSage, line 86: Do we want to set deleted=false to make sure the DAO enforces not being able to delete through an update?

test/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDaoTest.java
```diff
@@ -0,0 +1,194 @@
...
+        
+        // version 12, tag group1 == Subpops 3, 4
+        subpop = dao.getSubpopulationForUser(scheduleContext(12, "group1"));
+        assertEquals(SUBPOP_3, subpop.getName());
+        
+        // version 4, no tag == Subpops 2, 4
+        subpop = dao.getSubpopulationForUser(scheduleContext(4, null));
+        assertEquals(SUBPOP_2, subpop.getName());
+        
+        // version 4, tag group1 == Subpops 1,2,3,4, returns 2 in this case (most specific)
```
DwayneJengSage: Comment bug? Returns 1 instead of 2.

test/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDaoTest.java
```diff
@@ -0,0 +1,194 @@
...
+import org.sagebionetworks.bridge.models.schedules.ScheduleContext;
+import org.sagebionetworks.bridge.models.studies.Subpopulation;
+import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
+import org.sagebionetworks.bridge.models.studies.StudyIdentifierImpl;
+
+import com.google.common.collect.Sets;
+
+@ContextConfiguration("classpath:test-context.xml")
+@RunWith(SpringJUnit4ClassRunner.class)
+public class DynamoSubpopulationDaoTest {
```
DwayneJengSage, line 35: other unit tests?
- getSubpopulations() returns in sorted order
- test for createDefaultSubpopulation()
- getSubpopulationsForUser() returns no subpops

app/org/sagebionetworks/bridge/models/CriteriaUtils.java
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Utility classes for working with domain entities that should be matched by a growing list of 
+ * criteria, such as the version of the app making a request or the data groups associated to 
+ * a user. Matching is currently done through information passed in through the ScheduleContext, 
+ * a parameter object of values against which matching occurs.
+ */
+public class CriteriaUtils {
+    
+    public static Comparator<Criteria> SPECIFICITY_SORTER = new Comparator<Criteria>() {
```
DwayneJengSage: In-depth unit test for SPECIFICITY_SORTER?

Also, would this be better as a named inner class instead of an anonymous inner class?

test/org/sagebionetworks/bridge/models/CriteriaUtilsTest.java
```diff
@@ -0,0 +1,92 @@
...
+        // but this doesn't
+        assertFalse(CriteriaUtils.matchCriteria(context, new SimpleCriteria(Sets.newHashSet("group1", "group3"), Sets.newHashSet(), null, null)));
+    }
+    
+    @Test
+    public void noneOfGroupsMatch() {
+        ScheduleContext context = getContext(); // has group1, and group2
+        // Here, any group at all prevents a match.
+        assertFalse(CriteriaUtils.matchCriteria(context, new SimpleCriteria(Sets.newHashSet(), Sets.newHashSet("group3", "group1"), null, null)));
+    }
```
DwayneJengSage, line 89: Add a test that has a noneOf criteria, but still matches?

test/org/sagebionetworks/bridge/models/CriteriaUtilsTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+import java.util.Set;
+
+import org.junit.Test;
+
+import org.sagebionetworks.bridge.models.schedules.ScheduleContext;
+
+import com.google.common.collect.Sets;
+
+public class CriteriaUtilsTest {
```
DwayneJengSage, line 17: Is it worth adding a unit tests that has a context with no app version or data groups, just to make sure CriteriaUtils doesn't crash?

app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
```diff
@@ -37,6 +40,7 @@ private ScheduleContext(StudyIdentifier studyId, ClientInfo clientInfo, DateTime
         this.healthCode = healthCode;
         this.events = events;
         this.now = now;
+        this.userDataGroups = (userDataGroups == null) ? Sets.newHashSet() : userDataGroups;
```
DwayneJengSage: Immutable copy?

app/org/sagebionetworks/bridge/services/SubpopulationService.java
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     * Create subpopulation. 
+     * @param subpop
+     * @return
+     */
+    public Subpopulation createSubpopulation(Study study, Subpopulation subpop) {
+        checkNotNull(study);
+        checkNotNull(subpop);
+        
+        subpop.setGuid(BridgeUtils.generateGuid());
```
DwayneJengSage, line 39: Is it necessary for both the service and the DAO to set the guid? Would it be better for the service to set it and the DAO validates that it exists?

test/org/sagebionetworks/bridge/validators/SubpopulationValidatorTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.junit.Before;
+import org.junit.Test;
+
+import org.sagebionetworks.bridge.exceptions.InvalidEntityException;
+import org.sagebionetworks.bridge.models.studies.Subpopulation;
+
+import com.google.common.collect.Sets;
+
+public class SubpopulationValidatorTest {
```
DwayneJengSage, line 17: other unit tests?
- valid subpop with all params?
- valid subpop with no params?
- edge case: valid subpop with maxAppVersion equal to minAppVersion
- maxAppVersion less than minAppVersion
- null allOfGroup
- null noneOfGroup
- allOfGroup contains wrong group

test/org/sagebionetworks/bridge/validators/SubpopulationValidatorTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+    SubpopulationValidator validator;
+    
+    @Before
+    public void beofre() {
```
DwayneJengSage: Minor: typo "beofre"

test/org/sagebionetworks/bridge/services/SubpopulationServiceTest.java
```diff
@@ -0,0 +1,148 @@
...
+        subpop.setStudyIdentifier("junk-you-cannot-set");
+        subpop.setGuid("guid");
+        subpop.setRequired(false);
+        subpop.setDeleted(true);
+        
+        when(dao.getSubpopulation(any(), any())).thenReturn(Subpopulation.create());
+        
+        Subpopulation result = service.updateSubpopulation(study, subpop);
+        assertEquals("Name", result.getName());
+        assertEquals("guid", result.getGuid());
```
DwayneJengSage, line 108: Validate the study ID is set?

test/org/sagebionetworks/bridge/services/SubpopulationServiceTest.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+    @Test
+    public void getSubpopulations() {
+        Subpopulation subpop = Subpopulation.create();
+        // use the same twice, that's okay for this, just mocking a list
+        List<Subpopulation> list = Lists.newArrayList(subpop, subpop); 
+        when(dao.getSubpopulations(TEST_STUDY, false)).thenReturn(list);
+        
+        List<Subpopulation> results = service.getSubpopulations(TEST_STUDY);
+        assertEquals(2, results.size());
```
DwayneJengSage, line 124: Should check the list contains the results you expect.

test/org/sagebionetworks/bridge/services/SubpopulationServiceTest.java
```diff
@@ -0,0 +1,148 @@
...
+        verify(dao).getSubpopulation(TEST_STUDY, "AAA");
+    }
+    @Test
+    public void getSubpopulationForUser() {
+        // We test the matching logic in CriteriaUtilsTest as well as in the DAO. Here we just want
+        // to verify it is being carried through.
+        ScheduleContext context = new ScheduleContext.Builder()
+                .withClientInfo(ClientInfo.fromUserAgentCache("app/4")).build();
+        
+        service.getSubpopulationForUser(context);
```
DwayneJengSage: We should at least verify the return value.

app/org/sagebionetworks/bridge/play/controllers/SubpopulationController.java
```diff
@@ -0,0 +1,68 @@
...
+        this.subpopService = subpopService;
+    }
+
+    public Result getAllSubpopulations() {
+        UserSession session = getAuthenticatedSession(DEVELOPER);
+        Study study = studyService.getStudy(session.getStudyIdentifier());
+        
+        List<Subpopulation> subpopulations = subpopService.getSubpopulations(study);
+        return okResult(subpopulations);
+    }
```
DwayneJengSage, line 33: getAllSubpopulations() only needs the StudyIdentifier, not the full study.

app/org/sagebionetworks/bridge/play/controllers/SubpopulationController.java
```diff
@@ -0,0 +1,68 @@
...
+        subpop = subpopService.createSubpopulation(study, subpop);
+        
+        return createdResult(new GuidVersionHolder(subpop.getGuid(), subpop.getVersion()));
+    }
+    public Result updateSubpopulation(String guid) {
+        UserSession session = getAuthenticatedSession(DEVELOPER);
+        Study study = studyService.getStudy(session.getStudyIdentifier());
+        
+        Subpopulation subpop = parseJson(request(), Subpopulation.class);
+        subpop.setGuid(guid);
```
DwayneJengSage, line 48: By silently changing the guid, we could be masking a bug in the client code where they update the wrong subpop. It would be better to throw on a mismatched guid.

test/org/sagebionetworks/bridge/play/controllers/SubpopulationControllerTest.java
```diff
@@ -0,0 +1,190 @@
...
+        Http.Context.current.set(context);
+        
+        Result result = controller.getAllSubpopulations();
+        
+        assertEquals(200, result.status());
+        String json = Helpers.contentAsString(result);
+        JsonNode node = BridgeObjectMapper.get().readTree(json);
+        assertNotNull(node.get("items"));
+        assertNotNull(node.get("total"));
+        assertEquals("ResourceList", node.get("type").asText());
```
DwayneJengSage: We should validate the actual return values, not just that they exist.

test/org/sagebionetworks/bridge/play/controllers/SubpopulationControllerTest.java
```diff
@@ -0,0 +1,190 @@
...
+        assertTrue(created.isRequired());
+        assertEquals((Integer)2, created.getMinAppVersion());
+        assertEquals((Integer)10, created.getMaxAppVersion());
+        assertEquals(Sets.newHashSet("requiredGroup"), created.getAllOfGroups());
+        assertEquals(Sets.newHashSet("prohibitedGroup"), created.getNoneOfGroups());
+    }
+    
+    @Test
+    public void updateSubpopulation() throws Exception {
+        String json = "{\"name\":\"Name\",\"description\":\"Description\",\"required\":true,\"minAppVersion\":2,\"maxAppVersion\":10,\"allOfGroups\":[\"requiredGroup\"],\"noneOfGroups\":[\"prohibitedGroup\"]}";
```
DwayneJengSage, line 133: Test mismatched Guid in update? The guid is returned to the client, so they will very likely have it in their JSON when they call update()

test/org/sagebionetworks/bridge/models/CriteriaUtilsTest.java
```diff
@@ -0,0 +1,92 @@
...
+    @Test
+    public void matchesAppRange() {
+        ScheduleContext context = getContext();
+        assertTrue(CriteriaUtils.matchCriteria(context, new SimpleCriteria(Sets.newHashSet(), Sets.newHashSet(), null, 4)));
+        assertTrue(CriteriaUtils.matchCriteria(context, new SimpleCriteria(Sets.newHashSet(), Sets.newHashSet(), 1, null)));
+        assertTrue(CriteriaUtils.matchCriteria(context, new SimpleCriteria(Sets.newHashSet(), Sets.newHashSet(), 1, 4)));
+    }
+    
+    @Test
+    public void filtesAppRange() {
```
DwayneJengSage: Minor typo: filters instead of filtes

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulation.java
alxdarksage, line 59: The two operations that are common here are 1) select all sub pops for a study, and 2) select an individual sub pop. It seems the hashKey: study, range: guid approach does these both without any indices, which incur overhead. My understanding is that DDB handles hash/range queries well.

Regarding GUID vs. name, researchers will definitely interact with these by name, but internally it is much easier for us if we have an identifier we know to be unique. So this object has both a name and a guid, the name is editable even after creation, the guid is not.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulation.java
alxdarksage, line 120: Added setters to Subpopulation interface and copying sets when being set.

app/org/sagebionetworks/bridge/dao/SubpopulationDao.java
alxdarksage: We need to create a subpopulation as a migration step; in the future when creating new studies, it would be a convenience just as it's a convenience today to create a default consent to start with (in the new design, you have to create a default subpop to create a default consent). But also, a study isn't valid without these, all sorts of things can break until all these objects are created. 

I can 1) split the boolean up into two arguments, and 2) definitely can stand to be renamed then for clarity. In practice they vary together.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
alxdarksage, line 37: Added

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
alxdarksage: When done this way, submitting an existing subpopulation as a create causes it to be copied, which turns out to be very helpful. It doesn't destroy the existing subpopulation, and makes the option to create a "Copy" function in the researcher UI very easy.

Recently I changed the schedule plan API in the same way for the same reason, it made copies very easy.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
alxdarksage: It didn't seem harmful to update a deleted subpopulation, but I'm also preventing undeleting, for no real reason. I'll change to just check delete status and we can always change later if it comes up.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
alxdarksage: Changed

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
alxdarksage, line 104: No, for backwards compatibility with the existing consent documents, we're setting this to study ID. This means the default subpopulation will retrieve the existing consents, and existing studies won't lose this. Later subpopulations will use a different string, a GUID.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
alxdarksage: Changed

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
alxdarksage: I find this more readably when there are 2 or more operations to perform on a collection. Here we have to filter, sort, and covert to an immutable list. We're talking about collections of a handful of items so I don't think performance will be an issue. Only items that pass the filter are included in the sort and the immutable list, FWIW.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
alxdarksage, line 155: That's not what I intended! I did split up the flags so it's clearer what should happen (don't create default, do include deleted).

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
alxdarksage: Looking at the tests though, it would be helpful to actually copy the object so you can compare the before/after transformation of the object, so I'm adding that.

test/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDaoTest.java
alxdarksage, line 74: Adding test and adjusting so copy happens when calling create with an existing object.

test/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDaoTest.java
alxdarksage, line 86: Added test

test/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDaoTest.java
alxdarksage: Changed

test/org/sagebionetworks/bridge/models/CriteriaUtilsTest.java
alxdarksage, line 89: Added

test/org/sagebionetworks/bridge/models/CriteriaUtilsTest.java
alxdarksage, line 17: Added

app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
alxdarksage: Added

app/org/sagebionetworks/bridge/services/SubpopulationService.java
alxdarksage, line 39: Changed

test/org/sagebionetworks/bridge/validators/SubpopulationValidatorTest.java
alxdarksage, line 17: Hmm. I thought I had all these in the CriteriaUtilsTest, which now validates the Criteria values (to share with other entities that implement Criteria), but I didn't do that. I'll add them all there.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
alxdarksage: Forget it, I'm going to do as you say and just throw an exception if the object exists. This turns out to be a lot of code to allow a copy.

test/org/sagebionetworks/bridge/services/SubpopulationServiceTest.java
alxdarksage, line 108: Added

test/org/sagebionetworks/bridge/services/SubpopulationServiceTest.java
alxdarksage, line 124: changed

test/org/sagebionetworks/bridge/services/SubpopulationServiceTest.java
alxdarksage: added

app/org/sagebionetworks/bridge/play/controllers/SubpopulationController.java
alxdarksage, line 33: changed

app/org/sagebionetworks/bridge/play/controllers/SubpopulationController.java
alxdarksage, line 48: I feel like if they set the guid in the URL, which is to a resource, then that's what they're updating, and it's correct to ignore the guid in the JSON. It's not really even a documented property of the JSON object they are sending to us at that point, so this is just to prevent a security hole where they set it in a way we don't want. Like auto-marshaling a property to an object that's not supposed to be exposed.

test/org/sagebionetworks/bridge/play/controllers/SubpopulationControllerTest.java
alxdarksage: Added dummy list to verify it is passed back from the service to ResourceList

test/org/sagebionetworks/bridge/play/controllers/SubpopulationControllerTest.java
alxdarksage, line 133: Added guid property

test/org/sagebionetworks/bridge/models/CriteriaUtilsTest.java
alxdarksage: fixed

app/org/sagebionetworks/bridge/models/studies/Subpopulation.java
```diff
@@ -0,0 +1,43 @@
...
+    public String getGuid();
+    
+    public void setName(String name);
+    public String getName();
+    
+    public void setDescription(String description);
+    public String getDescription();
+    
+    public void setRequired(boolean required);
+    public boolean isRequired();
```
DwayneJengSage, line 35: What are the semantics of "required"? It looks like this is set only for the default subpop? iirc, we wanted to make the default subpop immutable, not just undeletable, so we can avoid the problem of no matching subpops.

We should comment this, since the semantics are unclear from the name.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -0,0 +1,148 @@
...
+    public Subpopulation updateSubpopulation(Subpopulation subpop) {
+        checkNotNull(subpop);
+        checkNotNull(subpop.getStudyIdentifier());
+        checkNotNull(subpop.getVersion());
+        checkNotNull(subpop.getGuid());
+        
+        Subpopulation existing = getSubpopulation(new StudyIdentifierImpl(subpop.getStudyIdentifier()), subpop.getGuid());
+        if (existing.isDeleted()) {
+            throw new EntityNotFoundException(Subpopulation.class);
+        }
```
DwayneJengSage, line 71: This prevents editing a deleted subpop, but it doesn't prevent deleting a subpop through an update. Is that intended?

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulation.java
```diff
@@ -0,0 +1,173 @@
...
+    @DynamoDBMarshalling(marshallerClass = StringSetMarshaller.class)
+    @DynamoDBAttribute
+    @Override
+    public Set<String> getAllOfGroups() {
+        return allOfGroups;
+    }
+    @Override
+    public void setAllOfGroups(Set<String> dataGroups) {
+        this.allOfGroups = (dataGroups == null) ? Sets.newHashSet() : Sets.newHashSet(dataGroups);
+    }
```
DwayneJengSage, line 120: Should the StringSets be mutable? (Here and in NoneOf and in the constructor.)

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -0,0 +1,148 @@
...
+    
+    @Override
+    public Subpopulation createSubpopulation(Subpopulation subpop) {
+        checkNotNull(subpop);
+        checkNotNull(subpop.getGuid());
+        checkNotNull(subpop.getStudyIdentifier());
+
+        if (subpop.getVersion() != null || subpop.isDeleted() || subpop.isRequired()) {
+            throw new BadRequestException("Subpopulation does not appear to be a new object");
+        }
```
DwayneJengSage, line 49: Unit test for attempting to create a required subpop?

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -0,0 +1,148 @@
...
+            return ImmutableList.of(subpop);
+        }
+        // Now filter out deleted subpopulations, if requested
+        return subpops.stream()
+            .filter(subpop -> includeDeleted || !subpop.isDeleted())
+            .collect(toImmutableList());
+    }
+    
+    @Override
+    public Subpopulation createDefaultSubpopulation(StudyIdentifier studyId) {
```
DwayneJengSage, line 101: Need unit test for createDefaultSubpopulation()

app/org/sagebionetworks/bridge/models/CriteriaUtils.java
```diff
@@ -0,0 +1,114 @@
...
+        private int specificity(Criteria crit) {
+            int specificity = 0;
+            specificity += (BridgeUtils.isEmpty(crit.getAllOfGroups())) ? 0 : 1;
+            specificity += (BridgeUtils.isEmpty(crit.getNoneOfGroups())) ? 0 : 1;
+            specificity += (crit.getMinAppVersion() == null) ? 0 : 1;
+            specificity += (crit.getMaxAppVersion() == null) ? 0 : 1;
+            return specificity;
+        }
+    };
+    */
```
DwayneJengSage: Remove commented code.

test/org/sagebionetworks/bridge/models/CriteriaUtilsTest.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @Test
+    public void noneOfGroupsDefinedButDontPreventMatch() {
+        ScheduleContext context = getContext(); // does not have group3, so it is matched
+        assertTrue(CriteriaUtils.matchCriteria(context, new SimpleCriteria(EMPTY_SET, Sets.newHashSet("group3"), null, null)));
+    }
+    
+    @Test
+    public void matchingWithMiminalContextDoesNotCrash() {
```
DwayneJengSage: Minor type: Miminal should be minimal

app/org/sagebionetworks/bridge/play/controllers/SubpopulationController.java
```diff
@@ -0,0 +1,68 @@
...
+    private SubpopulationService subpopService;
+    
+    @Autowired
+    public final void setSubpopulationService(SubpopulationService subpopService) {
+        this.subpopService = subpopService;
+    }
+
+    public Result getAllSubpopulations() {
+        UserSession session = getAuthenticatedSession(DEVELOPER);
+        Study study = studyService.getStudy(session.getStudyIdentifier());
```
DwayneJengSage: We're getting the study, but not using it.

app/org/sagebionetworks/bridge/models/studies/Subpopulation.java
alxdarksage, line 35: It is mutable. In some cases they will want and need to change the filter for it, particularly if they can't delete it. Added documentation.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
alxdarksage, line 71: Good point, Setting deleted to false before saving (setting it will have no effect for an update).

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulation.java
alxdarksage, line 120: Updated

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
alxdarksage, line 49: Added, also added test to verify that you can't set required or deleted on an update (so you can never set required or deleted when creating or updating. Required is internal and delete can only be changed through the delete call).

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
alxdarksage, line 101: Added variant of this that calls method directly

app/org/sagebionetworks/bridge/models/CriteriaUtils.java
alxdarksage: Removed

test/org/sagebionetworks/bridge/models/CriteriaUtilsTest.java
alxdarksage: Fixed

app/org/sagebionetworks/bridge/play/controllers/SubpopulationController.java
alxdarksage: Removed and using session.getStudyIdentifier() instead

PR #945, opened 2015-12-02
Splitting Redis keys in shared environments

alxdarksage:
Splitting Redis keys on shared environments by users, to avoid an issue we ran into yesterday where two machines were updating the same table.


app/org/sagebionetworks/bridge/redis/AbstractRedisKey.java
```diff
@@ -2,12 +2,21 @@
...
+import java.util.EnumSet;
+
+import org.sagebionetworks.bridge.config.BridgeConfig;
+import org.sagebionetworks.bridge.config.BridgeConfigFactory;
+import org.sagebionetworks.bridge.config.Environment;
+
 abstract class AbstractRedisKey implements RedisKey {
 
+    private final BridgeConfig config =  BridgeConfigFactory.getConfig();
+    private final EnumSet<Environment> SHARED_ENV = EnumSet.of(Environment.DEV, Environment.LOCAL);
```
DwayneJengSage, line 14: We create a lot of Redis keys. We should make these static final, so they won't be re-instantiated for every Redis key we create.

DwayneJengSage:
My concern here is that we've created something that's incredibly difficult (if not impossible) to test. At this point, it might be easier to create a RedisKeyFactor class that takes in the config (so we can mock it out) and refactor all the cases that use RedisKeys.

This is understandably a lot of up-front work (IntelliJ counts 72 uses of the RedisKey class), having something that can be programmatically tested will save us a lot of work in the long run.

DwayneJengSage:
An alternative, cheaper solution would to be just launch a separate Redis instance for Travis. Alternatively, https://sagebionetworks.jira.com/browse/BRIDGE-837 would also solve the problem.

PR #950, opened 2015-12-04
Studies have multiple subpopulations which each have their own consent stream.

alxdarksage:
Consents are now associated to subpopulations, which are associated to studies. Studies can end up having multiple consents.
- study will have a default subpopulation created if none exists, and this subpopulation's guid will be the study identifier. This associates the default subpopulation with the existing consents created and associated initially with a study. Subsequent subpopulations have a normal GUID
- user's enrollment date is still the first enrollment date to be submitted to the server
- user is consented if all required subpopulation consents have been signed
- user is up-to-date on consents if all required subpopulation consents were signed with the most recent versions of the consents
- we no longer send an email when the user updates their sharing status, because sharing isn't scoped to any specific consent

TODO: All tests are passing but I've made notes for additional tests (security around manipulating a subpopulation that's not in your test, for example, criteria filtering as applied to consents).


app/org/sagebionetworks/bridge/models/subpopulations/SubpopulationGuidImpl.java
```diff
@@ -0,0 +1,36 @@
+package org.sagebionetworks.bridge.models.subpopulations;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.Objects;
+
+public final class SubpopulationGuidImpl implements SubpopulationGuid {
```
DwayneJengSage: I really dislike this pattern. How much work are we actually saving by allowing Subpops and SubpopGuids to be used interchangeably? In contrast, using this pattern means we have an extra interface, and this interface is sometimes mutable and sometimes immutable, and if Subpops and SubpopGuids are interchangeable, we actually _lose_ a bit of type-safety.

Given that this pattern has already caused a lot of confusion for studies and surveys, I'd strongly prefer to not use this pattern here.

app/org/sagebionetworks/bridge/dynamodb/DynamoStudyConsent1.java
```diff
@@ -18,20 +18,20 @@
...
+    private String subpopGuid;
     private long createdOn;
     private boolean active;
     private String storagePath;
     private Long version;
 
     @Override
-    @DynamoDBHashKey
+    @DynamoDBHashKey(attributeName="studyKey")
     @JsonIgnore
```
DwayneJengSage: Do we want to expose the subpopGuid to users? Or do we still want to keep that under wraps like studyId?

app/org/sagebionetworks/bridge/dynamodb/DynamoStudyConsent1.java
```diff
@@ -91,14 +84,14 @@ public boolean equals(Object obj) {
         if (obj == null || getClass() != obj.getClass())
             return false;
         DynamoStudyConsent1 other = (DynamoStudyConsent1) obj;
-        return (Objects.equals(active, other.active) && Objects.equals(createdOn, other.createdOn) && 
-                Objects.equals(storagePath, other.storagePath) &&
-                Objects.equals(studyKey, other.studyKey) && Objects.equals(version, other.version));
+        return (Objects.equals(active, other.active) && Objects.equals(createdOn, other.createdOn)
+                && Objects.equals(storagePath, other.storagePath) && Objects.equals(subpopGuid, other.subpopGuid)
+                && Objects.equals(version, other.version));
     }
```
DwayneJengSage, line 89: Can we add a basic unit test for equalsVerifier?

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -104,15 +112,17 @@ public Subpopulation createDefaultSubpopulation(StudyIdentifier studyId) {
         subpop.setName("Default Consent Group");
         subpop.setMinAppVersion(0);
         subpop.setDefaultGroup(true);
+        // The first group is required until the study designers say otherwise
+        subpop.setRequired(true);
```
DwayneJengSage, line 116: Is it worth adding the required=true check to the unit test?

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -131,14 +141,15 @@ public Subpopulation getSubpopulation(StudyIdentifier studyId, String guid) {
...
-        Subpopulation subpop = getSubpopulation(studyId, guid);
+    public void deleteSubpopulation(StudyIdentifier studyId, SubpopulationGuid subpopGuid) {
+        Subpopulation subpop = getSubpopulation(studyId, subpopGuid);
         if (subpop == null || subpop.isDeleted()) {
             throw new EntityNotFoundException(Subpopulation.class);
         }
         if (subpop.isDefaultGroup()) {
             throw new BadRequestException("Cannot delete the default subpopulation for a study.");
         }
+        studyConsentDao.deleteAllConsents(subpopGuid);
```
DwayneJengSage, line 152: We should add a unit test to verify that consents are deleted when the subpop is deleted.

app/org/sagebionetworks/bridge/dynamodb/DynamoUserConsentDao.java
```diff
@@ -30,32 +30,26 @@
...
-        checkNotNull(studyConsent);
+        checkNotNull(subpopGuid);
         checkArgument(signedOn > 0L);
 
-        UserConsent activeConsent = getActiveUserConsent(healthCode, new StudyIdentifierImpl(studyConsent.getStudyKey()));
-        if (activeConsent != null && activeConsent.getConsentCreatedOn() == studyConsent.getCreatedOn()) {
-            throw new BridgeServiceException("Consent already exists.", HttpStatus.SC_CONFLICT);
+        UserConsent activeConsent = getActiveUserConsent(healthCode, subpopGuid);
+        if (activeConsent != null && activeConsent.getConsentCreatedOn() == consentCreatedOn) {
+            throw new EntityAlreadyExistsException(activeConsent);
```
DwayneJengSage: I think we originally used BridgeServiceException because EntityAlreadyExistsException exposes the UserConsent to the end-user, which we don't want to do.

test/org/sagebionetworks/bridge/models/accounts/ConsentStatusTest.java
```diff
@@ -0,0 +1,129 @@
...
+        statuses.add(REQUIRED_SIGNED_OBSOLETE);
+        statuses.add(OPTIONAL_UNSIGNED);
+        assertFalse(ConsentStatus.isConsentCurrent(statuses));
+        
+        statuses.clear();
+        statuses.add(REQUIRED_SIGNED_CURRENT);
+        statuses.add(REQUIRED_SIGNED_CURRENT);
+        statuses.add(OPTIONAL_UNSIGNED);
+        assertTrue(ConsentStatus.isConsentCurrent(statuses));
+    }
```
DwayneJengSage, line 120: Do we have the concept of optional and obselete? If so, do we want to add it to our test case here for completeness?

app/org/sagebionetworks/bridge/models/accounts/ConsentStatus.java
```diff
@@ -0,0 +1,112 @@
...
+    @JsonCreator
+    public ConsentStatus(@JsonProperty("name") String name, @JsonProperty("subpopulationGuid") String subpopulationGuid,
+            @JsonProperty("required") boolean isRequired, @JsonProperty("consented") boolean isConsented, 
+            @JsonProperty("mostRecentConsent") boolean isMostRecentConsent) {
+        this.name = checkNotNull(name);
+        this.subpopulationGuid = checkNotNull(subpopulationGuid);
+        this.required = isRequired;
+        this.consented = isConsented;
+        this.mostRecentConsent = isMostRecentConsent;
+    }
```
DwayneJengSage, line 67: Minor: We have two strings and three booleans. This is almost certainly more readable using the Builder pattern.

app/org/sagebionetworks/bridge/models/accounts/ConsentStatus.java
```diff
@@ -0,0 +1,112 @@
...
+        return required;
+    }
+
+    public boolean isConsented() {
+        return consented;
+    }
+    
+    public boolean isMostRecentConsent() {
+        return mostRecentConsent;
+    }
```
DwayneJengSage, line 87: For clarity, we should probably rename this to "hasSignedMostRecentConsent". "isMostRecentConsent" could be misread as meaning "the consent is the latest, but the user may not necessarily have signed it".

app/org/sagebionetworks/bridge/models/accounts/User.java
```diff
@@ -176,11 +153,35 @@ public boolean isInRole(Roles role) {
...
         return roleSet != null && !Collections.disjoint(roles, roleSet);
     }
+    
+    public List<ConsentStatus> getConsentStatuses() {
+        return consentStatuses;
+    }
+    
+    public void setConsentStatuses(List<ConsentStatus> consentStatuses) {
+        this.consentStatuses = (consentStatuses == null) ? ImmutableList.of() : ImmutableList.copyOf(consentStatuses);
+    }
```
DwayneJengSage, line 164: Should we make ConsentStatuses a map from subpopGuid to consentStatus? This will make it clear that there's only one consentStatus per subpop.

app/org/sagebionetworks/bridge/models/accounts/UserSessionInfo.java
```diff
@@ -26,35 +26,36 @@
...
         this.environment = ENVIRONMENTS.get(session.getEnvironment());
+        this.consentStatuses = session.getUser().getConsentStatuses();
     }
 
     public boolean isAuthenticated() {
         return authenticated;
     }
+    public List<ConsentStatus> getConsentStatuses() {
+        return consentStatuses;
+    }
```
DwayneJengSage, line 56: Similarly, a map here?

app/org/sagebionetworks/bridge/models/accounts/Withdrawal.java
```diff
@@ -19,5 +21,25 @@ public Withdrawal(@JsonProperty("reason") String reason) {
...
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj)
+            return true;
+        if (obj == null || getClass() != obj.getClass())
+            return false;
+        Withdrawal other = (Withdrawal) obj;
+        return (Objects.equals(reason, other.reason));
+    }
```
DwayneJengSage, line 38: Add equalsVerifier test?

app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
```diff
@@ -200,15 +202,16 @@ public Builder withContext(ScheduleContext context) {
...
             }
-            ScheduleContext context = new ScheduleContext(studyId, clientInfo, zone, endsOn, healthCode, events, now, userDataGroups);
-            // Not validating here. There are many tests to confirm that the scheduler will work with different 
-            // time windows, but the validator ensures the context object is within the declared allowable
-            // time window. This is validated in ScheduledActivityService.
-            //Validate.nonEntityThrowingException(VALIDATOR, context);
-            return context;
+            if (clientInfo == null) {
+                clientInfo = ClientInfo.UNKNOWN_CLIENT;
+            }
```
DwayneJengSage, line 213: Add a check for defaults (now, clientInfo) to the unit test?

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
```diff
@@ -55,9 +56,13 @@ public Result verifyEmail() throws Exception {
         JsonNode json = requestToJSON(request());
         EmailVerification emailVerification = parseJson(request(), EmailVerification.class);
         Study study = getStudyOrThrowException(json);
+        ScheduleContext context = new ScheduleContext.Builder()
+                .withStudyIdentifier(study)
+                .withClientInfo(getClientInfoFromUserAgentHeader()).build();
```
DwayneJengSage: What do we need from the ScheduleContext other than ClientInfo? As is, the ScheduleContext is incomplete, notably lacking dataGroups. I'm concerned someone in the future might forget that the ScheduleContext is incomplete and try to code against it and run into trouble because it's incomplete. For type safety, it would be better to just pass ClientInfo directly.

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
```diff
@@ -89,19 +94,22 @@ public Result resetPassword() throws Exception {
...
 
-        final JsonNode json = requestToJSON(request());
-        final SignIn signIn = parseJson(request(), SignIn.class);
-        final Study study = getStudyOrThrowException(json);
+        JsonNode json = requestToJSON(request());
+        SignIn signIn = parseJson(request(), SignIn.class);
+        Study study = getStudyOrThrowException(json);
+        ScheduleContext context = new ScheduleContext.Builder()
+                .withStudyIdentifier(study)
+                .withClientInfo(getClientInfoFromUserAgentHeader()).build();
```
DwayneJengSage: Similarly here.

app/org/sagebionetworks/bridge/services/AuthenticationServiceImpl.java
```diff
@@ -246,14 +257,23 @@ private UserSession getSessionFromAccount(final Study study, final Account accou
...
-        user.setSignedMostRecentConsent(consentService.hasUserSignedActiveConsent(study, user));
-        user.setConsent(consentService.hasUserConsentedToResearch(study, user));
 
+        // now that we know more about this user, we can expand on the request context.
+        context = new ScheduleContext.Builder()
+                .withContext(context)
+                .withHealthCode(healthCode)
+                .withUserDataGroups(user.getDataGroups())
+                .withStudyIdentifier(study.getIdentifier()) // probably already set
+                .build();
```
DwayneJengSage, line 272: Similarly, there are a lot of fields in ScheduleContext that aren't needed to get consent. This suggests to me that we should have a separate "ConsentContext" class that contains only the fields needed for consents.

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
```diff
@@ -25,36 +26,41 @@
...
     }
 
+    @Deprecated
     public Result getConsentSignature() throws Exception {
         final UserSession session = getAuthenticatedAndConsentedSession();
         final Study study = studyService.getStudy(session.getStudyIdentifier());
-
-        ConsentSignature sig = consentService.getConsentSignature(study, session.getUser());
-        return ok(ConsentSignature.SIGNATURE_WRITER.writeValueAsString(sig));
+        return getConsentSignatureV2(study.getIdentifier());
```
DwayneJengSage: getConsentSignatureV2() already calls getAuthenticatedAndConsentedSession() and already calls getStudy(). We can delete those calls from getConsentSignature() to avoid duplicate work.

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
```diff
@@ -25,36 +26,41 @@
...
     }
 
     @Deprecated
     public Result giveV1() throws Exception {
-        return giveConsentForVersion(1);
+        final UserSession session = getAuthenticatedSession();
+        final Study study = studyService.getStudy(session.getStudyIdentifier());
+        return giveConsentForVersion(1, SubpopulationGuid.create(study.getIdentifier()));
```
DwayneJengSage: Similarly here.

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
```diff
@@ -25,36 +26,41 @@
...
     }
 
+    @Deprecated
     public Result giveV2() throws Exception {
-        return giveConsentForVersion(2);
+        final UserSession session = getAuthenticatedSession();
+        final Study study = studyService.getStudy(session.getStudyIdentifier());
```
DwayneJengSage: Similarly here.

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
```diff
@@ -25,36 +26,41 @@
...
+        return giveV3(study.getIdentifier());
```
DwayneJengSage: Why call give3() which calls giveForVersion(2) when you can call giveForVersion(2) directly?

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
```diff
@@ -25,36 +26,41 @@
...
     }
 
+    @Deprecated
     public Result emailCopy() throws Exception {
         final UserSession session = getAuthenticatedAndConsentedSession();
         final Study study = studyService.getStudy(session.getStudyIdentifier());
```
DwayneJengSage: Similarly here.

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
```diff
@@ -74,18 +80,48 @@ public Result changeSharingScope() throws Exception {
         return changeSharingScope(sharing.getSharingScope(), "Data sharing has been changed.");
     }
     
+    @Deprecated
     public Result withdrawConsent() throws Exception {
         final UserSession session = getAuthenticatedAndConsentedSession();
+        final Study study = studyService.getStudy(session.getStudyIdentifier());
```
DwayneJengSage, line 82: Similarly here.

app/org/sagebionetworks/bridge/play/controllers/UserManagementController.java
```diff
@@ -34,8 +35,12 @@ public Result createUser() throws Exception {
         SignUp signUp = parseJson(request(), SignUp.class);
 
         boolean consent = JsonUtils.asBoolean(node, CONSENT_FIELD);
+        
+        ScheduleContext context = new ScheduleContext.Builder()
+                .withStudyIdentifier(session.getStudyIdentifier())
+                .withClientInfo(getClientInfoFromUserAgentHeader()).build();
```
DwayneJengSage: Similarly here.

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
```diff
@@ -74,18 +80,48 @@ public Result changeSharingScope() throws Exception {
...
         return okResult("User has been withdrawn from the study.");
     }
+    
+    public Result emailCopyV2(String guid) {
+        final UserSession session = getAuthenticatedAndConsentedSession();
+        final Study study = studyService.getStudy(session.getStudyIdentifier());
 
+        consentService.emailConsentAgreement(study, SubpopulationGuid.create(guid), session.getUser());
+        return okResult("Emailed consent.");
+    }
```
DwayneJengSage, line 119: Unit test for emailCopyV2()?

app/org/sagebionetworks/bridge/play/controllers/StudyConsentController.java
```diff
@@ -21,54 +24,132 @@
...
+        return addConsentV2(session.getStudyIdentifier().getIdentifier());
+    }
+
+    @Deprecated
+    public Result publishConsent(String createdOn) throws Exception {
+        UserSession session = getAuthenticatedSession(RESEARCHER);
+        return publishConsentV2(session.getStudyIdentifier().getIdentifier(), createdOn);
+    }
+    
+    // V2: consents associated to a subpopulation
```
DwayneJengSage, line 78: Unit tests for StudyConsentController V2 APIs?

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
```diff
@@ -50,6 +51,8 @@
 public class AuthenticationServiceImplTest {
```
DwayneJengSage, line 54: Tests for checking consent statuses in the user session in signIn and verifyEmail?

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -117,6 +121,11 @@ public void before() {
         SurveyResponseService surveyResponseService = mock(SurveyResponseService.class);
         when(surveyResponseService.createSurveyResponse(
             any(GuidCreatedOnVersionHolder.class), anyString(), any(List.class), anyString())).thenReturn(surveyResponse);
+
+        Subpopulation subpopulation = Subpopulation.create();
+        
+        SubpopulationService subpopService = mock(SubpopulationService.class);
+        when(subpopService.getSubpopulations(any())).thenReturn(Lists.newArrayList(subpopulation));
```
DwayneJengSage: For completeness, can we have this return multiple subpopulations, then verify that we correctly use the earliest consent?

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -82,7 +86,7 @@ public void before() {
         when(consent.getSignedOn()).thenReturn(ENROLLMENT.getMillis()); 
         
         userConsentDao = mock(UserConsentDao.class);
-        when(userConsentDao.getActiveUserConsent(any(String.class), any(StudyIdentifier.class))).thenReturn(consent);
+        when(userConsentDao.getActiveUserConsent(eq(HEALTH_CODE), any(SubpopulationGuid.class))).thenReturn(consent);
```
DwayneJengSage: This should check that it's the same as the SubpopGuid we return from our mock getSubpopulations()

app/org/sagebionetworks/bridge/services/SubpopulationService.java
```diff
@@ -41,11 +59,32 @@ public Subpopulation createSubpopulation(Study study, Subpopulation subpop) {
...
+    public Subpopulation createDefaultSubpopulation(Study study) {
+        SubpopulationGuid subpopGuid = SubpopulationGuid.create(study.getIdentifier());
+        // Migrating, studies will already have consents so don't create and publish a new one
+        // unless this is part of the creation of a new study after the introduction of subpopulations.
+        if (studyConsentService.getAllConsents(subpopGuid).isEmpty()) {
+            StudyConsentView view = studyConsentService.addConsent(subpopGuid, defaultConsentDocument);
+            studyConsentService.publishConsent(study, subpopGuid, view.getCreatedOn());
+        }
+        return subpopDao.createDefaultSubpopulation(study.getStudyIdentifier());
+    }
```
DwayneJengSage, line 87: Unit test for this? (For both cases, where consents already exist, and where consents don't already exist.)

app/org/sagebionetworks/bridge/models/subpopulations/SubpopulationGuidImpl.java
alxdarksage: Suggest an alternative. SubpopulationGuid as a final class, that can be returned from Subpopulation implementations as a key?

I feel that as long as SubpopGuid is used where only the key is needed, the type safety is correct... and obviously you can't go the other way and ask for the GUID object when you need to work on an entire subpopulation model. I haven't found this to present any issue with StudyIdentifier.

app/org/sagebionetworks/bridge/dynamodb/DynamoStudyConsent1.java
alxdarksage: Hmmm. Maybe expose it because there's more than one in a study. Will remove annotation and update docs. Also updated StudyConsentView which is the version users see.

app/org/sagebionetworks/bridge/dynamodb/DynamoStudyConsent1.java
alxdarksage, line 89: Added

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
alxdarksage, line 116: Added checks that it can be set/changed in CRUD operations, and that it's set to true when a default subpopulation is created.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
alxdarksage, line 152: Added

app/org/sagebionetworks/bridge/dynamodb/DynamoUserConsentDao.java
alxdarksage: Seems likely, but we've got an authenticated user who can retrieve their consent, so I don't know that we need to hide it like this. This may have been an overabundance of caution.

test/org/sagebionetworks/bridge/models/accounts/ConsentStatusTest.java
alxdarksage, line 120: It must be possible. Added.

app/org/sagebionetworks/bridge/models/accounts/ConsentStatus.java
alxdarksage, line 67: Added.

app/org/sagebionetworks/bridge/models/accounts/ConsentStatus.java
alxdarksage, line 87: Okay, changed to "signedMostRecentConsent" as in UserSession.

app/org/sagebionetworks/bridge/models/accounts/User.java
alxdarksage, line 164: We'll always iterate through the values of this map, but it won't harm anything. Changed.

app/org/sagebionetworks/bridge/models/accounts/UserSessionInfo.java
alxdarksage, line 56: Yes, added it all the way through to the API and changed the documentation.

app/org/sagebionetworks/bridge/models/accounts/Withdrawal.java
alxdarksage, line 38: Added

app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
alxdarksage, line 213: Added tests for defaults and requirement that study Id be set before construction.

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
alxdarksage: This is much better, removed ScheduleContext from the authentication methods.

app/org/sagebionetworks/bridge/models/subpopulations/SubpopulationGuidImpl.java
DwayneJengSage: As per IM conversation, we ran into a lot of trouble with this for both Studies and Surveys, specifically when we passed in the mutable full object when we actually needed an immutable ID holder, or when we unnecessarily loaded the full object when we only needed.

What we should do instead is have Subpop and SubpopGuid be separate objects, so you can't mistakenly use a Subpop for a SubpopGuid.

app/org/sagebionetworks/bridge/dynamodb/DynamoUserConsentDao.java
DwayneJengSage: As per IM conversation, we don't currently expose the UserConsent anywhere. I'm okay with exposing the UserConsent, but if we do, we'll need to do the work to make sure it serializes correctly and that it doesn't expose certain fields to the user, most notably healthCode and studyId

app/org/sagebionetworks/bridge/stormpath/StormpathAccount.java
```diff
@@ -53,15 +56,17 @@
...
-    private final String consentSignaturesKey;
     private final String oldHealthIdVersionKey;
     private final String oldConsentSignatureKey;
     private final Set<Roles> roles;
+    /**
+     * All signatures saved under a key that includes the subpopulation GUID. Don't
+     * use SubpopulationGuid interface as the key, because different implementations have different
+     * equals/hashCode values. Use the GUID string directly.
+     */
+    private Map<String, List<ConsentSignature>> allSignatures;
```
DwayneJengSage: This is another reason why having SubpopGuid being separate from Subpop is a good idea. If we can guarantee that SubpopGuid is immutable, we can use it as the map key.

app/org/sagebionetworks/bridge/stormpath/StormpathAccount.java
```diff
@@ -73,26 +78,28 @@
...
-                signatures.add(sig);
+        for (SubpopulationGuid subpopGuid : subpopGuids) {
+            List<ConsentSignature> signatures = decryptJSONFrom(subpopGuid.getGuid()+CONSENT_SIGNATURES_SUFFIX, CONSENT_SIGNATURES_TYPE);
+            if (signatures == null || signatures.isEmpty()) {
+                ConsentSignature sig = decryptJSONFrom(subpopGuid.getGuid()+CONSENT_SIGNATURE_SUFFIX, ConsentSignature.class);
+                if (sig != null) {
+                    getConsentSignatureHistory(subpopGuid).add(sig);
+                }
+            } else {
+                getConsentSignatureHistory(subpopGuid).addAll(signatures);
```
DwayneJengSage, line 89: Add unit tests for (a) reading consents for multiple subpops from CustomData and (b) writing consents for multiple subpops into CustomData.

app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
```diff
@@ -74,6 +79,10 @@ public void setStormpathClient(Client client) {
     public void setStudyService(StudyService studyService) {
         this.studyService = studyService;
     }
+    @Autowired
+    public void setSubpopulationService(SubpopulationService subpopService) {
```
DwayneJengSage: Minor: final

conf/logback.xml
```diff
@@ -14,7 +14,10 @@
   <logger name="org.sagebionetworks.bridge" level="DEBUG" additivity="false">
     <appender-ref ref="STDOUT" />
   </logger>
-  <logger name="org.springframework" level="INFO" additivity="false">
+  <logger name="org.sagebionetworks.bridge.play" level="ERROR" additivity="false">
```
DwayneJengSage: Our play controllers should have error level of INFO or lower. If we log it, we want to see it. If we don't want to see it, we shouldn't log it.

Specifically, this means we lose things like MetricsInterceptor.

app/org/sagebionetworks/bridge/services/AuthenticationServiceImpl.java
alxdarksage, line 272: Yes, absolutely. Once I turn to the schedules PR which also depends on some of the same information (which is being extracted out to the Criteria and CriteriaUtils classes), I'll reduce and/or change this. I'm not sure at this time what that will look like because we're still refactoring this. I think there's a CriteriaContext object with min/max app version\, clientInfo and data groups, and schedule context includes additional information. The consent matching may just use CriteriaContext.

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
alxdarksage: We need the user's study for backwards compatibility. Removed getting the study which we don't need.

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
alxdarksage: Changed all of these (removed study).

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
alxdarksage: Changed

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
alxdarksage, line 82: Removed study in both these places

app/org/sagebionetworks/bridge/play/controllers/UserManagementController.java
alxdarksage: Removed clientInfo entirely from this set of calls. You also can't, at this time, select anything but the default subpopulation when creating a user (as an admin, virtually always to set up a test). I can add this later.

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
alxdarksage, line 119: Added to ConsentControllerMockedTest.

test/org/sagebionetworks/bridge/models/accounts/UserTest.java
```diff
@@ -68,4 +79,63 @@ public void userIsInRoleSet() {
...
+        assertFalse(user.hasSignedMostRecentConsent());
+        
+        // A required consent is not consented
+        user = new User();
+        user.setConsentStatuses(Lists.newArrayList(
+            new ConsentStatus("Name", "guid1", true, true, false),
+            new ConsentStatus("Name", "guid2", true, false, false),
+            new ConsentStatus("Name", "guid3", false, false, false)
+        ));
+        assertFalse(user.hasSignedMostRecentConsent());
```
DwayneJengSage, line 140: For completeness, we should also add a "true" case here.

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerTest.java
```diff
@@ -103,6 +103,9 @@ public void testCode() throws Exception {
 
                 String output = jedisOps.get(sessionToken);
                 assertNull("Should no longer be session data", output);
+                } catch(Exception e) {
+                    e.printStackTrace();
+                }
```
DwayneJengSage: Remove this try-catch, as it converts test failures into successes.

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerTest.java
```diff
@@ -123,13 +126,15 @@ public void testCode() throws Exception {
...
+                    // Now, try and access the wrong study, you do not get it.
                     // There's actually no easy way to request another study at this point...
                     request = WS.url(TEST_BASE_URL + STUDIES_URL + secondStudy.getIdentifier());
                     request.setHeader(BridgeConstants.SESSION_TOKEN_HEADER, cookie.getValue());
                     response = request.get().get(TIMEOUT);
                     assertEquals("{\"message\":\"Caller does not have permission to access this service.\"}", response.getBody());
-                    
+
+                } catch(Exception e) {
+                    e.printStackTrace();
```
DwayneJengSage: Similarly here.

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerTest.java
```diff
@@ -165,6 +171,8 @@ public void testCode() throws Exception {
                     assertEquals(412, response.getStatus());
                     assertTrue(bodyString.contains("\"authenticated\":true"));
                     assertTrue(bodyString.contains("\"consented\":false"));
+                } catch(Exception e) {
+                    e.printStackTrace();
```
DwayneJengSage: Similarly here.

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerTest.java
```diff
@@ -172,10 +180,13 @@ public void testCode() throws Exception {
                 }
             }
         });
+        } catch(Exception e) {
+            e.printStackTrace();
+        }
```
DwayneJengSage: Similarly here.

app/org/sagebionetworks/bridge/play/controllers/StudyConsentController.java
alxdarksage, line 78: Added tests for all the methods that should accept a guid and pass it property to the services.

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
alxdarksage, line 54: Added a test

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: Added, set one to have a signedOn date that's high enough it breaks things (Long.MAX_VALUE). Test at the end calls that method and verifies it's taking the lower of the two numbers.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: In adding two different subpopulations, this was converted to look for specific arguments.

app/org/sagebionetworks/bridge/services/SubpopulationService.java
alxdarksage, line 87: Added tests that studyConsentService is/is not called depending on existence of consents.

app/org/sagebionetworks/bridge/dynamodb/DynamoStudyConsent1.java
alxdarksage: I'm removing the JsonIgnore annotation, and added a test to ensure StudyConsent serializes correctly because it's exposed in one API call where we return a list of consents. It looks the same as StudyConsentView, but doesn't have the contents of the consent (which is now documented).

app/org/sagebionetworks/bridge/dynamodb/DynamoUserConsentDao.java
alxdarksage: Reverting, we don't have the signature here

app/org/sagebionetworks/bridge/stormpath/StormpathAccount.java
alxdarksage: Changed to use SubpopulationGuid

app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
alxdarksage: All setters changed

conf/logback.xml
alxdarksage: Reverted...

test/org/sagebionetworks/bridge/models/accounts/UserTest.java
alxdarksage, line 140: added

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerTest.java
alxdarksage: Rethrowing (without the catch, you don't get a stack trace in the error logs, this seems to be something about the Play test infrastructure being used here).

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerTest.java
alxdarksage: Removed

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerTest.java
alxdarksage: Removed both try blocks

app/org/sagebionetworks/bridge/stormpath/StormpathAccount.java
alxdarksage, line 89: I think StormpathAccountTest.multipleConsentsAreMaintainedSeparately() does this? Or do you want this in the DAO test which will actually save to Stormpath?

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerTest.java
DwayneJengSage: As per IM conversation, Play does in fact produce a stack trace, so we should remove the catch block.

app/org/sagebionetworks/bridge/stormpath/StormpathAccount.java
DwayneJengSage, line 89: As per IM conversation, this test only tests the in-memory map. We need something that tests the marshalling to/from CustomData.

test/org/sagebionetworks/bridge/play/controllers/SubpopulationControllerTest.java
```diff
@@ -185,15 +190,60 @@ public void deleteSubpopulation() throws Exception {
...
     }
-
+    
+    @Test(expected = UnauthorizedException.class)
+    public void getAllSubpopulationsRequiresDeveloper() throws Exception {
+        User user = new User();
+        doReturn(user).when(session).getUser(); // no developer role
+        
+        controller.getAllSubpopulations();
+        verifyNoMoreInteractions(subpopService);
```
DwayneJengSage: FYI, none of your "verifyNoMoreInteractions()" will be called, since the controller always throws before it reaches this point. (Here and in the next N tests.)

alxdarksage: Removed

test/org/sagebionetworks/bridge/services/ConsentServiceImplTest.java
```diff
@@ -104,54 +119,37 @@ public void after() {
...
-        ConsentSignature returnedSig = consentService.getConsentSignature(testUser.getStudy(), testUser.getUser());
+        consentService.consentToResearch(testUser.getStudy(), subpopulation.getGuid(), testUser.getUser(), signature, SharingScope.NO_SHARING, false);
         
+        List<ConsentStatus> statuses = consentService.getConsentStatuses(context);
+        assertFalse(statuses.isEmpty());
+
+        ConsentSignature returnedSig = consentService.getConsentSignature(testUser.getStudy(), subpopulation.getGuid(), testUser.getUser());
         assertEquals("Eggplant McTester", returnedSig.getName());
         assertEquals("1970-01-01", returnedSig.getBirthdate());
         assertEquals(TestConstants.DUMMY_IMAGE_DATA, returnedSig.getImageData());
```
DwayneJengSage, line 167: The test you combined this with also had a check to make sure ParticipantOptions (sharing scope) was set, which isn't present in the new test. Should we add this back in?

test/org/sagebionetworks/bridge/services/ConsentServiceImplTest.java
```diff
@@ -104,54 +119,37 @@ public void after() {
...
         long signedOn = signature.getSignedOn();
         
-        consentService.consentToResearch(testUser.getStudy(), testUser.getUser(), signature, SharingScope.NO_SHARING, false);
-        assertTrue(consentService.hasUserConsentedToResearch(testUser.getStudy(), testUser.getUser()));
-
-        ConsentSignature returnedSig = consentService.getConsentSignature(testUser.getStudy(), testUser.getUser());
+        consentService.consentToResearch(testUser.getStudy(), subpopulation.getGuid(), testUser.getUser(), signature, SharingScope.NO_SHARING, false);
         
+        List<ConsentStatus> statuses = consentService.getConsentStatuses(context);
+        assertFalse(statuses.isEmpty());
```
DwayneJengSage: We should add a check to ensure the consent statuses are actually signed and not just present.

test/org/sagebionetworks/bridge/services/ConsentServiceImplTest.java
```diff
@@ -179,162 +188,194 @@ public void cannotConsentIfTooYoung() {
...
+        consentService.consentToResearch(testUser.getStudy(), subpopulation.getGuid(), testUser.getUser(), consent,
                 SharingScope.SPONSORS_AND_PARTNERS, false);
 
         // Consent exists, user is consented
-        assertTrue(consentService.hasUserConsentedToResearch(testUser.getStudy(), testUser.getUser()));
-        assertTrue(testUser.getUser().doesConsent());
-        assertNotNull(consentService.getConsentSignature(testUser.getStudy(), testUser.getUser()));
+        ConsentStatus status = consentService.getConsentStatuses(context).get(0);
+        assertTrue(status.isConsented());
+        assertNotNull(consentService.getConsentSignature(testUser.getStudy(), subpopulation.getGuid(), testUser.getUser()));
```
DwayneJengSage: You removed the check on user.doesConsent(). We should add this back into the test, to exercise the code where we update the user with their consent status, here and immediately below.

test/org/sagebionetworks/bridge/services/ConsentServiceImplTest.java
```diff
@@ -179,162 +188,194 @@ public void cannotConsentIfTooYoung() {
...
         
         // Now withdraw consent
-        consentService.withdrawConsent(testUser.getStudy(), testUser.getUser(), WITHDRAWAL, UNIX_TIMESTAMP);
+        consentService.withdrawConsent(testUser.getStudy(), subpopulation.getGuid(), testUser.getUser(), WITHDRAWAL, UNIX_TIMESTAMP);
         
         // Now user is not consented
-        assertFalse(consentService.hasUserConsentedToResearch(testUser.getStudy(), testUser.getUser()));
-        assertFalse(testUser.getUser().doesConsent());
+        status = consentService.getConsentStatuses(context).get(0);
+        assertFalse(status.isConsented());
```
DwayneJengSage: Restore check for user.doesConsent() here.

test/org/sagebionetworks/bridge/services/StudyConsentServiceImplTest.java
```diff
@@ -51,20 +52,22 @@
...
     public void before() {
+        String id = TestUtils.randomName(StudyConsentServiceImplTest.class);
+        
         study = new DynamoStudy();
-        study.setIdentifier(STUDY_ID.getIdentifier());
-        study.setName("A test study");
-        study.setSponsorName("A sponsor name");
+        study.setIdentifier(id);
+        study.setName(id);
+        study.setSponsorName(id);
```
DwayneJengSage: Minor: We probably shouldn't stuff the id into the name and sponsorName, just to make sure our tests don't "cross the streams", so to speak.

test/org/sagebionetworks/bridge/services/email/ParticipantRosterProviderTest.java
```diff
@@ -84,33 +84,30 @@ public void participantsCorrectlyDescribedInCSV() {
...
         assertEquals(output, provider.createParticipantTSV());
         
         participant.remove(UserProfile.SHARING_SCOPE_FIELD);
-        output = headerString + row("test@test.com","","Last","","false","abc","","false","AAA");
-        assertEquals(output, provider.createParticipantTSV());
-        
-        participant.setExternalId(null);
-        output = headerString + row("test@test.com","","Last","","false","","","false","AAA");
+        output = headerString + row("test@test.com","","Last","","false","abc","","false","AAA", "Default Consent Group");
         assertEquals(output, provider.createParticipantTSV());
```
DwayneJengSage, line 115: We accidentally removed the test case for removing externalId?

test/org/sagebionetworks/bridge/stormpath/StormpathAccountDaoMockTest.java
```diff
@@ -0,0 +1,154 @@
...
+        EmailVerification verification = new EmailVerification("tokenAAA");
+        
+        Client client = mock(Client.class);
+        Tenant tenant = mock(Tenant.class);
+        when(client.getCurrentTenant()).thenReturn(tenant);
+        dao.setStormpathClient(client);
+        dao.setSubpopulationService(mockSubpopService());
+        
+        dao.verifyEmail(study, verification);
+        verify(client).verifyAccountEmail("tokenAAA");
```
DwayneJengSage, line 64: We should also verify the return value.

test/org/sagebionetworks/bridge/stormpath/StormpathAccountDaoMockTest.java
```diff
@@ -0,0 +1,154 @@
...
+        dao.setStormpathClient(client);
+        dao.setSubpopulationService(mockSubpopService());
+        
+        String random = RandomStringUtils.randomAlphabetic(5);
+        String email = "bridge-testing+"+random+"@sagebridge.org";
+        SignUp signUp = new SignUp(random, email, PASSWORD, null, null);
+        dao.signUp(study, signUp, false);
+
+        ArgumentCaptor<com.stormpath.sdk.account.Account> argument = ArgumentCaptor.forClass(com.stormpath.sdk.account.Account.class);
+        verify(directory).createAccount(argument.capture(), anyBoolean());
```
DwayneJengSage, line 143: Also verify return value.

test/org/sagebionetworks/bridge/models/accounts/ConsentStatusTest.java
```diff
@@ -0,0 +1,159 @@
...
+    
+    @Test
+    public void toMapList() {
+        Map<SubpopulationGuid,ConsentStatus> map = ConsentStatus.toMap(TestConstants.REQUIRED_SIGNED_CURRENT);
+
+        SubpopulationGuid guid = SubpopulationGuid.create(TestConstants.REQUIRED_SIGNED_CURRENT.getSubpopulationGuid());
+        
+        assertEquals(TestConstants.REQUIRED_SIGNED_CURRENT, map.get(guid));
+    }
```
DwayneJengSage: toMapList() doesn't do what it says. It calls the varargs variant of toMap()

app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
```diff
@@ -42,189 +46,191 @@
...
+    public User consentToResearch(Study study, SubpopulationGuid subpopGuid, User user, ConsentSignature consentSignature, 
+            SharingScope sharingScope, boolean sendEmail) {
         checkNotNull(study, Validate.CANNOT_BE_NULL, "study");
+        checkNotNull(subpopGuid, Validate.CANNOT_BE_NULL, "subpopulationGuid");
         checkNotNull(user, Validate.CANNOT_BE_NULL, "user");
         checkNotNull(consentSignature, Validate.CANNOT_BE_NULL, "consentSignature");
         checkNotNull(sharingScope, Validate.CANNOT_BE_NULL, "sharingScope");
 
-        if (user.doesConsent()) {
+        ConsentStatus status = ConsentStatus.forSubpopulation(user.getConsentStatuses().values(), subpopGuid);
```
DwayneJengSage: Why iterate through the value set when you have a perfectly good map keyed off the subpopGuid that you already know?

app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
```diff
@@ -42,189 +46,191 @@
...
+        checkNotNull(context);
+        checkNotNull(context.getHealthCode());
+        
+        return subpopService.getSubpopulationForUser(context).stream().map(subpop -> {
+            boolean consented = userConsentDao.hasConsented(context.getHealthCode(), subpop.getGuid());
+            boolean mostRecent = hasUserSignedActiveConsent(context.getHealthCode(), subpop.getGuid());
+            return new ConsentStatus.Builder().withName(subpop.getName()).withGuid(subpop.getGuid())
+                    .withRequired(subpop.isRequired()).withConsented(consented).withSignedMostRecentConsent(mostRecent)
+                    .build();
+        }).collect(BridgeCollectors.toImmutableList());
```
DwayneJengSage: It would be trivially easy to convert this code into building a map instead of a list. That way, it's maps all the way down, and we don't have to do awkward conversions between maps and lists, and we don't have to use that awkward forSubpopulation() method.

app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
```diff
@@ -258,48 +269,38 @@ public void emailConsentAgreement(final Study study, final User user) {
...
+            for (ConsentStatus status : user.getConsentStatuses().values()) {
+                if (status.getSubpopulationGuid().equals(subpopGuid.getGuid())) {
+                    ConsentStatus updatedStatus = new ConsentStatus.Builder().withConsentStatus(status)
+                            .withConsented(consented).withSignedMostRecentConsent(consented).build();
+                    updatedStatuses.add(updatedStatus);
+                } else {
+                    updatedStatuses.add(status);
+                }
+            }
+            user.setConsentStatuses(ConsentStatus.toMap(updatedStatuses));
```
DwayneJengSage: I feel like this code would be much more elegant with a temporary list instead of a temporary map.

test/org/sagebionetworks/bridge/services/ConsentServiceImplTest.java
alxdarksage: Added ConsentStatus.isUserConsented() instead, also added check for user.doesConsent()

test/org/sagebionetworks/bridge/services/ConsentServiceImplTest.java
alxdarksage: Added back

test/org/sagebionetworks/bridge/services/ConsentServiceImplTest.java
alxdarksage: Added back

test/org/sagebionetworks/bridge/services/StudyConsentServiceImplTest.java
alxdarksage: changed

test/org/sagebionetworks/bridge/services/email/ParticipantRosterProviderTest.java
alxdarksage, line 115: Seems like it, added it back.

app/org/sagebionetworks/bridge/stormpath/StormpathAccount.java
alxdarksage, line 89: Added to this test to verify copy back-and-forth to the customData object. It's now checking the second object. But I still believe the DAO test is exercising the important stuff here: encryption and how Stormpath behaves.

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
alxdarksage: Changed here too

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
alxdarksage: Study also removed

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
alxdarksage: removed

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerTest.java
alxdarksage: Removed

test/org/sagebionetworks/bridge/services/ConsentServiceImplTest.java
alxdarksage, line 167: I cleaned up this test but didn't see the participant options test, I'll add now.

test/org/sagebionetworks/bridge/services/ConsentServiceImplTest.java
alxdarksage: Rewrote so the status objects in and out of the user, as well as the boolean calculated from the collection, are all tested every time in helper assertion methods.

test/org/sagebionetworks/bridge/services/ConsentServiceImplTest.java
alxdarksage: done

test/org/sagebionetworks/bridge/stormpath/StormpathAccountDaoMockTest.java
alxdarksage, line 64: added

test/org/sagebionetworks/bridge/stormpath/StormpathAccountDaoMockTest.java
alxdarksage, line 143: added

test/org/sagebionetworks/bridge/models/accounts/ConsentStatusTest.java
alxdarksage: No longer in code

app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
alxdarksage: changed

app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
alxdarksage: As ConsentStatus is now stored in a map, there are a number of these places where I've converted the logic to work on a map now

app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
alxdarksage: I think you meant map not list, it's now a map builder.

test/org/sagebionetworks/bridge/services/ConsentServiceImplTest.java
alxdarksage, line 167: Added test that sharing status is set and unset as you consent/withdraw.

app/org/sagebionetworks/bridge/models/accounts/User.java
```diff
@@ -30,16 +35,18 @@
     private String email;
     private String healthCode;
     private String studyKey;
-    private boolean consent;
-    private boolean signedMostRecentConsent;
     private SharingScope sharingScope;
     private Set<Roles> roles = Sets.newHashSet();
     private Set<String> dataGroups = Sets.newHashSet();
+    @JsonDeserialize(keyUsing = SubpopulationGuidDeserializer.class)
+    private Map<SubpopulationGuid,ConsentStatus> consentStatuses;
```
DwayneJengSage, line 39: Serialization test?

Also, where are we serializing this to JSON? This contains the health code and study Id, so we can't expose this to end users.

test/org/sagebionetworks/bridge/models/accounts/UserSessionInfoTest.java
```diff
@@ -52,9 +59,20 @@ public void userSessionInfoSerializesCorrectly() throws Exception {
...
+        assertTrue(consentStatus.get("required").asBoolean());
+        assertTrue(consentStatus.get("consented").asBoolean());
+        assertFalse(consentStatus.get("signedMostRecentConsent").asBoolean());
+        assertEquals("ConsentStatus", consentStatus.get("type").asText());
+        assertEquals(6, consentStatus.size());
         
         // ... and no things that shouldn't be there
-        assertEquals(11, node.size());
+        assertEquals(12, node.size());
     }
```
DwayneJengSage, line 76: Test for deserializing this from JSON back to a POJO?

test/org/sagebionetworks/bridge/services/AuthenticationServiceImplTest.java
```diff
@@ -330,4 +335,24 @@ public void requestResetPasswordLooksSuccessfulWhenNoAccount() throws Exception
...
+        
+        // This is the object we pass back to the user, we want to see the statuses copied or present
+        // all the way from the user to the sessionInfo. We test elsewhere that these are properly 
+        // serialized/deserialized (SubpopulationGuidDeserializer)
+        UserSessionInfo sessionInfo = new UserSessionInfo(user.getSession());
+        
+        ConsentStatus status = sessionInfo.getConsentStatuses().get(guid);
+        assertTrue(status.isConsented());
+        assertEquals(testUser.getStudyIdentifier().getIdentifier(), status.getSubpopulationGuid());
+    }
```
DwayneJengSage, line 359: Delete test user?

test/org/sagebionetworks/bridge/services/SubpopulationServiceTest.java
```diff
@@ -79,25 +109,67 @@ public void createSubpopulation() {
...
+        
+        SubpopulationGuid defaultGuid = SubpopulationGuid.create("test-study");
+        
+        ArgumentCaptor<StudyConsentForm> captor = ArgumentCaptor.forClass(StudyConsentForm.class);
+        
+        when(studyConsentService.getAllConsents(defaultGuid)).thenReturn(Lists.newArrayList());
+        when(studyConsentService.addConsent(eq(defaultGuid), any())).thenReturn(view);
+        
+        // No consents, so we add and publish one.
+        service.createDefaultSubpopulation(study);
```
DwayneJengSage: Verify return value?

test/org/sagebionetworks/bridge/services/SubpopulationServiceTest.java
```diff
@@ -79,25 +109,67 @@ public void createSubpopulation() {
...
+    @Test
+    public void createDefaultSubpopulationWhereConsentsExist() {
+        Study study = new DynamoStudy();
+        study.setIdentifier("test-study");
+        
+        SubpopulationGuid defaultGuid = SubpopulationGuid.create("test-study");
+        
+        when(studyConsentService.getAllConsents(defaultGuid)).thenReturn(Lists.newArrayList(new DynamoStudyConsent1()));
+        
+        service.createDefaultSubpopulation(study);
```
DwayneJengSage: Verify return value?

app/org/sagebionetworks/bridge/models/accounts/User.java
alxdarksage, line 39: No ConsentStatus doesn't contain healthCode, it's a summary of the status of consent to that consent. It only has the subpopulation guid. SubpopulationGuidDeserializerTest tests the serialization of this map.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulation.java
```diff
@@ -53,14 +59,25 @@ public void setStudyIdentifier(String studyIdentifier) {
...
+    @Override
+    @DynamoDBIgnore
+    @JsonIgnore
+    public SubpopulationGuid getGuid() {
+        return (guid == null) ? null : SubpopulationGuid.create(guid);
+    }    
+    @Override
+    public void setGuid(SubpopulationGuid subpopGuid) {
+        this.guid = (subpopGuid == null) ? null : subpopGuid.getGuid();
+    }
```
DwayneJengSage, line 80: Unit tests for get/setGuid and verify that get/setGuidString are updated?

app/org/sagebionetworks/bridge/models/accounts/User.java
DwayneJengSage, line 39: As per IM conversation, we have a JSON serialization annotation on the User. Question #1, are we serializing User? If yes, we need to test it and make sure we aren't accidentally exposing health codes to end users. If not, we should remove the annotation.

app/org/sagebionetworks/bridge/models/subpopulations/SubpopulationGuid.java
```diff
@@ -0,0 +1,38 @@
+package org.sagebionetworks.bridge.models.subpopulations;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.Objects;
+
+public final class SubpopulationGuid {
```
DwayneJengSage, line 7: Where did the unit tests (equals verifier, to string) go?

alxdarksage, line 7: SubpopulationGuidTest

test/org/sagebionetworks/bridge/models/accounts/UserSessionInfoTest.java
alxdarksage, line 76: We never do this. This is an object we create as a view to send to users. But I can write a test anyway.

app/org/sagebionetworks/bridge/models/subpopulations/SubpopulationGuid.java
DwayneJengSage, line 7: Doesn't exist

test/org/sagebionetworks/bridge/models/accounts/UserSessionInfoTest.java
DwayneJengSage, line 76: As per IM conversation, if we never deserialize it, can we remove the deserialize annotation and the now unused deserializer?

test/org/sagebionetworks/bridge/dynamodb/DynamoStudyConsentTest.java
```diff
@@ -0,0 +1,44 @@
...
+       
+       String json = BridgeObjectMapper.get().writeValueAsString(consent);
+       JsonNode node = BridgeObjectMapper.get().readTree(json);
+       
+       assertEquals("1970-01-01T00:00:00.123Z", node.get("createdOn").asText());
+       assertTrue(node.get("active").asBoolean());
+       assertEquals("ABC", node.get("subpopulationGuid").asText());
+       assertEquals("StudyConsent", node.get("type").asText());
+       assertEquals(4, node.size());
+    }
```
DwayneJengSage, line 42: Can we add a deserialization test here?

DwayneJengSage:
StormpathAccountTest.multipleConsentsAreMaintainedSeparately() (not shown because the diff is too big). This test tests marshalling one subpop into CustomData. It doesn't verify any of the other subpops. It also doesn't test unmarshalling back from CustomData into the account object.

app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
```diff
@@ -42,189 +47,208 @@
...
         checkNotNull(sharingScope, Validate.CANNOT_BE_NULL, "sharingScope");
 
-        if (user.doesConsent()) {
+        ConsentStatus status = user.getConsentStatuses().get(subpopGuid);
+        // There will be a status object for each subpopulation the user is mapped to. If 
+        // there's no status object, then in effect the subpopulation does not exist for 
+        // this user and they should get back a 404.
+        if (status == null) {
+            throw new EntityNotFoundException(Subpopulation.class);
+        }
```
DwayneJengSage, line 130: Can we write a unit test for this case?

test/org/sagebionetworks/bridge/models/accounts/UserSessionInfoTest.java
alxdarksage, line 76: I removed all of these and the tests passed, I will run the SDK against it as well since those tests span calls and require session support.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulation.java
alxdarksage, line 80: Added

app/org/sagebionetworks/bridge/models/subpopulations/SubpopulationGuid.java
alxdarksage, line 7: Added back (!)

test/org/sagebionetworks/bridge/dynamodb/DynamoStudyConsentTest.java
alxdarksage, line 42: Added deserialization test for StudyConsentForm, the object that gets submitted through the API (users don't actually create the DynamoStudyConsent1 object so far as I can see).

app/org/sagebionetworks/bridge/services/ConsentServiceImpl.java
alxdarksage, line 130: Added in the next commit: https://github.com/alxdarksage/BridgePF/commit/e12bbb1a0dd20043b63f646918e5cc1bf3ec85db

PR #952, opened 2015-12-10
Harvard

alxdarksage:


EricSiegNW:
This looks right to me. When it verifies the ID and then registers it, it also then sets the dataGroup to the football_players group.

syoung-smallwisdom:
looks good.

PR #953, opened 2015-12-10
Export external ID as part of the participant roster.

alxdarksage:
Surprised it's not there already, but it isn't. This would be the FPHS code in the FPHS study (for example).


app/org/sagebionetworks/bridge/models/studies/StudyParticipant.java
```diff
@@ -31,6 +31,12 @@ public String getEmail() {
     public void setEmail(String email) {
         put(UserProfile.EMAIL_FIELD, email);
     }
+    public String getExternalId() {
+        return getEmpty(UserProfile.EXTERNAL_ID_FIELD);
```
syoung-smallwisdom, line 35: Does getEmpty() return an empty string (instead of nil)?  
-- Edit --
looks like it based on tests.

syoung-smallwisdom:
looks good

PR #956, opened 2015-12-12
add Exporter attributes to Study

DwayneJengSage:
Study Exporter config for data access team (for permissions) and synapse project ID (so we know what project to export to). These used to live in a JSON file that was manually copied to each Bridge EX host. Now they will live in their proper Study in DDB.

Testing done:
- DynamoStudyTest
- manual local test, where I updated the new attributes in the API study, then GET'ed the study back and verified they were there

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
@@ -35,6 +35,8 @@ public final class DynamoStudy implements Study {
     private String identifier;
     private String stormpathHref;
     private String supportEmail;
+    private Long synapseDataAccessTeamId;
+    private String synapseProjectId;
     private String technicalEmail;
     private String consentNotificationEmail;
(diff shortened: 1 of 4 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
```diff
@@ -325,6 +351,8 @@ public int hashCode() {
         result = prime * result + Objects.hashCode(name);
         result = prime * result + Objects.hashCode(sponsorName);
         result = prime * result + Objects.hashCode(supportEmail);
+        result = prime * result + Objects.hashCode(synapseDataAccessTeamId);
+        result = prime * result + Objects.hashCode(synapseProjectId);
```
alxdarksage, line 355: we could convert to Objects.hash(prop1, prop2, prop3...)

alxdarksage:
Minor comment about refactoring hashCode method, we can merge this whenever.

PR #957, opened 2015-12-17
add logging and null check to unsubscribe API

DwayneJengSage:
This is to investigate the firehose of errors in Prod.

Testing done:
- EmailControllerTest
- manually launched local server, tested with an invalid token, verified error message in logs


DwayneJengSage:
Merging, since this is prod-urgent.

PR #967, opened 2016-01-06
Criteria schedule plan

alxdarksage:
- Changing calculation of tasks to return to user. Now the scheduler calculates these each time, and any persisted information is recovered from DDB into the results returned to the user. This approach allows the removal of criteria from the ScheduledActivity objects themselves. The criteria can get more complicated and will only change Subpopulation and CriteriaScheduleStrategy.ScheduleCriteria.
- Removed run key (no longer part of matching), as well as min/max app version from the activities, and the hidesOn value as well (persisted activities have to be returned regardless of state in order to reconcile with scheduled activities in the service).
- Introduced CriteriaScheduleStrategy, which allows a schedule plan to return different schedules to users based on scheduling criteria (user groups, min/max app version).
- Refactored ScheduleStrategy tests out of ScheduleStrategyTest and into the subclass tests (e.g. ABTestScheduleStrategyTest), as these continue to grow.


app/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategy.java
```diff
@@ -0,0 +1,198 @@
...
+            errors.popNestedPath();
+        }
+    }
+
+    @Override
+    public List<Schedule> getAllPossibleSchedules() {
+        return scheduleCriteria.stream().map(ScheduleCriteria::getSchedule).collect(BridgeCollectors.toImmutableList());
+    }        
+    
+    public static class ScheduleCriteria implements Criteria {
```
DwayneJengSage: Make ScheduleCriteria a top-level class.

app/org/sagebionetworks/bridge/dynamodb/DynamoScheduledActivityDao.java
```diff
@@ -62,43 +57,24 @@ public ScheduledActivity getActivity(String healthCode, String guid) {
...
-            .withQueryFilterEntry("hidesOn", condition)
-            .withHashKeyValues(hashKey);
-
-        PaginatedQueryList<DynamoScheduledActivity> queryResults = mapper.query(DynamoScheduledActivity.class, query);
+    public List<ScheduledActivity> getActivities(DateTimeZone timeZone, List<ScheduledActivity> activities) {
+        if (activities.isEmpty()) {
+            return ImmutableList.of();
+        }
+        List<Object> activitiesToLoad = new ArrayList<Object>(activities);
+        Map<String,List<Object>> resultMap = mapper.batchLoad(activitiesToLoad);
```
DwayneJengSage: Because of type erasure, you can always cast a List of ScheduledActivity to a List of Object. No need to create a copy of the list. (We do this in DynamoIndexHelper.)

app/org/sagebionetworks/bridge/play/controllers/ScheduleController.java
```diff
@@ -67,12 +72,18 @@ public Result getSchedules() throws Exception {
...
         
         List<SchedulePlan> plans = schedulePlanService.getSchedulePlans(clientInfo, studyId);
+
         List<Schedule> schedules = Lists.newArrayListWithCapacity(plans.size());
         for (SchedulePlan plan : plans) {
-            Schedule schedule = plan.getStrategy().getScheduleForUser(session.getStudyIdentifier(), plan, session.getUser());
-            schedules.add(schedule);
+            Schedule schedule = plan.getStrategy().getScheduleForUser(plan, context);
+            if (schedule != null) {
+                schedules.add(schedule);
```
alxdarksage, line 82: This could be an error, log it

test/org/sagebionetworks/bridge/dynamodb/DynamoScheduledActivityDaoTest.java
```diff
@@ -104,103 +102,92 @@ public void after() {
...
+        assertEquals("activities were created", activitiesToSchedule, savedActivities);
         
-        // Should not increase the number of activities
-        activities = activityDao.getActivities(context);
-        assertEquals("activities did not grow afer repeated getActivity()", collectionSize, activities.size());
-
         // Have activities gotten injected time zone? We have to do this during construction using the time zone
         // sent with this call/request.
-        assertEquals(DateTimeZone.UTC, ((DynamoScheduledActivity)activities.get(0)).getTimeZone());
+        assertEquals(DateTimeZone.UTC, ((DynamoScheduledActivity)savedActivities.get(0)).getTimeZone());
```
DwayneJengSage: What happens if you run the schedule with the ScheduleContext in one time zone, but then call getActivities() with a different timezone? Is this semantically meaningful? We should add a test case for this.

app/org/sagebionetworks/bridge/dynamodb/DynamoScheduledActivityDao.java
```diff
@@ -62,43 +57,24 @@ public ScheduledActivity getActivity(String healthCode, String guid) {
...
-
-        DynamoDBQueryExpression<DynamoScheduledActivity> query = new DynamoDBQueryExpression<DynamoScheduledActivity>()
-            .withQueryFilterEntry("hidesOn", condition)
-            .withHashKeyValues(hashKey);
-
-        PaginatedQueryList<DynamoScheduledActivity> queryResults = mapper.query(DynamoScheduledActivity.class, query);
+    public List<ScheduledActivity> getActivities(DateTimeZone timeZone, List<ScheduledActivity> activities) {
+        if (activities.isEmpty()) {
+            return ImmutableList.of();
+        }
```
DwayneJengSage, line 64: Is it worth having a unit test case on calling this with an empty list? Or is this expected to never happen?

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
```diff
@@ -0,0 +1,348 @@
...
+        user.setDataGroups(Sets.newHashSet("group3"));
+        ScheduleContext context = new ScheduleContext.Builder()
+                .withClientInfo(ClientInfo.fromUserAgentCache("app/44"))
+                .withUser(user).build();
+        
+        // First two don't match because app version is wrong and missing required groups
+        // The last is returned because the context has no prohibited groups
+        Schedule schedule = strategy.getScheduleForUser(plan, context);
+        assertEquals(strategyWithProhibitedDataGroups, schedule);
+    }
```
DwayneJengSage, line 242: Add a test case where multiple criteria match?

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
```diff
@@ -0,0 +1,348 @@
...
+        setUpStrategyWithAppVersions();
+        setUpStrategyWithOneRequiredDataGroup();
+        setUpStrategyWithProhibitedDataGroups();
+
+        List<Schedule> schedules = strategy.getAllPossibleSchedules();
+        assertEquals(3, schedules.size());
+    }
+    
+    @Test
+    public void validates() {
```
DwayneJengSage: Add test cases to validates:
- criteria w/o a schedule
- criteria w/ invalid schedule
- happy case: CriteriaScheduleStrategy that passes validation

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
```diff
@@ -0,0 +1,348 @@
...
+
+    @Test
+    public void canGetAllPossibleScheduled() {
+        setUpStrategyWithAppVersions();
+        setUpStrategyWithOneRequiredDataGroup();
+        setUpStrategyWithProhibitedDataGroups();
+
+        List<Schedule> schedules = strategy.getAllPossibleSchedules();
+        assertEquals(3, schedules.size());
+    }
```
DwayneJengSage, line 255: How difficult is it to verify the contents of the schedule list in addition to just verifying the size?

app/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategy.java
```diff
@@ -0,0 +1,198 @@
...
+        return null;        
+    }
+
+    @Override
+    public void validate(Set<String> dataGroups, Set<String> taskIdentifiers, Errors errors) {
+        for (int i=0; i < scheduleCriteria.size(); i++) {
+            ScheduleCriteria criteria = scheduleCriteria.get(i);
+            errors.pushNestedPath("scheduleCriteria["+i+"]");
+            if (criteria.getSchedule() == null){
+                errors.rejectValue("schedule", "is required");
```
DwayneJengSage, line 59: Can schedule even be null? It looks like ScheduleCriteria.Builder.build() validates this,

app/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategy.java
```diff
@@ -0,0 +1,198 @@
...
+            public Builder addProhibitedGroup(String... groups) {
+                for (String group : groups) {
+                    this.noneOfGroups.add(group);    
+                }
+                return this;
+            }
+            public ScheduleCriteria build() {
+                checkNotNull(schedule);
+                return new ScheduleCriteria(schedule, minAppVersion, maxAppVersion, allOfGroups, noneOfGroups);
+            }
```
DwayneJengSage: What happens if a researcher submits a ScheduleCriteria without a schedule? Will they get a 500 error?

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
```diff
@@ -0,0 +1,348 @@
...
+    private Schedule strategyNoCriteria = makeValidSchedule("Strategy No Criteria");
+    
+    private CriteriaScheduleStrategy strategy;
+    private SchedulePlan plan;
+    private SchedulePlanValidator validator;
+    
+    @Test
+    public void hashCodeEquals() {
+        EqualsVerifier.forClass(CriteriaScheduleStrategy.class).allFieldsShouldBeUsed().verify();
+    }
```
DwayneJengSage, line 52: Should add EqualsVerified for ScheduleCriteria.

app/org/sagebionetworks/bridge/models/schedules/ScheduleStrategy.java
```diff
@@ -15,7 +13,8 @@
 @JsonTypeInfo( use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
 @JsonSubTypes({
     @Type(name="SimpleScheduleStrategy", value=SimpleScheduleStrategy.class),
-    @Type(name="ABTestScheduleStrategy", value=ABTestScheduleStrategy.class)
+    @Type(name="ABTestScheduleStrategy", value=ABTestScheduleStrategy.class),
+    @Type(name="CriteriaScheduleStrategy", value=CriteriaScheduleStrategy.class)
 })
```
DwayneJengSage, line 18: How do these JSON annotations work? Do we have any tests to verify they work as expected?

app/org/sagebionetworks/bridge/play/controllers/ScheduleController.java
```diff
@@ -47,8 +48,12 @@ public Result getSchedulesV3() throws Exception {
         List<SchedulePlan> plans = schedulePlanService.getSchedulePlans(clientInfo, studyId);
         List<Schedule> schedules = Lists.newArrayListWithCapacity(plans.size());
         for (SchedulePlan plan : plans) {
-            Schedule schedule = plan.getStrategy().getScheduleForUser(session.getStudyIdentifier(), plan, session.getUser());
-            schedules.add(schedule);
+            ScheduleContext context = new ScheduleContext.Builder()
+                    .withUser(session.getUser()).withClientInfo(clientInfo).build();
```
DwayneJengSage: ScheduleContext should be constructed outside the loop, since it doesn't vary with the loop.

app/org/sagebionetworks/bridge/play/controllers/ScheduleController.java
```diff
@@ -47,8 +48,12 @@ public Result getSchedulesV3() throws Exception {
...
+            Schedule schedule = plan.getStrategy().getScheduleForUser(plan, context);
+            if (schedule != null) {
+                schedules.add(schedule);    
+            }
```
DwayneJengSage: Lots of duplicated code between here and getSchedules(). Should refactor into a shared method.

app/org/sagebionetworks/bridge/services/ScheduledActivityOperations.java
```diff
@@ -0,0 +1,54 @@
...
+                results.add(dbActivity);
+            } else {
+                // but don't save expired tasks, there is no point
+                if (activity.getStatus() != ScheduledActivityStatus.EXPIRED) {
+                    saves.add(activity);    
+                }
+                results.add(activity);
+            }
+        }
+    }
```
DwayneJengSage: A few notes:

(1) results always ends up being the same as scheduledActivities. This means you don't have to generate the results list, and you can easily turn this into a method that takes in scheduledActivites and dbActivities and just returns saves.

(2) This should be simple enough to turn this into a static utility method (either in its own class or in ScheduledActivityService), or make it non-static if we need to mock it.

(3) Minor: A comment clarifying that scheduledActivities is always a superset of dbActivities.

test/org/sagebionetworks/bridge/services/ScheduledActivityOperationsTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.junit.Test;
+
+import org.sagebionetworks.bridge.models.schedules.ScheduledActivity;
+
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
+
+public class ScheduledActivityOperationsTest {
```
DwayneJengSage: Add unit test to make sure getResults() sorts?

test/org/sagebionetworks/bridge/dynamodb/DynamoScheduledActivityDaoMockTest.java
```diff
@@ -147,53 +157,15 @@ public void testOfFirstPeriod() throws Exception {
...
-        verify(mapper).query((Class<DynamoScheduledActivity>) any(Class.class),
-                        (DynamoDBQueryExpression<DynamoScheduledActivity>) any(DynamoDBQueryExpression.class));
+        verify(mapper).batchLoad((List<Object>)any());
         verifyNoMoreInteractions(mapper);
     }
     
     @SuppressWarnings("unchecked")
-    @Test
-    public void testOfSecondPeriodWithDifferentStartTime() throws Exception {
-        DateTime endsOn = NOW.plus(Period.parse("P4D"));
```
DwayneJengSage, line 193: It looks like you deleted testOfSecondPeriodWithDifferentStartTime(). Does this test have any value?

test/org/sagebionetworks/bridge/models/schedules/ABTestScheduleStrategyTest.java
```diff
@@ -3,17 +3,62 @@
...
+
+import nl.jqno.equalsverifier.EqualsVerifier;
+import nl.jqno.equalsverifier.Warning;
 
+/**
+ * Further tests for these strategy objects are in ScheduleStrategyTest.
+ */
 public class ABTestScheduleStrategyTest {
 
+    private BridgeObjectMapper mapper = BridgeObjectMapper.get();
```
DwayneJengSage: Nitpick: static final

test/org/sagebionetworks/bridge/models/schedules/ABTestScheduleStrategyTest.java
```diff
@@ -26,4 +71,86 @@ public void testScheduleCollector() {
         assertEquals("Schedule 3", schedules.get(2).getLabel());
         assertTrue(schedules instanceof ImmutableList);
     }
+    
+    @Test
+    public void canRountripABTestingPlan() throws Exception {
+        DynamoSchedulePlan plan = createABSchedulePlan();
+        String output = new BridgeObjectMapper().writeValueAsString(plan);
```
DwayneJengSage: Nitpick: Why create a new mapper when we can re-use the existing one?

test/org/sagebionetworks/bridge/models/schedules/ABTestScheduleStrategyTest.java
```diff
@@ -26,4 +71,86 @@ public void testScheduleCollector() {
...
+        // and they should not be in order...
+        Map<String, Integer> countsByLabel = Maps.newHashMap();
+        for (Schedule schedule : schedules) {
+            Integer count = countsByLabel.get(schedule.getLabel());
+            if (count == null) {
+                countsByLabel.put(schedule.getLabel(), 1);
+            } else {
+                countsByLabel.put(schedule.getLabel(), ++count);
+            }
+        }
```
DwayneJengSage, line 107: Minor: A Guava Multiset will make this much simpler https://code.google.com/p/guava-libraries/wiki/NewCollectionTypesExplained#Multiset

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
```diff
@@ -0,0 +1,348 @@
...
+import nl.jqno.equalsverifier.EqualsVerifier;
+
+public class CriteriaScheduleStrategyTest {
+    
+    private Schedule strategyWithAppVersions = makeValidSchedule("Strategy With App Versions");
+    private Schedule strategyWithOneRequiredDataGroup = makeValidSchedule("Strategy With One Required Data Group");
+    private Schedule strategyWithRequiredDataGroups = makeValidSchedule("Strategy With Required Data Groups");
+    private Schedule strategyWithOneProhibitedDataGroup = makeValidSchedule("Strategy With One Prohibited Data Group");
+    private Schedule strategyWithProhibitedDataGroups = makeValidSchedule("Strategy With One Prohibited Data Groups");
+    private Schedule strategyNoCriteria = makeValidSchedule("Strategy No Criteria");
```
DwayneJengSage: Minor: Can we make makeValidSchedule() static and all of these schedules static final?

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
```diff
@@ -0,0 +1,348 @@
...
+    
+    private CriteriaScheduleStrategy strategy;
+    private SchedulePlan plan;
+    private SchedulePlanValidator validator;
```
DwayneJengSage: Minor: Similarly, can we make plan and validator static final?

test/org/sagebionetworks/bridge/models/schedules/SimpleScheduleStrategyTest.java
```diff
@@ -5,14 +5,86 @@
...
+
+import nl.jqno.equalsverifier.EqualsVerifier;
+import nl.jqno.equalsverifier.Warning;
 
+/**
+ * Further tests for these strategy objects are in ScheduleStrategyTest.
+ */
 public class SimpleScheduleStrategyTest {
 
+    private BridgeObjectMapper mapper = BridgeObjectMapper.get();
```
DwayneJengSage: Nitpick: Same here.

test/org/sagebionetworks/bridge/models/schedules/SimpleScheduleStrategyTest.java
```diff
@@ -5,14 +5,86 @@
...
+        Schedule schedule = TestUtils.getSchedule("AAA");
+        SimpleScheduleStrategy strategy = new SimpleScheduleStrategy();
+        strategy.setSchedule(schedule);
+
+        DynamoSchedulePlan plan = new DynamoSchedulePlan();
+        plan.setModifiedOn(DateUtils.getCurrentMillisFromEpoch());
+        plan.setStudyKey(study.getIdentifier());
+        plan.setStrategy(strategy);
+
+        String output = new BridgeObjectMapper().writeValueAsString(plan);
```
DwayneJengSage: Nitpick: same here

app/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategy.java
alxdarksage: Broke this out, and broke out ScheduleGroup as ABTestGroup.

app/org/sagebionetworks/bridge/play/controllers/ScheduleController.java
alxdarksage, line 82: Logging this. Added the user ID so we can log it along with the plan name, so we can trace back if need be.

test/org/sagebionetworks/bridge/dynamodb/DynamoScheduledActivityDaoTest.java
alxdarksage: In a single request you can't do this and I'm not sure the results would be accurate or desirable. The ScheduledActivityServiceTest.retrievalActivitiesAcrossTimeAndTimeZones() test examines calling getActivities() from different time zones, e.g. if the user took a plane to a different time zone and opened the app, and the behavior is consistent. It shifts to the user's current time zone and recalculates.

app/org/sagebionetworks/bridge/dynamodb/DynamoScheduledActivityDao.java
alxdarksage, line 64: No, it could happen. Added test.

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
alxdarksage, line 242: Added, tests both that different criteria are matched, and that the first valid schedule in the list is returned.

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
alxdarksage: Added

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
alxdarksage, line 255: Not hard, added

app/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategy.java
alxdarksage, line 59: Yes, I had to remove it in order to get the case where it is null to show up in the validation errors. Otherwise you get a null pointer exception/a 500 error. So now this is verified as part of validation.

app/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategy.java
alxdarksage: Yeah, see above. I've removed that check and now they should get InvalidEntityException indicating the schedule is missing.

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
alxdarksage, line 52: added (also had to mark that class final).

app/org/sagebionetworks/bridge/models/schedules/ScheduleStrategy.java
alxdarksage, line 18: It uses the value of the "type" annotation to determine which class to use to deserialize (so this is one of those cases where the client has to provide the type property. Added tests to ScheduleStrategyTest to verify this works.

app/org/sagebionetworks/bridge/play/controllers/ScheduleController.java
alxdarksage: Moved

app/org/sagebionetworks/bridge/play/controllers/ScheduleController.java
alxdarksage: Combined.

app/org/sagebionetworks/bridge/services/ScheduledActivityOperations.java
alxdarksage: The replacement of some scheduled activities with db activities is needed to pass state (started and finished) into the results returned to the user. So I think we need two lists here. I will incorporate this code back into the service as we discussed.

Moved this into ScheduledActivityService update scheduledActivities in place with persisted activities, and calculate and return the list of activities to save.

test/org/sagebionetworks/bridge/services/ScheduledActivityOperationsTest.java
alxdarksage: Added ScheduledActivityServiceMockTest.orderActivitieFiltersAndSorts() to test filtering and sorting.

test/org/sagebionetworks/bridge/dynamodb/DynamoScheduledActivityDaoMockTest.java
alxdarksage, line 193: I think so because the DAO logic has gotten much simpler. Before we were filtering based on the app version as recorded in the task, now we are just fetching a set list of activities. So there's much less to test (sorting, filtering, filtering the query based on hidesOn are all gone now).

test/org/sagebionetworks/bridge/models/schedules/ABTestScheduleStrategyTest.java
alxdarksage: Changed both of these

test/org/sagebionetworks/bridge/models/schedules/ABTestScheduleStrategyTest.java
alxdarksage, line 107: Cool. Changed to use a multiset.

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
alxdarksage: Changed and renamed SCHEDULE_FOR_STRATEGY_\* for clarity.

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
alxdarksage: Changed

test/org/sagebionetworks/bridge/models/schedules/SimpleScheduleStrategyTest.java
alxdarksage: Changed

test/org/sagebionetworks/bridge/models/schedules/SimpleScheduleStrategyTest.java
alxdarksage: Changed

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
```diff
@@ -0,0 +1,416 @@
...
+        setUpStrategyWithAppVersions(); // certainly should not match this one, although criteria are valid
+        
+        User user = getUser();
+        user.setDataGroups(Sets.newHashSet("req1", "req2")); // both are required
+        ScheduleContext context = new ScheduleContext.Builder()
+                .withClientInfo(ClientInfo.fromUserAgentCache("app/6")) // in range
+                .withUser(user).build();
+        
+        // First two don't match because app version is wrong and missing required groups
+        // The last is returned because the context has no prohibited groups
```
DwayneJengSage: Minor: Comment bug

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
```diff
@@ -0,0 +1,416 @@
...
+        EqualsVerifier.forClass(CriteriaScheduleStrategy.class).allFieldsShouldBeUsed().verify();
+    }
+    
+    @Before
+    public void before() {
+        strategy = new CriteriaScheduleStrategy();
+
+        PLAN.setLabel("Schedule plan label");
+        PLAN.setStudyKey(TEST_STUDY_IDENTIFIER);
+        PLAN.setStrategy(strategy);
```
DwayneJengSage, line 59: Minor: PLAN.set\* could be moved to a static initializer so we don't have to make these redundant calls with each unit test.

alxdarksage, line 59: I don't like static initializers that much, I think you don't like static initializers that much, I would rather just leave it.

app/org/sagebionetworks/bridge/dynamodb/DynamoScheduledActivityDao.java
```diff
@@ -57,48 +52,30 @@ public ScheduledActivity getActivity(String healthCode, String guid) {
         if (dbActivity == null) {
             throw new EntityNotFoundException(ScheduledActivity.class);
         }
+        dbActivity.setTimeZone(timeZone);
```
DwayneJengSage, line 55: Do we have a unit test to test setting the time zone? Otherwise, how do we know that it works? If this is in a pre-existing test, how did that test pass before this change was made?

app/org/sagebionetworks/bridge/models/schedules/ScheduleCriteria.java
```diff
@@ -0,0 +1,108 @@
...
+    private ScheduleCriteria(@JsonProperty("schedule") Schedule schedule, 
+            @JsonProperty("minAppVersion") Integer minAppVersion, 
+            @JsonProperty("maxAppVersion") Integer maxAppVersion, 
+            @JsonProperty("allOfGroups") Set<String> allOfGroups, 
+            @JsonProperty("noneOfGroups") Set<String> noneOfGroups) {
+        this.schedule = schedule;
+        this.minAppVersion = minAppVersion;
+        this.maxAppVersion = maxAppVersion;
+        this.allOfGroups = (allOfGroups != null) ? allOfGroups : Sets.newHashSet();
+        this.noneOfGroups = (noneOfGroups != null) ? noneOfGroups : Sets.newHashSet();
```
DwayneJengSage, line 29: Unit test to test ScheduleCriteria with null allOf and noneOf groups and make sure they're properly converted to empty sets and pass validation?

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
DwayneJengSage, line 59: I actually love static initializers. They do exactly what I need them to do. Plus in this case, a static initializer would be much cleaner than using a non-static method to (redundantly) initialize a static variable.

app/org/sagebionetworks/bridge/models/schedules/ScheduleCriteria.java
alxdarksage, line 29: Added

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
alxdarksage, line 59: moved

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
alxdarksage: updated

app/org/sagebionetworks/bridge/dynamodb/DynamoScheduledActivityDao.java
alxdarksage, line 55: The only place this was used before was to load an activity in order to update its start/end times. And in fact, that's all we're using it for now, as well. But it seemed more consistent to have it always set the timezone, rather than do it in one method and not the other, in case someone ever calls this besides the update code (where it's set to null because it doesn't matter).

Time zones are tested in ScheduledActivityServiceTest.retrievalActivitiesAcrossTimeAndTimeZones at length. I've added similar tests to DynamoScheduledActivityDaoTest for this method. It relies on the same code though (the ScheduledActivity implementation).

app/org/sagebionetworks/bridge/models/schedules/ScheduleCriteria.java
alxdarksage: Added

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
alxdarksage: updated

app/org/sagebionetworks/bridge/models/schedules/ScheduleStrategy.java
alxdarksage, line 18: Incidentally, added tests in the SDK too as this is a case where the type attribute has to be sent. The SDK sends it.

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
```diff
@@ -0,0 +1,412 @@
...
+            "Strategy With Required Data Groups");
+    private static final Schedule SCHEDULE_FOR_STRATEGY_WITH_ONE_PROHIBITED_DATA_GROUP = makeValidSchedule(
+            "Strategy With One Prohibited Data Group");
+    private static final Schedule SCHEDULE_FOR_STRATEGY_WITH_PROHIBITED_DATA_GROUPS = makeValidSchedule(
+            "Strategy With One Prohibited Data Groups");
+    private static final Schedule SCHEDULE_FOR_STRATEGY_NO_CRITERIA = makeValidSchedule("Strategy No Criteria");
+    private static final Schedule SCHEDULE_FOR_STRATEGY_WITH_ALL_REQUIREMENTS = makeValidSchedule(
+            "Strategy with all requirements");
+    
+    private static final CriteriaScheduleStrategy strategy = new CriteriaScheduleStrategy();
```
DwayneJengSage: Given that each test modifies strategy, I think this should stay as a non-static member that gets reset in the Before method.

PR #968, opened 2016-01-08
Adding data groups to the creation of health data records.

alxdarksage:
Of note: DDB does not support empty string sets (I believe it throws an error), but rather than use the StringSetMarshaller and serialize this to a JSON array "[]", I am only setting data groups in the health record if the set is non-empty. So the reference is either null, or a non-empty set. None of the tests actually test against DDB to verify this works, shall we add a non-mock DAO test for this?

### DIFF
--- a/app/org/sagebionetworks/bridge/models/healthdata/HealthDataRecordBuilder.java
+++ b/app/org/sagebionetworks/bridge/models/healthdata/HealthDataRecordBuilder.java
@@ -1,6 +1,9 @@
 package org.sagebionetworks.bridge.models.healthdata;
 
 import com.fasterxml.jackson.databind.JsonNode;
+
+import java.util.Set;
+
 import org.joda.time.LocalDate;
 
 import org.sagebionetworks.bridge.dao.ParticipantOption;
@@ -26,6 +29,7 @@ public abstract class HealthDataRecordBuilder {
     private LocalDate uploadDate;
     private String uploadId;
     private String userExternalId;
+    private Set<String> userDataGroups;
     private ParticipantOption.SharingScope userSharingScope;
     private Long version;
 
@@ -42,6 +46,7 @@ public abstract class HealthDataRecordBuilder {
         uploadDate = record.getUploadDate();
         uploadId = record.getUploadId();
         userExternalId = record.getUserExternalId();
+        userDataGroups = record.getUserDataGroups();
         userSharingScope = record.getUserSharingScope();
         version = record.getVersion();
         return this;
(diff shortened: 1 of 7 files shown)


app/org/sagebionetworks/bridge/upload/TranscribeConsentHandler.java
```diff
@@ -39,9 +45,15 @@ public void handle(@Nonnull UploadValidationContext context) {
 
         // Also get external ID
         String userExternalId = options.get(ParticipantOption.EXTERNAL_IDENTIFIER);
+        
+        // And get user data groups
+        String dataGroupsSer = options.get(ParticipantOption.DATA_GROUPS);
+        Set<String> userDataGroups = (isNotBlank(dataGroupsSer))
+                ? Sets.newHashSet(Splitter.on(",").split(dataGroupsSer)) : null;
```
DwayneJengSage, line 52: We had a bug where the externalId wasn't being propagated to the record because of a bug in the copy-constructor. To make sure we don't have a similar bug here, I highly recommend also addeding data groups to https://github.com/Sage-Bionetworks/BridgePF/blob/develop/test/org/sagebionetworks/bridge/upload/TranscribeConsentHandlerTest.java

This is admittedly non-obvious and won't scale, so long-term, we'll need to figure out a better solution for this.

app/org/sagebionetworks/bridge/models/healthdata/HealthDataRecordBuilder.java
```diff
@@ -42,6 +46,7 @@ public HealthDataRecordBuilder copyOf(HealthDataRecord record) {
         uploadDate = record.getUploadDate();
         uploadId = record.getUploadId();
         userExternalId = record.getUserExternalId();
+        userDataGroups = record.getUserDataGroups();
```
DwayneJengSage, line 49: Similarly, we should add to the copy-constructor test in HealthDataRecordTest

DwayneJengSage:
Two tests I would recommend adding, but otherwise looks good.

app/org/sagebionetworks/bridge/upload/TranscribeConsentHandler.java
alxdarksage, line 52: Okay, added some more checks in HealthDataRecordTest and UploadArtifactsHandlerTest that values are being copied/transferred.

app/org/sagebionetworks/bridge/models/healthdata/HealthDataRecordBuilder.java
alxdarksage, line 49: added

PR #974, opened 2016-01-13
rate-limit ParticipantRosterGenerator

DwayneJengSage:
ParticipantRosterGenerator doesn't have rate-limiting, which means it calls DDB as fast as Stormpath will vend accounts. This is problematic, as it means we cannot predict how much DDB throughput we need.

Adding rate limiting by sleeping between each account for 60ms. This is enough to handle up to 60k accounts per hour.

Testing done:
- ParticipantRosterGeneratorTest
- manually tested ParticipantRosterGenerator on my local server

### DIFF
--- a/app/org/sagebionetworks/bridge/services/ParticipantRosterGenerator.java
+++ b/app/org/sagebionetworks/bridge/services/ParticipantRosterGenerator.java
@@ -90,7 +90,7 @@ public class ParticipantRosterGenerator implements Runnable {
                 try {
                     Thread.sleep(60);
                 } catch (InterruptedException ex) {
-                    // Interruptios should never happen. Set interrupt status for hygiene.
+                    // Interruptions should never happen. Set interrupt status for hygiene.
                     Thread.currentThread().interrupt();
                 }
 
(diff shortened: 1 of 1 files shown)


app/org/sagebionetworks/bridge/services/ParticipantRosterGenerator.java
```diff
@@ -84,7 +84,16 @@ public void run() {
...
             while (accounts.hasNext()) {
                 Account account = accounts.next();
-                
+
+                // Rate limit, so we don't end up consuming DDB as fast as possible. 60 millisecond sleep is
+                // ~16.7 accounts/sec. This will allow us to process 60k accounts in an hour.
+                try {
+                    Thread.sleep(60);
+                } catch (InterruptedException ex) {
+                    // Interruptios should never happen. Set interrupt status for hygiene.
```
alxdarksage: Minor typo: "Interruptions"

alxdarksage:
Looking at this, we're calling subpopulationService for every user, which is dumb... the set of subpopulations won't change for a study and this is study scoped. It would greatly reduce DDB usage if we got those once. If you have a JIRA item for this, assign it to me after this PR is merged and I'll fix (or you can if you really want to).

DwayneJengSage:
JIRA for caching the subpop call (or anything other than blindly calling it a jillion times) https://sagebionetworks.jira.com/browse/BRIDGE-1081

JIRA for merging the 3 ParticipantOption DB scans into one https://sagebionetworks.jira.com/browse/BRIDGE-1082

PR #976, opened 2016-01-14
add dedupe to Upload API

DwayneJengSage:
Dedupes if it's the same health code, same schema (activity) within a second. Initial version is in "log and observe" mode, so we can observe how often this happens before we actually start deduping.

Testing done:
- unit tests
- manual tested original upload, duplicate upload (0.8 seconds later) non-duplicate upload (2 seconds later)


app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDedupeDao.java
```diff
@@ -0,0 +1,63 @@
...
+        // Hash key comes from health code and schema
+        DynamoUploadDedupe hashKey = new DynamoUploadDedupe();
+        hashKey.setHealthCode(healthCode);
+        hashKey.setSchemaKey(schemaKey.toString());
+
+        // range key, anything within 1 second is valid
+        Condition createdOnCondition = new Condition().withComparisonOperator(ComparisonOperator.BETWEEN)
+                .withAttributeValueList(
+                        new AttributeValue().withN(String.valueOf(createdOn - CREATED_ON_DELTA_MILLIS)),
+                        new AttributeValue().withN(String.valueOf(createdOn + CREATED_ON_DELTA_MILLIS)));
```
alxdarksage, line 42: I'm surprised duplicates can have different createdOn timestamps... we really have no other way to determine if it's a previous upload? MD5 or something? It'll never be the case that a client sends files with createdOn values that are close to one another that are nevertheless different content?

app/org/sagebionetworks/bridge/upload/DedupeHandler.java
```diff
@@ -0,0 +1,57 @@
...
+            long createdOn = recordBuilder.getCreatedOn();
+            String healthCode = recordBuilder.getHealthCode();
+            UploadSchemaKey schemaKey = new UploadSchemaKey.Builder().withStudyId(recordBuilder.getStudyId())
+                    .withSchemaId(recordBuilder.getSchemaId()).withRevision(recordBuilder.getSchemaRevision()).build();
+
+            // Is this a duplicate?
+            boolean isDupe = uploadDedupeDao.isDuplicate(createdOn, healthCode, schemaKey);
+            if (isDupe) {
+                // For now, log. Once we observe this to be running reliably, we can make it filter out dupes.
+                LOG.info("DedupeHandler detected a dupe: Upload ID " + uploadId);
```
alxdarksage, line 46: But all we do is log?

alxdarksage, line 46: Besides some questions about how this works, looks fine. Can merge when you're ready.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDedupeDao.java
DwayneJengSage, line 42: Feedback from researchers is that the apps will sometimes upload the exact same set of files with a different createdOn timestamp for whatever reason. MD5 doesn't work, since if the createdOn is different, info.json will be different, and hence the MD5 will be different.

Based on discussions with the researcher, it's generally impossible for a user to complete the same survey or same activity more than once in a second, so keying off schema and createdOn should allow us to dedupe. The only concern is passive data collection, but if passive data is sending multiple results in a short time frame, it's almost certainly a bug in the client.

app/org/sagebionetworks/bridge/upload/DedupeHandler.java
DwayneJengSage, line 46: Yes, we want to observe this for a while to (a) see how many dupes we're catching and (b) make sure we don't have false positives.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDedupeDao.java
alxdarksage, line 42: Makes sense.

alxdarksage:
I can merge whenever you're ready, will work on build issues this morning.

PR #981, opened 2016-01-25
Limiting the number of scheduled activities (tasks) we will attempt to delete to keep tasks in sync with schedule plans

alxdarksage:
In tests and development of studies, it's very useful to have them cleaned up, but in production, this easily exceeds throughput (I've tested this with very large values... at best we could do it if we rate-limited this and created a separate thread, but this is probably not necessary in production situations). So, limit to a threshold count indicative of development rather than production.


alxdarksage:
Don't like this solution. Going to possibly remove deletion all the time.

PR #982, opened 2016-01-25
Updating to Stormpath SDK 1.0.RC8.1

alxdarksage:
This fixed an issue I saw locally where creating a Stormpath directory was not updating the verify email template (now iterating through all of them and updating them).

### DIFF
--- a/app/org/sagebionetworks/bridge/stormpath/StormpathDirectoryDao.java
+++ b/app/org/sagebionetworks/bridge/stormpath/StormpathDirectoryDao.java
@@ -177,9 +177,8 @@ public class StormpathDirectoryDao implements DirectoryDao {
     private void adjustPasswordPolicies(Study study, Directory directory) {
         PasswordPolicy passwordPolicy = directory.getPasswordPolicy();
         
-        for (ModeledEmailTemplate template : passwordPolicy.getResetEmailTemplates()) {
-            updateTemplate(study, template, study.getResetPasswordTemplate(), "resetPassword");
-        }
+        ModeledEmailTemplate template = passwordPolicy.getResetEmailTemplates().single();
+        updateTemplate(study, template, study.getResetPasswordTemplate(), "resetPassword");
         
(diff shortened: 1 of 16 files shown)


app/org/sagebionetworks/bridge/stormpath/StormpathDirectoryDao.java
```diff
@@ -177,9 +177,9 @@ private static AccountStoreMapping getApplicationMapping(String href, Applicatio
     private void adjustPasswordPolicies(Study study, Directory directory) {
         PasswordPolicy passwordPolicy = directory.getPasswordPolicy();
         
-        // According to the documentation, there is only one...
-        ModeledEmailTemplate template = passwordPolicy.getResetEmailTemplates().iterator().next();
-        updateTemplate(study, template, study.getResetPasswordTemplate(), "resetPassword");
+        for (ModeledEmailTemplate template : passwordPolicy.getResetEmailTemplates()) {
+            updateTemplate(study, template, study.getResetPasswordTemplate(), "resetPassword");
+        }
```
DwayneJengSage: In StormpathDirectoryDaoTest, we have asserts that there's only 1 reset password template and only 1 account verification email template.

This is cause for concern because (a) it means the fix we tried to make doesn't match up with the outcome we're seeing and (b) the unit tests don't test the change that we made.

DwayneJengSage: After following up with Alx, we suspect that hasNext() has a side effect, so we want to do this in a loop rather than calling iterator.next().

That said, our unit tests assume there's only ever 1 template, and our implicit assumption is that there's only ever 1 template. We should make this assumption explicit, check for the number of templates, and log an error if there isn't.

alxdarksage:
I've reverted to the retrieval of one template because this isn't an issue with multiple templates. I've not been able to find any satisfactory workaround to this issue so far this morning. I have sent all I know to Stormpath, I'm sure at this point that the problem is on their side.

PR #986, opened 2016-01-26
Once cached, any user could retrieve a survey. Fixed.

alxdarksage:
Added the requestor's study to the key. When retrieved from the db, a requestor outside the study will get an error and no caching, thus a check each time, while someone in the study, who is successful, will get the survey cached under a key with that study in it. Users outside the study can thus never retrieve a cached version of the survey.

https://sagebionetworks.jira.com/browse/BRIDGE-1103

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/SurveyController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/SurveyController.java
@@ -58,15 +58,6 @@ public class SurveyController extends BaseController {
         return okResult(surveys);
     }
     
-    public Result getAllSurveysMostRecentVersion2() throws Exception {
-        UserSession session = getAuthenticatedSession(DEVELOPER);
-        StudyIdentifier studyId = session.getStudyIdentifier();
-
-        List<Survey> surveys = surveyService.getAllSurveysMostRecentVersion(studyId);
-        verifySurveyIsInStudy(session, surveys);
-        return okResult(surveys);
-    }
-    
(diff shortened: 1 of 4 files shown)


DwayneJengSage:
Generally looks good. 2 minor non-blocking issues:
1. The get_() and get_ForUser() methods have more or less identical viewCache code. Is it worth consolidating those to call shared methods to reduce the amount of duplicate code?
2. Is it worth adding a get() to the delete, update, version, and publish unit tests to make sure the cache is invalidated and we hit the service?

alxdarksage:
Added tests to verify the cache is being cleared; reduced duplication in the controller through the liberal use of lambdas.

DwayneJengSage:
I think there was a misunderstanding of my comment. There is still a lot of duplicate code between getSurveyForUser() and getSurvey(), as well as between getSurveyMostRecentlyPublishedVersionForUser() and getSurveyMostRecentlyPublishedVersion()

alxdarksage:
Updated

DwayneJengSage:
2 minor comments:
1. Is it also worth combining getSurveyMostRecentlyPublishedVersionForUser() and getSurveyMostRecentlyPublishedVersion()? There's 5 lines in common between them
2. studyId is trivially derived from session. Do we need to pass studyId in addition to session into getSurveyInternal()?

Also, there's a merge conflict in this pull request.

alxdarksage:
Removed one getSurvey method that was redundant, and further consolidated the code in the other controllers, with some renaming of *Internal methods for clarity. Removed studyId from a couple of methods that also took the session.

PR #988, opened 2016-01-27
Update Stormpath directory for study when study changes

alxdarksage:
When study developers update information that's used to fill out email templates, or is used for email addresses in the templates, the Stormpath directory has to be updated. Right now only checking if the templates have changed. Checking for a wider variety of changes to determine when to update the directory; adding tests to verify. BRIDGE-1099.


DwayneJengSage:
Looks good. Ready to merge whenever you're ready.

PR #989, opened 2016-01-27
Multi-Choice ints with more than 20 options should be attachments

DwayneJengSage:
Schema field type INLINE_JSON_BLOB has a char limt of 100 because of Synapse column and table width limitations. This is used for enum types in surveys (Multi-Choice, Single-Choice in iOS, MultiValueConstraints in Bridge).

We originally assumed an enum with int values would always fit in that char limit. However, it only takes 37 ints for this to exceed 100 chars, even with all whitespaced removed from the JSON. In FPHS, we have an enum with 39 ints.

I'm adding a limit of 20 ints before the enum gets exported as an attachment instead of as inline JSON.

Testing done:
- With the old code, created survey with an int enum with 40 options. Uploaded a response with all of the options selected. Verified this gets truncated in Synapse.
- With the new code, created a new revision of the survey. Uploaded a response against the new version of the survey. Verified this becomes an attachment in Synapse.
- For backwards compatibility, re-uploaded the v1 answers and verified that nothing breaks.
- DynamoUploadSchemaDaoTest


alxdarksage:
Looks good, we can merge when you're in.

DwayneJengSage:
Ready to merge whenever you are.

PR #992, opened 2016-01-27
Consent a newly created user to all required consents

alxdarksage:
...not just the default consent. Removes one source of errors if the test study has another leftover required subpopulation. Our intention is to get the user around 412 errors for testing.

https://sagebionetworks.jira.com/browse/BRIDGE-1092

### DIFF
--- a/app/org/sagebionetworks/bridge/services/UserAdminService.java
+++ b/app/org/sagebionetworks/bridge/services/UserAdminService.java
@@ -1,6 +1,7 @@
 package org.sagebionetworks.bridge.services;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static org.sagebionetworks.bridge.dao.ParticipantOption.SharingScope.NO_SHARING;
 
 import java.util.Iterator;
 
@@ -10,11 +11,11 @@ import org.sagebionetworks.bridge.cache.CacheProvider;
 import org.sagebionetworks.bridge.dao.AccountDao;
 import org.sagebionetworks.bridge.dao.DistributedLockDao;
 import org.sagebionetworks.bridge.dao.HealthIdDao;
-import org.sagebionetworks.bridge.dao.ParticipantOption.SharingScope;
 import org.sagebionetworks.bridge.exceptions.BridgeServiceException;
 import org.sagebionetworks.bridge.json.DateUtils;
 import org.sagebionetworks.bridge.models.ClientInfo;
 import org.sagebionetworks.bridge.models.accounts.Account;
+import org.sagebionetworks.bridge.models.accounts.ConsentStatus;
 import org.sagebionetworks.bridge.models.accounts.SignIn;
 import org.sagebionetworks.bridge.models.accounts.SignUp;
 import org.sagebionetworks.bridge.models.accounts.User;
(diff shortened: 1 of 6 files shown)


test/org/sagebionetworks/bridge/services/StormPathUserAdminServiceTest.java
```diff
@@ -16,17 +20,23 @@
...
 import org.sagebionetworks.bridge.models.ClientInfo;
+import org.sagebionetworks.bridge.models.accounts.ConsentStatus;
 import org.sagebionetworks.bridge.models.accounts.SignIn;
 import org.sagebionetworks.bridge.models.accounts.SignUp;
 import org.sagebionetworks.bridge.models.accounts.User;
 import org.sagebionetworks.bridge.models.accounts.UserSession;
 import org.sagebionetworks.bridge.models.studies.Study;
+import org.sagebionetworks.bridge.models.subpopulations.Subpopulation;
+
+import scala.collection.immutable.Map;
```
DwayneJengSage: This is the wrong Map class. (It also doesn't seem to be used at all in this class.)

app/org/sagebionetworks/bridge/services/UserAdminService.java
```diff
@@ -132,14 +133,20 @@ public UserSession createUser(SignUp signUp, Study study, SubpopulationGuid subp
...
-                        : SubpopulationGuid.create(study.getIdentifier());
-
-                consentService.consentToResearch(study, consentTo,
-                        newUserSession.getUser(), consent, SharingScope.NO_SHARING, false);
+                
+                User user = newUserSession.getUser();
+                if (subpopGuid != null) {
+                    consentService.consentToResearch(study, subpopGuid, user, signature, NO_SHARING, false);
+                } else {
+                    for (ConsentStatus consentStatus : newUserSession.getUser().getConsentStatuses().values()) {
```
DwayneJengSage: Nitpick: You already have newUserSession.getUser() saved in var user, so you don't need to call it again here.

test/org/sagebionetworks/bridge/services/StormPathUserAdminServiceTest.java
```diff
@@ -150,4 +163,31 @@ public void testDeleteUserThatHasBeenDeleted() {
         userAdminService.deleteUser(study, session.getUser().getEmail());
         assertNull(authService.getSession(session.getSessionToken()));
     }
+    
+    @Test
+    public void creatingUserConsentsToAllRequiredConsents() {
+        Study study = null;
+        try {
+            study = TestUtils.getValidStudy(StormPathUserAdminServiceTest.class);
+            studyService.createStudy(study);
```
DwayneJengSage: Creating a study is a very heavy-weight operation. Is there any way we can run this test without creating and destroying a study every time?

test/org/sagebionetworks/bridge/services/StormPathUserAdminServiceTest.java
```diff
@@ -150,4 +163,31 @@ public void testDeleteUserThatHasBeenDeleted() {
...
+            subpop.setName("Another subpopulation");
+            subpop.setRequired(true);
+            subpopService.createSubpopulation(study, subpop);
+            
+            UserSession session = userAdminService.createUser(signUp, study, null, true, true);
+            List<ConsentStatus> statuses = new ArrayList<ConsentStatus>(session.getUser().getConsentStatuses().values());
+            
+            // There are two consents, the user is consented to both of them, the session reflects this
+            assertTrue(statuses.get(0).isConsented());
+            assertTrue(statuses.get(1).isConsented());
```
DwayneJengSage: We should check user.doesConsent() just to make sure this fix covers what we need it to cover.

DwayneJengSage:
Overall looks good. Some minor comments and some comments related to test.

test/org/sagebionetworks/bridge/services/StormPathUserAdminServiceTest.java
alxdarksage: Removed

app/org/sagebionetworks/bridge/services/UserAdminService.java
alxdarksage: Yes, changed

test/org/sagebionetworks/bridge/services/StormPathUserAdminServiceTest.java
alxdarksage: I just moved this test to use mocks. The problem is that the mock consentService isn't going to update the consent status records, nor should it in the test. So all this test will test is that consentService.consentToResearch() is called for all the consents (or that if a sub pop guid is provided, that is used instead).

PR #997, opened 2016-01-29
Changing sign up logic so we don't get successes where nothing happens on the server.

alxdarksage:


app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -176,17 +176,25 @@ public void signUp(Study study, SignUp signUp, boolean isAnonSignUp) {
...
-                requestResetPassword(study, email);
-                logger.info("Sign up attempt for existing email address in study '"+study.getIdentifier()+"'");
-            } else {
-                throw e;
+                accountDao.requestResetPassword(study, email);
+                return;
+            } catch(EntityNotFoundException noEmailException) {
+                logger.info("Email not found, must notify user");
+                // But say the user name exists, not the account
+                throw new EntityAlreadyExistsException(exception.getEntity(), "Username already exists.");
```
DwayneJengSage, line 197: As noted in the meeting, the apps currently all use email for the username. What happens if I try to sign up with a username that matches someone else's email address, but with an email address that's my own (unused) email address? Could this be abused to enumerate email addresses?

alxdarksage, line 197: Yes, in this case, you will get a username is already taken error. 

Basically I feel like this is a weakness of the clients, not us or the API. It may mean that this problem can't be fixed... although it shouldn't come up in the clients.

Leave this PR and I'll send an email describing the issue and why it's probably unwise to fix. We may just have to close this as well as the JIRA as a won't fix.

alxdarksage:
Going to eliminate username altogether. Stay tuned.

PR #1000, opened 2016-02-02
Email verification

alxdarksage:
- service to add and verify an email address as a verified sender at SES (for consent emails);
- can query to find the status or request to verify the support email address of a study through the API (verification skipped if the address is already verified)
- tied into creating and saving studies. 

I envision using the APIs to give some UI to study developers similar to the AWS console for this, where the status is retrieved and shown separately next to the support email address, and if it is "UNVERIFIED" there will be a button to verify, if pending a button to resend the request.


app/org/sagebionetworks/bridge/cache/CacheProvider.java
```diff
@@ -207,7 +209,31 @@ public void removeString(String cacheKey) {
...
+            String cacheKey = RedisKey.STUDY_EMAIL_STATUS.getRedisKey(email);
+            String result = jedisOps.setex(cacheKey, BridgeConstants.BRIDGE_STUDY_EMAIL_STATUS_IN_SECONDS, status.name());
+            if (!"OK".equals(result)) {
+                throw new BridgeServiceException("View storage error");
+            }
+        } catch (Throwable e) {
+            promptToStartRedisIfLocal(e);
+            throw new BridgeServiceException(e);
+        }
+    }
```
DwayneJengSage: It looks like email verification calls are fairly cheap. It's probably not worth building cache logic for this.

test/org/sagebionetworks/bridge/play/controllers/StudyControllerTest.java
```diff
@@ -107,7 +139,54 @@ public void canSendEmailRoster() throws Exception {
...
-        verify(userProfileService).sendStudyParticipantRoster(study);
+        verify(mockUserProfileService).sendStudyParticipantRoster(mockStudy);
+    }
+    
+    @Test
+    public void getEmailStatusNoCache() throws Exception {
+        controller.getEmailStatus();
+        
+        verify(mockVerificationService).getEmailStatus(EMAIL_ADDRESS);
+    }
```
DwayneJengSage: We should verify the return value, not just that the backing service is called.

test/org/sagebionetworks/bridge/services/EmailVerificationServiceTest.java
```diff
@@ -0,0 +1,125 @@
...
+    
+    @Test
+    public void getEmailStatus() {
+        mockSession("Success");
+        
+        EmailVerificationStatus status = service.getEmailStatus(EMAIL_ADDRESS);
+        
+        verify(sesClient).getIdentityVerificationAttributes(any());
+        assertEquals(EmailVerificationStatus.VERIFIED, status);
+    }
```
DwayneJengSage, line 115: Unit test for null (unverified) status?

app/org/sagebionetworks/bridge/services/EmailVerificationStatus.java
```diff
@@ -0,0 +1,22 @@
...
+
+public enum EmailVerificationStatus {
+    UNVERIFIED,
+    PENDING,
+    VERIFIED;
+    
+    public static final EmailVerificationStatus fromSesVerificationStatus(String status) {
+        // From the SES docs, this string can be : "Pending", "Success", "Failed", or "TemporaryFailure". 
+        // We also call Pending == Verified and Sending == Pending. Everything else is unverified 
+        // and is eligible to trigger the verification workflow again.
```
DwayneJengSage: Comment says Pending == Verified and Sending == Pending, which is confusing.

app/org/sagebionetworks/bridge/cache/CacheProvider.java
alxdarksage: Okay, I removed it.

test/org/sagebionetworks/bridge/play/controllers/StudyControllerTest.java
alxdarksage: added

test/org/sagebionetworks/bridge/services/EmailVerificationServiceTest.java
alxdarksage: As set up, studies can't be saved without a support email and that's what we pass in. So there's no validation of a submitted value here because no value is submitted from the user. There's emailDoesntExistRequestVerification() which tests what happens if SES has not seen the email before (you get back a null value for value in the attribute map).

app/org/sagebionetworks/bridge/services/EmailVerificationStatus.java
alxdarksage: Updated

test/org/sagebionetworks/bridge/services/EmailVerificationServiceTest.java
alxdarksage, line 115: Added these. We check for a couple of nulls that don't happen when testing against SES, plus we have tests for them.

PR #1001, opened 2016-02-02
Username is removed from the system

alxdarksage:
- only email and password now required for sign up/sign in (we use email for username in Stormpath, where it is required information);
- username continues to be accepted silently (on sign up we silently ignore it and use the email address; on sign in, you can submit it and we will treat it as if you submitted the value under the "email" property.... it does need to be the email, so we will have to migrate developer, researcher, and admin accounts);
- user profile, session still have "username" property (it's the email address of the user), in case someone looks at this;
- this fixes the case where an existing username/new email either fails silently, confusing the user, or throws an error that exposes to the caller that the username is in use. Since our apps all already set the username to be people's email addresses, this was not acceptable (it's a session enumeration vulnerability).

This should be backwards compatible IF the username of existing accounts use the email address. Will adjust non-app accounts before we roll this out to each environment.

### DIFF
--- a/app/org/sagebionetworks/bridge/models/accounts/SignUp.java
+++ b/app/org/sagebionetworks/bridge/models/accounts/SignUp.java
@@ -60,7 +60,7 @@ public final class SignUp implements BridgeEntity {
 
     @Override
     public String toString() {
-        return "SignUp [email=" + email + ", password=" + password + ", roles=" + roles + ", dataGroups="+ dataGroups +"]";
+        return "SignUp [email=" + email + ", password=[redacted], roles=" + roles + ", dataGroups="+ dataGroups +"]";
     }
 
 }
--- a/app/org/sagebionetworks/bridge/models/accounts/SignUp.java
+++ b/app/org/sagebionetworks/bridge/models/accounts/SignUp.java
@@ -12,25 +12,18 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 
 public final class SignUp implements BridgeEntity {
 
-    private final String username;
     private final String email;
     private final String password;
     private final Set<Roles> roles;
     private final Set<String> dataGroups;
 
     @JsonCreator
-    public SignUp(@JsonProperty("username") String username, @JsonProperty("email") String email, 
-            @JsonProperty("password") String password, @JsonProperty("roles") Set<Roles> roles, 
-            @JsonProperty("dataGroups") Set<String> dataGroups) {
-        this.username = username;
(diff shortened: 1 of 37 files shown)


app/org/sagebionetworks/bridge/services/UserDataDownloadViaSqsService.java
```diff
@@ -52,14 +52,14 @@ public final void setSqsClient(AmazonSQSClient sqsClient) {
...
-        String username = user.getUsername();
+        String email = user.getEmail();
         String startDateStr = dateRange.getStartDate().toString();
         String endDateStr = dateRange.getEndDate().toString();
 
         // construct message as string-string map
         Map<String, String> requestMap = new HashMap<>();
         requestMap.put(REQUEST_KEY_STUDY_ID, studyId);
-        requestMap.put(REQUEST_KEY_USERNAME, username);
+        requestMap.put(REQUEST_KEY_USERNAME, email);
```
alxdarksage, line 62: Are there any consequences to this change? Production users have the same value for username and email, so it should appear no different after deployment.

app/org/sagebionetworks/bridge/models/accounts/SignUp.java
```diff
@@ -12,27 +12,20 @@
...
     private final Set<Roles> roles;
     private final Set<String> dataGroups;
 
     @JsonCreator
-    public SignUp(@JsonProperty("username") String username, @JsonProperty("email") String email, 
-            @JsonProperty("password") String password, @JsonProperty("roles") Set<Roles> roles, 
-            @JsonProperty("dataGroups") Set<String> dataGroups) {
-        this.username = username;
+    public SignUp(@JsonProperty("email") String email, @JsonProperty("password") String password,
+            @JsonProperty("roles") Set<Roles> roles, @JsonProperty("dataGroups") Set<String> dataGroups) {
```
DwayneJengSage, line 22: Add a unit test where the JSON contains both username and email. This will ensure that when old clients pass in username, nothing breaks.

app/org/sagebionetworks/bridge/services/UserDataDownloadViaSqsService.java
DwayneJengSage, line 62: As long as all usernames are the same as email addresses, this should work fine.

As a sanity test, I tested this with my test account. It works fine.

app/org/sagebionetworks/bridge/models/accounts/SignUp.java
alxdarksage, line 22: Added for both SignIn and SignUp

PR #1008, opened 2016-02-09
stub implementation for mPower visualization

DwayneJengSage:
See https://sagebionetworks.jira.com/wiki/display/BRIDGE/mPower+Visualization

Stub implementation just returns random values, to allow early integration by app developers. Next step is to fill in the implementation.

Testing done:
- unit tests
- manually test normal case, param validation, and 401s for not signed in users

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/MpowerVisualizationController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/MpowerVisualizationController.java
@@ -1,0 +1,67 @@
+package org.sagebionetworks.bridge.play.controllers;
+
+import java.util.Set;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.google.common.collect.ImmutableSet;
+import org.joda.time.LocalDate;
+import org.joda.time.Period;
+import org.joda.time.PeriodType;
+import org.springframework.stereotype.Controller;
+import play.mvc.Result;
+
+import org.sagebionetworks.bridge.exceptions.BadRequestException;
+import org.sagebionetworks.bridge.json.BridgeObjectMapper;
+
+/**
+ * Controller for mPower visualization. This is mPower-specific for now to get things out the door. We'll figure out if
+ * and how to generalize this later.
+ */
+@Controller
+public class MpowerVisualizationController extends BaseController {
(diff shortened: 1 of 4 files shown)


app/org/sagebionetworks/bridge/play/controllers/MpowerVisualizationController.java
```diff
@@ -0,0 +1,67 @@
...
+        } catch (RuntimeException ex) {
+            throw new BadRequestException("Invalid end date " + endDate);
+        }
+
+        // To prevent browning out the back end, the date range must be <= 45 days.
+        Period dateRange = new Period(startDateObj, endDateObj, PeriodType.days());
+        if (dateRange.getDays() > 45) {
+            throw new BadRequestException("Date range cannot exceed 45 days, startDate=" + startDate + ", endDate=" +
+                    endDate);
+        }
```
alxdarksage, line 50: If they specify nothing, can they just get the last 45 days?

conf/routes
```diff
@@ -135,6 +135,11 @@ GET    /fphs/externalId   @org.sagebionetworks.bridge.play.controllers.FPHSContr
 POST   /fphs/externalId   @org.sagebionetworks.bridge.play.controllers.FPHSController.registerExternalIdentifier
 POST   /fphs/externalIds  @org.sagebionetworks.bridge.play.controllers.FPHSController.addExternalIdentifiers
 
+# MPOWER (PARKINSON'S) ---------------------------------------------------------------------------------------
+
+# Visualization
+GET /mPower/visualization @org.sagebionetworks.bridge.play.controllers.MpowerVisualizationController.getVisualization(startDate: String ?= null, endDate: String ?= null)
```
alxdarksage: I would, for consistency, call this /parkinson/visualization since that's the study identifier.

app/org/sagebionetworks/bridge/play/controllers/MpowerVisualizationController.java
DwayneJengSage, line 50: I think this is a great idea for the real implementation. I don't think it's worth adding this to the stub implementation, only to throw it out and re-implement it for the real implementation.

conf/routes
DwayneJengSage: Changed

PR #1010, opened 2016-02-10
Expire user session

liujoshua:


app/org/sagebionetworks/bridge/services/UserAdminService.java
```diff
@@ -163,9 +163,25 @@ public UserSession createUser(SignUp signUp, Study study, SubpopulationGuid subp
...
     /**
+     * Sign out a user's session.
+     *
+     * @param study
+     *              study of target user
+     * @param email
+     *              target user
+     */
+    public void invalidateUserSession(Study study, String email) {
+        checkNotNull(study);
```
alxdarksage, line 175: I believe the ID is already in session.getUser().getId()

conf/routes
```diff
@@ -48,6 +48,7 @@ POST   /v3/consents/:timestamp/publish   @org.sagebionetworks.bridge.play.contro
 # Users
 POST   /v3/users                        @org.sagebionetworks.bridge.play.controllers.UserManagementController.createUser
 DELETE /v3/users                        @org.sagebionetworks.bridge.play.controllers.UserManagementController.deleteUser(email: java.lang.String ?= null)
+POST   /v3/users/signout                @org.sagebionetworks.bridge.play.controllers.UserManagementController.invalidateUserSession(email: java.lang.String ?= null)
```
alxdarksage: /signOut for consistency with /v3/auth/signOut

app/org/sagebionetworks/bridge/services/UserAdminService.java
```diff
@@ -163,9 +163,25 @@ public UserSession createUser(SignUp signUp, Study study, SubpopulationGuid subp
...
+        Preconditions.checkArgument(StringUtils.isNotBlank(email));
```
alxdarksage: nitpick: checkArgument can be statically imported like checkNotNull

app/org/sagebionetworks/bridge/services/UserAdminService.java
liujoshua, line 175: Not sure what this comment means

alxdarksage, line 175: I don't think you need to retrieve the user's account again, as in the controller when you retrieve the session, it has a user, and the user has the same ID you're retrieving to invalidate the session. You might change this signature to take the ID directly.

PR #1011, opened 2016-02-10
Repairing incomplete consent records.

alxdarksage:
After authentication, look at the user's signatures. If one exists for a valid and active consent, but there's no consent record in DynamoDB, create it. We've seen a small (~5?) number of people from the early days of production where this has been the case.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/AuthenticationService.java
+++ b/app/org/sagebionetworks/bridge/services/AuthenticationService.java
@@ -5,7 +5,6 @@ import static org.sagebionetworks.bridge.dao.ParticipantOption.DATA_GROUPS;
 import static org.sagebionetworks.bridge.dao.ParticipantOption.SHARING_SCOPE;
 import static org.sagebionetworks.bridge.dao.ParticipantOption.SharingScope;
 
-import java.util.List;
 import java.util.Map;
 
 import org.sagebionetworks.bridge.BridgeUtils;
@@ -260,11 +259,11 @@ public class AuthenticationService {
         
         Map<SubpopulationGuid,ConsentStatus> statuses = session.getUser().getConsentStatuses();
         for (Map.Entry<SubpopulationGuid,ConsentStatus> entry : statuses.entrySet()) {
-            ConsentSignature signature = account.getActiveConsentSignature(entry.getKey());
+            ConsentSignature activeSignature = account.getActiveConsentSignature(entry.getKey());
             ConsentStatus status = entry.getValue();
-            
-            if (signature != null && !status.isConsented()) {
-                repairConsent(session, entry.getKey(), signature);
+
+            if (activeSignature != null && !status.isConsented()) {
+                repairConsent(session, entry.getKey(), activeSignature);
                 repaired = true;
             }
         }
@@ -279,9 +278,9 @@ public class AuthenticationService {
      * If the signature exists but consentStatus says the user is not consented, then the record is missing in
      * DDB. We need to create it.
      */
-    private void repairConsent(UserSession session, SubpopulationGuid subpopGuid, ConsentSignature signature) {
+    private void repairConsent(UserSession session, SubpopulationGuid subpopGuid, ConsentSignature activeSignature) {
         logger.error("Signature found without a matching user consent record. Adding consent for " + session.getUser().getId());
-        long signedOn = signature.getSignedOn();
+        long signedOn = activeSignature.getSignedOn();
         if (signedOn == 0L) {
             signedOn = DateTimeUtils.currentTimeMillis(); // this is so old we did not record a signing date...
         }
--- a/app/org/sagebionetworks/bridge/services/AuthenticationService.java
+++ b/app/org/sagebionetworks/bridge/services/AuthenticationService.java
@@ -5,6 +5,7 @@ import static org.sagebionetworks.bridge.dao.ParticipantOption.DATA_GROUPS;
 import static org.sagebionetworks.bridge.dao.ParticipantOption.SHARING_SCOPE;
 import static org.sagebionetworks.bridge.dao.ParticipantOption.SharingScope;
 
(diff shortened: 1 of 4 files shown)


app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -238,6 +254,40 @@ public void resetPassword(PasswordReset passwordReset) throws BridgeServiceExcep
         accountDao.resetPassword(passwordReset);
     }
     
+    private void repairConsents(Study study, Account account, UserSession session, ClientInfo clientInfo){
```
DwayneJengSage, line 266: Minor: Add some clarifying comments on why we need to do this, so future us don't look back on this code and get confused.

Some explanation about migrating from UserConsent2 to UserConsent3 and that wasn't completed successfully and how that left a bunch of accounts that have consent sigs in Stormpath but not in DDB, and that account.getActiveConsentSignature() gets consent sigs from Stormpath and how user.getConsentStatuses() gets consent sigs from DDB.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -238,6 +254,40 @@ public void resetPassword(PasswordReset passwordReset) throws BridgeServiceExcep
...
+        boolean repaired = false;
+        
+        Map<SubpopulationGuid,ConsentStatus> statuses = session.getUser().getConsentStatuses();
+        for (Map.Entry<SubpopulationGuid,ConsentStatus> entry : statuses.entrySet()) {
+            ConsentSignature activeSignature = account.getActiveConsentSignature(entry.getKey());
```
DwayneJengSage, line 271: Verify if account.getActiveConsentSignature() works for Stormpath accounts that have the old consent_signature key but not the new consent_signatureS key.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -238,6 +254,40 @@ public void resetPassword(PasswordReset passwordReset) throws BridgeServiceExcep
...
+            if (activeSignature != null && !status.isConsented()) {
+                repairConsent(session, entry.getKey(), activeSignature);
+                repaired = true;
+            }
+        }
+        
+        // These are incorrect since they are based on looking up DDB records, so re-create them.
+        if (repaired) {
+            updateConsentStatuses(study, session.getUser(), clientInfo);
+        }
```
DwayneJengSage, line 283: Add a comment or rename the method to make it clear that this is updating the consent statuses in the session, which is a cache of the consents in DDB.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -238,6 +254,40 @@ public void resetPassword(PasswordReset passwordReset) throws BridgeServiceExcep
...
+    }
+    
+    /**
+     * If the signature exists but consentStatus says the user is not consented, then the record is missing in
+     * DDB. We need to create it.
+     */
+    private void repairConsent(UserSession session, SubpopulationGuid subpopGuid, ConsentSignature activeSignature) {
+        logger.error("Signature found without a matching user consent record. Adding consent for " + session.getUser().getId());
```
DwayneJengSage: Is this an error, a warning, or info? Error suggests this indicates that something is wrong with the server and devs should look into it. Warning suggests that a few of these is not a problem, but a lot of them indicates a systemic problem. Info suggests it's worth looking at it, but it shouldn't raise any alarms.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -238,6 +254,40 @@ public void resetPassword(PasswordReset passwordReset) throws BridgeServiceExcep
...
+        long signedOn = activeSignature.getSignedOn();
+        if (signedOn == 0L) {
+            signedOn = DateTimeUtils.currentTimeMillis(); // this is so old we did not record a signing date...
+        }
```
DwayneJengSage, line 295: Is it possible to test this specific case? Evidence suggests that all affected users do not have a signedOn in their Stormpath consent sig. Ergo, this is the most important case to test.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage, line 283: renamed to updateInMemoryConsentStatuses

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage, line 266: Added

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage: I'd like to know if this happens. I don't think we need to do anything about it, we'd just do what we're automating here. But I would like to look and see if users are still being created in this state, or if this is only a historical problem. That is, some users have had a DynamoUserConsent2 record, but some have not.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage, line 295: In the test for this, I zeroed out the signature, and deleted the history putting the signature in the singular field slot, as was done before implementing consent withdrawal. The repair method still repairs the account correctly.

DwayneJengSage:
The log message in AuthenticationService.repairConsent() should be an info, not an error. Errors and warnings will trigger alarms, and it sounds like we don't need or want to trigger alarms when that happens.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage, line 271: Added to test

PR #1012, opened 2016-02-11
Use the verifyEmailEnabled status flag to enable/disable the workflow…

alxdarksage:
… in Stormpath.

We've already verified setting/updating the property in the study, and we verify that we update Stormpath when the value changes. I have manually changed the value and verified that it is changed in Stormpath (this changes workflow outside the application that we have no way to test).

### DIFF
--- a/app/org/sagebionetworks/bridge/stormpath/StormpathDirectoryDao.java
+++ b/app/org/sagebionetworks/bridge/stormpath/StormpathDirectoryDao.java
@@ -201,7 +201,8 @@ public class StormpathDirectoryDao implements DirectoryDao {
         ModeledEmailTemplate template = policy.getAccountVerificationEmailTemplates().single();
         updateTemplate(study, template, study.getVerifyEmailTemplate(), "verifyEmail");
 
-        policy.setVerificationEmailStatus(EmailStatus.ENABLED);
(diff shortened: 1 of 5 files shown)


app/org/sagebionetworks/bridge/stormpath/StormpathDirectoryDao.java
```diff
@@ -201,7 +201,8 @@ private void adjustVerifyEmailPolicies(Study study, Directory directory) {
         ModeledEmailTemplate template = policy.getAccountVerificationEmailTemplates().single();
         updateTemplate(study, template, study.getVerifyEmailTemplate(), "verifyEmail");
 
-        policy.setVerificationEmailStatus(EmailStatus.ENABLED);
+        EmailStatus verifyEmailStatus = study.isEmailVerificationEnabled() ? EmailStatus.ENABLED : EmailStatus.DISABLED;
+        policy.setVerificationEmailStatus(verifyEmailStatus);
```
DwayneJengSage, line 205: Would it be better to have this test in StormpathDirectoryDaoTest, so that the code is 1:1 with the test?

alxdarksage, line 205: Moved the test for this to StormpathDirectoryDaoTest.java.

PR #1013, opened 2016-02-11
implement mPower Visualization, backed by DDB

DwayneJengSage:
Replace the stub implementation with a real implementation, backed by DDB.

I was originally going to gate this on studyId=parkinson, but then I decided to make it easier to test, I'd allow it in any study. healthCodes aren't shared across studies, so this should be fine.

Testing done:
- added new unit tests
- manual tests: specified date range, default date range (yesterday), date range larger than available data

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoMpowerVisualizationDao.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoMpowerVisualizationDao.java
@@ -1,0 +1,55 @@
+package org.sagebionetworks.bridge.dynamodb;
+
+import java.util.List;
+import javax.annotation.Resource;
+
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBQueryExpression;
+import com.amazonaws.services.dynamodbv2.model.AttributeValue;
+import com.amazonaws.services.dynamodbv2.model.ComparisonOperator;
+import com.amazonaws.services.dynamodbv2.model.Condition;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.joda.time.LocalDate;
+import org.springframework.stereotype.Component;
+
+import org.sagebionetworks.bridge.dao.MpowerVisualizationDao;
+import org.sagebionetworks.bridge.json.BridgeObjectMapper;
+
+/** DDB implementation of mPower visualization. */
+@Component
+public class DynamoMpowerVisualizationDao implements MpowerVisualizationDao {
+    private DynamoDBMapper mapper;
+
+    /** DDB mapper for the mPower visualization table. */
+    @Resource(name = "mpowerVisualizationDdbMapper")
+    final void setDdbMapper(DynamoDBMapper mapper) {
+        this.mapper = mapper;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public JsonNode getVisualization(String healthCode, LocalDate startDate, LocalDate endDate) {
+        // hash key comes from study and health code
+        DynamoMpowerVisualization hashKey = new DynamoMpowerVisualization();
+        hashKey.setHealthCode(healthCode);
+
+        // range key is between start date and end date
+        Condition dateCondition = new Condition().withComparisonOperator(ComparisonOperator.BETWEEN)
+                .withAttributeValueList(new AttributeValue().withS(startDate.toString()),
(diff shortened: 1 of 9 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoMpowerVisualizationDao.java
```diff
@@ -0,0 +1,55 @@
...
+    @Resource(name = "mpowerVisualizationDdbMapper")
+    final void setDdbMapper(DynamoDBMapper mapper) {
+        this.mapper = mapper;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public JsonNode getVisualization(String healthCode, LocalDate startDate, LocalDate endDate) {
+        // hash key comes from study and health code
+        DynamoMpowerVisualization hashKey = new DynamoMpowerVisualization();
```
alxdarksage, line 34: As there's nothing specific to mPower, what about renaming this to something more generic?

app/org/sagebionetworks/bridge/services/MpowerVisualizationService.java
```diff
@@ -0,0 +1,79 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import org.sagebionetworks.bridge.dao.MpowerVisualizationDao;
+import org.sagebionetworks.bridge.exceptions.BadRequestException;
+import org.sagebionetworks.bridge.json.DateUtils;
+
+/** mPower visualization service. */
+@Component
+public class MpowerVisualizationService {
```
alxdarksage, line 16: Same thing, how about VisualizationService

app/org/sagebionetworks/bridge/services/MpowerVisualizationService.java
```diff
@@ -0,0 +1,79 @@
...
+     *
+     * @param healthCode
+     *         user's health code
+     * @param startDateStr
+     *         start date for visualization
+     * @param endDateStr
+     *         end date for visualization
+     * @return raw JSON containing visualization data
+     */
+    public JsonNode getVisualization(String healthCode, String startDateStr, String endDateStr) {
```
alxdarksage: I would expect types to be converted by the time a service is called. IMO the dates are only strings because they were retrieved via HTTP, they should be parsed in the controller and all HTTP conversion dealt with before the service. This allows the service to be used elsewhere without spurious conversions to strings.

app/org/sagebionetworks/bridge/dynamodb/DynamoMpowerVisualizationDao.java
```diff
@@ -0,0 +1,55 @@
...
+
+        // query
+        DynamoDBQueryExpression<DynamoMpowerVisualization> query =
+                new DynamoDBQueryExpression<DynamoMpowerVisualization>().withHashKeyValues(hashKey)
+                        .withRangeKeyCondition("date", dateCondition);
+        List<DynamoMpowerVisualization> vizList = mapper.query(DynamoMpowerVisualization.class, query);
+
+        // convert to JSON object
+        ObjectNode vizColNode = BridgeObjectMapper.get().createObjectNode();
+        for (DynamoMpowerVisualization oneViz : vizList) {
```
liujoshua, line 50: Order of the keys for a JSON object aren't supposed to be meaningful, but here they are dates which have a natural order for iteration. It might be better to have an array so make iterating the data points easier

DwayneJengSage, line 50: I can see it either way. I think ultimately this should be driven by what the clients need. I've added a comment to https://sagebionetworks.jira.com/browse/BRIDGE-1115 to ask which one they need.

Note that the response might not come back before we merge this pull request, so we may need to go back and change it.

app/org/sagebionetworks/bridge/dynamodb/DynamoMpowerVisualizationDao.java
DwayneJengSage, line 34: Discussed this on IM. The code isn't specific to mPower, but the feature definitely is. We can't know for sure whether this will fit other use cases until we have other use cases, and given the timeline and constraints, we don't have time to design this in a generic way. As such, I'd prefer clearly labeling this as mPower-specific until we know it's not.

We could always rename it later.

app/org/sagebionetworks/bridge/services/MpowerVisualizationService.java
DwayneJengSage: Moved to controller.

PR #1021, opened 2016-02-17
Separate selection criteria to a different table

alxdarksage:
Criteria for filtering in Subpopulation and SchedulePlan are moved out into a separate Criteria table. 
Subpopulation currently implements criteria and when manipulated, we create the criteria object out of it as a step toward migrating this. ScheduleCriteria has been reworked to contain a Criteria object, rather than implementing the Criteria interface (this is not in use yet). After migration the same update will be made to subpopulation.

Tests: I have updated the Java SDK and run it against this version of BridgePF and the tests pass (updated to Criteria is a member object of Subpopulation and CriteriaScheduleStrategy). I'll check in the SDK tests after this is integrated.

The researcher UI will be broken until it is updated.


app/org/sagebionetworks/bridge/dynamodb/DynamoCriteriaDao.java
```diff
@@ -0,0 +1,77 @@
...
+    
+    @Resource(name = "criteriaMapper")
+    final void setCriteriaMapper(DynamoDBMapper criteriaMapper) {
+        this.criteriaMapper = criteriaMapper;
+    }
+    
+    // Until criteria as an interface is shifted off of SchedulePlan/Subpopulation, 
+    // we must copy from an interface to an implementation.
+    @Override
+    public Criteria copyCriteria(String key, Criteria criteria) {
```
liujoshua: Seems like this method could be either
a) made not public because getCriteria and createOrUpdateCreateCriteria will always call it
b) be moved to Criteria instead of CriteriaDao (copy seems separate from DAO and would help clean up ScheduleCriteria's build method)

Or a combination of a and b

test/org/sagebionetworks/bridge/dynamodb/DynamoCriteriaDaoTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+import org.sagebionetworks.bridge.models.Criteria;
+
+import com.google.common.collect.Sets;
+
+@ContextConfiguration("classpath:test-context.xml")
+@RunWith(SpringJUnit4ClassRunner.class)
+public class DynamoCriteriaDaoTest {
```
DwayneJengSage, line 25: Is it worth adding a test to verify that we can create/update from a ScheduleCriteria or a Subpopulation?

app/org/sagebionetworks/bridge/dynamodb/DynamoCriteriaDao.java
```diff
@@ -0,0 +1,77 @@
...
+    public void deleteCriteria(String key) {
+        checkArgument(isNotBlank(key));
+        
+        DynamoCriteria hashKey = new DynamoCriteria();
+        hashKey.setKey(key);
+        
+        Criteria criteria = criteriaMapper.load(hashKey);
+        if (criteria != null) {
+            criteriaMapper.delete(hashKey);
+        }
```
DwayneJengSage: Do we want to error if the key doesn't exist? Or silently return? Should we add a unit test case for this?

app/org/sagebionetworks/bridge/dynamodb/DynamoCriteriaDao.java
liujoshua: Maybe getCriteria(key) could be overloaded with getCriteria(key, defaultValue) or something

app/org/sagebionetworks/bridge/models/schedules/ScheduleCriteria.java
```diff
@@ -5,60 +5,68 @@
...
 import com.google.common.collect.Sets;
 
+@JsonDeserialize(builder=ScheduleCriteria.Builder.class)
 public final class ScheduleCriteria implements Criteria {
     private final Schedule schedule;
-    private final Integer minAppVersion;
-    private final Integer maxAppVersion;
-    private final Set<String> allOfGroups;
-    private final Set<String> noneOfGroups;
+    private final Criteria criteria;
```
DwayneJengSage, line 12: Looking at DDB, it's trivial to see that no one is currently using CriteriaScheduleStrategy. As such, we have a lot more leeway into changing this. Furthermore, we control all the studies currently using schedules, so it's trivially easy to ensure this remains true.

For one, if the goal is to consolidate all of the Criteria objects into a single class, it doesn't make sense for ScheduleCriteria to be a subclass of Criteria. Instead of having ScheduleCriteria contain both a Criteria and a Schedule, it would make more sense for CriteriaScheduleStrategy to instead contain a list of Criteria and a map from Criteria keys to schedules. This avoids a lot of the intermediate complexity in ScheduleCriteria.

The only complication is the integration tests, but I don't foresee that being difficult to work around.

alxdarksage, line 12: Very true, in this case I can just get rid of all the intermediate stuff and remove the interface.

app/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDao.java
```diff
@@ -88,39 +95,100 @@ public SchedulePlan getSchedulePlan(StudyIdentifier studyIdentifier, String guid
...
+        if (plan.getStrategy() instanceof CriteriaScheduleStrategy) {
+            CriteriaScheduleStrategy strategy = (CriteriaScheduleStrategy)plan.getStrategy();
+            for (int i=0; i < strategy.getScheduleCriteria().size(); i++) {
+                ScheduleCriteria scheduleCriteria = strategy.getScheduleCriteria().get(i);
+                
+                // Add the key... builder makes this verbose
+                scheduleCriteria = new ScheduleCriteria.Builder()
+                        .withSchedule(scheduleCriteria.getSchedule())
+                        .withCriteria(scheduleCriteria.getCriteria())
+                        .withKey(getKey(plan, i)).build();
```
DwayneJengSage: What's the goal here? If the goal is to change or create the criteria key, the copyCriteria() method should do the trick. If the goal is to pass both the key and the criteria to the consumer, Java 8 allows lambdas with multiple args. (You might have to create a functional interface to represent this, but ultimately it'll be cleaner.)

If the goal is the preserve the "schedule" field, as noted in another comment, since CriteriaScheduleStrategy currently isn't in use, so we can refactor this to not include the "schedule" field.

app/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDao.java
```diff
@@ -88,39 +95,100 @@ public SchedulePlan getSchedulePlan(StudyIdentifier studyIdentifier, String guid
...
+    // and save that. We then set that criteria object on the scheduleCriteria. Once all models have a criteria 
+    // object, this will be removed along with the fields on the scheduleCriteria object.
+    private Criteria persistCriteria(ScheduleCriteria scheduleCriteria) {
+        Criteria criteria = scheduleCriteria.getCriteria();
+        Criteria makeCopyOf = (criteria == null) ? scheduleCriteria : criteria;
+
+        Criteria copy = criteriaDao.copyCriteria(scheduleCriteria.getKey(), makeCopyOf);
+        criteriaDao.createOrUpdateCriteria(copy);
+        return copy;
+    }
```
DwayneJengSage, line 174: createOrUpdateCriteria() already calls copy() internally. After the aforementioned refactor, this could be as simple as a passthrough to createOrUpdateCriteria.

app/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDao.java
```diff
@@ -88,39 +95,100 @@ public SchedulePlan getSchedulePlan(StudyIdentifier studyIdentifier, String guid
...
+        checkNotNull(plan);
         
         plan.setStudyKey(studyIdentifier.getIdentifier());
         plan.setGuid(BridgeUtils.generateGuid());
         plan.setModifiedOn(DateUtils.getCurrentMillisFromEpoch());
+        
+        forEachCriteria(plan, scheduleCriteria -> persistCriteria(scheduleCriteria));
         mapper.save(plan);
         return plan;
     }
```
DwayneJengSage, line 116: It's probably worth thinking about how we store this in DDB. I feel like it would be cleanest if the Schedule Plan in DDB stored criteria as a list of criteria IDs (or maybe even a count of criteria, since the IDs all follow the same pattern). When we read it from DDB, we query the Criteria table for each criteria ID and populate it into the Schedule Plan.

When we create or save CriteriaSchedulePlanStrategies, the JSON includes the criteria in place, and we strip out the Criteria, save them individually to the Criteria table, and write the list of criteria IDs (or counts) into the Strategy.

This way, we can avoid saving the same thing to DDB twice.

app/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDao.java
```diff
@@ -88,39 +95,100 @@ public SchedulePlan getSchedulePlan(StudyIdentifier studyIdentifier, String guid
...
+                        .withCriteria(scheduleCriteria.getCriteria())
+                        .withKey(getKey(plan, i)).build();
+
+                Criteria criteria = consumer.apply(scheduleCriteria);
+
+                // Now add the new criteria object and set new object in list
+                scheduleCriteria = new ScheduleCriteria.Builder()
+                        .withSchedule(scheduleCriteria.getSchedule())
+                        .withCriteria(criteria).build();
+                strategy.getScheduleCriteria().set(i, scheduleCriteria);
```
DwayneJengSage: The save and the delete don't really need this block, since we don't really need to save criteria into the SchedulePlans table, and delete will just ignore it anyway. It might be cleaner to just have separate loops and helper methods for get, save, and delete.

app/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDao.java
```diff
@@ -88,39 +95,100 @@ public SchedulePlan getSchedulePlan(StudyIdentifier studyIdentifier, String guid
...
+        return copy;
+    }
+
+    private Criteria loadCriteria(ScheduleCriteria scheduleCriteria) {
+        Criteria criteria = criteriaDao.getCriteria(scheduleCriteria.getKey());
+        if (criteria == null) {
+            criteria = criteriaDao.copyCriteria(scheduleCriteria.getKey(), scheduleCriteria);
+        }
+        return criteria;
+    }
```
DwayneJengSage, line 182: Since no one currently uses CriteriaScheduleStrategy, after the refactor, this can be a direct call to getCriteria.

test/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDaoMockTest.java
```diff
@@ -0,0 +1,213 @@
...
+    @Before
+    public void before() {
+        mapper = mock(DynamoDBMapper.class);
+        criteriaDao = mock(CriteriaDao.class);
+        dao = new DynamoSchedulePlanDao();
+        dao.setSchedulePlanMapper(mapper);
+        dao.setCriteriaDao(criteriaDao);
+        
+        // We need the copy constructor to work in order to verify CriteriaDao works.
+        when(criteriaDao.copyCriteria(any(), any())).thenAnswer(invocation -> {
```
DwayneJengSage: You're looking for .thenCallRealMethod()

DwayneJengSage: Alternatively/additionally, move the copy constructor to the Criteria class, since it doesn't actually depend on implementation. (Implementation details are abstracted away behind Criteria.create().)

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulation.java
```diff
@@ -49,6 +51,16 @@ public DynamoSubpopulation() {
...
     
     @JsonIgnore
+    @DynamoDBIgnore
+    public String getKey() {
+        return "subpopulation:"+guid;
+    }
+    
+    @Override
+    public void setKey(String key) {
+    }
```
DwayneJengSage, line 61: I think it would be cleaner to have Subpopulation no longer implement the Criteria interface. This ensures that we stop passing in Subpopulation when we actually want a Criteria, so we can cleanly finish the refactor. This also means we don't have to do silly things like implement getKey() and setKey(), when we really don't need them.

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -164,9 +191,36 @@ public void deleteAllSubpopulations(StudyIdentifier studyId) {
...
+
+    // Load the criteria object. If it doesn't exist, assemble it from the criteria data on the subpopulation
+    // object. Once all models have a criteria object, this will be removed along with the fields on subpopulation.
+    private void loadCriteria(Subpopulation subpop) {
+        Criteria criteria = criteriaDao.getCriteria(getKey(subpop));
+        if (criteria == null) {
+            criteria = criteriaDao.copyCriteria(getKey(subpop), subpop);
+        }
+        subpop.setCriteria(criteria);
+    }
```
DwayneJengSage, line 233: If as per the previous comment, we make Subpop not implement Criteria, we won't be able to use the copy constructor and may need to duplicate a bit of logic. I think that's fine though, since this is a transitionary period, and that logic will go away.

Alternatively, there are exactly 2 non-trivial Subpopulations in production right now. We could manually migrate those, delete the test subpops, and add some logic such that if Subpop.getCriteria() is null, we trivially match that criteria.

test/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDaoMockTest.java
```diff
@@ -0,0 +1,202 @@
...
+import org.sagebionetworks.bridge.models.CriteriaContext;
+import org.sagebionetworks.bridge.models.subpopulations.Subpopulation;
+import org.sagebionetworks.bridge.models.subpopulations.SubpopulationGuid;
+
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper;
+import com.amazonaws.services.dynamodbv2.datamodeling.PaginatedQueryList;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
+
+public class DynamoSubpopulationDaoMockTest {
```
DwayneJengSage, line 42: No unit test for updateSubpop()

DwayneJengSage, line 42: No unit test for getSubpopulations() (plural)

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -113,7 +131,13 @@ public Subpopulation createDefaultSubpopulation(StudyIdentifier studyId) {
...
         subpop.setDefaultGroup(true);
         // The first group is required until the study designers say otherwise
-        subpop.setRequired(true); 
+        subpop.setRequired(true);
+        
+        // We know in this case that criteria do not exist
+        Criteria criteria = criteriaDao.copyCriteria(getKey(subpop), subpop);
+        criteriaDao.createOrUpdateCriteria(criteria);
+        subpop.setCriteria(criteria);
+
```
DwayneJengSage, line 144: Instead of adding the criteria to the Subpop and then copying it to the criteria, it would be cleaner to create the criteria directly.

test/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDaoMockTest.java
```diff
@@ -0,0 +1,202 @@
...
+            Criteria actualCriteria = Criteria.create();
+            actualCriteria.setKey(key);
+            if (criteria != null) {
+                actualCriteria.setMinAppVersion(criteria.getMinAppVersion());
+                actualCriteria.setMaxAppVersion(criteria.getMaxAppVersion());
+                actualCriteria.setAllOfGroups(criteria.getAllOfGroups());
+                actualCriteria.setNoneOfGroups(criteria.getNoneOfGroups());
+            }
+            return actualCriteria;        
+        });
```
DwayneJengSage: Similarly, .thenCallRealMethod() or refactor copyCriteria() to somewhere where it can be shared.

test/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDaoMockTest.java
DwayneJengSage, line 42: Is it worth adding a unit test for deleteAllSubpopulations()?

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDao.java
```diff
@@ -164,9 +191,36 @@ public void deleteAllSubpopulations(StudyIdentifier studyId) {
...
+    // and save that. Set that on subpopulation. Once all models have a criteria object, this will be removed 
+    // along with the fields on subpopulation.
+    private void persistCriteria(Subpopulation subpop) {
+        Criteria criteria = subpop.getCriteria();
+        Criteria makeCopyOf = (criteria == null) ? subpop : criteria;
+
+        Criteria copy = criteriaDao.copyCriteria(getKey(subpop), makeCopyOf);
+        criteriaDao.createOrUpdateCriteria(copy);
+        subpop.setCriteria(copy);
+    }
```
DwayneJengSage, line 222: If we refactor such that null criteria means match everything and manually migrate the 2 non-trivial criteria, we could simplify or remove the back-compat logic here.

test/org/sagebionetworks/bridge/dynamodb/DynamoCriteriaDaoTest.java
```diff
@@ -0,0 +1,79 @@
...
+    @Resource
+    DynamoCriteriaDao criteriaDao;
+
+    @Test
+    public void deleteAndGetAreQuiet() {
+        // Before creating an object, create and delete do not throw errors and do something sensible
+        criteriaDao.deleteCriteria("key"); // no exception
+        Criteria retrieved = criteriaDao.getCriteria("key"); // just return null
+        assertNull(retrieved);
+    }
```
DwayneJengSage, line 39: If for any reason, multiple unit tests are running at the same time, there's a chance that this might collide and not do what you expect it to. Would it be better to use a random key?

app/org/sagebionetworks/bridge/dynamodb/DynamoCriteriaDao.java
alxdarksage: This has been moved to the criteria interface as a generic copy function and removed entirely from DAO.

app/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDao.java
alxdarksage, line 116: I think we can go further and have a strategy table, but that needs to be incremental. Pulling out criteria simplifies this particular strategy object and it will make a later migration easier. But since DDB doesn't support joins it's also risky to query multiple tables for the data, so this might be better done if/when we move to a RDBMS.

app/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDao.java
alxdarksage: Setting the key in case the items have been reordered. This code is much simpler now,

app/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulation.java
alxdarksage, line 61: We use this (the fact that these fields are set on the subpopulation and are columns in the table). We have to migrate. After deployment, saving each subpopulation should save a criteria object, then we can remove the interface, the fields, etc. and redeploy.

app/org/sagebionetworks/bridge/models/schedules/ScheduleCriteria.java
alxdarksage, line 12: I've updated so ScheduleStrategy just includes a schedule and criteria, and doesn't implement the Criteria interface. That made building these simpler (new ScheduleStrategy(schedule, criteria)), so I removed the builder, and adjusted all accordingly.

test/org/sagebionetworks/bridge/dynamodb/DynamoCriteriaDaoTest.java
alxdarksage, line 39: Changing to random key

test/org/sagebionetworks/bridge/dynamodb/DynamoCriteriaDaoTest.java
alxdarksage, line 25: Added test using subpopulation which is still implementing the Criteria interface. Also some tests that a criteria object is created and updated in the course of CRUDing a subpopulation.

test/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDaoMockTest.java
```diff
@@ -0,0 +1,189 @@
...
+import org.sagebionetworks.bridge.models.subpopulations.Subpopulation;
+import org.sagebionetworks.bridge.models.subpopulations.SubpopulationGuid;
+
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper;
+import com.amazonaws.services.dynamodbv2.datamodeling.PaginatedQueryList;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
+
+@RunWith(MockitoJUnitRunner.class)
+public class DynamoSubpopulationDaoMockTest {
```
DwayneJengSage, line 42: Missing unit tests for:
- updateSubpopulation()
- getSubpopulations() (plural)
- deleteAllSubpopulations()

test/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDaoMockTest.java
```diff
@@ -0,0 +1,198 @@
...
+import org.sagebionetworks.bridge.models.schedules.ScheduleCriteria;
+import org.sagebionetworks.bridge.models.schedules.SchedulePlan;
+
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper;
+import com.amazonaws.services.dynamodbv2.datamodeling.QueryResultPage;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
+
+@RunWith(MockitoJUnitRunner.class)
+public class DynamoSchedulePlanDaoMockTest {
```
DwayneJengSage, line 39: What happens if we create or update a CriteriaSchedulePlan that has no criteria in its schedule? Is that prevented further up the the stack? Should we create a unit test to test for that?

app/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDao.java
```diff
@@ -88,39 +95,103 @@ public SchedulePlan getSchedulePlan(StudyIdentifier studyIdentifier, String guid
...
+     */
+    private Criteria loadCriteria(ScheduleCriteria scheduleCriteria) {
+        String key = scheduleCriteria.getCriteria().getKey();
+        Criteria criteria = criteriaDao.getCriteria(key);
+        if (criteria == null) {
+            criteria = Criteria.create();
+            criteria.setKey(key);
+        }
+        return criteria;
+    }
```
DwayneJengSage, line 182: When will the null case happen? Persist criteria ensures we always create a criteria, and there are no pre-existing CriteriaScheduleStrategies in DDB, so there's no back-compat case to handle.

app/org/sagebionetworks/bridge/models/Criteria.java
```diff
@@ -2,14 +2,74 @@
...
+        DynamoCriteria crit = new DynamoCriteria();
+        if (criteria != null) {
+            crit.setKey(criteria.getKey());
+            crit.setMinAppVersion(criteria.getMinAppVersion());
+            crit.setMaxAppVersion(criteria.getMaxAppVersion());
+            crit.setNoneOfGroups(criteria.getNoneOfGroups());
+            crit.setAllOfGroups(criteria.getAllOfGroups());
+        }
+        return crit;
+    }
```
DwayneJengSage, line 35: Unit test for copy function?

app/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategy.java
```diff
@@ -53,16 +53,22 @@ public Schedule getScheduleForUser(SchedulePlan plan, ScheduleContext context) {
...
                 errors.rejectValue("schedule", "is required");
             } else {
                 errors.pushNestedPath("schedule");
-                new ScheduleValidator(taskIdentifiers).validate(criteria.getSchedule(), errors);
+                new ScheduleValidator(taskIdentifiers).validate(schCriteria.getSchedule(), errors);
+                errors.popNestedPath();
+            }
+            if (schCriteria.getCriteria() == null) {
+                errors.rejectValue("criteria", "is required");
+            } else {
```
DwayneJengSage, line 67: Unit test for validation with missing criteria?

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
```diff
@@ -260,55 +264,32 @@ public void validatesInvalidScheduleCriteria() {
...
         // We're looking here specifically for errors generated by the strategy, not the plan
         // it's embedded in or the schedules. I've made those valid so they don't add errors.
-        strategy.addCriteria(new ScheduleCriteria.Builder()
-                .withMinAppVersion(-2)
-                .withMaxAppVersion(-10)
-                .withSchedule(SCHEDULE_FOR_STRATEGY_WITH_ONE_REQUIRED_DATA_GROUP).build());
-        
-        // Has no schedule
-        strategy.addCriteria(new ScheduleCriteria.Builder().build());
+        strategy.addCriteria(new ScheduleCriteria(SCHEDULE_FOR_STRATEGY_WITH_ONE_REQUIRED_DATA_GROUP, criteria));
```
DwayneJengSage, line 274: Unit test for ScheduleCriteria without a schedule?

app/org/sagebionetworks/bridge/models/schedules/ScheduleCriteria.java
```diff
@@ -1,92 +1,37 @@
...
-    private ScheduleCriteria(@JsonProperty("schedule") Schedule schedule, 
-            @JsonProperty("minAppVersion") Integer minAppVersion, 
-            @JsonProperty("maxAppVersion") Integer maxAppVersion, 
-            @JsonProperty("allOfGroups") Set<String> allOfGroups, 
-            @JsonProperty("noneOfGroups") Set<String> noneOfGroups) {
+    public ScheduleCriteria(@JsonProperty("schedule") Schedule schedule, 
+            @JsonProperty("criteria") Criteria criteria) {
+        checkNotNull(schedule);
+        checkNotNull(criteria);
+
```
DwayneJengSage: This is user-submitted. If the user submits a ScheduleCriteria without a schedule or a criteria, this will throw a 500 Internal Server Error. It would be better to move this logic to the validator.

app/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDao.java
```diff
@@ -88,39 +95,103 @@ public SchedulePlan getSchedulePlan(StudyIdentifier studyIdentifier, String guid
...
+     * Save the criteria object if it exists. If not, return an empty criteria object which will return the 
+     * criteria stubbed out in the JSON representation of the schedule plan.
+     */
+    private Criteria persistCriteria(ScheduleCriteria scheduleCriteria) {
+        Criteria criteria = scheduleCriteria.getCriteria();
+        if (criteria != null) {
+            criteriaDao.createOrUpdateCriteria(criteria);
+        }
+        return (criteria != null) ? criteria : Criteria.create();
+    }
```
DwayneJengSage, line 174: What does it mean if criteria is null? Can this ever happen?

alxdarksage: You could submit a schedule plan where the criteria property is null, in order to indicate there are no criteria. That's okay, we do set an empty criteria object though to reduce the possibility of errors.

app/org/sagebionetworks/bridge/models/schedules/ScheduleCriteria.java
alxdarksage: The presence of both are validated in CriteriaScheduleStrategy.validate(), which is called as part of validation. So a user should get a 400 if these are missing. Removed these so they don't interfere with validation.

app/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDao.java
alxdarksage, line 174: Again for this object, which hasn't so far been used, we can know this exists, so removing the null checks.

test/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategyTest.java
alxdarksage, line 274: We check with both validation and preconditions that a ScheduleCriteria object has a schedule and a criteria object, and both of those are validated. Added tests to verify that missing the property or having it set to null don't break things.

app/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategy.java
alxdarksage, line 67: Added

app/org/sagebionetworks/bridge/models/Criteria.java
alxdarksage, line 35: Added for all interface functions

app/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDao.java
alxdarksage, line 182: This shouldn't happen because as you say, no existing cases. removing

test/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDaoMockTest.java
alxdarksage, line 39: If there are none, the schedule object has an empty collection and that's fine. I've verified even submitting no "scheduleCriteria" property, you still get an empty collection. We validate that any scheduleCriteria object that is provided will have both a schedule and a criteria. The plan without any of these does nothing, but it's an intermediate step that's likely to be saved while editing and should be possible.

test/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDaoMockTest.java
alxdarksage, line 42: Added

app/org/sagebionetworks/bridge/models/schedules/CriteriaScheduleStrategy.java
```diff
@@ -52,18 +52,28 @@ public Schedule getScheduleForUser(SchedulePlan plan, ScheduleContext context) {
...
-        for (int i=0; i < scheduleCriteria.size(); i++) {
-            ScheduleCriteria criteria = scheduleCriteria.get(i);
-            errors.pushNestedPath("scheduleCriteria["+i+"]");
-            if (criteria.getSchedule() == null){
-                errors.rejectValue("schedule", "is required");
-            } else {
-                errors.pushNestedPath("schedule");
-                new ScheduleValidator(taskIdentifiers).validate(criteria.getSchedule(), errors);
+        if (scheduleCriteria == null) {
+            errors.rejectValue("scheduleCriteria", "is required");
```
DwayneJengSage: Minor: scheduleCriteria can't ever actually be null.

PR #1027, opened 2016-02-22
Hide studyKey on schedule plan from the JSON in the API.

alxdarksage:


test/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanTest.java
```diff
@@ -51,6 +61,9 @@ public void canSerializeDynamoSchedulePlan() throws Exception {
         assertEquals(plan.getGuid(), plan2.getGuid());
         assertEquals(plan.getLabel(), plan2.getLabel());
         assertEquals(plan.getModifiedOn(), plan2.getModifiedOn());
+        
+        ScheduleStrategy retrievedStrategy = plan.getStrategy();
+        assertEquals(retrievedStrategy, strategy);
```
DwayneJengSage, line 67: Minor: Do we want to assert that study key is null? Also, do we want to write a study key into the JSON to verify that fromJson() ignores it?

test/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDaoTest.java
```diff
@@ -71,6 +71,7 @@ public void canSerializeAndDeserializeSchedulePlan() throws Exception {
         String output = mapping.writeValueAsString(abPlan);
         
         SchedulePlan newPlan = mapping.readValue(output, SchedulePlan.class);
+        newPlan.setStudyKey(abPlan.getStudyKey()); // not serialized
         
         assertEquals("Schedule plans are equal", abPlan.hashCode(), newPlan.hashCode());
```
DwayneJengSage: Minor: Can we use equals() instead of hashCode()? If equals() and hashCode() are implemented correctly (which we test for in DynamoSchedulePlanTest), then equals() and hashCode() should be equivalent, and equals() is a stronger test.

test/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanTest.java
alxdarksage, line 67: We do test assertNull(node.get("studyKey")); above. Will add test that JSON studyKey is ignored

test/org/sagebionetworks/bridge/dynamodb/DynamoSchedulePlanDaoTest.java
alxdarksage: Not sure why this doesn't just assertEquals(). Fixed both cases of this in test.

PR #1029, opened 2016-02-24
Making changes so it's easier to determine what test is leaving a survey in the db

alxdarksage:
And not cleaning up correctly (I found a lot of "blood pressure" surveys in the UAT tables, so this will replace that with IDs that can be traced back to a specific test).


DwayneJengSage:
Looks good. Ready to merge whenever you're ready.

PR #1031, opened 2016-02-25
remove Redis dependency from unit tests

DwayneJengSage:
See https://sagebionetworks.jira.com/browse/BRIDGE-837

Our builds frequently fail due to connection issues with Redis. This change refactors the Spring configs so that in unit tests, we mock out Redis.

There was a previous attempt to use https://github.com/50onRed/mock-jedis, however, 50onRed/mock-jedis doesn't support a lot of the use cases we use Jedis for.

We also had to split out the Play modules, since some of the unit tests use test server, which loads the full Play module.

JedisTransactionTest was deleted, because the test is only meaningful when backed by a real Redis instance, and we want to remove our test dependency on Redis.

Testing done:
- updated unit tests
- ran integration test against local server to verify basic functionality


conf/test-context.xml
```diff
@@ -1,18 +1,16 @@
...
-    <import resource="file:conf/application-context.xml" />
+    <import resource="file:conf/shared-config.xml" />
+
+    <bean class="org.sagebionetworks.bridge.config.BridgeTestSpringConfig" />
 
     <bean id="testUserAdminHelper" class="org.sagebionetworks.bridge.TestUserAdminHelper" depends-on="defaultStudyBootstrapper">
         <property name="userAdminService" ref="userAdminService" />
         <property name="authService" ref="authenticationService" />
         <property name="studyService" ref="studyService" />
     </bean>
```
alxdarksage, line 15: Can this now be added to BridgeTestSpringConfig?

PR #1032, opened 2016-02-26
Persist language preferences from Accept-Language header

alxdarksage:
Once we retrieve the Accept-Language header, we save it in the user's session and in ParticipantOptions for export with the data set (might not need this initially, since different languages will go to different Synapse tables). Thereafter, we use the languages saved with the user rather than the Accept-Language header, so changes to language outside the application do not change the user's subpopulation and potentially, their consent status in a way that would be mysterious to app users.

There is no external API change for end users. Language selection is not exposed through UserSessionInfo and cannot be changed by the user.

If/when we work on the fuller localization, I would allow users to change language choice through the UserProfile, but with that comes 1) the need to declare what the possible language options are for a study, and 2) users in the app would probably need to be warned that they will have to reconsent if they do this, so it's not 100% just something that happens on the server.

See: https://sagebionetworks.jira.com/browse/BRIDGE-1119


app/org/sagebionetworks/bridge/models/accounts/User.java
```diff
@@ -90,11 +89,11 @@ public void setHealthCode(String healthCode) {
     }
 
     public String getEncryptedHealthCode() {
-        return encryptor.encrypt(healthCode);
+        return ENCRYPTOR.encrypt(healthCode);
     }
 
     public void setEncryptedHealthCode(String healthCode) {
-        this.healthCode = encryptor.decrypt(healthCode);
+        this.healthCode = ENCRYPTOR.decrypt(healthCode);
```
DwayneJengSage, line 96: Should this be static or should it be injected? What happens in unit tests?

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -310,10 +311,18 @@ private UserSession getSessionFromAccount(Study study, CriteriaContext context,
         
         user.setSharingScope(optionsService.getEnum(healthCode, SHARING_SCOPE, SharingScope.class));
         user.setDataGroups(optionsService.getStringSet(healthCode, DATA_GROUPS));
+        user.setLanguages(optionsService.getOrderedStringSet(healthCode, LANGUAGES));
```
DwayneJengSage, line 314: JIRA or separate task for optionsService.getAllOptions() or something equivalent. This allows us to make one call to DDB that gets all options for a single user.

test/org/sagebionetworks/bridge/BridgeUtilsTest.java
```diff
@@ -52,24 +52,24 @@ public void periodsNotInterpretedAsRegex() {
...
-        set = BridgeUtils.commaListToSet(" a");
-        assertEquals(Sets.newHashSet("a"), set);
+        set = BridgeUtils.commaListToOrderedSet(" a");
+        assertEquals(TestUtils.newLinkedHashSet("a"), set);
         
         // Does not produce a null value.
-        set = BridgeUtils.commaListToSet("a,,b");
-        assertEquals(Sets.newHashSet("a","b"), set);
+        set = BridgeUtils.commaListToOrderedSet("a,,b");
+        assertEquals(TestUtils.newLinkedHashSet("a","b"), set);
```
DwayneJengSage: Is it worth adding a test with the same elements but in a different order, just to verify that order is preserved?

Similarly, for setToCommaList(), where we pass in two different linked (or immutable) hash sets with the same elements in different orders and verify that the order is preserved?

test/org/sagebionetworks/bridge/services/AuthenticationServiceTest.java
```diff
@@ -401,4 +398,27 @@ public void userWithSignatureAndNoDbRecordIsRepaired() throws Exception {
         }
     }
     
+    @Test
+    public void existingLanguagePreferencesAreLoaded() {
```
DwayneJengSage, line 402: How about a unit test to get the language from the CriteriaContext?

app/org/sagebionetworks/bridge/models/accounts/User.java
alxdarksage, line 96: I know what's going on here, this class is User. It's not a managed bean. It could be a prototype bean, but then we'd have to know about spring's application context to instantiate a new user. However this could be passed in through the constructor.

test/org/sagebionetworks/bridge/BridgeUtilsTest.java
alxdarksage: assertEquals(set1, set2) doesn't actually check the order of the sets are the same. Added a method to assert that as well, added additional test to change order of arguments.

test/org/sagebionetworks/bridge/services/AuthenticationServiceTest.java
alxdarksage, line 402: Added

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage, line 314: https://sagebionetworks.jira.com/browse/BRIDGE-1137

PR #1033, opened 2016-02-26
Filter criteria using the language in CriteriaContext.

alxdarksage:
This is the simple version of filtering. I will look at the more detailed version of filtering but absent a concrete project or concrete case of simple filtering not being sufficient, I don't think this is high priority.


DwayneJengSage:
Looks good. Ready to merge.

PR #1037, opened 2016-03-01
Better participant options api

alxdarksage:
Let's do an in person review of this, it's big.

1) Consolidates type safety code for accessing ParticipantOptions table
2) Simplifies API so that you do not need to make multiple calls to retrieve multiple options for a user
3) Reduces the number of calls in the API.

This is all possible since we're loading the JSON entirely whenever we retrieve an option, so now we return an object allowing us to parse out any of the values for a record. Then refactored code to use these.


app/org/sagebionetworks/bridge/models/accounts/ParticipantOptionsLookup.java
```diff
@@ -0,0 +1,64 @@
...
+    
+    public ParticipantOptionsLookup(Map<String,String> options) {
+        checkNotNull(options);
+        this.options = options;
+    }
+    
+    public String getString(ParticipantOption option) {
+        if (options == null) {
+            return option.getDefaultValue();
+        }
```
DwayneJengSage: Redundant null check.

test/org/sagebionetworks/bridge/models/accounts/ParticipantOptionsLookupTest.java
```diff
@@ -0,0 +1,149 @@
...
+        LinkedHashSet<String> set = TestUtils.newLinkedHashSet("en","fr","es");
+        assertEquals(set, lookup.getOrderedStringSet(LANGUAGES));
+    }
+
+    @Test
+    public void getBooleanWithValue() {
+        ParticipantOptionsLookup lookup = setupLookup(EMAIL_NOTIFICATIONS, "false");
+        
+        assertEquals(Boolean.FALSE, lookup.getBoolean(EMAIL_NOTIFICATIONS));
+    }
```
DwayneJengSage, line 100: Boolean.valueOf() is asymmetrical https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#valueOf%28java.lang.String%29 I'd recommend testing both true and false in unit tests.

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceTest.java
```diff
@@ -82,122 +81,130 @@ public void setString() {
...
         
-        Map<ParticipantOption,OptionLookup> result = service.getAllOptionsForAllStudyParticipants(TEST_STUDY);
-        assertEquals(map, result);
+        AllParticipantOptionsLookup result = service.getOptionsForAllParticipants(TEST_STUDY);
+        assertEquals(allLookup, result);
         
-        verify(mockDao).getAllOptionsForAllStudyParticipants(TEST_STUDY);
+        verify(mockDao).getOptionsForAllParticipants(TEST_STUDY);
         verifyNoMoreInteractions(mockDao);
     }
```
DwayneJengSage, line 189: Minor: This test is now redundant with the previous test.

app/org/sagebionetworks/bridge/models/accounts/ParticipantOptionsLookup.java
alxdarksage: removed

test/org/sagebionetworks/bridge/models/accounts/ParticipantOptionsLookupTest.java
alxdarksage, line 100: added

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceTest.java
alxdarksage, line 189: removed

PR #1038, opened 2016-03-03
Case insensitive lang match

alxdarksage:
Language codes are not case sensitive and we don't enforce case anywhere... make sure the language matching is case insensitive as well.


app/org/sagebionetworks/bridge/models/CriteriaUtils.java
```diff
@@ -73,6 +71,21 @@ public static void validate(Criteria criteria, Set<String> dataGroups, Errors er
...
+    private static boolean languageDoesNotMatch(Set<String> preferredLanguages, String targetLanguage) {
+        if (targetLanguage == null) {
+            return false;
+        }
+        for (String prefLang : preferredLanguages) {
+            if (targetLanguage.equalsIgnoreCase(prefLang)) {
+                return false;
+            }
+        }
+        // It doesn't match if 1) target language has been specified and 2) user doesn't ask for it
```
alxdarksage: need to fix this comment

PR #1043, opened 2016-03-09
Adding API to get paged summary records of accounts.

alxdarksage:


app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
```diff
@@ -0,0 +1,51 @@
...
+
+@Controller
+public class ParticipantController extends BaseController {
+    
+    private ParticipantService participantService;
+    
+    @Autowired
+    public void setParticipantService(ParticipantService participantService) {
+        this.participantService = participantService;
+    }
```
DwayneJengSage, line 28: Minor: final on the setter

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
```diff
@@ -0,0 +1,51 @@
...
+    
+    private int getIntOrDefault(String value, int defaultValue) {
+        if (StringUtils.isBlank(value)) {
+            return defaultValue;
+        }
+        try {
+            return parseInt(value);
+        } catch(NumberFormatException e) {
+            return defaultValue;
+        }
```
DwayneJengSage, line 49: Should malformed ints degrade to default values? Or should we throw a BadRequest?

test/org/sagebionetworks/bridge/stormpath/StormpathAccountDaoTest.java
```diff
@@ -80,7 +83,47 @@ public void getAllAccounts() {
...
+            if (accounts.getTotal() < 6) {
+                for (int i=0; i < (6-accounts.getTotal()); i++) {
+                    String random = RandomStringUtils.randomAlphabetic(5);
+                    String email = "bridge-testing+"+random+"@sagebridge.org";
+                    SignUp signUp = new SignUp(email, PASSWORD, Sets.newHashSet(TEST_USERS), null);
+                    accountDao.signUp(study, signUp, false);
+                    newAccounts.add(email);
+                }
+                accounts = accountDao.getPagedAccountSummaries(study, 0, 5);
+            }
```
DwayneJengSage, line 100: As written, this test will fetch 5 accounts, then see if it's less than 6, which is always true. This means we'll create test accounts even when there are already 6 test accounts.

It would be better to first get 6 accounts and verify we have 6, then separately make the call for page size 5.

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
alxdarksage, line 49: Changed to throw BadRequestException

test/org/sagebionetworks/bridge/stormpath/StormpathAccountDaoTest.java
alxdarksage, line 100: Yes, that's not right. Changed to verify at least six first, and then retrieve 5 accounts only.

alxdarksage:
Also added .orderByEmail() to criteria.

PR #1048, opened 2016-03-12
Fetch participant

alxdarksage:
Retrieve an individual participant by email address. Along with the paged access to participant email addresses, this is going to replace the participant roster, and the service code in this PR is adapted from the ParticipantRoster generator code. However this version also includes a complete consent history for the user.

The roster code will be removed once this is completely deployed and I have scripts to user it to produce a roster, or we have a UI to do it, or something like that.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/ParticipantService.java
+++ b/app/org/sagebionetworks/bridge/services/ParticipantService.java
@@ -1,28 +1,122 @@
 package org.sagebionetworks.bridge.services;
 
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static org.sagebionetworks.bridge.BridgeConstants.API_MAXIMUM_PAGE_SIZE;
 import static org.sagebionetworks.bridge.BridgeConstants.API_MINIMUM_PAGE_SIZE;
+import static org.sagebionetworks.bridge.dao.ParticipantOption.DATA_GROUPS;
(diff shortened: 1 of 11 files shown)


app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -1,30 +1,124 @@
...
+        
+        Account account = accountDao.getAccount(study, email);
+        String healthCode = getHealthCode(account);
+
+        List<Subpopulation> subpopulations = subpopService.getSubpopulations(study.getStudyIdentifier());
+        for (Subpopulation subpop : subpopulations) {
+            // Create an empty history if there's no health Code.
+            List<UserConsentHistory> history = NO_HISTORY;
+            if (healthCode != null) {
+                history = consentService.getUserConsentHistory(study, subpop.getGuid(), healthCode, email);
```
alxdarksage: Does this always return a list? If not you should retrieve first.

test/org/sagebionetworks/bridge/models/accounts/StudyParticipant2Test.java
```diff
@@ -0,0 +1,78 @@
...
+import com.google.common.collect.Sets;
+
+public class StudyParticipant2Test {
+
+    private static final Set<String> DATA_GROUPS = Sets.newHashSet("group1","group2");
+    private static final Map<String,String> ATTRIBUTES = Maps.newHashMap();
+    static {
+        ATTRIBUTES.put("A", "B");
+        ATTRIBUTES.put("C", "D");
+    }
```
alxdarksage: ImmutableMap has a nicer Builder

PR #1052, opened 2016-03-15
Limit use of absolute schedule window

alxdarksage:
Do not use the absolute scheduling window to winnow out tasks based on their event timestamp, only use it to filter tasks by their scheduledOn timestamp.

This is being used by FPHS now, and the behavior was very counter-intuitive because it was filtering tasks out if a) the event timestamp was outside the window or b) the scheduled start time was outside the window. Just options (b) is more useful and easier to use, so switching to that.

### DIFF
--- a/test/org/sagebionetworks/bridge/models/schedules/CronActivitySchedulerTest.java
+++ b/test/org/sagebionetworks/bridge/models/schedules/CronActivitySchedulerTest.java
@@ -154,8 +154,7 @@ public class CronActivitySchedulerTest {
         
         // We'd expect, based on that schedule, to have a task:
         scheduledActivities = schedule.getScheduler().getScheduledActivities(plan, getContext(now.plusDays(4)));
-        // 1 or 2 activities depending on the time of the week (4 day window creates some overlap)
-        assertTrue(scheduledActivities.size() == 1 || scheduledActivities.size() == 2);
+        assertEquals(1, scheduledActivities.size());
         
         DateTimeUtils.setCurrentMillisSystem();
     }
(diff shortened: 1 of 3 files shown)


test/org/sagebionetworks/bridge/models/schedules/CronActivitySchedulerTest.java
```diff
@@ -133,6 +135,54 @@ public void recurringCronScheduleAgainstCalculatedEventWorks() {
...
+        schedule.setExpires("P7D");
+        
+        events.clear();
+        events.put("enrollment", DateTime.parse("2016-03-02T00:04:37.000Z"));
+        events.put("two_weeks_before_enrollment", events.get("enrollment").minusWeeks(2));
+        
+        // We'd expect, based on that schedule, to have a task:
+        scheduledActivities = schedule.getScheduler().getScheduledActivities(plan, getContext(now.plusDays(4)));
+        // 1 or 2 activities depending on the time of the week (4 day window creates some overlap)
+        assertTrue(scheduledActivities.size() == 1 || scheduledActivities.size() == 2);
```
DwayneJengSage: Isn't there exactly 1 Monday between Mar 14 and Mar 20? And don't we know the exact schedule start dates and end dates? Shouldn't we be able to determinsitically say how many events are in the schedule?

test/org/sagebionetworks/bridge/models/schedules/CronActivitySchedulerTest.java
```diff
@@ -133,6 +135,54 @@ public void recurringCronScheduleAgainstCalculatedEventWorks() {
...
+        schedule.setExpires("P7D");
+        
+        events.clear();
+        events.put("enrollment", DateTime.parse("2016-03-15T00:04:37.000Z"));
+        events.put("two_weeks_before_enrollment", events.get("enrollment").minusWeeks(2));
+        
+        // We'd expect, based on that schedule, to have a task:
+        scheduledActivities = schedule.getScheduler().getScheduledActivities(plan, getContext(now.plusDays(4)));
+        // 1 or 2 activities depending on the time of the week (4 day window creates some overlap)
+        assertTrue(scheduledActivities.size() == 1 || scheduledActivities.size() == 2);
```
DwayneJengSage: Similarly, don't we know there are exactly 2 activities here?

PR #1055, opened 2016-03-17
Filter summary accounts on email

alxdarksage:
Adding a substring, case-insensitive string filter for email address to the getPagedAccounts API.


liujoshua:
LGTM

PR #1056, opened 2016-03-17
Update user's participant options

alxdarksage:
Researcher API to update a user's participant options. Can send some or all options in the JSON for this endpoint. Includes capability of endpoint to update user's dataGroups, and may replace it.

It may be possible to remove DataGroups object and the dataGroups API for researchers, as this is a superset of that functionality.

Other PRs have to come first before this PR is merged.


app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -137,6 +142,17 @@ public StudyParticipant2 getParticipant(Study study, String email) {
         return accountDao.getPagedAccountSummaries(study, offsetBy, pageSize);
     }
     
+    public void updateParticipantOptions(Study study, String email, ParticipantOptions options) {
+        checkNotNull(options);
+        
+        Account account = accountDao.getAccount(study, email);
+        String healthCode = getHealthCode(account);
```
DwayneJengSage, line 152: You need a null-check between account and healthCode.

app/org/sagebionetworks/bridge/models/accounts/ParticipantOptions.java
```diff
@@ -0,0 +1,61 @@
...
+ * Object to deserialize JSON update of participant options. Some or all of the 
+ * values can be missing in the JSON and we will only update values that are non-null.
+ */
+public final class ParticipantOptions {
+
+    private final @Nullable SharingScope sharingScope;
+    private final @Nullable Boolean notifyByEmail;
+    private final @Nullable String externalId;
+    private final @Nullable Set<String> dataGroups;
+    private final @Nullable LinkedHashSet<String> languages;
```
DwayneJengSage: As is, if we add a new Participant Option, we'll need to update this class, the DAO, and every test that uses the ParticipantOptions constructor. This means that adding a new Participant Option is quadratically difficult with a rather large coefficient.

I feel like instead of top-level attributes, we should just have a string-to-string map. This way, we only need to update the enum and the direct consumer of the option.

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage, line 152: Added

app/org/sagebionetworks/bridge/models/accounts/ParticipantOptions.java
alxdarksage: We would still need to know for specific properties, if they are coming in as booleans or arrays, and deserialize them correctly into strings. This seems like the easiest way to do it since all the work is left to Jackson.

alxdarksage: We discussed, we can not use Jackson and add a deserializer to each enum type, to convert these into a Map{String,String}, skipping entries if property doesn't exist. That can be saved directly in the corresponding DDB table.

DwayneJengSage: Discussed this offline with Alx. The main concern here is that we have a quadratically growing list of things to change when we update ParticipantOptions, which makes this a maintainability nightmare.

On the flip side, we need to be able to parse the incoming request based on specific options and types. I think the best way to handle this would be to have the enum itself contain the parsing logic. This way, it's easy to maintain, type-safe, and the ParticipantOptions.java could still just be a map.

app/org/sagebionetworks/bridge/dao/ParticipantOption.java
```diff
@@ -1,22 +1,74 @@
...
 
-    SHARING_SCOPE(SharingScope.NO_SHARING.name()),
-    EMAIL_NOTIFICATIONS(Boolean.TRUE.toString()),
-    EXTERNAL_IDENTIFIER(null),
-    DATA_GROUPS(null),
-    LANGUAGES(null);
+    SHARING_SCOPE(SharingScope.NO_SHARING.name(), "sharingScope") {
+        public String deserialize(JsonNode node) {
+            checkNotNull(node);
+            return SharingScope.valueOf(node.asText().toUpperCase()).name();
```
DwayneJengSage: This is user input. If the caller passes in an invalid sharing scope, Bridge server will throw a 500. We should add a try-catch here or in ParticipantController to wrap the exception as a BadRequest.

app/org/sagebionetworks/bridge/dao/ParticipantOption.java
```diff
@@ -1,22 +1,74 @@
...
+            return arrayNodeToOrderedString(node);
+        }
+    };
 
     private final String defaultValue;
+    private final String fieldName;
+    
+    private static String arrayNodeToOrderedString(JsonNode node) {
+        Set<String> results = new LinkedHashSet<>();
+        ArrayNode array = (ArrayNode)node;
```
DwayneJengSage, line 66: Similarly here, if the caller doesn't pass in an ArrayNode, Bridge server will throw a 500.

DwayneJengSage:
Looks mostly good. Just a few comments w.r.t user inputs and robustness.

alxdarksage:
Added checks for JSON type errors as well as incorrect SharingScope enum names. All throw BadRequestException.

DwayneJengSage:
Looks good. Ready to merge?

PR #1059, opened 2016-03-21
Participant API to update user profile

alxdarksage:
Add ability for researcher to update the fields of a user profile that can be updated (first/last name and custom user profile attributes).


liujoshua:
Looks good

PR #1064, opened 2016-03-22
Add the email filter to the PagedResourceList

alxdarksage:
As we do for the pageSize, offsetBy, etc.


alxdarksage:
As discussed on IM, I've renamed filter --> emailFilter.

PR #1065, opened 2016-03-23
Do not return a 500 when an Accept-Language header is malformed.

alxdarksage:
Just log and continue.


DwayneJengSage:
Looks good. Ready to merge?

PR #1066, opened 2016-03-23
Participant api cleanup

alxdarksage:
- removed updating data groups for a user (this can be done through the options update call);
- but, added validation of data groups to the update options service method;
- StudyParticipant2 --> StudyParticipant (original SP class has been deleted)
- moved signing a user out to the participant controller and service
- added tests and exceptions for more error conditions

Basically all the calls that 1) require a researcher and 2) act on a study user are now done through the ParticipantController/ParticipantService. More are coming.


liujoshua:
Looks good

PR #1067, opened 2016-03-24
calling upload complete on an already completed upload won't trigger validation again

DwayneJengSage:
See https://sagebionetworks.jira.com/browse/BRIDGE-1236

Some apps, notably Diabetes, will create a bunch of duplicate uploads in parallel, then call upload complete on these uploads in parallel. This causes some the validation pipeline to run in parallel for the same records and occasionally causes concurrent modification exceptions in DDB. This is bad because (a) redundant validation (b) errors in the logs.

This change moves validation from the controller to the service, where the "can be validated" gate is being checked and actually being honored.

Testing done:
- updated unit tests
- ran upload integration tests


BiomedicalProgrammer:
LGTM

app/org/sagebionetworks/bridge/play/controllers/UploadController.java
```diff
@@ -70,10 +62,7 @@ public Result uploadComplete(String uploadId) throws Exception {
 
         // mark upload as complete
```
alxdarksage: minor: comment still relevant?

PR #1069, opened 2016-03-25
Add AccountStatus to the update of a user's profile by a researcher

alxdarksage:
DON'T MERGE. Going to refactor UserProfile/StudyParticipant first.
- Add AccountStatus to the update of a user's profile by a researcher. 
- Verify that users cannot change their own account status when updating their profile. 
- Test cleanup of UserProfileControllerTest

NOTE: This is messy and we've identified the design reasons as to why. We're going to simplify this code by consolidating all the editable fields of ParticipantOptions and Account on the UserProfile, which will be how you update these values as a user on your own account. Researchers will have further fields they can edit on the StudyParticipant object (status), which is a sub type of UserProfile that includes some read-only information like consent statuses and healthCode (for studies configured to support it).


liujoshua:
Looks good

PR #1070, opened 2016-03-28
Add a flag to turn stricter external ID validation on or off

alxdarksage:
Prelude to refactoring of external ID tables. This is just a flag, which we'll use to ignore all the additional external validation stuff for studies that don't need or use it (which includes all but FPHS and Lilly right now).

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
@@ -13,7 +13,6 @@ import org.sagebionetworks.bridge.models.studies.Study;
 import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
 import org.sagebionetworks.bridge.models.studies.StudyIdentifierImpl;
 
-import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBAttribute;
 import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBHashKey;
 import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBIgnore;
 import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMarshalling;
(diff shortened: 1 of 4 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
```diff
@@ -380,10 +391,12 @@ public String toString() {
                             + "synapseProjectId=%s, technicalEmail=%s, consentNotificationEmail=%s, "
                             + "version=%s, userProfileAttributes=%s, taskIdentifiers=%s, dataGroups=%s, passwordPolicy=%s, "
                             + "verifyEmailTemplate=%s, resetPasswordTemplate=%s, strictUploadValidationEnabled=%s, "
-                            + "healthCodeExportEnabled=%s, emailVerificationEnabled=%s, minSupportedAppVersions=%s]",
-            name, active, sponsorName, identifier, stormpathHref, minAgeOfConsent, maxNumOfParticipants,
-            supportEmail, synapseDataAccessTeamId, synapseProjectId, technicalEmail, consentNotificationEmail, version,
-            profileAttributes, taskIdentifiers, dataGroups, passwordPolicy, verifyEmailTemplate, resetPasswordTemplate,
-            strictUploadValidationEnabled, healthCodeExportEnabled, emailVerificationEnabled, minSupportedAppVersions);
+                            + "healthCodeExportEnabled=%s, emailVerificationEnabled=%s, externalIdValidationEnabled=%s, "
```
liujoshua, line 394: General question not specific to this PR. Do we need toString to produce this exact format, or could we use Guava's MoreObjects.ToStringHelper?

liujoshua:
Looks good, ready for merge

app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
alxdarksage, line 394: We use commons lang ToString builder in the SDK and I like it a lot. We could convert these over, but currently I've just been doing what's consistent with the rest of the project. Or we could switch all to Guava's... we include commons.lang3 because it has some better string validation functions.

PR #1071, opened 2016-03-28
DynamoDB external ID DAO and model objects

alxdarksage:


app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDao.java
```diff
@@ -0,0 +1,296 @@
...
+    private DynamoDBQueryExpression<DynamoExternalIdentifier> createCountQuery(StudyIdentifier studyId,
+            String idFilter, Boolean assignmentFilter) {
+
+        DynamoDBQueryExpression<DynamoExternalIdentifier> query = new DynamoDBQueryExpression<DynamoExternalIdentifier>();
+        if (idFilter != null) {
+            // You cannot filter a query on a hash key, so we copy this value to another column where we can filter
+            query.withQueryFilterEntry(FILTERABLE_EXTERNAL_ID, new Condition()
+                    .withAttributeValueList(new AttributeValue().withS(idFilter))
+                    .withComparisonOperator(CONTAINS));
+        }
```
DwayneJengSage, line 255: I think we can do this with a RangeKeyCondition.

Range keys only support "begins with", not "contains". Is this enough? How useful would this feature be, realistically?

app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDao.java
```diff
@@ -0,0 +1,296 @@
...
+        checkNotNull(studyId);
+        checkArgument(isNotBlank(externalId));
+        checkArgument(isNotBlank(healthCode));
+        
+        DynamoExternalIdentifier keyObject = new DynamoExternalIdentifier(studyId, externalId);
+        
+        DynamoExternalIdentifier identifier = mapper.load(keyObject);
+        if (identifier == null) {
+            throw new EntityNotFoundException(ExternalIdentifier.class);
+        }
```
DwayneJengSage, line 176: Minor: unit test for assigning an ID that doesn't exist?

app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDao.java
```diff
@@ -0,0 +1,296 @@
...
+            // You cannot filter a query on a hash key, so we copy this value to another column where we can filter
+            query.withQueryFilterEntry(FILTERABLE_EXTERNAL_ID, new Condition()
+                    .withAttributeValueList(new AttributeValue().withS(idFilter))
+                    .withComparisonOperator(CONTAINS));
+        }
+        if (assignmentFilter == Boolean.TRUE) {
+            query.withQueryFilterEntry(HEALTH_CODE, new Condition().withComparisonOperator(NOT_NULL));
+        } else if (assignmentFilter == Boolean.FALSE) {
+            query.withQueryFilterEntry(HEALTH_CODE, new Condition().withComparisonOperator(NULL));
+        }
```
DwayneJengSage, line 258: Should the assignmentFilter also account for reserved external IDs? Would that be more useful that assigned external IDs? Would a researcher ever accidentally get a reserved but not assigned ID? Or is this unlikely because of how the API call will include both the reservation and assignment, so this will never come up?

app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDao.java
```diff
@@ -0,0 +1,296 @@
...
+        DynamoDBSaveExpression saveExpression = new DynamoDBSaveExpression();
+        saveExpression.withConditionalOperator(ConditionalOperator.AND);
+        saveExpression.setExpected(map);
+        return saveExpression;
+    }
+    
+    /**
+     * Save the record with the user's healthCode IF the healthCode is not yet set. If calling code calls 
+     * the reservation method first, this should not happen, but we do prevent it.  
+     */
```
DwayneJengSage, line 303: Minor: comment should say we do _not_ prevent it (assigning with reserving)

app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdentifier.java
```diff
@@ -0,0 +1,85 @@
...
+@DynamoDBTable(tableName = "ExternalIdentifier")
+public class DynamoExternalIdentifier implements ExternalIdentifier {
+
+    private String identifier;
+    /**
+     * DynamoDB will not filter a query on the value of the hash key, so we copy this 
+     * value to another column where we can filter on it. This is not exposed outside
+     * of the DAO.
+     */
+    private String filterableIdentifier;
```
DwayneJengSage: See above. We can use a range key condition for beginsWith, if that satisfies the use case.

DwayneJengSage:
Overall looks good. Just a few minor comments. The biggest open question is whether we need a contains query or if we can get away with a range key lookup using beginsWith

app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDao.java
alxdarksage, line 176: Added

app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDao.java
alxdarksage, line 258: You're right, filtering to unassigned codes should not return reserved codes. That could encourage conflicts. Added this along with tests so reserved codes are treated as assigned.

app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDao.java
alxdarksage, line 303: Fixed

app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdentifier.java
alxdarksage: I'll switch to use of range key. If that's not good enough, we can probably change later.

app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDao.java
alxdarksage, line 255: Made the switch to range key. This seems useful enough and we could change it later if it's not.

alxdarksage:
Switched to range key filter.

PR #1076, opened 2016-04-01
Extid service

alxdarksage:
- consistently use "offsetKey" and get rid of "lastKey" synonym
- validate that the external IDs contain alphanumeric characters (only). no spaces, for example
- validate that the new list of IDS doesn't contain duplicates, because AWS throws an interesting error over this (submitting IDs that have already been saved just works, which we want, so users can re-import lists if they succeed only partially)
- switch to using a validator for all this stuff
- return an identifier object that includes the assignment status of the code
- fixed bug where the offsetKey does not match the rangeKey, throwing a DDB validation exception


liujoshua:
Looks good

PR #1077, opened 2016-04-05
Adding new, cleaner API to create and update user information through…

alxdarksage:
… the participant API (requires researcher).

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
@@ -60,6 +60,35 @@ public class ParticipantController extends BaseController {
         return okResult(page);
     }
     
+    public Result createParticipant() throws Exception {
+        UserSession session = getAuthenticatedSession(RESEARCHER);
+        Study study = studyService.getStudy(session.getStudyIdentifier());
+        
+        StudyParticipant participant = parseJson(request(), StudyParticipant.class);
+        
+        participantService.createParticipant(study, participant);
+        
+        return createdResult("Participant created.");
+    }
+    
+    public Result updateParticipant(String email) {
+        UserSession session = getAuthenticatedSession(RESEARCHER);
+        Study study = studyService.getStudy(session.getStudyIdentifier());
+        if (isBlank(email)) {
+            throw new BadRequestException(EMAIL_REQUIRED);
+        }
+        StudyParticipant participant = parseJson(request(), StudyParticipant.class);
+        // Just stop right here because something is wrong
+        if (participant.getEmail() != null && !participant.getEmail().equals(email)) {
+            throw new BadRequestException("Email in JSON does not match URI of request");
+        }
+        participantService.updateParticipant(study, email, participant);
+        
+        return okResult("Participant updated.");
+    }
+    
+    // Already. will remove as soon as replacement is rolled out as I've already implemented this in SDK/Researcher UI
+    @Deprecated
     public Result updateParticipantOptions(String email) {
         UserSession session = getAuthenticatedSession(RESEARCHER);
         Study study = studyService.getStudy(session.getStudyIdentifier());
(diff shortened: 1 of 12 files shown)


app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
```diff
@@ -60,6 +60,35 @@ public Result getParticipants(String offsetByString, String pageSizeString, Stri
...
+    public Result updateParticipant(String email) {
+        UserSession session = getAuthenticatedSession(RESEARCHER);
+        Study study = studyService.getStudy(session.getStudyIdentifier());
+        if (isBlank(email)) {
+            throw new BadRequestException(EMAIL_REQUIRED);
+        }
+        StudyParticipant participant = parseJson(request(), StudyParticipant.class);
+        // Just stop right here because something is wrong
+        if (participant.getEmail() != null && !participant.getEmail().equals(email)) {
+            throw new BadRequestException("Email in JSON does not match URI of request");
```
alxdarksage: "does not match email query parameter"

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -158,6 +160,87 @@ public StudyParticipant getParticipant(Study study, String email) {
...
+        for(String attribute : study.getUserProfileAttributes()) {
+            String value = participant.getAttributes().get(attribute);
+            account.setAttribute(attribute, value);
+        }
+        accountDao.updateAccount(study, account);
+        
+        externalIdService.assignExternalId(study, participant.getExternalId(), healthCode);
+    }
+    
+    public void updateParticipant(Study study, String email, StudyParticipant participant) {
```
liujoshua, line 172: createParticipant and updateParticipant are pretty similar, can we refactor?

Are there reasons for subtle differences such as
- calling assignExternalId in a different order
- updateParticipant checks isExternalIdValidationEnabled, createParticipant does not
- updateParticipant requires email and healthCode

Seems like there's an intended different use case that I'm missing (not just create vs update, but additional rules/validation)

app/org/sagebionetworks/bridge/validators/StudyParticipantValidator.java
```diff
@@ -0,0 +1,87 @@
...
+    @Override
+    public boolean supports(Class<?> clazz) {
+        return StudyParticipant.class.isAssignableFrom(clazz);
+    }
+
+    @Override
+    public void validate(Object object, Errors errors) {
+        StudyParticipant participant = (StudyParticipant)object;
+        
+        if (isNew) {
```
liujoshua, line 35: The isNew flag here seems to answer my question about the differences between createParticipant and updateParticipant are. Could we structure the service like this method? Here, it's a lot clearly what's going on and why

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage, line 172: External IDs have some rules, if we are managing them (if not, it's just a string you can add/update however the client wants):
- you can add them at creation or on update, but once added, they cannot be changed or removed. (Actually for new accounts where this is enabled, I've written it to enforce supplying an external ID... this might be too strict for some studies but seems right for the Lilly case).
- an ID should be assigned to one and only one person. The external ID table is used to make sure we don't write the same ID to two different records in the ParticipantOptions table.

When creating an account, we reserve the ID and then only take it if the account is created successfully. This ensures that even if a client double submits a creation request, it won't be assigned the same ID. Or if Stormpath fails, as it sometimes does, that we'll eventually be able to assign the ID again.

In the update case we can't enforce this, but we can throw an error entirely if the ID is already taken, so we do it first. Then again, the account exists, so it's not as much of an issue. I am about to restrict this further however, so that it only updates if there's no external ID and you're adding one.

Regarding checking the study flag, you'll notice that the externalIdService does this as well. I could do it here just for clarity though.

Finally... healthCode is only given the second time because one of the points of the reservation is that we don't have a healthCode yet. We're just saying "don't assign this ID in the .5-1.5 seconds it takes us to create an account with Stormpath".

liujoshua:
Looks good

PR #1084, opened 2016-04-07
Removing original routes to update user (participant).

alxdarksage:
These are already removed in SDK integration tests.


app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -161,6 +156,14 @@ public StudyParticipant getParticipant(Study study, String email) {
...
         return accountDao.getPagedAccountSummaries(study, offsetBy, pageSize, emailFilter);
     }
+    
+    public void signUserOut(Study study, String email) {
+        checkNotNull(study);
+        checkArgument(isNotBlank(email));
+        
+        Account account = getAccountThrowingException(study, email);
+        cacheProvider.removeSessionByUserId(account.getId());
+    }
```
alxdarksage, line 166: This was just moved above private methods.

PR #1091, opened 2016-04-13
Upload DAO should catch DDB ConditionalCheckException and throw Bridge ConcurrentModificationException

DwayneJengSage:
See https://sagebionetworks.jira.com/browse/BRIDGE-1251

This adds a catch in the DAO to catch the DDB exception and rethrow a Bridge exception, so the request errors out with a 409 instead of a 500.

Testing done:
- Added a 10 second sleep to test repro'ing the 500 error. After the fix, this returns a 409 instead.
- Added unit tests.


DwayneJengSage:
Reviewed by Joshua.

PR #1100, opened 2016-04-22
Updating the unsubscribe webhook that we provide to MailChimp

alxdarksage:
Updating the unsubscribe webhook that we provide to MailChimp for mailings. The hook seems sound but had to be adjusted because we look up users by IDs now, and that we don't get from MailChimp.

I believe the errors were originally because users without health codes were submitting unsubscribes.. ? This should not happen now with the way this has been restructured. We create health codes if they don't exist just so we can record this setting.

Tested this using curl and a POST body as described by MailChimp.

To test further I will deploy to develop, and then create a test list in MailChimp pointed at development, and send myself a mass mailing, unsubscribe, etc. This is the only way I can see to test what MailChimp sends beyond their docs.

### DIFF
--- a/app/org/sagebionetworks/bridge/dao/AccountDao.java
+++ b/app/org/sagebionetworks/bridge/dao/AccountDao.java
@@ -122,4 +122,12 @@ public interface AccountDao {
      *      about the request and the total number of records.
      */
     public PagedResourceList<AccountSummary> getPagedAccountSummaries(Study study, int offsetBy, int pageSize, String emailFilter);
+    
+    /**
+     * For MailChimp, and other external systems, we need a way to get a healthCode for a given email.
+     * @param study
+     * @param email
+     * @return
+     */
+    public String getHealthCodeForEmail(Study study, String email);
 }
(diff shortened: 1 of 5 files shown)


app/org/sagebionetworks/bridge/dao/AccountDao.java
```diff
@@ -122,4 +122,12 @@
      *      about the request and the total number of records.
      */
     public PagedResourceList<AccountSummary> getPagedAccountSummaries(Study study, int offsetBy, int pageSize, String emailFilter);
+    
+    /**
+     * For MailChimp, and other external systems, we need a way to get a healthCode for a given email.
+     * @param study
+     * @param email
+     * @return
+     */
```
DwayneJengSage, line 110: Minor: We don't need the @param and @return in the comments if we aren't actually going to fill them in.

DwayneJengSage:
Overall looks good with a minor comments issue.

That said, it looks like this code change doesn't address the issue that happened last December (https://sagebionetworks.jira.com/browse/BRIDGE-1059). Reading from the stack trace, it looks like the error came from the part of the code that has no functional changes, and not from the health code stuff as theorized in the description.

PR #1108, opened 2016-04-25
Replace signup

alxdarksage:
The first of two PRs to use StudyParticipant and remove SignUp. As a result, when onboarding, clients will be able to save any information about a user when signing up (not just data groups, as supported today). This is easy to replace because SignUp is never returned in the API and we pluck the information we need out of the model object to create an account.
- Remove SignUp and substitute StudyParticipant
- Consolidate SignUpValidator and StudyParticipantValidate (latter is a superset of the former already)
- Improved immutability of StudyParticipant so it can be compared to other StudyParticipant objects using equals()


app/org/sagebionetworks/bridge/models/accounts/StudyParticipant.java
```diff
@@ -112,6 +111,32 @@ public String getId() {
...
     
+    @Override
+    public int hashCode() {
+        return Objects.hash(attributes, consentHistories, createdOn, dataGroups, email, 
+                externalId, firstName, healthCode, id, languages, lastName, notifyByEmail, 
+                password, roles, sharingScope, status);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
DwayneJengSage, line 121: Add equalsVerifier test?

app/org/sagebionetworks/bridge/models/accounts/StudyParticipant.java
```diff
@@ -122,14 +147,33 @@ public String getId() {
         private boolean notifyByEmail;
         private Set<String> dataGroups = ImmutableSet.of();
         private String healthCode;
-        private Map<String,String> attributes = Maps.newHashMap();
-        private Map<String,List<UserConsentHistory>> consentHistories = Maps.newHashMap();
-        private Set<Roles> roles = Sets.newHashSet();
+        private Map<String,String> attributes = ImmutableMap.of();
+        private Map<String,List<UserConsentHistory>> consentHistories = ImmutableMap.of();
+        private Set<Roles> roles = ImmutableSet.of();
```
DwayneJengSage: If you want to make these immutable, you should also have the setters or build() make immutable copies.

test/org/sagebionetworks/bridge/TestUserAdminHelper.java
```diff
@@ -46,17 +46,25 @@
...
-            this.roles = Sets.newHashSet(signUp.getRoles());
+        public TestUser(String email, String password, Set<Roles> roles, Set<String> dataGroups, Study study,
+                UserSession session) {
+            this.email = email;
+            this.password = password;
+            this.roles = Sets.newHashSet(TEST_USERS);
+            if (roles != null) {
+                this.roles.addAll(roles);    
+            }
             this.roles.add(TEST_USERS);
```
DwayneJengSage: Minor: You create a set with TEST_USERS, add a bunch of roles, and then add TEST_USER.

test/org/sagebionetworks/bridge/stormpath/StormpathAccountDaoTest.java
```diff
@@ -135,13 +136,6 @@ public void getStudyPagedAccounts() {
             accounts = accountDao.getPagedAccountSummaries(study, 0, 20, "bridgeit@");
             assertEquals(1, accounts.getItems().size());
             assertEquals("bridgeit@sagebase.org", accounts.getItems().get(0).getEmail());
-            
-            accounts = accountDao.getPagedAccountSummaries(study, 0, 20, "bridge-testing+SADT");
-            assertTrue(accounts.getItems().size() > 0);
-            for (AccountSummary summary : accounts.getItems()) {
-                assertNull(summary.getFirstName());
-                assertNull(summary.getLastName());
-            }
```
DwayneJengSage, line 144: Is this deleted test code still needed?

alxdarksage, line 144: Yes. This almost seems like an error I made when merging, this shouldn't be removed. Will restore.

app/org/sagebionetworks/bridge/models/accounts/StudyParticipant.java
alxdarksage, line 121: Added

app/org/sagebionetworks/bridge/models/accounts/StudyParticipant.java
alxdarksage: Shuffled this around so collections are copied into immutable collections before constructor is called.

test/org/sagebionetworks/bridge/TestUserAdminHelper.java
alxdarksage: Cleaned up

test/org/sagebionetworks/bridge/stormpath/StormpathAccountDaoTest.java
alxdarksage, line 144: Added this back.

PR #1109, opened 2016-04-25
Remove old APIs based on using email to get/update participants.

alxdarksage:
I duplicated tests when creating the side-by-side ID-based methods, so removing the email-based tests now. Removed routes and controller methods for email-based calls. Re-organized StormpathDao.


app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
```diff
@@ -278,7 +278,27 @@ public Account getAccount(Study study, String identifier) {
         checkNotNull(study);
```
alxdarksage, line 278: In this class, moving methods around to cleanup, not deleting or adding.

PR #1115, opened 2016-04-29
Scheduling updates

alxdarksage:
- Fixing scheduling for studies where users do not have to consent to use the server. When no enrollment event exists, we use the timestamp of the creation of the user's stormpath account (which should always exist).
- Removing references to survey responses in the scheduling system to simplify the code.

We are no longer looking for consent records if a really old account doesn't have an enrollment event. The createdOn timestamp of the user's account is close enough for the purposes of scheduling and far simpler to retrieve.

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/ScheduledActivityController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/ScheduledActivityController.java
@@ -6,13 +6,18 @@ import java.util.List;
 import org.apache.commons.lang3.StringUtils;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
+
+import org.sagebionetworks.bridge.dao.AccountDao;
 import org.sagebionetworks.bridge.exceptions.BadRequestException;
 import org.sagebionetworks.bridge.json.DateUtils;
 import org.sagebionetworks.bridge.models.ClientInfo;
 import org.sagebionetworks.bridge.models.ResourceList;
+import org.sagebionetworks.bridge.models.accounts.Account;
+import org.sagebionetworks.bridge.models.accounts.User;
 import org.sagebionetworks.bridge.models.accounts.UserSession;
 import org.sagebionetworks.bridge.models.schedules.ScheduleContext;
 import org.sagebionetworks.bridge.models.schedules.ScheduledActivity;
+import org.sagebionetworks.bridge.models.studies.Study;
 import org.sagebionetworks.bridge.services.ScheduledActivityService;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Controller;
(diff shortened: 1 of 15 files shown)


app/org/sagebionetworks/bridge/play/controllers/ScheduledActivityController.java
```diff
@@ -94,13 +111,31 @@ public Result updateScheduledActivities() throws Exception {
...
         ClientInfo clientInfo = getClientInfoFromUserAgentHeader();
         
+        DateTime accountCreatedOn = session.getUser().getAccountCreatedOn();
+        if (accountCreatedOn == null) {
+            Study study = studyService.getStudy(session.getStudyIdentifier());
+            User user = session.getUser();
+            // Everyone should have an ID at this point... otherwise sessions are hanging out for over a week.
+            if (user.getId() == null) {
+                accountCreatedOn = DateTime.now();
+                LOG.debug("neither accountCreatedOn nor ID exist in session, using current time");
```
DwayneJengSage, line 120: Minor: Sessions can live on forever (see https://sagebionetworks.jira.com/browse/BRIDGE-1024), so not everyone may have the user ID in their session. Worst case scenario, we'll need to flush everyone's sessions.

test/org/sagebionetworks/bridge/play/controllers/ScheduledActivityControllerTest.java
```diff
@@ -186,4 +229,24 @@ public void mustBeAuthenticated() throws Exception {
...
+    @Test
+    public void oldSessionsWithNoIdAndNoAccountCreatedOn() throws Exception {
+        session.getUser().setAccountCreatedOn(null); // these are not currently in the session
+        session.getUser().setId(null);
+        
+        controller.getScheduledActivities(null, "-07:00", "3");
+        verify(scheduledActivityService).getScheduledActivities(any(), contextCaptor.capture());
+        ScheduleContext context = contextCaptor.getValue();
+        assertNotNull(context.getAccountCreatedOn()); // this is a timestamp, so
+    }
```
DwayneJengSage, line 259: Should we also have a test case where the accountCreatedOn is in the session, and we verify it's being passed correctly to the scheduledActivityService?

app/org/sagebionetworks/bridge/validators/ScheduleContextValidator.java
```diff
@@ -30,6 +30,9 @@ public void validate(Object object, Errors errors) {
         if (context.getCriteriaContext().getHealthCode() == null) {
             errors.rejectValue("healthCode", "is required");
         }
+        if (context.getAccountCreatedOn() == null) {
+            errors.rejectValue("accountCreatedOn", "is required");
+        }
```
DwayneJengSage, line 35: Test case in ScheduleContextValidatorTest?

DwayneJengSage:
Overall looks good. A few comments, which should be quick to address.

test/org/sagebionetworks/bridge/play/controllers/ScheduledActivityControllerTest.java
alxdarksage, line 259: Added.

app/org/sagebionetworks/bridge/validators/ScheduleContextValidator.java
alxdarksage, line 35: Added

PR #1121, opened 2016-05-03
Return session after updates

alxdarksage:
Now that we're recalculating sessions after user updates, return this session to the user. For the admin create method, the ID is useful to delete the user later; for the participant update method, it is useful so the client can see whether or not the account is still consented. We don't return a session for the general participant update API so if you edit yourself through that API, it may be out of sync. However we only expect researchers to use that API through the web app.


alxdarksage:
reviewed by joshua

PR #1122, opened 2016-05-04
Always create health code

alxdarksage:
This started with the desire to make accounts that are created in the Stormpath UI fully editable in the Bridge Study Manager. It should also make it possible for developers we set up to use their accounts in the apps while testing, which they sometimes do. 

But then I also wanted to eliminate the code in several places to check and create the healthId/healthCode mapping if it doesn't exist. This was all highly redundant and once you ensure an account always has a health code in AccountDao, there's no reason not to return the healthCode itself as part of the account. Then it's not necessary to use healthCodeService in most places.

The iterator methods now return AccountSummary which is much faster to retrieve than an account, and then like the paged methods, you can retrieve an individual, fully-initialized account object.

In short: if you have a reference to an account it should always have a health code.

### DIFF
--- a/app/org/sagebionetworks/bridge/models/accounts/AccountSummary.java
+++ b/app/org/sagebionetworks/bridge/models/accounts/AccountSummary.java
@@ -1,32 +1,51 @@
 package org.sagebionetworks.bridge.models.accounts;
 
+import static org.sagebionetworks.bridge.BridgeConstants.STORMPATH_NAME_PLACEHOLDER_STRING;
+
 import java.util.Objects;
 
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
+
+import org.sagebionetworks.bridge.BridgeUtils;
+import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 public final class AccountSummary {
 
+    public static final AccountSummary create(StudyIdentifier studyId, com.stormpath.sdk.account.Account acct) {
+        java.util.Date javaDate = acct.getCreatedAt();
+        DateTime createdOn = (javaDate != null) ? new DateTime(javaDate) : null;
+        String id = BridgeUtils.getIdFromStormpathHref(acct.getHref());
+        
+        String firstName = (STORMPATH_NAME_PLACEHOLDER_STRING.equals(acct.getGivenName())) ? null : acct.getGivenName();
+        String lastName = (STORMPATH_NAME_PLACEHOLDER_STRING.equals(acct.getSurname())) ? null : acct.getSurname();
+        return new AccountSummary(firstName, lastName, acct.getEmail(), id, createdOn,
+                AccountStatus.valueOf(acct.getStatus().name()), studyId);
+    }
+    
     private final String firstName;
     private final String lastName;
     private final String email;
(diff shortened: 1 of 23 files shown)


app/org/sagebionetworks/bridge/models/accounts/AccountSummary.java
```diff
@@ -1,32 +1,51 @@
...
-            @JsonProperty("createdOn") DateTime createdOn, @JsonProperty("status") AccountStatus status) {
+            @JsonProperty("createdOn") DateTime createdOn, @JsonProperty("status") AccountStatus status,
+            @JsonProperty("studyIdentifier") StudyIdentifier studyIdentifier) {
         this.firstName = firstName;
         this.lastName = lastName;
         this.email = email;
         this.id = id;
         this.createdOn = (createdOn == null) ? null : createdOn.withZone(DateTimeZone.UTC);
         this.status = status;
+        this.studyIdentifier = studyIdentifier;
```
alxdarksage, line 48: Add to SDK.

app/org/sagebionetworks/bridge/services/backfill/ConsentAuditAndRepairBackfill.java
```diff
@@ -163,10 +168,13 @@ private long getSignedOnDate(StringBuilder sb, DateTime enrollment, long synthet
...
+     */
     private String getHealthCode(BackfillTask task, BackfillCallback callback, Account account) {
-        HealthId mapping = healthCodeService.getMapping(account.getHealthId());
-        if (mapping != null) {
-            return mapping.getCode();
+        if (account.getHealthCode() != null) {
+            return account.getHealthCode();
         }
         callback.newRecords(getBackfillRecordFactory().createOnly(task, "No health code found: " + account.getId()));
         return null;
```
alxdarksage: Check again: this can be removed.

app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
```diff
@@ -392,15 +363,33 @@ public void deleteAccount(Study study, String id) {
...
+    private Account constructAccount(StudyIdentifier studyId, com.stormpath.sdk.account.Account acct) {
+        List<SubpopulationGuid> subpopGuids = getSubpopulationGuids(studyId);
+        StormpathAccount account = new StormpathAccount(studyId, subpopGuids, acct, encryptors);
+        HealthId healthId = null;
+        if (account.getHealthCode() == null) {
+            healthId = healthCodeService.getMapping(account.getHealthId());
+        }
+        if (healthId == null) {
+            healthId = healthCodeService.createMapping(studyId);
+            account.setHealthId(healthId);
```
alxdarksage: redundant

test/org/sagebionetworks/bridge/stormpath/StormpathAccountDaoMockTest.java
```diff
@@ -318,28 +286,46 @@ public void updatingAccountWithRemovedGroups() {
...
+        when(mockAccount.getRoles()).thenReturn(newGroupSet);
+
+        // execute
+        dao.updateAccount(mockAccount);
+
+        // return this, so the caller can verify back-end mock calls
+        return mockSpAccount;
     }
 
 }
```
alxdarksage: Test that the user does NOT have the health ID in the Stormpath account. The DAO should call the healthCodeService and set this up.

PR #1124, opened 2016-05-05
Show 400 errors back to user on password reset screen. 

alxdarksage:
Add password validation to the reset password pathway. Factored out validation code (based on PasswordPolicy) to a utility method so we can also validate in ParticipantStudyValidator.

Communicate 400 errors back to the reset password UI. Otherwise users have no idea if there are password constraints they are not meeting, they just get a generic error message back.

### DIFF
--- a/app/views/resetPassword.scala.html
+++ b/app/views/resetPassword.scala.html
@@ -44,7 +44,6 @@ location.search.substr(1).split("&").forEach(function(item) {
 $("*[name='password']").focus();
 
 function success() {
-	console.log(arguments);
     $("#resetPasswordSection").html("<p class='success'>Your password has been successfully updated.</p>");
 }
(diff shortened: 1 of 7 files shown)


app/views/resetPassword.scala.html
```diff
@@ -44,15 +44,21 @@
 $("*[name='password']").focus();
 
 function success() {
+	console.log(arguments);
```
DwayneJengSage: Will this log passwords to the logs? If so, we should definitely remove this.

alxdarksage: No, but should be removed

DwayneJengSage:
Other than the one comment about logs, this looks good.

alxdarksage:
r=dwayne

PR #1129, opened 2016-05-06
Signup refactor

alxdarksage:
- Use the StudyParticipant object being submitted in signUp and when an admin creates a user, to call the ParticipantService. SignUp API now lets you save everything (language, dataGroups, externalId, etc.), and this should be easier to maintain going forward.
- removed the distributed lock when signing up. I don't believe there's any crisis condition in trying to create an account with the same credentials concurrently. Only one will be registered with Stormpath and that account will send an email verification, it will have the active health code etc.
- as in the SDK, moved immutable collection construction in the StudyParticipant into the constructor;


app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -184,12 +185,10 @@ public IdentifierHolder saveParticipant(Study study, Set<Roles> callerRoles, Str
             if (isNotBlank(participant.getExternalId())) {
                 externalIdService.reserveExternalId(study, participant.getExternalId());    
             }
-            account = accountDao.signUp(study, participant, study.isEmailVerificationEnabled());
+            account = accountDao.constructAccount(study, participant.getEmail(), participant.getPassword());
```
alxdarksage, line 188: Possible to remove this and do new Account() or Account.create() or something instead?

PR #1132, opened 2016-05-07
add granular uploadedOn timestamp to health data record

DwayneJengSage:
Add a new timestamp to health data records to support both hourly export and on-demand export.

We use a compound key with the studyId as the hash and uploadedOn as the range key. This is because (1) we need to do range queries on the uploadedOn for it to be useful and (2) this feature is configured on a per-study basis anyway.

Testing done:
- added/updated unit tests
- deleted tables, then verified table is created with the new index
- manually tested on local server, verified uploadedOn
- UploadTest in integration tests

### DIFF
--- a/app/org/sagebionetworks/bridge/models/healthdata/HealthDataRecord.java
+++ b/app/org/sagebionetworks/bridge/models/healthdata/HealthDataRecord.java
@@ -20,7 +20,7 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 @BridgeTypeName("HealthData")
 @JsonDeserialize(as = DynamoHealthDataRecord.class)
 public interface HealthDataRecord extends BridgeEntity {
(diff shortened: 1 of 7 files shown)


app/org/sagebionetworks/bridge/models/healthdata/HealthDataRecord.java
```diff
@@ -20,7 +20,7 @@
 @BridgeTypeName("HealthData")
 @JsonDeserialize(as = DynamoHealthDataRecord.class)
 public interface HealthDataRecord extends BridgeEntity {
-    public static final ObjectWriter PUBLIC_RECORD_WRITER = new BridgeObjectMapper().writer(
+    ObjectWriter PUBLIC_RECORD_WRITER = new BridgeObjectMapper().writer(
```
alxdarksage, line 23: Why remove static final?

alxdarksage:
Looks good to me. One comment.

PR #1133, opened 2016-05-10
added usesCustomExportSchedule (hourly export schedule) to study

DwayneJengSage:
Added a usesCustomExportSchedule flag to Study. By default, all studies are exported using the default nightly schedule. Some studies may need custom schedules for hourly or on-demand exports. To prevent this study from being exported twice (once by the custom schedule, once by the default schedule), we need a flag that tells the exporter not to export this this study as part of the default schedule.

Testing done:
- added unit tests
- manually tested flipping this flag back and forth


alxdarksage:
Looks fine to merge but right now anyone can change this via the API, should this be an admin-specific setting since it requires coordination with the exporter?

DwayneJengSage:
This is no different from the Synapse project and data access team flags, which can also currently be set by developers. That said, one could argue that all of those settings should only be set by admins.

PR #1134, opened 2016-05-11
Delete user

alxdarksage:
Delete the user object, which was used to initialize sessions and then was reference a lot from the session. Instead sessions are instantiated using a participant from the participant API

UserSessionInfo is essentially a view onto the UserSession object for the API, may still be refactored to a Jackson serializer.

There is code to deserialize the old session JSON from Redis into a new UserSession object.


test/org/sagebionetworks/bridge/cache/CacheProviderTest.java
```diff
@@ -219,7 +211,82 @@ public void addAndRemoveViewFromCacheProvider() throws Exception {
...
+                "'sessionToken':'ccea2978-f5b9-4377-8194-f887a3e2a19b','internalSessionToken':"+
+                "'4f0937a5-6ebf-451b-84bc-fbf649b9e93c','user':{'id':'6gq4jGXLmAxVbLLmVifKN4',"+
+                "'firstName':'Bridge','lastName':'IT','email':'bridgeit@sagebase.org','studyKey':'api',"+
+                "'sharingScope':'no_sharing','accountCreatedOn':'2016-04-21T16:48:22.386Z',"+
+                "'roles':['admin'],'dataGroups':['group1'],'consentStatuses':{'api':{'name':"+
+                "'Default Consent Group','subpopulationGuid':'api','required':true,'consented':false,"+
+                "'signedMostRecentConsent':true,'type':'ConsentStatus'}},'languages':['en','fr'],"+
+                "'encryptedHealthCode':'TFMkaVFKPD48WissX0bgcD3esBMEshxb3MVgKxHnkXLSEPN4FQMKc01tDbBA"+
+                "VcXx94kMX6ckXVYUZ8wx4iICl08uE+oQr9gorE1hlgAyLAM=','type':'User'},'studyIdentifier':"+
+                "{'identifier':'api','type':'StudyIdentifier'},'type':'UserSession'}");
```
DwayneJengSage: Without indentation, this is very difficult to read. Can we use a style like https://github.com/Sage-Bionetworks/BridgePF/blob/develop/test/org/sagebionetworks/bridge/models/upload/UploadSchemaTest.java#L120 ?

app/org/sagebionetworks/bridge/models/accounts/UserSessionInfo.java
```diff
@@ -1,93 +1,91 @@
...
-        return dataGroups;
-    }
-    public String getId() {
-        return id;
+    public static JsonNode toJSON(UserSession session) {
+        ObjectNode node = JsonNodeFactory.instance.objectNode();
+        node.put(AUTHENTICATED, session.isAuthenticated());
+        node.put(SESSION_TOKEN, session.getSessionToken());
+        node.put(ENVIRONMENT, ENVIRONMENTS.get(session.getEnvironment()));
+        node.put(DATA_SHARING, session.getParticipant().getSharingScope() != SharingScope.NO_SHARING);
```
DwayneJengSage: Do we want just the isDataSharing boolean? Or do we also want the sharing scope as well?

app/org/sagebionetworks/bridge/models/accounts/UserSessionInfo.java
```diff
@@ -1,93 +1,91 @@
...
+        node.put(USERNAME, session.getParticipant().getEmail());
```
DwayneJengSage: Do apps actually check for "username"? Wouldn't "email" be more accurate?

test/org/sagebionetworks/bridge/cache/CacheProviderTest.java
alxdarksage: Adding

app/org/sagebionetworks/bridge/models/accounts/UserSessionInfo.java
```diff
@@ -1,93 +1,91 @@
...
+                // do nothing
+            } else if (child.isTextual()) {
+                node.put(fieldName, child.asText());    
+            } else if (child.isBoolean()) {
+                node.put(fieldName, child.asBoolean());
+            } else if (child.isArray()) {
+                node.putArray(fieldName).addAll((ArrayNode)child);
+            } else if (child.isObject()) {
+                node.putPOJO(fieldName, child);
+            }
```
DwayneJengSage, line 77: ObjectNode.set() should simplify this logic and make it future-proof.

alxdarksage, line 77: much better, changed this

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -91,50 +92,71 @@ final void setExternalIdService(ExternalIdService externalIdService) {
...
+        if (callerRoles.contains(RESEARCHER)){
+            Map<String,List<UserConsentHistory>> consentHistories = Maps.newHashMap();
+            List<Subpopulation> subpopulations = subpopService.getSubpopulations(study.getStudyIdentifier());
+            for (Subpopulation subpop : subpopulations) {
+                // always returns a list, even if empty
+                List<UserConsentHistory> history = consentService.getUserConsentHistory(study, subpop.getGuid(),
+                        account.getHealthCode(), account.getId());
+                consentHistories.put(subpop.getGuidString(), history);
+            }
+            builder.withConsentHistories(consentHistories);
```
DwayneJengSage, line 154: Previously, if a non-researcher called getParticipant to get their own participant info, they would get their own consent history. Now, this is locked to researchers only. Is this an intentional change?

Incidentally, there are no tests in ParticipantService test for a non-researcher, non-developer, non-admin user getting their own participant info.

app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
```diff
@@ -411,7 +411,7 @@ private Account getAccountWithEmail(Study study, String email) {
     }
     
     private void rethrowResourceException(ResourceException e, Account account) {
-        logger.info(String.format("Stormpath error: %s: %s", e.getCode(), e.getMessage()));
+        //logger.info(String.format("Stormpath error: %s: %s", e.getCode(), e.getMessage()));
```
DwayneJengSage: Was this intentionally commented out? Should we delete this log message?

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -91,50 +92,71 @@ final void setExternalIdService(ExternalIdService externalIdService) {
...
+     * consent and healthCode information. 
+     */
+    public StudyParticipant getParticipantForSession(Study study, Account account) {
+        checkNotNull(study);
+        checkNotNull(account);
+        
+        StudyParticipant.Builder builder = getParticipantBuilder(study, NO_CALLER_ROLES, account);
+        builder.withHealthCode(account.getHealthCode());
+        return builder.build();
+    }
```
DwayneJengSage, line 110: Is it worth adding a unit test just to make sure this code path is being exercised? It should probably have parity with UserTest.initializedFromAccount(), which it replaces.

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerMockTest.java
```diff
@@ -208,11 +207,10 @@ public void signUpWithCompleteUserData() throws Exception {
         
         verify(authenticationService).signUp(eq(study), participantCaptor.capture());
         
-        StudyParticipant originalParticipant = TestConstants.PARTICIPANT;
+        StudyParticipant originalParticipant = TestUtils.getStudyParticipant(AuthenticationControllerMockTest.class);
         StudyParticipant persistedParticipant = participantCaptor.getValue();
         assertEquals(originalParticipant.getFirstName(), persistedParticipant.getFirstName());
         assertEquals(originalParticipant.getLastName(), persistedParticipant.getLastName());
-        assertEquals(originalParticipant.getEmail(), persistedParticipant.getEmail());
```
DwayneJengSage, line 215: Do we still need to assert on email?

test/org/sagebionetworks/bridge/play/controllers/UserDataDownloadControllerTest.java
```diff
@@ -54,7 +51,7 @@ public void test() throws Exception {
 
         // validate args sent to mock service
         ArgumentCaptor<DateRange> dateRangeCaptor = ArgumentCaptor.forClass(DateRange.class);
-        verify(mockService).requestUserData(eq(studyIdentifier), same(user), dateRangeCaptor.capture());
+        verify(mockService).requestUserData(eq(studyIdentifier), any(String.class), dateRangeCaptor.capture());
```
DwayneJengSage: Since we changed from passing in the user to passing in the email from the participant, we should update the test to use Participant and verify the email from Participant is being passed into requestUserData().

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage, line 154: This method is called from a controller that only allows researchers access. But if a developer were to call this, no, they shouldn't get the consent history as I understand it right now (because it contains birthdate and signature).

Added test that non-researcher does not get healthCode or history, and researcher gets healthCode and history.

app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
alxdarksage: Added back (seems useful to have in the logs, didn't mean to leave it commented out)

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage, line 110: Added a test that verifies healthCode is included in the call, along with some other fields you'd expect from full initialization of the StudyParticipant.

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerMockTest.java
alxdarksage, line 215: Changed test so this assertion could be added back in (just need to not call getParticipant() twice as it creates a dummy participant with a random email.

test/org/sagebionetworks/bridge/play/controllers/UserDataDownloadControllerTest.java
alxdarksage: Added participant with email to mockSession and verify that's what is used in call.

PR #1139, opened 2016-05-17
A test and a comment to verify BRIDGE-1350 is fixed.

alxdarksage:
Changes to address recent failures in production:
- ParticipantService changed so participant is always returned with health code. On error was calling getParticipant() and getting a participant record with a health code that was being saved in the session. Instead for the appropriate method, we remove healthCode if the study has disabled healthCode export.
- null pointer precondition check for healthCode before setting session in cache;
- when initializing session, if a session exists, keep the session tokens but update the session. Another issue was a session being set from the verifyEmail call which comes from a web page, not the client application, and lacks the User-Agent header to properly construct this session. consenting to research call fixes the session now.
- always update session cookie when setting the session in the cache. Not known to cause a problem but possibly only because no one looks at the cookie for this.
- fixed code path when creating accounts that may have been creating a healthCode and then throwing it away rather than saving it. Probably did not cause an issue since on sign in/consent it would be fixed in the course of constructing an account. But still.
- improved cache provider test based on examination of real session in production
- refactored tests as appropriate (e.g. moved the check that health code was not being included from the service to the controller).


DwayneJengSage:
Does this change re-apply https://github.com/Sage-Bionetworks/BridgePF/pull/1134 ?

Would it be cleaner to re-apply that pull request in a separate pull request, and then make a separate pull request with the fix on top of that? This would make reviewing the fix much much much easier.

DwayneJengSage:
While the new code was deployed to prod yesterday, we were seeing errors that suggest either the participant was missing from the session or the health code was missing from the participant. Was this bug fixed?

This is critical to fix, as it prevents users from uploading.

DwayneJengSage:
Filing https://sagebionetworks.jira.com/browse/BRIDGE-1353 as a blocker on this pull request

test/org/sagebionetworks/bridge/stormpath/StormpathAccountDaoMultipleSubpopulationsTest.java
```diff
@@ -1,109 +0,0 @@
...
-import org.sagebionetworks.bridge.models.accounts.StudyParticipant;
-import org.sagebionetworks.bridge.models.studies.Study;
-import org.sagebionetworks.bridge.models.subpopulations.ConsentSignature;
-import org.sagebionetworks.bridge.models.subpopulations.Subpopulation;
-import org.sagebionetworks.bridge.services.StudyService;
-import org.sagebionetworks.bridge.services.SubpopulationService;
-
-@ContextConfiguration("classpath:test-context.xml")
-@RunWith(SpringJUnit4ClassRunner.class)
-public class StormpathAccountDaoMultipleSubpopulationsTest {
```
alxdarksage, line 29: Test here was moved to a different class, still exists (see below).

DwayneJengSage, line 29: Actually, this specific change reverts the fix for ConsentServiceTest. I know for a fact that the version of this test in StormpathAccountDaoTest will break ConsentServiceTest, which is why it was moved to its own file in the first place.

test/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDaoTest.java
```diff
@@ -4,15 +4,15 @@
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
```
DwayneJengSage: This change undoes the ConsentServiceTest fix. Revert the entire file.

test/org/sagebionetworks/bridge/dynamodb/DynamoParticipantOptionsDaoMockTest.java
```diff
@@ -1,28 +0,0 @@
...
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.verify;
-
-import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper;
-import com.google.common.collect.ImmutableMap;
-import org.junit.Test;
-
-import org.sagebionetworks.bridge.TestConstants;
-
-public class DynamoParticipantOptionsDaoMockTest {
```
DwayneJengSage, line 14: This undoes the ConsentServiceTest fix. Restore the entire file.

test/org/sagebionetworks/bridge/dynamodb/DynamoParticipantOptionsDaoTest.java
```diff
@@ -178,6 +183,28 @@ public void updateAllOptions() {
         assertEquals(DATA_GROUPS_SET, lookup.getStringSet(DATA_GROUPS));
         assertEquals(LANGUAGES_ORDERED_SET, lookup.getOrderedStringSet(LANGUAGES));
     }
+    
+    @Test
+    public void updateNoOptions() {
```
DwayneJengSage: This change undoes the ConsentServiceTest fix. Revert the entire file.

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
```diff
@@ -159,17 +159,30 @@ public void oddParametersUseDefaults() throws Exception {
 
     @Test
     public void getParticipant() throws Exception {
-        StudyParticipant studyParticipant = new StudyParticipant.Builder().withFirstName("Test").build();
-        
-        when(participantService.getParticipant(STUDY, CALLER_ROLES, ID)).thenReturn(studyParticipant);
+        STUDY.setHealthCodeExportEnabled(true);
```
DwayneJengSage: Modifying a static member will persist that change across tests. This is bad. If you need to modify the study, create a new study for each test instead of modifying the old one.

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
```diff
@@ -159,17 +159,30 @@ public void oddParametersUseDefaults() throws Exception {
...
+        String json = Helpers.contentAsString(result);
+        StudyParticipant retrievedParticipant = BridgeObjectMapper.get().readValue(json, StudyParticipant.class);
+        
         assertEquals("Test", retrievedParticipant.getFirstName());
+        assertEquals("healthCode", retrievedParticipant.getHealthCode());
+    }
+    
+    @Test
+    public void getParticipantWithNoHealthCode() throws Exception {
+        STUDY.setHealthCodeExportEnabled(false);
```
DwayneJengSage: Same here.

test/org/sagebionetworks/bridge/services/AuthenticationServiceTest.java
```diff
@@ -319,12 +320,12 @@ public void invalidDataGroupsAreRejected() throws Exception {
...
-        AuthenticationService authServiceSpy = spy(authService);
+        authService = spy(authService);
         
         testUser = helper.getBuilder(AuthenticationServiceTest.class)
                 .withConsent(false).withSignIn(false).build();
-        authServiceSpy.signUp(testUser.getStudy(), testUser.getStudyParticipant());
-        verify(authServiceSpy).requestResetPassword(any(Study.class), any(Email.class));
+        authService.signUp(testUser.getStudy(), testUser.getStudyParticipant());
+        verify(authService).requestResetPassword(any(Study.class), any(Email.class));
     }
```
DwayneJengSage, line 324: This modifies member variable authService. Past experience tells me that JUnit and TestNG don't reset member variables between tests. This could cause corruption across multiple tests. Revert this test to use a local variable.

test/org/sagebionetworks/bridge/services/HealthCodeServiceMockTest.java
```diff
@@ -1,45 +0,0 @@
...
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verifyZeroInteractions;
-import static org.mockito.Mockito.when;
-
-import org.junit.Test;
-
-import org.sagebionetworks.bridge.dao.HealthIdDao;
-import org.sagebionetworks.bridge.models.accounts.HealthId;
-
-public class HealthCodeServiceMockTest {
```
DwayneJengSage, line 14: This reverts the ConsentServiceTest fix. Restore the entire file.

test/org/sagebionetworks/bridge/services/HealthCodeServiceTest.java
```diff
@@ -24,6 +34,16 @@
...
 
+    @Before
+    public void before() {
+        clearDynamo();
+    }
+
+    @After
+    public void after() {
+        clearDynamo();
+    }
```
DwayneJengSage: There is no need to clear dynamo before and after each test run. The test runs fine with other entries in the table.

Clearing Dynamo means our local test accounts lose their healthcodes, which makes local testing a pain.

test/org/sagebionetworks/bridge/services/HealthCodeServiceTest.java
```diff
@@ -34,4 +54,41 @@ public void createAndRetrieveHealthId() {
...
+        // valid
+        HealthId id1 = healthCodeService.getMapping("123");
+        assertEquals("123", id1.getId());
+        assertEquals("abc", id1.getCode());
+        
+        // invalid... should return no object at all.
+        HealthId id2 = healthCodeService.getMapping("456");
+        assertNull(id2);
+        
+    }
```
DwayneJengSage: This reverts the ConsentServiceTest fix. Revert the entire file.

test/org/sagebionetworks/bridge/services/StudyServiceTest.java
```diff
@@ -56,18 +55,15 @@
...
     
     private Study study;
     
     @Before
     public void before() {
-        mockCache = mock(CacheProvider.class);
-        studyService.setCacheProvider(mockCache);
+        cache = mock(CacheProvider.class);
+        studyService.setCacheProvider(cache);
     }
```
DwayneJengSage, line 71: This reverts the ConsentServiceTest fix. Revert the entire file.

test/org/sagebionetworks/bridge/stormpath/StormpathAccountDaoMultipleSubpopulationsTest.java
DwayneJengSage, line 29: Restore the entire file.

test/org/sagebionetworks/bridge/stormpath/StormpathAccountDaoTest.java
```diff
@@ -307,7 +310,62 @@ public void canResendEmailVerification() throws Exception {
             }
         }
     }
+    
+    @Test
+    public void canSetAndRetrieveConsentsForMultipleSubpopulations() {
```
DwayneJengSage: This test breaks ConsentServiceTest. Revert the entire file.

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
```diff
@@ -43,7 +43,7 @@ public Result getSelfParticipant() {
         UserSession session = getAuthenticatedSession();
         Study study = studyService.getStudy(session.getStudyIdentifier());
         
-        StudyParticipant participant = participantService.getParticipant(study, NO_CALLER_ROLES, session.getId());
+        StudyParticipant participant = participantService.getParticipant(study, session.getId(), false);
```
DwayneJengSage, line 46: This exposes health code to end users. Should we filter here?

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
```diff
@@ -257,11 +269,11 @@ public void updateParticipantRequiresIdMatch() throws Exception {
     public void getSelfParticipant() throws Exception {
         StudyParticipant studyParticipant = new StudyParticipant.Builder().withFirstName("Test").build();
         
-        when(participantService.getParticipant(STUDY, NO_CALLER_ROLES, ID)).thenReturn(studyParticipant);
+        when(participantService.getParticipant(STUDY, ID, false)).thenReturn(studyParticipant);
 
         Result result = controller.getSelfParticipant();
         
-        verify(participantService).getParticipant(STUDY, NO_CALLER_ROLES, ID);
+        verify(participantService).getParticipant(STUDY, ID, false);
```
DwayneJengSage: Do we need a check to make sure we're filtering out health code?

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
```diff
@@ -273,7 +285,7 @@ public void getSelfParticipant() throws Exception {
     public void updateSelfParticipant() throws Exception {
         // All values should be copied over here.
         StudyParticipant participant = TestUtils.getStudyParticipant(ParticipantControllerTest.class);
-        doReturn(participant).when(participantService).getParticipant(STUDY, NO_CALLER_ROLES, ID);
+        doReturn(participant).when(participantService).getParticipant(STUDY, ID, false);
```
DwayneJengSage: Add a check to verify that healthcode is being saved into the session.

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
```diff
@@ -115,12 +120,6 @@ public Result updateParticipant(String userId) {
             throw new BadRequestException("ID in JSON does not match email in URL.");
         }
         participantService.updateParticipant(study, session.getParticipant().getRoles(), userId, participant);
-        
-        // Push changes to the user's session, including consent statuses.
-        CriteriaContext context = new CriteriaContext.Builder()
-                .withStudyIdentifier(study.getStudyIdentifier()).build();
-        session = authenticationService.updateSession(study, context, userId);
-        updateSession(session);
```
DwayneJengSage, line 123: Do we no longer need to update the session?

test/org/sagebionetworks/bridge/play/controllers/UserProfileControllerTest.java
```diff
@@ -158,7 +158,7 @@ public void getUserProfile() throws Exception {
     @Test
     public void updateUserProfile() throws Exception {
         StudyParticipant participant = new StudyParticipant.Builder().withExternalId("originalId").build();
-        doReturn(participant).when(participantService).getParticipant(study, Sets.newHashSet(), ID);
+        doReturn(participant).when(participantService).getParticipant(study, ID, false);
```
DwayneJengSage, line 166: Add check to make sure healthCode was added to session?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -402,8 +401,8 @@ public void getStudyParticipant() {
         assertTrue(retrievedHistory2.isEmpty());
         
         // A non-researcher will not get the healthCode
-        participant = participantService.getParticipant(STUDY, Sets.newHashSet(DEVELOPER), ID);
-        assertNull(participant.getHealthCode());
+        participant = participantService.getParticipant(STUDY, ID, false);
+        assertEquals(HEALTH_CODE, participant.getHealthCode());
```
DwayneJengSage: Minor: Comment is not accurate, should say something about always getting healthcodes regardless of history.

app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
```diff
@@ -392,9 +392,11 @@ private Account constructAccount(StudyIdentifier studyId, com.stormpath.sdk.acco
         }
         if (healthId == null) {
             healthId = healthCodeService.createMapping(studyId);
+            account.setHealthId(healthId);
             updateAccount(account);
```
DwayneJengSage, line 396: Add unit test to ensure we're updating the account with a healthId?

test/org/sagebionetworks/bridge/cache/CacheProviderTest.java
```diff
@@ -306,12 +309,13 @@ private void assertSession(String json) {
         assertEquals(DateTime.parse("2016-04-21T16:48:22.386Z"), participant.getCreatedOn());
         assertEquals(Sets.newHashSet(Roles.ADMIN), participant.getRoles());
         assertEquals(Sets.newHashSet("en","fr"), participant.getLanguages());
-        assertNotNull(participant.getHealthCode());
-        assertNotEquals("TFMkaVFKPD48WissX0bgcD3esBMEshxb3MVgKxHnkXLSEPN4FQMKc01tDbBA"+
-                "VcXx94kMX6ckXVYUZ8wx4iICl08uE+oQr9gorE1hlgAyLAM=", participant.getHealthCode());
+        assertEquals("ABC", participant.getExternalId());
         
-        SubpopulationGuid apiGuid = SubpopulationGuid.create("api");
+        Encryptor ENCRYPTOR = new AesGcmEncryptor(BridgeConfigFactory.getConfig().getProperty("bridge.healthcode.redis.key"));
```
DwayneJengSage: We can make the encryptor a static member var.

DwayneJengSage:
Done reviewing. It also looks like you need to resolve some merge conflicts.

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
alxdarksage, line 46: It's marked @JsonIgnore, however, the encryptedHealthCode is being exposed. Let's get this PR fixed and merged and I'll create a follow-up PR to hide encryptedHealthCode in the API.

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
alxdarksage, line 123: Yes, this can cause errors because we're updating another user's session when we don't know their app version, and this could be used to determine consent status. I would rather that administrators purge the user's session if they want the user to pick up the changes, and that will force the user to sign in again.

app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
alxdarksage, line 396: StormpathAccountDaoMockTest.getAccountCreatesHealthCode() tests this. If you remove that setHealthId now that it's in an if/else statement, the test breaks. If you comment out the other branch, the authenticate test fails because it verifies that we initialize an existing healthCode with a healthId.

test/org/sagebionetworks/bridge/cache/CacheProviderTest.java
alxdarksage: moved

test/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDaoTest.java
alxdarksage: Should revert on update

test/org/sagebionetworks/bridge/dynamodb/DynamoParticipantOptionsDaoMockTest.java
alxdarksage, line 14: Should be restored on update

test/org/sagebionetworks/bridge/dynamodb/DynamoParticipantOptionsDaoTest.java
alxdarksage: Should revert on update

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
alxdarksage: Converted to instance property.

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
alxdarksage: Changed everywhere

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
alxdarksage: Added

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
alxdarksage: Added

test/org/sagebionetworks/bridge/play/controllers/UserProfileControllerTest.java
alxdarksage, line 166: Added

test/org/sagebionetworks/bridge/services/AuthenticationServiceTest.java
alxdarksage, line 324: reverted

test/org/sagebionetworks/bridge/services/HealthCodeServiceMockTest.java
alxdarksage, line 14: Added back

test/org/sagebionetworks/bridge/services/HealthCodeServiceTest.java
alxdarksage: removed (this must have been removed by your consent services changes, then added back)

test/org/sagebionetworks/bridge/services/HealthCodeServiceTest.java
alxdarksage: reverted

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Removed. This tests the same thing as above and was originally because roles were being switched in the test, which is no longer the case.

test/org/sagebionetworks/bridge/services/StudyServiceTest.java
alxdarksage, line 71: reverted

test/org/sagebionetworks/bridge/stormpath/StormpathAccountDaoMultipleSubpopulationsTest.java
alxdarksage, line 29: restored

test/org/sagebionetworks/bridge/stormpath/StormpathAccountDaoTest.java
alxdarksage: removed

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
alxdarksage, line 123: We could automate this. Currently though there's a button and an API to do it as a separate step. If you sign people out, you have to do work on the client to determine if this was an edit of the current user's session, etc, which is error-prone.

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
alxdarksage, line 46: For now I'm removing healthCode and encryptedHealthCode from the JSON just to be sure this happens. This can be replaced with a Jackson filter in a follow-on update.

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
```diff
@@ -100,8 +98,14 @@ public Result getParticipant(String userId) {
...
-        StudyParticipant participant = participantService.getParticipant(study,
-                session.getParticipant().getRoles(), userId);
+        StudyParticipant participant = participantService.getParticipant(study, userId, true);
+        
+        // if not enabled, do not include the health code in data returned from this API. 
+        if (!study.isHealthCodeExportEnabled()) {
+            participant = new StudyParticipant.Builder().copyOf(participant)
+                    .withHealthCode(null).build();
+        }
+
```
alxdarksage: For now I'm removing healthCode and encryptedHealthCode from the JSON just to be sure this happens. This can be replaced with a Jackson filter in a follow-on update.

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
```diff
@@ -257,23 +272,28 @@ public void updateParticipantRequiresIdMatch() throws Exception {
...
         Result result = controller.getSelfParticipant();
         
-        verify(participantService).getParticipant(STUDY, NO_CALLER_ROLES, ID);
+        verify(participantService).getParticipant(study, ID, false);
         
         StudyParticipant deserParticipant = BridgeObjectMapper.get()
                 .readValue(Helpers.contentAsString(result), StudyParticipant.class);
 
         assertEquals("Test", deserParticipant.getFirstName());
+        assertNull(deserParticipant.getHealthCode());
```
DwayneJengSage, line 307: This is a vacuous test. The participant we return from the mock already doesn't contain the health code, so we're not exercising the code that filters out health code.

app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
DwayneJengSage, line 396: getAccountCreatesHealthCode() only tests that createMapping() was called. It doesn't test that setHealthId() was called before updateAccount(). We should have a test to exercise this bug fix.

DwayneJengSage:
It looks like github swallowed one of my comments. On StormpathAccountDao line 395 "getAccountCreatesHealthCode() only tests that createMapping() was called. It doesn't test that setHealthId() was called before updateAccount(). We should have a test to exercise this bug fix."

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
```diff
@@ -45,7 +45,10 @@ public Result getSelfParticipant() {
         
         StudyParticipant participant = participantService.getParticipant(study, session.getId(), false);
         
-        return okResult(participant);
+        // Do not return healthCode. This will shortly be handled with Jackson filters.
+        StudyParticipant updated = new StudyParticipant.Builder().copyOf(participant)
+                .withHealthCode(null).withEncryptedHealthCode(null).build();
```
DwayneJengSage, line 50: Is it necessary to clear out both health code and encrypted health code? They're both backed by the same member var. Or is this a "just in case" kind of thing, in case we need to change the implementation?

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
DwayneJengSage, line 307: Since we explicitly clear out both health code and encrypted health code, we should also explicitly check both health code and encrypted health code here.

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
```diff
@@ -146,42 +149,55 @@ public void getParticipants() throws Exception {
...
+        StudyParticipant studyParticipant = new StudyParticipant.Builder().withFirstName("Test").withHealthCode("healthCode").build();
+        when(participantService.getParticipant(study, ID, true)).thenReturn(studyParticipant);
         
-        verify(participantService).getParticipant(STUDY, CALLER_ROLES, ID);
+        Result result = controller.getParticipant(ID);
+        String json = Helpers.contentAsString(result);
+        StudyParticipant retrievedParticipant = BridgeObjectMapper.get().readValue(json, StudyParticipant.class);
+        
+        assertEquals("Test", retrievedParticipant.getFirstName());
+        assertNull(retrievedParticipant.getHealthCode());
```
DwayneJengSage, line 196: Similarly, do we want to explicitly test both getHealthCode() and getEncryptedHealthCode()?

app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
alxdarksage, line 396: added verification of customData being called with ID and then saved

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
alxdarksage, line 307: I've changed these so both health code and encrypted health code are in the participant returned from the mock, and both are verified to be absent as returned from the controller.

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
alxdarksage, line 50: This is a temporary "just in case", I'm going to replace this with Jackson filters which is a little cleaner.

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
alxdarksage, line 196: Added this to the test (yes we want to test this until we can change it).

PR #1151, opened 2016-05-25
Adding ability for researcher to trigger a reset password request for a user.

alxdarksage:
Start a reset of a user's password for a user that has been locked out of the study, or for support of an administrative user like a developer. It will be possible in the UI to just trigger one of these when creating a user with a role.


npm-debug.log
```diff
@@ -0,0 +1,22 @@
+0 info it worked if it ends with ok
```
DwayneJengSage: Remove this file from the diff, add it to .gitignore

DwayneJengSage:
Changes look good. Feel free to merge after removing npm-debug.log.

npm-debug.log
alxdarksage: Done

PR #1155, opened 2016-06-03
Report apis

alxdarksage:
APIs for general reports (either at a study level or for individuals in a study).


app/org/sagebionetworks/bridge/dynamodb/DynamoReportData.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+    @Override
+    public void setKey(String key) {
+        this.key = key;
+    }
+    @DynamoDBMarshalling(marshallerClass = LocalDateMarshaller.class)
+    @DynamoDBRangeKey
+    @JsonSerialize(using = LocalDateToStringSerializer.class)
+    @Override
+    public LocalDate getDate() {
```
DwayneJengSage, line 39: Do we have a unit test for serializing and deserializing DynamoReportData?

test/org/sagebionetworks/bridge/dynamodb/DynamoReportDataDaoTest.java
```diff
@@ -0,0 +1,101 @@
...
+    public void canCrud() {
+        ReportData report1 = createReport(LocalDate.parse("2016-03-30"), "a", "b");
+        ReportData report2 = createReport(LocalDate.parse("2016-03-31"), "c", "d");
+        
+        // These are not in date order. They should come back in date order.
+        dao.saveReportData(report2);
+        dao.saveReportData(report1);
+        
+        DateRangeResourceList<? extends ReportData> results = dao.getReportData(
+                reportDataKey, START_DATE, END_DATE);
```
DwayneJengSage, line 80: Is it worth adding a report outside of the date range so make sure the date range query works? What about a report with a different key?

app/org/sagebionetworks/bridge/dynamodb/DynamoReportDataDao.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    public void deleteReportData(ReportDataKey key) {
+        checkNotNull(key);
+        
+        DynamoDBScanExpression scan = new DynamoDBScanExpression()
+                .withFilterConditionEntry("key", new Condition()
+                        .withComparisonOperator(ComparisonOperator.EQ)
+                        .withAttributeValueList(new AttributeValue(key.getKeyString())));
```
DwayneJengSage: Don't use a scan. Scan will read the entire table, regardless of which and how many entries match your filters. Instead, use a query.

app/org/sagebionetworks/bridge/models/DateRangeResourceList.java
```diff
@@ -0,0 +1,31 @@
+package org.sagebionetworks.bridge.models;
+
+import java.util.List;
+
+import org.joda.time.LocalDate;
+
+public class DateRangeResourceList<T> {
```
DwayneJengSage, line 7: Unit test to make sure this serializes into the format we expect?

app/org/sagebionetworks/bridge/services/ReportService.java
```diff
@@ -0,0 +1,113 @@
...
+        reportDataDao.deleteReportData(key);
+    }
+    
+    private ReportDataKey makeKey(String healthCode, String identifier, StudyIdentifier studyId, ReportType reportType) {
+        return new ReportDataKey.Builder()
+                .withHealthCode(healthCode)
+                .withReportType(reportType)
+                .withIdentifier(identifier)
+                .withStudyIdentifier(studyId).build();
+    }
```
DwayneJengSage, line 108: This defeats the whole purpose of using the builder pattern. The point of the builder pattern is that when we add a new parameter, we don't have to update N^2 constructors and constructor usages. If we anticipate adding new keys (which I suspect is likely), we should remove this faux-constructor and just use the builder directly. If we are sure we'll never add new report keys, then we should get rid of the builder and just use the constructor directly.

test/org/sagebionetworks/bridge/models/reports/ReportDataKeyTest.java
```diff
@@ -0,0 +1,50 @@
...
+import static org.junit.Assert.assertNull;
+import static org.sagebionetworks.bridge.TestConstants.TEST_STUDY;
+
+import org.junit.Test;
+
+import org.sagebionetworks.bridge.json.BridgeObjectMapper;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+public class ReportDataKeyTest {
```
DwayneJengSage, line 15: EqualsVerifier test?

app/org/sagebionetworks/bridge/dynamodb/DynamoReportData.java
alxdarksage, line 39: Adding.

test/org/sagebionetworks/bridge/play/controllers/ReportControllerTest.java
```diff
@@ -0,0 +1,206 @@
...
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
+
+import play.mvc.Result;
+import play.test.Helpers;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ReportControllerTest {
```
DwayneJengSage, line 49: Is it worth adding tests where start and end date are null, to make sure our service doesn't crash in that scenario?

app/org/sagebionetworks/bridge/play/controllers/ReportController.java
```diff
@@ -0,0 +1,142 @@
...
+    
+    final void setParticipantService(ParticipantService participantService) {
+        this.participantService = participantService;
+    }
+    
+    /**
+     * Reports for specific individuals accessible only to consented study participants.
+     */
+    public Result getParticipantReport(String identifier, String startDateString, String endDateString) {
+        UserSession session = getAuthenticatedAndConsentedSession();
```
DwayneJengSage, line 49: Will we run into the same issue as mPower with multiple consents? Or do we plan to address that a different way?

test/org/sagebionetworks/bridge/dynamodb/DynamoReportDataDaoTest.java
alxdarksage, line 80: Added both cases.

app/org/sagebionetworks/bridge/dynamodb/DynamoReportDataDao.java
alxdarksage: I think at one point I was trying to delete all reports for all users at once, and gave up on this. Changing.

test/org/sagebionetworks/bridge/services/ReportServiceTest.java
```diff
@@ -0,0 +1,306 @@
...
+
+    @Test(expected = BadRequestException.class)
+    public void startDateAfterEndDate() {
+        service.getParticipantReport(TEST_STUDY, IDENTIFIER, HEALTH_CODE, END_DATE, START_DATE);
+    }
+
+    @Test(expected = BadRequestException.class)
+    public void dateRangeTooWide() {
+        service.getParticipantReport(TEST_STUDY, IDENTIFIER, HEALTH_CODE, START_DATE, START_DATE.plusDays(46));
+    }
```
DwayneJengSage, line 222: Should we also have the date validation tests and default date tests for Study reports as well?

app/org/sagebionetworks/bridge/models/DateRangeResourceList.java
alxdarksage, line 7: Added serialization test. Not deserialization because we never deserialize it on the server, so didn't want to add @JsonCreator et al when we don't do it, to support the test.

app/org/sagebionetworks/bridge/services/ReportService.java
alxdarksage, line 108: I really don't care either way, removed.

test/org/sagebionetworks/bridge/models/reports/ReportDataKeyTest.java
alxdarksage, line 15: Added

test/org/sagebionetworks/bridge/play/controllers/ReportControllerTest.java
alxdarksage, line 49: Yes, but in ReportServiceTest where dates are defaulted. I'll add null test here though too.

app/org/sagebionetworks/bridge/play/controllers/ReportController.java
alxdarksage, line 49: Reports built against this service need to pass the correct headers in, and that's a client side issue. That's a bug in the existing mPower 1.3 app. Here we expect consented users and that's correct. They must be consented.

test/org/sagebionetworks/bridge/services/ReportServiceTest.java
alxdarksage, line 222: Added.

test/org/sagebionetworks/bridge/dynamodb/DynamoReportDataTest.java
```diff
@@ -0,0 +1,50 @@
...
+        objNode.put("a", true);
+        objNode.put("b", "string");
+        objNode.put("c", 10);
+        
+        reportData.setKey(key.getKeyString());
+        reportData.setDate(DATE);
+        reportData.setData(objNode);
+        
+        String json = MAPPER.writeValueAsString(reportData);
+        ReportData deser = MAPPER.readValue(json, ReportData.class);
```
DwayneJengSage: We want to validate the JSON to make sure there are no weird serialized JSON.

alxdarksage: Okay, added that too.

test/org/sagebionetworks/bridge/play/controllers/ReportControllerTest.java
```diff
@@ -115,8 +132,16 @@ public void getParticipantReportDataNoDates() throws Exception {
         assertResult(result);
     }
     
+    @Test(expected = BadRequestException.class)
+    public void getParticipantReportNoUserAgent() throws Exception {
+        setupContext("Bad Request", "en-US");
+
+        controller.getParticipantReport("foo", START_DATE.toString(), END_DATE.toString());
+    }
+
```
DwayneJengSage, line 168: Test case to also verify language?

alxdarksage, line 168: Added

PR #1159, opened 2016-06-09
An index of report identifiers

alxdarksage:
Adding a table to index the identifiers used for reports. You need to list these out so that research administrators can navigate to, and add or view, report data for an individual or the study as a whole.

(I am going to add to the researcher UI so we can test reports even before integration with an analysis in Synapse is created.)

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/ReportController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/ReportController.java
@@ -53,7 +53,7 @@ public class ReportController extends BaseController {
      * Get a list of the identifiers used for participant reports in this study.
      */
     public Result getParticipantReportIndices() throws Exception {
-        UserSession session = getAuthenticatedAndConsentedOrResearcherSession();
+        UserSession session = getAuthenticatedSession();
         
         List<? extends ReportIndex> indices = reportService.getReportIndices(
                 session.getStudyIdentifier(), ReportType.PARTICIPANT);
@@ -65,7 +65,6 @@ public class ReportController extends BaseController {
      */
     public Result getParticipantReport(String identifier, String startDateString, String endDateString) {
         UserSession session = getAuthenticatedAndConsentedOrResearcherSession();
-        checkHeaders();
         
         LocalDate startDate = parseDateHelper(startDateString);
         LocalDate endDate = parseDateHelper(endDateString);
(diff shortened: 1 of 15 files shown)


app/org/sagebionetworks/bridge/play/controllers/ReportController.java
```diff
@@ -45,10 +50,21 @@ final void setAccountDao(AccountDao accountDao) {
...
+                session.getStudyIdentifier(), ReportType.PARTICIPANT);
+        return okResult(indices);
+    }
+    
+    /**
+     * Reports for specific individuals accessible to either consented study participants, or to researchers.
      */
     public Result getParticipantReport(String identifier, String startDateString, String endDateString) {
-        UserSession session = getAuthenticatedAndConsentedSession();
+        UserSession session = getAuthenticatedAndConsentedOrResearcherSession();
```
DwayneJengSage, line 67: This method gets the report for the current user's health code, which means even though Researchers can access this, they can't do anything useful with it.

DwayneJengSage:
Also in-memory caching for writing report indices.

test/org/sagebionetworks/bridge/services/ReportServiceTest.java
```diff
@@ -158,6 +174,10 @@ public void saveStudyReportData() {
         
         service.saveStudyReport(TEST_STUDY, IDENTIFIER, someData);
         
+        verify(mockReportIndexDao).addIndex(new ReportDataKey.Builder()
+                .withStudyIdentifier(TEST_STUDY)
+                .withReportType(ReportType.STUDY)
+                .withIdentifier(IDENTIFIER).build());
```
DwayneJengSage: Do we also want to add a test for deleting study report data / index?

DwayneJengSage:
Overall looks good. Just a few comments.

DwayneJengSage:
We need a ReportService unit test to verify that when you delete a report it also deletes the index.

test/org/sagebionetworks/bridge/models/reports/ReportDataKeyTest.java
```diff
@@ -25,17 +26,32 @@ public void constructParticipantKey() {
...
         assertEquals("report", key.getIdentifier());
         assertEquals(ReportType.PARTICIPANT, key.getReportType());
     }
     
     @Test
+    public void canIncludeDateValidation() {
+        try {
+            // This is tested in tests for the validator, but date is actually validated in the 
+            // build method.
+            new ReportDataKey.Builder().validateWithDate(null).build();
```
DwayneJengSage, line 41: We need a success case test for validateWithDate().

Also, since you moved some of the validation logic to the builder, we need a test to validate that the builder validates the ReportDataKey.

alxdarksage, line 41: Added date to a method that constructs and tests a valid key.

PR #1161, opened 2016-06-13
Report API updates

alxdarksage:
Tweaking permissions of some endpoints to allow administration of reports in the research manager;

Adding APIs to delete individual report records (again mostly for admin purposes). 

Split some routes for permissions reasons (e.g. only participants need to submit consent-related headers).

Revised the report URL namespace because some of the URLs were definitely incorrect and I think participant reports are a logical sub-resource of participants, and the URLs are clearer if they are.

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/ReportController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/ReportController.java
@@ -6,7 +6,6 @@ import static org.sagebionetworks.bridge.Roles.WORKER;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
 import java.util.LinkedHashSet;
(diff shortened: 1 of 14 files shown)


app/org/sagebionetworks/bridge/play/controllers/ReportController.java
```diff
@@ -50,21 +66,25 @@ final void setAccountDao(AccountDao accountDao) {
     }
     
     /**
-     * Get a list of the identifiers used for participant reports in this study.
+     * Get a list of the identifiers used for reports in this study. If the value is missing or invalid,
```
liujoshua, line 69: Identifiers or indices? Or are they interchangeable names?

PR #1162, opened 2016-06-15
Updates to Jedis connection pool based on advice from Redis Clound.

alxdarksage:
Their approach is to test when the connections are created, but just let them throw errors afterward rather than doing these additional ping checks (which aren't guaranteed anyway... you could still have an error between the ping and the actual request). But they do take up resources. Also increasing the pool size (limit is 256 in production).

See https://sagebionetworks.jira.com/browse/BRIDGE-1387


DwayneJengSage:
Looks good. Let me know if you want me to merge this or if you want me to hold off.

PR #1167, opened 2016-06-18
field and choice name sanitization; allow other for multi-choice fields; better schema field name validation

DwayneJengSage:
A few changes here:
- sanitization for schema field names, choice enumeration, and submitted upload values (https://sagebionetworks.jira.com/browse/BRIDGE-1392)
- allowOtherChoices in multi-choice schema fields - This is needed to combine Mutable Schemas and the new Upload Fields with Surveys (https://sagebionetworks.jira.com/browse/BRIDGE-1289)
- improved validation for schema field name collisions

Testing done:
- updated unit tests
- added sanitization scenarios to the UploadHandlersEndToEndTest
- ran integ tests UploadSchemaTest and UploadTest against local server
- manually tested schema with allowOther and basic field name sanitization

### DIFF
--- a/app/org/sagebionetworks/bridge/validators/UploadSchemaValidator.java
+++ b/app/org/sagebionetworks/bridge/validators/UploadSchemaValidator.java
@@ -1,19 +1,27 @@
 package org.sagebionetworks.bridge.validators;
 
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.TreeSet;
 
 import com.google.common.base.Strings;
 
 import org.springframework.validation.Errors;
 import org.springframework.validation.Validator;
+
+import org.sagebionetworks.bridge.BridgeUtils;
 import org.sagebionetworks.bridge.models.upload.UploadFieldDefinition;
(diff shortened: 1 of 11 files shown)


app/org/sagebionetworks/bridge/validators/UploadSchemaValidator.java
```diff
@@ -101,16 +107,50 @@ public void validate(Object target, Errors errors) {
...
+                // alphabetical order.
+                Set<String> seenFieldNameSet = new HashSet<>();
+                Set<String> dupeFieldNameSet = new TreeSet<>();
+                for (String oneFieldName : fieldNameList) {
+                    if (seenFieldNameSet.contains(oneFieldName)) {
+                        dupeFieldNameSet.add(oneFieldName);
+                    } else {
+                        seenFieldNameSet.add(oneFieldName);
+                    }
+                }
```
alxdarksage, line 148: This is happening after sanitization? So #foo and !foo which I think both become _foo are detected as dups?

DwayneJengSage, line 148: I believe so. This is because sanitization is happening in the builder, and validation can only be called on a built object.

I'll add a test to verify this.

PR #1172, opened 2016-06-22
Withdraw a user entirely from a study

alxdarksage:
Provide a 'withdraw from all consents in a study' method for both individuals and researchers, acting on their behalf.

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoUserConsentDao.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoUserConsentDao.java
@@ -64,12 +64,22 @@ public class DynamoUserConsentDao implements UserConsentDao {
         checkNotNull(subpopGuid);
         checkArgument(withdrewOn > 0L);
         
-        DynamoUserConsent3 activeConsent = (DynamoUserConsent3)getActiveUserConsent(healthCode, subpopGuid);
-        if (activeConsent == null) {
+        // In case a conflict has occurred where two consents are active for a single subpopulation, find 
+        // all of them and withdraw from all of them.
+        DynamoUserConsent3 hashKey = new DynamoUserConsent3(healthCode, subpopGuid);
+        DynamoDBQueryExpression<DynamoUserConsent3> query = new DynamoDBQueryExpression<DynamoUserConsent3>()
+            .withScanIndexForward(false)
+            .withHashKeyValues(hashKey)
+            .withQueryFilterEntry("withdrewOn", new Condition().withComparisonOperator(ComparisonOperator.NULL));
+        
+        List<DynamoUserConsent3> results = mapper.query(DynamoUserConsent3.class, query);
+        if (results.isEmpty()) {
             throw new EntityNotFoundException(UserConsent.class);
         }
-        activeConsent.setWithdrewOn(withdrewOn);
-        mapper.save(activeConsent);
+        for (DynamoUserConsent3 consent : results) {
+            consent.setWithdrewOn(withdrewOn);
+            mapper.save(consent);
+        }
     }
 
     @Override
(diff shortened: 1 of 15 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoUserConsentDao.java
```diff
@@ -89,6 +89,7 @@ public UserConsent getActiveUserConsent(String healthCode, SubpopulationGuid sub
 
         DynamoDBQueryExpression<DynamoUserConsent3> query = new DynamoDBQueryExpression<DynamoUserConsent3>()
             .withScanIndexForward(false)
+            .withLimit(1)
             .withQueryFilterEntry("withdrewOn", new Condition().withComparisonOperator(ComparisonOperator.NULL))
             .withHashKeyValues(hashKey);
```
DwayneJengSage, line 104: Is it possible for someone to consent, consent again, and then withdraw from the second consent but still have an active first consent? If that's not possible, would it be better to just always get the last consent without the filter, then check if it's withdrawn or not?

DwayneJengSage, line 104: If it _is_ possible to withdraw from the second consent but not the first, can we have a test for that to make sure this query works as expected?

app/org/sagebionetworks/bridge/services/ConsentService.java
```diff
@@ -48,6 +51,8 @@
 @Component
 public class ConsentService {
 
+    private final Logger LOGGER = LoggerFactory.getLogger(HealthCodeService.class);
```
DwayneJengSage: Minor: The class should be ConsentService.class

conf/routes
```diff
@@ -14,18 +14,19 @@ POST   /v3/auth/verifyEmail             @org.sagebionetworks.bridge.play.control
...
+POST   /v3/participants/:userId                              @org.sagebionetworks.bridge.play.controllers.ParticipantController.updateParticipant(userId: String)
+GET    /v3/participants/:userId/activities                   @org.sagebionetworks.bridge.play.controllers.ParticipantController.getActivityHistory(userId: String, offsetKey: String ?= null, pageSize: String ?= null)
+DELETE /v3/participants/:userId/activities                   @org.sagebionetworks.bridge.play.controllers.ParticipantController.deleteActivities(userId: String)
+POST   /v3/participants/:userId/signOut                      @org.sagebionetworks.bridge.play.controllers.ParticipantController.signOut(userId: String)
+POST   /v3/participants/:userId/requestResetPassword         @org.sagebionetworks.bridge.play.controllers.ParticipantController.requestResetPassword(userId: String)
+POST   /v3/participants/:userId/resendEmailVerification      @org.sagebionetworks.bridge.play.controllers.ParticipantController.resendEmailVerification(userId: String)
+POST   /v3/participants/:userId/consents/withdraw			 @org.sagebionetworks.bridge.play.controllers.ParticipantController.withdrawFromAllConsents(userId: String)
```
DwayneJengSage: Minor: tabs instead of spaces

conf/routes
```diff
@@ -14,18 +14,19 @@ POST   /v3/auth/verifyEmail             @org.sagebionetworks.bridge.play.control
...
+POST   /v3/participants/:userId/consents/:guid/resendConsent @org.sagebionetworks.bridge.play.controllers.ParticipantController.resendConsentAgreement(userId: String, guid: String)
```
DwayneJengSage, line 29: Minor: Was this intentionally renamed?

DwayneJengSage:
Overall looks good. Just a few minor comments.

app/org/sagebionetworks/bridge/dynamodb/DynamoUserConsentDao.java
alxdarksage, line 104: We check twice in the consent path to make sure they aren't creating a duplicate, including right before creating the user consent record. So it should not be possible but you could contrive it to cause it to happen. However, externally it will appear you are withdrawn because you can only have one signed and active signature. This is a bucket of worms, I can write a test to see if, for example, you are considered consented at this point or not.

If we just moved the consentCreatedOn value into the signature, we could get rid of the UserConsent table and then there would be no danger of inconsistency. They would either have an active signature in their history, or not.

alxdarksage, line 104: Okay, I created a test that directly creates two consent records for one subpopulation, and verified that withdrawing didn't work correctly. Then I fixed it by iterating through all consents and withdrawing all of them.

PR #1205, opened 2016-07-12
Do not sign user in after verifying email. 

alxdarksage:
UPDATE: want to confirm with Mike in person that we don't want this, since it will foreclose on a particular user experience working on desktop.

This request is not sent from the client and does not have the correct HTTP headers to accurately determine consent status. It is also one of the reasons we can't perform more validation of the CriteriaContext object.


liujoshua:
Ready to merge

PR #1207, opened 2016-07-13
Active consent in subpopulation

alxdarksage:
- Moving the consentCreatedOn value of the active consent to the subpopulation. We're using it to find the active consent once it has been set (need to run migration). Now there can only be one active consent because it is treated as a foreign key, not a flag we need to scan for in the StudyConsent table
- Pushed UserSession out of the ConsentService, as well as updating the session in the service. This is now done in the controller after methods that logically update the session. This makes refactoring ConsentService easier, and I don't believe session should be passed to any service.
- Added a migration to find and copy the active consentCreatedOn value to the subpopulations.
- moved StudyConsentService out of the ConsentEmailProvider, which simplifies that provider and was made possible by the above changes.

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
@@ -111,7 +111,7 @@ public class ConsentController extends BaseController {
         consentService.withdrawConsent(study, SubpopulationGuid.create(guid), session.getParticipant(), withdrawal,
                 withdrewOn);
         
-        UserSession updatedSession = authenticationService.updateSession(study, getCriteriaContext(session));
+        UserSession updatedSession = authenticationService.getSession(study, getCriteriaContext(session));
         if (!updatedSession.doesConsent()) {
             updateSharingStatusAndSession(study, updatedSession, SharingScope.NO_SHARING);
         } else {
@@ -171,7 +171,7 @@ public class ConsentController extends BaseController {
         consentService.consentToResearch(study, subpopGuid, session.getParticipant(), consentSignature,
                 sharing.getSharingScope(), true);
         
-        UserSession updatedSession = authenticationService.updateSession(study, getCriteriaContext(session));
+        UserSession updatedSession = authenticationService.getSession(study, getCriteriaContext(session));
         updateSession(updatedSession);
         return createdResult("Consent to research has been recorded.");
     }
--- a/app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
@@ -4,12 +4,15 @@ import static org.sagebionetworks.bridge.dao.ParticipantOption.SHARING_SCOPE;
 
 import org.joda.time.DateTime;
 import org.sagebionetworks.bridge.dao.ParticipantOption.SharingScope;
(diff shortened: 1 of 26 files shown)


app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
```diff
@@ -105,9 +108,15 @@ public Result withdrawConsentV2(String guid) throws Exception {
         final Study study = studyService.getStudy(session.getStudyIdentifier());
         final long withdrewOn = DateTime.now().getMillis();
         
-        consentService.withdrawConsent(study, SubpopulationGuid.create(guid), session, withdrawal, withdrewOn);
-        updateSession(session);
+        consentService.withdrawConsent(study, SubpopulationGuid.create(guid), session.getParticipant(), withdrawal,
+                withdrewOn);
         
+        UserSession updatedSession = authenticationService.updateSession(study, getCriteriaContext(session));
```
alxdarksage: rename: getSession()

app/org/sagebionetworks/bridge/services/ConsentService.java
```diff
@@ -114,80 +118,77 @@ public ConsentSignature getConsentSignature(Study study, SubpopulationGuid subpo
...
-        ConsentStatus status = session.getConsentStatuses().get(subpopGuid);
-        // There will be a status object for each subpopulation the user is mapped to. If 
-        // there's no status object, then in effect the subpopulation does not exist for 
-        // this user and they should get back a 404.
-        if (status == null) {
-            throw new EntityNotFoundException(Subpopulation.class);
-        }
-        if (status.isConsented()) {
+        Account account = accountDao.getAccount(study, participant.getId());
+        ConsentSignature active = account.getActiveConsentSignature(subpopGuid);
```
alxdarksage: This will not allow you to re-consent to a more up-to-date consent version. We should check and only throw exception if the active consent timestamp is later than the signature's.

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
```diff
@@ -134,30 +143,36 @@ Result changeSharingScope(SharingScope sharingScope, String message) {
...
+        
+        consentService.consentToResearch(study, subpopGuid, session.getParticipant(), consentSignature,
                 sharing.getSharingScope(), true);
         
-        StudyParticipant participant = new StudyParticipant.Builder().copyOf(session.getParticipant())
-                .withSharingScope(sharing.getSharingScope()).build();
-        session.setParticipant(participant);
-        updateSession(session);
+        UserSession updatedSession = authenticationService.updateSession(study, getCriteriaContext(session));
+        updateSession(updatedSession);
```
alxdarksage, line 175: Maybe it would be clearer if we had BaseController.refreshSession(session) that did all of this.

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
alxdarksage: Renamed

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
alxdarksage, line 175: Replaced updateSession() with one method that does all this (gets new session, sets it everywhere), and used that everywhere instead.

PR #1212, opened 2016-07-14
Handle concurrent modifications to upload status

liujoshua:


app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
```diff
@@ -66,14 +71,16 @@ public Upload getUpload(@Nonnull String uploadId) {
...
 
         // TODO: If we globalize Bridge, we'll need to make this timezone configurable.
         upload2.setUploadDate(LocalDate.now(BridgeConstants.LOCAL_TIME_ZONE));
         try {
             mapper.save(upload2);
         } catch (ConditionalCheckFailedException ex) {
-            throw new ConcurrentModificationException("Upload " + upload.getUploadId() + " is already complete");
+            // the only other modification of upload object is during validation, so this upload must have been already marked complete
+            LOG.info("Concurrent modification of upload " + upload.getUploadId() + " while marking upload complete");
         }
```
DwayneJengSage, line 83: You'll need to update DynamoUploadDaoTest to account for this change.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
```diff
@@ -93,6 +100,10 @@ public void writeValidationStatus(@Nonnull Upload upload, @Nonnull UploadStatus
         upload2.setRecordId(recordId);
 
         // persist
-        mapper.save(upload2);
+        try {
+            mapper.save(upload2);
+        } catch (ConditionalCheckFailedException ex) {
+            throw new ConcurrentModificationException("Unable to write validation status for upload " + upload.getUploadId());
+        }
```
DwayneJengSage: Similarly, we should add a test to DynamoUploadDaoTest to test this.

DwayneJengSage: Also, this change is probably unnecessary, since this is only being called by the async upload validation pipeline. This means that the error is never bubbled up to the service response, so there's no difference between a ConditionalCheckFailedException and a ConcurrentModificationException.

If we do keep this change, we want to preserve the original error message and stack trace as well.

DwayneJengSage:
Looks good. Feel free to merge whenever you're ready.

PR #1216, opened 2016-07-18
Adding four new fields to the Upload table

alxdarksage:
All for reporting. None are being used but the Upload object serializes correctly for future APIs. 

I believe we'll be able to add two GSIs to this table:

healthCode + requestedOn (replacing existing GIS), and 

studyId + requestedOn (can be added when we can use it, I think the idea is to create an overall report for a certain time period of the upload situation for a study).

### DIFF
--- a/app/org/sagebionetworks/bridge/BridgeConstants.java
+++ b/app/org/sagebionetworks/bridge/BridgeConstants.java
@@ -73,6 +73,5 @@ public class BridgeConstants {
     
     public static final String STORMPATH_NAME_PLACEHOLDER_STRING = "<EMPTY>";
     
-    public static final Set<Roles> NO_CALLER_ROLES = ImmutableSet.of();
-    
+    public static final Set<Roles> NO_CALLER_ROLES = ImmutableSet.of();    
 }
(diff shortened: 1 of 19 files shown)


app/org/sagebionetworks/bridge/BridgeConstants.java
```diff
@@ -74,4 +74,9 @@
     public static final String STORMPATH_NAME_PLACEHOLDER_STRING = "<EMPTY>";
     
     public static final Set<Roles> NO_CALLER_ROLES = ImmutableSet.of();
+    
+    public static final String UPLOAD_FINISHED_BY_WORKER = "s3 listener";
+    
+    public static final String UPLOAD_FINISHED_BY_CLIENT = "client";
```
DwayneJengSage: Do we expect to use a freeform string in the future? Would it be cleaner and more type-safe to use an enum?

test/org/sagebionetworks/bridge/play/controllers/UploadControllerTest.java
```diff
@@ -138,7 +139,7 @@ public void differentUserInSameStudyCannotCompleteUpload() throws Exception {
         } catch(UnauthorizedException e) {
             
         }
-        verify(uploadService, never()).uploadComplete(any(), any());
+        verify(uploadService, never()).uploadComplete(any(), eq(BridgeConstants.UPLOAD_FINISHED_BY_CLIENT), any());
```
DwayneJengSage: Minor: Should this by any(), so we can verify that no one can complete this?

DwayneJengSage:
Overall looks good. Just a couple minor (non-blocking) comments.

alxdarksage:
Updated based on comments. Thanks for review!

DwayneJengSage:
Looks good. There's a merge conflict that needs to be resolved.

PR #1223, opened 2016-07-21
Adding ability for researches to see upload validation status.

alxdarksage:
There are a number of controller methods with these "either/or" conditions for access, and I believe I'm closer to having a general purpose method for testing all these in one line, but my attempt earlier today to refactor ended up touching a lot of files and getting very messy. So just changing this controller and I might try and clean up the user of getSession\* methods at a later time.


liujoshua:
Good to go

PR #1227, opened 2016-07-25
Stop renewing session every time it is accessed.

alxdarksage:
- remove migration code that's no longer necessary.
- change session lease to 12 hours (still asking around about desirable time for this, I would make it 6 or 8 hours)

Testing:
- Moved existing CacheProviderTest to CacheProviderMockTest for mock tests of provider
- Made the session timeout value an injected value in the CacheProvider, for testing,
- Added CacheProviderTest which creates a real JedisOps object, and using a very short TTL, verifies that a session set to expire at 4 seconds, does expire at 4 seconds, even when (in the middle of the period), we get and set the session again. TTL from Redis is correct to use as the EXPIRES value when calling "setter", and "get" does not update TTL value. 
- (also verifies the initial setex call sets an expiration period)

### DIFF
--- a/test/org/sagebionetworks/bridge/cache/CacheProviderMockTest.java
+++ b/test/org/sagebionetworks/bridge/cache/CacheProviderMockTest.java
@@ -1,0 +1,326 @@
+package org.sagebionetworks.bridge.cache;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyInt;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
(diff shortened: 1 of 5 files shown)


DwayneJengSage:
CacheProvider.setUserSession() also extends the expiration time on sessions. This is called for consent, external IDs, data groups, and update participant self (in addition to signIn). This can be used to potentially extend a session indefinitely. Is there a way we can work around this?

Also, how do we intend to test this?

alxdarksage:
Tests added to verify neither get nor set update the TTL of the session object in Redis. Requires redis.url point to a running instance of redis when testing locally, as mocking this would only prove that CacheProvider works against my understanding of Redis' API, not that it actually works.

test/org/sagebionetworks/bridge/cache/CacheProviderMockTest.java
```diff
@@ -0,0 +1,326 @@
+package org.sagebionetworks.bridge.cache;
```
alxdarksage, line 1: This class is unchanged, I moved CacheProviderTest --> CacheProviderMockTest

test/org/sagebionetworks/bridge/cache/CacheProviderTest.java
```diff
@@ -1,326 +1,104 @@
 package org.sagebionetworks.bridge.cache;
```
alxdarksage, line 1: ... and this test class is entirely new, despite the diff.

PR #1228, opened 2016-07-25
Handle articipants signing in with unknown clients

liujoshua:
Since some Bridge functionality depends on knowing the app version, it's problematic to have participant apps we cannot identify.
- Log these cases
- Update criteria so that criteria requiring either a min or max will not match apps with unknown version


alxdarksage:
What will we do with these warnings in the logs? This is not actually an abnormal situation for back end clients like the Bridge Study Manager, so there will be calls like this.

PR #1233, opened 2016-07-28
Study uploads reporting

alxdarksage:


app/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -209,5 +213,16 @@ public static String getIdFromStormpathHref(String href) {
...
+    public static DateTime getDateTimeOrDefault(String value, DateTime defaultValue) {
+        if (isBlank(value)) {
+            return defaultValue;
+        }
+        try {
+            return DateTime.parse(value);
+        } catch(Exception e) {
+            throw new BadRequestException(value + " is not an ISO timestamp");
+        }
+    }
```
DwayneJengSage: This should be in DateUtils. Also, we should use DateUtils.parseISODateTime().

Also, unit tests?

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
```diff
@@ -98,6 +107,20 @@ public Upload getUpload(@Nonnull String uploadId) {
...
+        RangeKeyCondition condition = new RangeKeyCondition("requestedOn").between(
+                startTime.getMillis(), endTime.getMillis());
+        
+        List<DynamoUpload2> keysToGet = studyIdRequestedOnIndex.queryKeys(DynamoUpload2.class, "studyId",
+                studyId.getIdentifier(), condition);
+
+        return keysToGet.stream().map(key -> {
+            return mapper.load(key);
+        }).collect(Collectors.toList());
+    }
```
DwayneJengSage, line 109: DynamoIndexHelper.query() instead of queryKeys() will save you the work of re-querying the table yourself.

test/org/sagebionetworks/bridge/play/controllers/StudyControllerTest.java
```diff
@@ -167,6 +180,29 @@ public void userCannotAccessCurrentStudy() throws Exception {
         testRoleAccessToCurrentStudy(null);
     }
     
+    @Test
+    public void canGetUploadsForStudy() throws Exception {
+        doReturn(mockSession).when(controller).getAuthenticatedSession(DEVELOPER);
+        
+        DateTime startTime = DateTime.parse("2010-01-01T00:00:00.000Z").withZone(DateTimeZone.UTC);
+        DateTime endTime = DateTime.parse("2010-01-02T00:00:00.000Z").withZone(DateTimeZone.UTC);
```
DwayneJengSage: Nitpick: withZone() is unnecessary here, since nothing is keyed off timezone.

app/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -220,6 +220,29 @@ public Upload getUpload(@Nonnull String uploadId) {
...
+     * <p>Get uploads for an entire study in a time window. Start and end time are optional. If neither are provided, they 
+     * default to the last day of uploads. If end time is not provided, the query ends at the time of the request. If the 
+     * start time is not provided, it defaults to a day before the end time. The time window is constrained to two days 
+     * of uploads (though those days can be any period in time). </p>
+     */
+    public DateTimeRangeResourceList<? extends UploadView> getStudyUploads(@Nonnull StudyIdentifier studyId,
+            @Nullable DateTime startTime, @Nullable DateTime endTime) {
+        checkNotNull(studyId);
+
+        return getUploads(startTime, endTime, false, (start, end)-> {
```
DwayneJengSage: Why don't we want to include the schema info for the study view? I imagine this would be very useful.

app/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -236,12 +259,10 @@ public Upload getUpload(@Nonnull String uploadId) {
             throw new BadRequestException("Query window cannot be longer than two days: " + startTime + "-" + endTime);
         }
```
DwayneJengSage, line 259: It would be preferable to do something like "startTime.plusDays(2).isBefore(endTime)". Otherwise, really weird things happen with leap years and leap seconds.

app/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -236,12 +259,10 @@ public Upload getUpload(@Nonnull String uploadId) {
...
         
-        List<? extends Upload> results = uploadDao.getUploads(healthCode, startTime, endTime);
-        
-        List<UploadView> views = results.stream().map(upload -> {
+        List<UploadView> views = supplier.get(startTime, endTime).stream().map(upload -> {
             UploadView.Builder builder = new UploadView.Builder();
             builder.withUpload(upload);
-            if (upload.getStatus() == UploadStatus.SUCCEEDED) {
+            if (includeSchemaInfo && upload.getStatus() == UploadStatus.SUCCEEDED) {
                 UploadValidationStatus status = getUploadValidationStatus(upload.getUploadId());
```
DwayneJengSage: Instead of calling getUploadValidationStatus(), you could call getRecordById() in either HealthDataService or HealthDataDao. This will save you the redundant call to getUploads(). You can also key on the presence or absence of recordId, which is equivalent to UploadStatus.SUCCEEDED, but more direct.

app/org/sagebionetworks/bridge/BridgeUtils.java
alxdarksage: Was original in StudyController where it was tested, moved to DateUtils and added separate tests there.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
alxdarksage, line 109: Cool. I've updated both uses of index class.

test/org/sagebionetworks/bridge/play/controllers/StudyControllerTest.java
alxdarksage: removed

app/org/sagebionetworks/bridge/services/UploadService.java
alxdarksage: I hadn't thought about it but creating a report of uploads by their schema does seem useful, I'll add.

app/org/sagebionetworks/bridge/services/UploadService.java
alxdarksage, line 259: Changed. Changed tests as well to test participant/study uploads calls separately.

app/org/sagebionetworks/bridge/services/UploadService.java
alxdarksage: changed

DwayneJengSage:
Looks good. Ready to merge whenever.

PR #1236, opened 2016-08-29
Integrate Surveys with Mutable Schemas; relax rules for Mutable Schemas

DwayneJengSage:
Last code change needed to finish off https://sagebionetworks.jira.com/browse/BRIDGE-1289

This change integrate surveys with mutable schemas. We also relax some of the rules for mutable schemas to make them easier to work with.

Testing done:
- unit tests updated
- integ tests (survey, upload schema, upload)
- manual tests

Manual test cases:
- create a new survey schema
- compatible update to survey schema
- incompatible update to survey schema
- explicitly ask for a new schema rev for a survey schema
- attempt to modify a schema field with an incompatible change
- successfully modify a schema field with a compatible change
- attempt to add a required field without minAppVersion to a schema
- successfully add a required field with minAppVersion to a schema

Note: https://github.com/Sage-Bionetworks/BridgeIntegrationTests/pull/76 needs to be pushed out before we can check this in.

TODO: Additional integ tests to test the updated functionality.

### DIFF
--- a/app/org/sagebionetworks/bridge/dao/SurveyDao.java
+++ b/app/org/sagebionetworks/bridge/dao/SurveyDao.java
@@ -43,7 +43,7 @@ public interface SurveyDao {
      *         true if you want to cut a new survey schema, false if you should (attempt to) modify the existing one
      * @return published survey
      */
-    Survey publishSurvey(StudyIdentifier study, GuidCreatedOnVersionHolder keys, boolean newSchemaRev);
+    public Survey publishSurvey(StudyIdentifier study, GuidCreatedOnVersionHolder keys, boolean newSchemaRev);
 
     /**
      * Delete this survey. Survey still exists in system and can be retrieved by direct reference
--- a/app/org/sagebionetworks/bridge/dao/SurveyDao.java
+++ b/app/org/sagebionetworks/bridge/dao/SurveyDao.java
@@ -31,13 +31,19 @@ public interface SurveyDao {
     public Survey versionSurvey(GuidCreatedOnVersionHolder keys);
 
     /**
-     * Make this version of this survey available for scheduling. One scheduled for publishing, 
-     * a survey version can no longer be changed (it can still be the source of a new version).  
+     * Make this version of this survey available for scheduling. One scheduled for publishing,
+     * a survey version can no longer be changed (it can still be the source of a new version).
      * There can be more than one published version of a survey.
+     *
+     * @param study
+     *         study that the survey lives in
      * @param keys
-     * @return
+     *         survey ID and created-on timestamp, which identifies the survey you want to publish
+     * @param newSchemaRev
+     *         true if you want to cut a new survey schema, false if you should (attempt to) modify the existing one
+     * @return published survey
      */
-    public Survey publishSurvey(StudyIdentifier study, GuidCreatedOnVersionHolder keys);
+    Survey publishSurvey(StudyIdentifier study, GuidCreatedOnVersionHolder keys, boolean newSchemaRev);
 
     /**
      * Delete this survey. Survey still exists in system and can be retrieved by direct reference
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
@@ -5,8 +5,10 @@ import javax.annotation.Resource;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
@@ -16,7 +18,9 @@ import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBQueryExpression;
 import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBSaveExpression;
 import com.amazonaws.services.dynamodbv2.model.ConditionalCheckFailedException;
 import com.amazonaws.services.dynamodbv2.model.ExpectedAttributeValue;
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 
 import com.google.common.collect.Maps;
@@ -25,19 +29,26 @@ import org.springframework.stereotype.Component;
 import org.sagebionetworks.bridge.BridgeUtils;
 import org.sagebionetworks.bridge.dao.UploadSchemaDao;
 import org.sagebionetworks.bridge.exceptions.BadRequestException;
+import org.sagebionetworks.bridge.exceptions.BridgeServiceException;
 import org.sagebionetworks.bridge.exceptions.ConcurrentModificationException;
 import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
 import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
 import org.sagebionetworks.bridge.models.surveys.Constraints;
 import org.sagebionetworks.bridge.models.surveys.DataType;
 import org.sagebionetworks.bridge.models.surveys.MultiValueConstraints;
+import org.sagebionetworks.bridge.models.surveys.NumericalConstraints;
 import org.sagebionetworks.bridge.models.surveys.StringConstraints;
 import org.sagebionetworks.bridge.models.surveys.Survey;
 import org.sagebionetworks.bridge.models.surveys.SurveyQuestion;
+import org.sagebionetworks.bridge.models.surveys.SurveyQuestionOption;
+import org.sagebionetworks.bridge.models.surveys.Unit;
 import org.sagebionetworks.bridge.models.upload.UploadFieldDefinition;
 import org.sagebionetworks.bridge.models.upload.UploadFieldType;
 import org.sagebionetworks.bridge.models.upload.UploadSchema;
 import org.sagebionetworks.bridge.models.upload.UploadSchemaType;
+import org.sagebionetworks.bridge.upload.UploadUtil;
+import org.sagebionetworks.bridge.validators.UploadSchemaValidator;
+import org.sagebionetworks.bridge.validators.Validate;
 
 /** DynamoDB implementation of the {@link org.sagebionetworks.bridge.dao.UploadSchemaDao} */
 @Component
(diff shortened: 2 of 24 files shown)


alxdarksage:
Something that has come up is: can you take a required field and make it optional, without also adding a min app version.

Also the min/max app version thing: we are in the process of changing this so they can be set by platform because we have the first Android app coming along. This issue will potentially apply to field definitions as well, don't know what you want to do about this.

app/org/sagebionetworks/bridge/dao/SurveyDao.java
```diff
@@ -31,13 +31,19 @@
...
+     *         study that the survey lives in
      * @param keys
-     * @return
+     *         survey ID and created-on timestamp, which identifies the survey you want to publish
+     * @param newSchemaRev
+     *         true if you want to cut a new survey schema, false if you should (attempt to) modify the existing one
+     * @return published survey
      */
-    public Survey publishSurvey(StudyIdentifier study, GuidCreatedOnVersionHolder keys);
+    Survey publishSurvey(StudyIdentifier study, GuidCreatedOnVersionHolder keys, boolean newSchemaRev);
```
alxdarksage: Do you want to decrease visibility of all the methods in this interface to be consistent?

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
```diff
@@ -136,125 +156,297 @@ public void setStudyIdIndex(DynamoIndexHelper studyIdIndex) {
...
+
+        // Validate schema. There's nowhere else in the call path that validates it, so we have to do it here.
+        Validate.entityThrowingException(UploadSchemaValidator.INSTANCE, schemaToCreate);
+
+        // Short-cut: If we specify we want a new schema rev, then skip directly to creating a new schema rev.
+        if (!newSchemaRev) {
+            // Get the current rev.
+            String studyId = studyIdentifier.getIdentifier();
+            DynamoUploadSchema oldUploadSchema = getUploadSchemaNoThrow(studyId, schemaId);
+            if (oldUploadSchema != null && oldUploadSchema.getSchemaType() == UploadSchemaType.IOS_SURVEY) {
```
alxdarksage, line 185: Are you attempting to update the existing schema only if it's a survey schema? Do you plan to try this for task schemas as well, as client devs have tried to make fields optional before for these as well.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
```diff
@@ -136,125 +156,297 @@ public void setStudyIdIndex(DynamoIndexHelper studyIdIndex) {
...
-                    // an inline_json_blob.
-                    return UploadFieldType.INLINE_JSON_BLOB;
+                uploadFieldType = UploadFieldType.MULTI_CHOICE;
+                fieldDefBuilder.withAllowOtherChoices(multiValueConstraints.getAllowOther());
+
+                // convert the survey answer option list to a list of possible multi-choice answers
+                List<String> fieldAnswerList = new ArrayList<>();
+                //noinspection Convert2streamapi
+                for (SurveyQuestionOption oneSurveyOption : multiValueConstraints.getEnumeration()) {
+                    fieldAnswerList.add(oneSurveyOption.getValue());
```
alxdarksage, line 252: Value (server value) can be null. When it's null, the survey engine is supposed to return the label for the option. There are definitely surveys in production that do this.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
```diff
@@ -136,125 +156,297 @@ public void setStudyIdIndex(DynamoIndexHelper studyIdIndex) {
...
-                // TODO: revisit this in the new upload data format
-                return UploadFieldType.INLINE_JSON_BLOB;
+                uploadFieldType = UploadFieldType.SINGLE_CHOICE;
+
+                // Unfortunately, survey questions don't know their own length. Fortunately, we can determine this
+                // by iterating over all answers.
+                int maxLength = 0;
+                //noinspection Convert2streamapi
+                for (SurveyQuestionOption oneSurveyOption : multiValueConstraints.getEnumeration()) {
+                    maxLength = Math.max(maxLength, oneSurveyOption.getValue().length());
```
alxdarksage, line 263: Same, value can be null, then use label

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
```diff
@@ -136,125 +156,297 @@ public void setStudyIdIndex(DynamoIndexHelper studyIdIndex) {
...
+        fieldDefBuilder.withType(uploadFieldType);
+        fieldDefList.add(fieldDefBuilder.build());
+
+        // NumericalConstraints (integer, decimal, duration) have units. We want to write the unit into Synapse in case
+        // (a) the survey question changes the units or (b) we add support for app-specified units.
+        if (constraints instanceof NumericalConstraints) {
+            UploadFieldDefinition unitFieldDef = new DynamoUploadFieldDefinition.Builder()
+                    .withName(fieldName + UploadUtil.UNIT_FIELD_SUFFIX).withType(UploadFieldType.STRING)
+                    .withRequired(false).withMaxLength(Unit.MAX_STRING_LENGTH).build();
+            fieldDefList.add(unitFieldDef);
```
alxdarksage, line 295: Sorry, where is the selected unit included in the name?

app/org/sagebionetworks/bridge/play/controllers/SurveyController.java
```diff
@@ -239,13 +239,13 @@ public Result updateSurvey(String surveyGuid, String createdOnString) throws Exc
...
-    public Result publishSurvey(String surveyGuid, String createdOnString) throws Exception {
+    public Result publishSurvey(String surveyGuid, String createdOnString, String newSchemaRev) throws Exception {
         UserSession session = getAuthenticatedSession(DEVELOPER);
         StudyIdentifier studyId = session.getStudyIdentifier();
          
         Survey survey = getSurveyWithoutCacheInternal(surveyGuid, createdOnString, session);
-        
-        survey = surveyService.publishSurvey(studyId, survey);
+
+        survey = surveyService.publishSurvey(studyId, survey, Boolean.parseBoolean(newSchemaRev));
```
alxdarksage: In terms of backwards compatibility, does this return false if newSchemaRev is null?

app/org/sagebionetworks/bridge/validators/SurveyValidator.java
```diff
@@ -205,6 +208,41 @@ private void doValidateConstraintsType(Errors errors, UIHint hint, MultiValueCon
             rejectField(errors, "uiHint",
                     "doesn't allow multiples but the '%s' UI hint gathers more than one answer", hintName);
         }
+
+        // must have an enumeration (list of question options)
+        List<SurveyQuestionOption> optionList = mcon.getEnumeration();
+        if (optionList == null || optionList.isEmpty()) {
+            rejectField(errors, "enumeration", "must have non-null, non-empty choices list");
```
alxdarksage, line 215: This will prevent people from saving between creating such a question, and definition some options. Is it absolutely necessary?

app/org/sagebionetworks/bridge/validators/SurveyValidator.java
```diff
@@ -205,6 +208,41 @@ private void doValidateConstraintsType(Errors errors, UIHint hint, MultiValueCon
...
+                } else {
+                    dupeSet.add(optionValue);
+                }
+
+                errors.popNestedPath();
+            }
+
+            // values must be unique
+            if (!dupeSet.isEmpty()) {
+                rejectField(errors, "enumeration", "must have unique values");
```
alxdarksage, line 243: This will prevent showing a user different labels that for all intents and purposes, have the same label (I guess an example might be drug names that all map to the same generic drug name.... if that list is obnoxiously long and they want to make separate entries, rather than one with A/B/C/D, they will no longer be able to do it. Is it absolutely necessary?

test/org/sagebionetworks/bridge/models/surveys/SurveyQuestionOptionTest.java
```diff
@@ -0,0 +1,48 @@
...
+        assertEquals("test-detail", option.getDetail());
+        assertEquals("test-value", option.getValue());
+        assertEquals(DUMMY_IMAGE, option.getImage());
+
+        String optionString = option.toString();
+        assertTrue(optionString.contains(option.getLabel()));
+        assertTrue(optionString.contains(option.getDetail()));
+        assertTrue(optionString.contains(option.getValue()));
+        assertTrue(optionString.contains(option.getImage().toString()));
+    }
```
alxdarksage, line 26: test of SurveyQuestionOption(String label) constructor?

DwayneJengSage:
> Something that has come up is: can you take a required field and make it optional, without also adding a min app version.

Yes. UploadUtilTest.isCompatibleFieldDefRequired() explicitly tests that.

> Also the min/max app version thing: we are in the process of changing this so they can be set by platform because we have the first Android app coming along. This issue will potentially apply to field definitions as well, don't know what you want to do about this.

As per IM conversation, this would introduce a lot of extra complexity for very little gain. At this point, it would be better off to remove min/maxAppVersion from fields altogether. In this case, you can only add optional fields, and you can only flip fields from required to optional.

This code change will reflect that change in logic. A future change will be needed to remove min- and maxAppVersion altogether.

app/org/sagebionetworks/bridge/dao/SurveyDao.java
DwayneJengSage: As per IM conversation, I don't want to add unnecessary code churn to this pull request. I've reverted this particular change and filed https://sagebionetworks.jira.com/browse/BRIDGE-1460 to clean this up.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
DwayneJengSage: As per IM conversation, this is necessary for surveys, since survey editors don't have direct control over schemas. This is also simpler for surveys since there are lots of assumptions we can make if we know it's a survey.

While we _could_ implement this for schemas, it would be far more complicated, and it wouldn't be necessary, since study developers have direct control over schemas.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
DwayneJengSage: As discussed on IM, the change to SurveyQuestionOption makes this unnecessary.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
DwayneJengSage: When you create an Int question "foo", we create schema fields "foo" and "foo_unit".

app/org/sagebionetworks/bridge/validators/SurveyValidator.java
DwayneJengSage: As per IM conversation, this is necessary because otherwise, we get duplicate column names in Synapse.

While we could implement dedupe logic, this adds extra complexity, is very unlikely to be relevant, and researchers and trivially work around this.

app/org/sagebionetworks/bridge/validators/SurveyValidator.java
DwayneJengSage: [EDIT: Replied to the wrong comment]

As per IM conversation, this is necessary because otherwise, we get duplicate column names in Synapse.

While we could implement dedupe logic, this adds extra complexity, is very unlikely to be relevant, and researchers can trivially work around this.

app/org/sagebionetworks/bridge/validators/SurveyValidator.java
DwayneJengSage: Yes, this is necessary, because we can't create a survey schema if the MultiValue questions don't have the enumerations. See https://sagebionetworks.jira.com/browse/BRIDGE-1462 to have separate validation steps for "syntactically correct" and "complete and publishable"

test/org/sagebionetworks/bridge/models/surveys/SurveyQuestionOptionTest.java
DwayneJengSage: As per IM conversation, the goal is to test the changes made in the code review. Exhaustively testing SurveyQuestionOption is beyond the scope of this change.

PR #1237, opened 2016-08-30
Add platform support to the Criteria object

alxdarksage:
Subpopulations (aka consent groups) and criteria-based schedule plans contain a "criteria" object that is data about how these objects should be filtered based on a request. We do language, data group and app version filtering of these objects, now the app version filtering can be declared for any arbitray operating system name:

```
CardioHealth/1 (iPhone 6.0; iPhone OS/9.0.2) BridgeSDK/10
```

If the User-Agent string doesn't contain that "iPhone OS" information, then no filtering is done. This is opt-in.

Note that while the server will accept any string for the OS name, I wrote the SDK to have a more typesafe enumeration of the values ("Android" and "iPhone OS" only). That's because in the BSM UI, those are the only OS names I allow you to set versions for, so I wanted to help implementers align these (seems really easy to get this wrong).
- criteria object now has a map of operating system names to integer min/max app versions (two we specifically expect are "Android" and "iPhone OS");
- Find the os name in the ClientInfo object and use it to filter by app version (if no os name is specified... don't filter. This is our default behavior currently)
- the existing min/max app versions are still retrieved from DDB but transferred and saved in the map as the values for IOS (which they are by default). Eventually this code can be removed when the relevant objects are re-saved (which essentially migrates them). I plan to do this manually, confirm, and then remove this code from DynamoCriteria as part of migration.

### DIFF
--- a/app/org/sagebionetworks/bridge/BridgeUtils.java
+++ b/app/org/sagebionetworks/bridge/BridgeUtils.java
@@ -210,5 +210,17 @@ public class BridgeUtils {
         }
         return null;
     }
-
+    
+    /**
+     * Converts a string to an error key friendly string, e.g. "iPhone OS" is converted to "iphone_os".
+     * 
+     * @throws IllegalArgumentException
+     *             if the string cannot be converted to an error key.
+     */
+    public static String textToErrorKey(String text) {
+        if (StringUtils.isBlank(text)) {
+            throw new IllegalArgumentException("String is not translatable to an error key: " + text);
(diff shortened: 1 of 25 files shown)


syoung-smallwisdom:
Note: In answer to the question "but what about iPad?", Apple uses "iPhone OS" for all iOS devices. It is a holdover to when there was _only_ iPhones using that operating system.

alxdarksage:
reviewer=joshua

PR #1240, opened 2016-09-02
Allow caller to set a minimum number of tasks to return for each schedule

alxdarksage:
API will now return the greater of N days of tasks, or a minimum number of tasks. Only applies to repeating schedules (interval or cron). Intended to be used to show weekly, monthly, or even less frequent tasks to the user (which would not be in the time window).

Also removing code that ensured an account createdOn timestamp existed for a user... all the sessions have long since expired or been flushed and should be loaded with this now (it always exists in a Stormpath record).


app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -115,5 +115,24 @@ protected DateTime getFirstEventDateTime(ScheduleContext context, String eventId
...
         }
         return eventDateTime;
-    }    
+    }
+    
+    /**
+     * If scheduling hasn't reached the end time, or hasn't accumulated the minimum number of tasks, returns true, or 
+     * false otherwise. 
+     */
+    protected boolean continueScheduling(ScheduleContext context, DateTime scheduledTime, List<ScheduledActivity> scheduledActivities) {
```
liujoshua: I'd prefer a name like shouldContinueScheduling or shouldScheduleAdditionalActivities

PR #1243, opened 2016-09-07
remove min/maxAppVersion from schema fields

DwayneJengSage:
It was decided that supporting min/maxAppVersion on individual schema fields adds an unsustainable amount of complexity, so we've decided to remove this feature altogether.

This change removes min/maxAppVersion from schema fields as well as the logic for parsing app version from uploads and calculating required fields based on app version. It's a big huge amount of deleted code.

Note: This should only be merged after the following are merged:
- https://github.com/Sage-Bionetworks/BridgeJavaSDK/pull/308
- https://github.com/Sage-Bionetworks/BridgeIntegrationTests/pull/81

Testing done:
- updated unit tests
- ran upload and schema integ tests
- manual tests

Manual tests:
- back-compat: fetch a schema that currently has min/maxAppVersion, verify nothing breaks
- write a schema, verified min/maxAppVersion are ignored, verify it's written to DDB with nulls removed, etc
- get schema back to make sure it can be written and read round-trip


alxdarksage:
Looks good. Merging. Will remove from UI today as well.

PR #1248, opened 2016-09-13
Upload dedupe logic for completed uploads

DwayneJengSage:
Part of https://sagebionetworks.jira.com/browse/BRIDGE-468

Dedupe logic currently only applies to requested incomplete uploads. In some cases, we have duplicate uploads for uploads that are already complete. This change allows us to dedupe uploads that have already been completed.

This works by, when we create an upload, if we detect it as a dupe (based on healthCode and MD5 hash within the last 7 days), we'll tag the upload with the dupe upload ID. Then when the upload is completed, if it's a dupe, we ignore it. This allows apps to continue their existing workflow of request-upload-complete, but we don't end up with duplicate uploads.

Testing done:
- ran updated Upload unit tests
- upload integ test
- manual tests

Manual test cases:
- not a dupe
- dupe of requested incomplete upload
- dupe of completed upload
- dupe of completed upload, but passes study whitelist

### DIFF
--- a/app/org/sagebionetworks/bridge/services/UploadService.java
+++ b/app/org/sagebionetworks/bridge/services/UploadService.java
@@ -11,11 +11,13 @@ import javax.annotation.Resource;
 import java.net.URL;
 import java.util.Date;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import com.amazonaws.AmazonClientException;
 import com.google.common.base.Strings;
 
+import com.google.common.collect.ImmutableSet;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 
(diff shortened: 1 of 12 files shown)


app/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -148,7 +157,8 @@ public UploadSession createUpload(StudyIdentifier studyId, StudyParticipant part
             uploadId = originalUploadId;
         } else {
             // This is a new upload.
-            Upload upload = uploadDao.createUpload(uploadRequest, studyId, participant.getHealthCode());
+            Upload upload = uploadDao.createUpload(uploadRequest, studyId, participant.getHealthCode(),
+                    originalUploadId);
             uploadId = upload.getUploadId();
 
             if (originalUploadId != null) {
```
liujoshua, line 169: Is there a reason we shouldn't set UploadStatus.DUPLICATE here, and wait for upload complete? 

If we set it in upload complete, could we call it UploadStatus.VALIDATION_SKIPPED (assigned on upload complete) to distinguish this from duplicate detection (assigned during upload creation)? Right now looking at the Upload object would confuse me to see one with with a non-null duplicateUploadId yet not be marked UploadStatus.DUPLICATE, and a rename would keep meaning of the UploadStatuses independent of whether it's a duplicate

liujoshua (approved):

PR #1251, opened 2016-09-15
Survey Answer Choice Validation Improvements, and other minor changes

DwayneJengSage:
Main JIRA: https://sagebionetworks.jira.com/browse/BRIDGE-1482

Changes:
- survey answer choices can now be keywords, since they get the question name pre-pended to them
- improving the error messages for survey and schema validation, to include the path and the offending string
- since we're validating field names instead of sanitizing, sanitizing logic has been removed from schema field def builder
- single-choice fields now default to length 100, to prevent cases where a minor update to a single-choice question triggers a full schema refresh (longer single-choice fields will use unbounded strings)
- "true" and "false" have been added to reserved keywords

Testing done:
- updated and ran relevant unit tests
- ran integ tests for surveys, schemas, and survey schemas

Manual Tests
- survey question name and answer choice validation
- schema field name and answer choice validation
- survey with single-choice fields (short and unbounded)
- schema with keywords as choice
- true/false illegal schema field name

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
@@ -76,6 +76,21 @@ public class DynamoUploadSchemaDao implements UploadSchemaDao {
                     .put(DataType.DATETIME, UploadFieldType.TIMESTAMP)
                     .build();
 
+    // Default string length for single_choice field. This is needed to prevent single_choice questions from changing
+    // in length a lot. Package-scoped to facilitate unit tests.
+    static final int SINGLE_CHOICE_DEFAULT_LENGTH = 100;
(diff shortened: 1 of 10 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java
```diff
@@ -76,6 +76,21 @@
                     .put(DataType.DATETIME, UploadFieldType.TIMESTAMP)
                     .build();
 
+    // Default string length for single_choice field. This is needed to prevent single_choice questions from changing
+    // in length a lot. Package-scoped to facilitate unit tests.
+    static final int SINGLE_CHOICE_DEFAULT_LENGTH = 100;
```
alxdarksage, line 81: Should the editor default string fields to a length of 100? From the code below I see it's an internal optimization but will it change the maxLength value that the user sees at a later time?

PR #1256, opened 2016-09-19
Deleting uploads when deleting a user.

alxdarksage:


app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
```diff
@@ -153,4 +155,16 @@ public void writeValidationStatus(@Nonnull Upload upload, @Nonnull UploadStatus
         // persist
         mapper.save(upload2);
     }
+    
+    @Override
+    public void deleteUploadsForHealthCode(@Nonnull String healthCode) {
+        List<? extends Upload> uploadsToDelete = healthCodeRequestedOnIndex.query(
+                DynamoUpload2.class, "healthCode", healthCode, null);
```
DwayneJengSage, line 162: Can the batchDelete() API be done on key objects? Or do we need full objects? If it can, then we can use DynamoIndexHelper.queryKeys() instead of query(), which will save us a second DDB batch read.

DwayneJengSage (changes requested):
Overall looks good. One minor comment about DynamoIndexHelper usage.

DwayneJengSage (approved):

PR #1258, opened 2016-09-21
Save information about user's request on sign in and when retrieving activities.

alxdarksage:
JIRA: https://sagebionetworks.jira.com/browse/BRIDGE-1421

NOTE: this will fail until the following PR is merged:
https://github.com/Sage-Bionetworks/bridge-base/pull/27
(I accidentally merged the BridgeTestUtils update of the in memory Jedis implementation so that'll also fail until bridge-base is merged.)

This is saved in Redis at this point... although the information could be lost when we flush Redis, it's not particularly high value. It's useful when troubleshooting a user's issues and we can always ask them to sign out/sign in again to see what their device is sending.

I can also use the information to show in the researcher UI which subpopulation(s) the user belongs to, and from the schedule plans, what activities they should be getting back from the server. This would also be helpful given the questions I sometimes receive about this.

Another follow-on TODO: there's a lot of static constants that duplicate the same information (language set, user group set, user id, etc.) between the tests, these could be moved to TestConstants. I only did USER_DATA_GROUPS in this commit.


app/org/sagebionetworks/bridge/models/RequestInfo.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    public Set<String> getUserDataGroups() {
+        return userDataGroups;
+    }
+
+    public DateTime getActivitiesAccessedOn() {
+        return (timeZone != null && activitiesAccessedOn != null) ?
+                activitiesAccessedOn.withZone(timeZone) : activitiesAccessedOn;
+    }
```
DwayneJengSage, line 75: We should probably add tests to make sure getActivitiesAccessedOn behaves reasonably when the DateTime differs from the TimeZone (not likely with getActivitiesAccessedOn, but very likely for getSignedInOn). We should also test to make sure the behavior is reasonable for _no_ timezone.

test/org/sagebionetworks/bridge/models/RequestInfoTest.java
```diff
@@ -0,0 +1,81 @@
...
+public class RequestInfoTest {
+
+    private static final StudyIdentifier STUDY_ID = new StudyIdentifierImpl("test-study");
+    private static final String USER_ID = "userId";
+    private static final ClientInfo CLIENT_INFO = ClientInfo.parseUserAgentString("app/20");
+    private static final String USER_AGENT_STRING = "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/53.0.2785.116 Safari/537.36";
+    private static final LinkedHashSet<String> LANGUAGES = TestUtils.newLinkedHashSet("en", "fr");
+    private static final DateTimeZone PST = DateTimeZone.forOffsetHours(-7);
+    private static final DateTime ACTIVITIES_REQUESTED_ON = DateUtils.getCurrentDateTime();
+    private static final DateTime SIGNED_IN_ON = ACTIVITIES_REQUESTED_ON.minusHours(4);
```
DwayneJengSage: Is it worth creating ACTIVITIES_REQUESTED_ON and SIGNED_IN_ON in timezones other than UTC to make sure we're preserving the timezone properly? Is it worth doing the test below to make sure timezones are being preserved?

IIRC, by default, Joda/Jackson _doesn't_ preserve timezone.

DwayneJengSage (changes requested):
Overall looks good. Just a few questions about date-times and timezones.

test/org/sagebionetworks/bridge/models/RequestInfoTest.java
alxdarksage: I'll create some tests for these. This use of time zones is informational only (I'm assuming the researcher would be interested in the time in the user's local time, though that may not be true and we might eventually enforce UTC instead). 

In JavaScript, the tools for manipulating dates are very poor and it isn't easy to show a timestamp in an alternate timezone, even if I have it. If we don't have it, it's okay to show the times in UTC and I can enforce this in case the server is in yet another time zone.

alxdarksage: Ack you're right, Jackson doesn't preserve the time zone on serialization! It's an ISO timestamp. Will try and fix.

alxdarksage: Added serializers that preserve timezone, changed tests to introduce a server time zone of MST, verifying either the time zone of the user is used, or else defaults to UTC.

DwayneJengSage (approved):

DwayneJengSage:
Changes look good. Feel free to merge whenever you're ready.

PR #1259, opened 2016-09-21
backfill for survey guid and createdOn in schemas

DwayneJengSage:
We want to tag all survey schemas with the survey guid and createdOn. This will at least allow us to identify the schema as a survey schema and link it back to the original survey. This is useful for the Study Manager UI.

Testing done:
- added unit tests for backfill
- manual tested backfill, 3 case: (1) survey without schema (2) survey schema already has fields (3) survey schema needs to be backfilled

See also https://sagebionetworks.jira.com/browse/BRIDGE-1297


alxdarksage (commented):
This looks good. Thinking out loud about why it should work:

Right now, every published survey (there can be multiple published versions with the same GUID) gets its schema using the survey identifier as the schema identifier, and the schema revision in that version of the survey. So published survey --> schema is covered.

And on the list of schemas, the most recent schema for a survey will be for the most recently published survey, so getting all of the most recently published surveys should work to update the auto-generated schemas for surveys on the schemas page.

PR #1264, opened 2016-09-30
re-enable New Relic

DwayneJengSage:
Restores New Relic.

Unfortunately, I have no way of testing it, as my personal heroku environment is unable to connect to Redis. (Following up with Redis support on this.) We'll just have to test this in devo.


alxdarksage:
Looks good to me. Merge when ready. And IMO merge however often you need to get it working, it may take a few tries.

PR #1265, opened 2016-09-30
add extra logging to AsyncBackfillTemplateTest

DwayneJengSage:
This test seems to fail a lot, often enough to be noticeable and force retries in Travis.

This change adds extra logging so that when it does fail, we'll have some idea of what went wrong.

### DIFF
--- a/test/org/sagebionetworks/bridge/services/backfill/AsyncBackfillTemplateTest.java
+++ b/test/org/sagebionetworks/bridge/services/backfill/AsyncBackfillTemplateTest.java
@@ -30,8 +30,11 @@ import org.sagebionetworks.bridge.models.backfill.BackfillTask;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class AsyncBackfillTemplateTest {
+    private static final Logger LOG = LoggerFactory.getLogger(AsyncBackfillTemplateTest.class);
 
     private static final ObjectMapper MAPPER = new ObjectMapper();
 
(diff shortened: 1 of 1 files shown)


test/org/sagebionetworks/bridge/services/backfill/AsyncBackfillTemplateTest.java
```diff
@@ -148,7 +151,15 @@ public boolean matches(Object item) {
                 long expireInMillis = TestBackfillService.EXPIRE * 1000L;
                 // Make sure the time point after which we look for the list of backfill tasks
                 // goes back for the duration of lock expiration
-                return beforeBackfill - expireInMillis < since && since < afterBackfill - expireInMillis;
+                long lower = beforeBackfill - expireInMillis;
+                long upper = afterBackfill - expireInMillis;
```
liujoshua, line 155: Could you explain why upper = afterBackfill - expireInMillis, and not upper = afterBackfill + expireInMillis?

DwayneJengSage: We call getTasks() with now - expireInMillis. The test doesn't mock out "now", so we don't know exactly when getTasks will be called.

However, we do keep track of when we started the backfill and when we finished, so we can bound now to between beforeBackfill and afterBackill. So we can bound when getTasks() to between (beforeBackfill - expireInMillis) and (afterBackfill - expireInMillis).

liujoshua:
Makes more sense. Why can't getTask() have happened between (afterBackfill
- expireInMillis, afterBackfill)? Could that happen if backfill completes
  in a shorter duration than expireInMillis?

DwayneJengSage:
I'm not sure I understand the question. getTasks() is definitely called sometime between beforeBackfill and afterBackfill. This means the timestamp we pass in is definitely between (beforeBackfill - expireInMillis) and (afterBackfill - expireInMillis).

liujoshua:
Looks good. I looked at the code under test and all makes sense

liujoshua (approved):

PR #1266, opened 2016-10-06
new api to update health record with exporter status

zaclinsage:
create new api to update health record by upsert synapse exporter status. create corresponding controller, api in service and related unit tests

### DIFF
--- a/app/org/sagebionetworks/bridge/models/healthdata/RecordExportStatusRequest.java
+++ b/app/org/sagebionetworks/bridge/models/healthdata/RecordExportStatusRequest.java
@@ -1,0 +1,24 @@
+package org.sagebionetworks.bridge.models.healthdata;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import org.sagebionetworks.bridge.json.JsonUtils;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+
+/**
+ * POJO to wrap synapse exporter status infomation, including a list of record ids and a exporter status
+ */
+public class RecordExportStatusRequest {
+
+    private @Nonnull List<String> recordIds;
+    private @Nonnull HealthDataRecord.ExporterStatus synapseExporterStatus;
+
+    public List<String> getRecordIds() { return this.recordIds; }
+
+    public HealthDataRecord.ExporterStatus getSynapseExporterStatus() { return this.synapseExporterStatus; }
+
+    public void setRecordIds(List<String> recordIds) { this.recordIds = recordIds; }
+
+    public void setSynapseExporterStatus(HealthDataRecord.ExporterStatus synapseExporterStatus) { this.synapseExporterStatus = synapseExporterStatus; }
+}
--- a/app/org/sagebionetworks/bridge/models/healthdata/RecordExportStatusRequest.java
+++ b/app/org/sagebionetworks/bridge/models/healthdata/RecordExportStatusRequest.java
@@ -1,15 +1,14 @@
 package org.sagebionetworks.bridge.models.healthdata;
 
-import com.fasterxml.jackson.databind.JsonNode;
-import org.sagebionetworks.bridge.json.JsonUtils;
+import org.sagebionetworks.bridge.models.BridgeEntity;
 
 import javax.annotation.Nonnull;
 import java.util.List;
 
 /**
- * POJO to wrap synapse exporter status infomation, including a list of record ids and a exporter status
+ * POJO to wrap synapse exporter status information, including a list of record ids and a exporter status
  */
-public class RecordExportStatusRequest {
+public class RecordExportStatusRequest implements BridgeEntity {
 
     private @Nonnull List<String> recordIds;
     private @Nonnull HealthDataRecord.ExporterStatus synapseExporterStatus;
--- a/app/org/sagebionetworks/bridge/play/controllers/HealthDataController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/HealthDataController.java
@@ -22,13 +22,13 @@ public class HealthDataController extends BaseController {
         this.healthDataService = healthDataService;
     }
 
-    public Result setRecordsStatus() throws JsonProcessingException{
+    public Result updateRecordsStatus() throws JsonProcessingException{
         UserSession session = getAuthenticatedSession(WORKER);
 
         RecordExportStatusRequest recordExportStatusRequest = parseJson(request(), RecordExportStatusRequest.class);
 
         List<String> updatedRecordIds = healthDataService.updateRecordsWithExporterStatus(recordExportStatusRequest);
 
-        return okResult("update exporter status to: " + updatedRecordIds + " complete!");
+        return okResult("Update exporter status to: " + updatedRecordIds + " complete.");
     }
 }
--- a/app/org/sagebionetworks/bridge/play/controllers/HealthDataController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/HealthDataController.java
@@ -1,0 +1,34 @@
+package org.sagebionetworks.bridge.play.controllers;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import org.sagebionetworks.bridge.models.accounts.UserSession;
+import org.sagebionetworks.bridge.models.healthdata.RecordExportStatusRequest;
+import org.sagebionetworks.bridge.services.HealthDataService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Controller;
+import play.mvc.Result;
+
+import java.util.List;
+
+import static org.sagebionetworks.bridge.Roles.WORKER;
+
+@Controller
+public class HealthDataController extends BaseController {
+
+    private HealthDataService healthDataService;
+
+    @Autowired
+    final void setHealthDataService(HealthDataService healthDataService) {
+        this.healthDataService = healthDataService;
+    }
+
+    public Result setRecordsStatus() throws JsonProcessingException{
+        UserSession session = getAuthenticatedSession(WORKER);
+
+        RecordExportStatusRequest recordExportStatusRequest = parseJson(request(), RecordExportStatusRequest.class);
+
+        List<String> updatedRecordIds = healthDataService.updateRecordsWithExporterStatus(recordExportStatusRequest);
+
+        return okResult("update exporter status to: " + updatedRecordIds + " complete!");
+    }
+}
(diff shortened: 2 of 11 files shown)


app/org/sagebionetworks/bridge/models/healthdata/RecordExportStatusRequest.java
```diff
@@ -0,0 +1,24 @@
+package org.sagebionetworks.bridge.models.healthdata;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import org.sagebionetworks.bridge.json.JsonUtils;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+
+/**
+ * POJO to wrap synapse exporter status infomation, including a list of record ids and a exporter status
```
alxdarksage: typo: "information"

app/org/sagebionetworks/bridge/play/controllers/HealthDataController.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    public Result setRecordsStatus() throws JsonProcessingException{
+        UserSession session = getAuthenticatedSession(WORKER);
+
+        RecordExportStatusRequest recordExportStatusRequest = parseJson(request(), RecordExportStatusRequest.class);
+
+        List<String> updatedRecordIds = healthDataService.updateRecordsWithExporterStatus(recordExportStatusRequest);
+
+        return okResult("update exporter status to: " + updatedRecordIds + " complete!");
```
alxdarksage: Minor: Capitalize "update", no exclamation mark (just a period).

app/org/sagebionetworks/bridge/play/controllers/HealthDataController.java
```diff
@@ -0,0 +1,34 @@
...
+public class HealthDataController extends BaseController {
+
+    private HealthDataService healthDataService;
+
+    @Autowired
+    final void setHealthDataService(HealthDataService healthDataService) {
+        this.healthDataService = healthDataService;
+    }
+
+    public Result setRecordsStatus() throws JsonProcessingException{
```
alxdarksage: name something that doesn't look like a setter (updateRecordStatuses, for example).

app/org/sagebionetworks/bridge/services/HealthDataService.java
```diff
@@ -156,4 +158,36 @@ public HealthDataAttachmentBuilder getAttachmentBuilder() {
...
+     *         POJO contains: a lit of health record ids, not upload ids and
+     *         an Synapse Exporter Status with value either NOT_EXPORTED or SUCCEEDED
+     * @return updated health record ids list
+     */
+    public List<String> updateRecordsWithExporterStatus(RecordExportStatusRequest recordExportStatusRequest) {
+        List<String> healthRecordIds = recordExportStatusRequest.getRecordIds();
+        HealthDataRecord.ExporterStatus synapseExporterStatus = recordExportStatusRequest.getSynapseExporterStatus();
+        if (recordExportStatusRequest.getRecordIds().isEmpty() || recordExportStatusRequest.getSynapseExporterStatus()== null) {
+            throw new InvalidEntityException(String.format(Validate.CANNOT_BE_BLANK, "healthRecordIds/synapseExporterStatus"));
+        }
```
alxdarksage: Assert that recordExportStatusRequest is not null. It'll fail immediately here but it's always helpful to be told in the error message what is wrong.

Check these fields separately and throw separate BadRequestExceptions, unless you are going to make this an entity. If you're going to make this an entity (and it probably should be, because it's in the API), you need to create a validator that throws InvalidEntityException with the proper field-level errors.

The error below about the maximum number of records would still be separate and would still be a BadRequestException, because the entity is valid but the call didn't meet the preconditions of the API.

app/org/sagebionetworks/bridge/services/HealthDataService.java
```diff
@@ -156,4 +158,36 @@ public HealthDataAttachmentBuilder getAttachmentBuilder() {
...
+
+        if (healthRecordIds.size() > MAX_NUM_RECORD_IDS) {
+            throw new InvalidEntityException("Size of the record ids list exceeds the limit.");
+        }
+
+        List<String> updatedRecordIds = new ArrayList<String>();
+
+        healthRecordIds.forEach(id->{
```
alxdarksage: .map().collect(toList()) would be more concise here

conf/routes
```diff
@@ -136,6 +136,9 @@ GET    /v3/scheduleplans/:guid     @org.sagebionetworks.bridge.play.controllers.
 POST   /v3/scheduleplans/:guid     @org.sagebionetworks.bridge.play.controllers.SchedulePlanController.updateSchedulePlan(guid: String)
 DELETE /v3/scheduleplans/:guid     @org.sagebionetworks.bridge.play.controllers.SchedulePlanController.deleteSchedulePlan(guid: String)
 
+# Bridge Exporter
+POST /v3/recordExportStatuses      @org.sagebionetworks.bridge.play.controllers.HealthDataController.setRecordsStatus
```
alxdarksage: Please follow the convention for upload statuses and use POST /v3/exportstatuses (all lowercase)

test/org/sagebionetworks/bridge/play/controllers/HealthDataControllerTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+        // spy controller
+        HealthDataController controller = spy(new HealthDataController());
+        controller.setHealthDataService(healthDataService);
+        doReturn(mockSession).when(controller).getAuthenticatedSession(anyVararg());
+
+        // execute and validate
+        Result result = controller.setRecordsStatus();
+        assertEquals(result.status(), 200);
+    }
```
alxdarksage, line 81: validate that this controller called the mock service and passed a correctly deserialized payload. The deserialization of the export status object should be tested separately as well to make sure that all works.

app/org/sagebionetworks/bridge/models/healthdata/RecordExportStatusRequest.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+public class RecordExportStatusRequest {
```
alxdarksage: Shouldn't have verbs in entity names or any references to HTTP (request/response). So this is ExportStatuses, ExportJobStatus, or something similar.

test/org/sagebionetworks/bridge/services/HealthDataServiceTest.java
```diff
@@ -123,4 +140,43 @@ public void getRecordsForUploadDateSuccess() {
...
+        // mock dao
+        HealthDataDao mockDao = mock(HealthDataDao.class);
+        when(mockDao.createOrUpdateRecord(record)).thenReturn(TEST_RECORD_ID);
+
+        HealthDataService svc = new HealthDataService();
+        svc.setHealthDataDao(mockDao);
+
+        // execute and validate
+        String retVal = svc.createOrUpdateRecord(record);
+        assertEquals(TEST_RECORD_ID, retVal);
```
alxdarksage, line 194: You are only validating that you mocked this, you can remove.

test/org/sagebionetworks/bridge/services/HealthDataServiceTest.java
```diff
@@ -123,4 +140,43 @@ public void getRecordsForUploadDateSuccess() {
...
+
+        // then create a mock json request
+        RecordExportStatusRequest recordExportStatusRequest = createMockRecordExportStatusRequest();
+        when(mockDao.getRecordById(TEST_RECORD_ID)).thenReturn(record);
+
+        // finally call service method and assert
+        svc.updateRecordsWithExporterStatus(recordExportStatusRequest);
+        HealthDataRecord recordAfter = svc.getRecordById(TEST_RECORD_ID);
+        assertNotNull(recordAfter.getSynapseExporterStatus());
+        assertEquals(HealthDataRecord.ExporterStatus.SUCCEEDED, recordAfter.getSynapseExporterStatus());
```
alxdarksage, line 206: I don't think this is verifying correctly. You just mocked this return value you are verifying, which only tells you that your mock worked. You want to verify() that the DAO was called with the correct object, that is, that this service performed the right call through to the DAO.

Also, please write tests to validate that all the error conditions you are checking for in the service are correctly being validated, this is the key work it is doing: validating all the inputs and communicating errors back to the user that allow them to fix the problem.

test/org/sagebionetworks/bridge/services/HealthDataServiceTest.java
```diff
@@ -123,4 +140,43 @@ public void getRecordsForUploadDateSuccess() {
...
+
+    }
+
+    private RecordExportStatusRequest createMockRecordExportStatusRequest() throws Exception {
+        RecordExportStatusRequest request = new RecordExportStatusRequest();
+        request.setRecordIds(Arrays.asList(TEST_RECORD_ID));
```
alxdarksage: Might consider putting two records in here just to make sure any iteration that needs to occur is occurring (true for the controller test as well).

alxdarksage (commented):
Comments for some updates. Thanks!

app/org/sagebionetworks/bridge/models/healthdata/RecordExportStatusRequest.java
DwayneJengSage: Following up with Alx offline about this. It looks like there's no agreed-upon naming convention, particularly for verb-like operations. The name is fine for now, and is descriptive enough that there won't be any confusion, and can be easily changed in the future.

conf/routes
DwayneJengSage: Similarly, (other than capitalization), this name is fine as is for now, and can be easily changed in the future if it becomes a problem.

test/org/sagebionetworks/bridge/services/HealthDataServiceTest.java
DwayneJengSage: Following up with Alx offline, but this is not only fine to keep as is, but it's an important part of testing.

Part of the implementation is passing data from the DAO to the output of the server. It's important to test this code, to ensure you're not returning the wrong object.

test/org/sagebionetworks/bridge/services/HealthDataServiceTest.java
DwayneJengSage: This is a little more complicated. Yes, it's important to test that the Service is correctly setting the ExportStatus on the record. However, it looks like the implementation just sets the status on the record and throws it away without saving it to DDB.

It also demonstrates the importance of manual tests. (Or integration tests if you're modifying an existing API, but that comes later.) Your unit tests and mocks make certain assumptions about how a system works. However, the real system might work under different assumptions. As such, it's important to test things manually, at a high level, just to make sure everything works they way you expected them to.

app/org/sagebionetworks/bridge/models/healthdata/RecordExportStatusRequest.java
zaclinsage: yes, it is a 'request' in noun about health record's exporter status

PR #1275, opened 2016-11-08
This works, going to look at setting events to always use UTC.

alxdarksage:
This might originally have been prevented by putting all event timestamps in UTC, however, it turns out to be very challenging to keep DateTime in UTC as every time you transform it, you have to reset it (at least in testing). In any event, now that there are duplicate one-time tasks in the database, this alone will not fix the issue.

So the strategy here is to do the following:

1) From now on, when creating one-time tasks, set their scheduledOn time portion to midnight of the day of the event, not whatever the time portion is of the event timestamp;

2) During scheduling, fix these persisted one-time tasks to use the same midnight time value, which fixes the GUIDs of the scheduled activities as well, and then makes it possible to identify and remove duplicates.

3) People rarely (never?) set an expiration for one-time tasks but if they do, enforce it being at least 24 hours in the future because otherwise, the expiration will be confusing.


alxdarksage:
The consequences of this approach, which removes duplicates, seeps into persistent tasks, into one-time tasks scheduled against events other than enrollment, etc.

May try instead to ensure that the event timestamps are in UTC, or local time if the times are specified. This still has issues for "travel around the world" users but that already exists. This should be more robust.

PR #1276, opened 2016-11-09
add CompoundActivity to schedule models

DwayneJengSage:
Adds CompoundActivity to schedule models. This new type of activity holds a list of schema references and survey references. This allows the client to (a) know what schemas correspond to an activity and (b) allow a single task identifier to correspond to a list of schemas and surveys.

This is just for models and hooking up to the schedules, plans, and activity APIs. Next steps:
* Add CompoundActivityDefinitions, so study developers don't need to specify the same list of 4 schemas multiple times.
* Scheduler to fill in schema revs and survey createdOn at activity request time.

Testing done:
* relevant unit tests for activities and activity validator
* integ tests for ScheduledActivityTest, SchedulePlanTest, ScheduleTest
* manual tests

Tested with this schedule:
{
  "label":"Schedule w/ Compound Activity",
  "strategy":{
    "type":"SimpleScheduleStrategy",
    "schedule":{
      "scheduleType":"ONCE",
      "activities":[{
        "label":"Test Compound Activity",
        "compoundActivity":{
          "taskIdentifier":"task:CCC",
          "schemaList":[
            {"id":"schema-1"},
            {"id":"schema-2"}
          ],
          "surveyList":[
            {"guid":"survey-guid-A"},
            {"guid":"survey-guid-B"}
          ]
        }
      }]
    }
  }
}

Tested the following operations:
POST /v3/scheduleplans
GET /v3/scheduleplans
GET /v4/schedules
GET /v3/activities?offset=-8&daysAhead=3


liujoshua (approved):

app/org/sagebionetworks/bridge/models/schedules/CompoundActivity.java
```diff
@@ -0,0 +1,118 @@
...
+    private final List<SchemaReference> schemaList;
+    private final List<SurveyReference> surveyList;
+    private final String taskIdentifier;
+
+    /** Private constructor, use builder to construct. */
+    private CompoundActivity(List<SchemaReference> schemaList, List<SurveyReference> surveyList,
+            String taskIdentifier) {
+        this.schemaList = schemaList;
+        this.surveyList = surveyList;
+        this.taskIdentifier = taskIdentifier;
```
alxdarksage, line 23: Just a note that I don't think there would be a case where there was a surveyList AND a taskIdentifier (but if there's a taskIdentifier there should probably be at least one schema in the list), but that's just something to check when validating I think.

PR #1277, opened 2016-11-10
Fix onetime task scheduling

alxdarksage:
Let's try again, this change (task == scheduled activity):

- enforces all event tasks to be in UTC time zone (they should be already);
- carries over schedules in scheduled tasks in the scheduler, to infer after-the-fact whether or not tasks are derived from non-repeating schedules (once, persistent);
- then scheduling without converting the event-based timestamps to the user's time zone. This is only done by binding late when creating a task to return to the user (this is probably the source of duplication when the time zone changed);
- instead of non-recurring tasks using the event timestamp's time portion, use midnight, when scheduling, and when loading persisted tasks;
- that in turn allows us to de-duplicate existing duplicated tasks (non-recurring tasks);

Added significant new tests to verify behavior, in particular ScheduledActivityServiceDuplicateTest, which verifies all this behavior.

The key is that scheduling occurs against a time in UTC, which is invariant, but the tasks are scheduled using a non-specific time of day, and joined late with the user's requesting time zone. The only place where the user's time zone is relevant to scheduling, is in the scheduling window and endsOn values, that is, whether given the user's requesting time zone, a task should be seen or not.

**On the subject of changing timezones:** we calculate these once/persistent tasks when the user asks for them in a given time zone, so setting it to midnight on their day always works. Yes, if they then fly to the prior day,the task will disappear until the next day, and that's okay. If they fly forward to the next day, a once/persistent task will still be there, so that also is fine. If it's a task that is expired in short order like a day, that might cause an issue, but I've actually never seen anyone create such a schedule and I think it's an acceptable risk.

Friday afternoon: merged recent changes from Dwayne, tests pass.


test/org/sagebionetworks/bridge/dynamodb/DynamoActivityEventDaoTest.java
```diff
@@ -53,27 +54,27 @@ public void canCrudEvent() {
...
-        assertEquals(time1.minusMonths(2), map.get("two_months_before_enrollment"));
-        assertEquals(time2, map.get("survey:AAA-BBB-CCC:finished"));
-        assertEquals(time3, map.get("question:DDD-EEE-FFF:answered=someValue"));
-        assertEquals(time6, map.get("activity:AAA-BBB-CCC:finished"));
+        assertEquals(time1.withZone(DateTimeZone.UTC), map.get("enrollment"));
+        assertEquals(time1.withZone(DateTimeZone.UTC).minusWeeks(2), map.get("two_weeks_before_enrollment"));
+        assertEquals(time1.withZone(DateTimeZone.UTC).minusMonths(2), map.get("two_months_before_enrollment"));
+        assertEquals(time2.withZone(DateTimeZone.UTC), map.get("survey:AAA-BBB-CCC:finished"));
+        assertEquals(time3.withZone(DateTimeZone.UTC), map.get("question:DDD-EEE-FFF:answered=someValue"));
+        assertEquals(time6.withZone(DateTimeZone.UTC), map.get("activity:AAA-BBB-CCC:finished"));
```
DwayneJengSage, line 64: Can we define the times in a timezone other than UTC, so we can test we're actually forcing the times into UTC?

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -42,21 +41,36 @@ protected DateTime getScheduledTimeBasedOnEvent(ScheduleContext context) {
     protected void addScheduledActivityForAllTimes(List<ScheduledActivity> scheduledActivities,
```
DwayneJengSage: Maybe in order to make this easier to reason about, we should have this method take in a LocalDate instead of a DateTime. For schedules without times, we can add midnight to make it a LocalDateTime. For schedules with, we use those times to make LocalDateTimes. Since tasks are scheduled using LocalDateTimes, this method and tasks work the same way, and we don't have to deal with timezones at all under this call chain.

Conceptually, this means that Interval and Cron scheduler can only be used to pick the day of the activity (which de facto true once we make this change), and in order to select the specific times of day, you need to specify the list of times.

This should be stable as long as the LocalDates we're passing in are stable. Which should be true if all events are stored in UTC and handled in UTC, which is true after this change.

There are only 2 uses of add...ForAllTimes() and only 3 uses of add...ForTime(), so it shouldn't be too time-consuming to refactor.

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -42,21 +41,36 @@ protected DateTime getScheduledTimeBasedOnEvent(ScheduleContext context) {
...
+        // Normalize to UTC minus one day from enrollment, to ensure no matter where user is at the time of request,
+        // the one-time activity is available.
+        if (Schedule.isScheduleWithoutTimes(schedule)) {
+            return Schedule.eventToMidnight(scheduledTime);
+        }
+        return scheduledTime;
+    }
+    
     protected void addScheduledActivityAtTime(List<ScheduledActivity> scheduledActivities, SchedulePlan plan,
             ScheduleContext context, DateTime scheduledTime) {
```
DwayneJengSage: Likewise add...AtTime() should take a LocalDateTime, since we ignore its timezone anyway.

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -115,7 +130,7 @@ protected DateTime getFirstEventDateTime(ScheduleContext context, String eventId
             String[] eventIds = eventIdsString.trim().split("\\s*,\\s*");
             for (String thisEventId : eventIds) {
                 if (context.getEvent(thisEventId) != null) {
-                    eventDateTime = context.getEvent(thisEventId).withZone(context.getZone());
+                    eventDateTime = context.getEvent(thisEventId);
```
DwayneJengSage, line 124: Are we guaranteed that events in the context are always in UTC? Do we want to force UTC here as well?

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -127,8 +142,13 @@ protected DateTime getFirstEventDateTime(ScheduleContext context, String eventId
...
      */
-    protected boolean shouldContinueScheduling(ScheduleContext context, DateTime scheduledTime, List<ScheduledActivity> scheduledActivities) {
-        boolean boundaryNotMet = scheduledTime.isBefore(context.getEndsOn()) || hasNotMetMinimumCount(context, scheduledActivities.size());
+    protected boolean shouldContinueScheduling(ScheduleContext context, DateTime scheduledTime,
+            List<ScheduledActivity> scheduledActivities) {
+        DateTime scheduledTimeInZone = scheduledTime.withZone(context.getZone());
+        DateTime endsOnInZone = context.getEndsOn().withZone(context.getZone());
+        
+        boolean boundaryNotMet = scheduledTimeInZone.isBefore(endsOnInZone) || 
+                hasNotMetMinimumCount(context, scheduledActivities.size());
```
DwayneJengSage: Given that our timezones are using offsets and not names like "PST", I believe this change is actually not necessary. Two datetimes will be in the same relative order no matter what offset your convert them to, so forcing them into the user's local timezone doesn't actually change anything.

test/org/sagebionetworks/bridge/models/schedules/ScheduleTest.java
```diff
@@ -159,4 +159,30 @@ public void scheduleWithDelayNotPersistent() {
...
+        assertTrue(Schedule.isScheduleWithoutTimes(schedule));
+    }
+
+    @Test
+    public void eventToPriorUTCMidnight() {
+        DateTime dateTime = DateTime.parse("2016-11-06T04:32.123-07:00");
+        DateTime midnight = Schedule.eventToMidnight(dateTime);
+        
+        assertEquals("2016-11-06T00:00:00.000Z", midnight.toString());
+    }
```
DwayneJengSage: Is it worth adding test cases that cross the date (late night with a negative offset that's tomorrow UTC, early morning with a positive offset that's yesterday UTC) to make sure we fully understand this behavior?

app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
```diff
@@ -195,7 +195,7 @@ public Builder withContext(ScheduleContext context) {
             this.events = context.events;
             this.now = context.now;
             this.minimumPerSchedule = context.minimumPerSchedule;
-            this.accountCreatedOn = context.accountCreatedOn;
+            this.accountCreatedOn = new DateTime(context.accountCreatedOn, DateTimeZone.UTC);
```
DwayneJengSage: Add a unit test to verify we're forcing UTC?

test/org/sagebionetworks/bridge/validators/ScheduleValidatorTest.java
```diff
@@ -31,6 +31,21 @@ public void before() {
...
+        schedule.addActivity(TestConstants.TEST_3_ACTIVITY);
+        try {
+            Validate.entityThrowingException(validator, schedule);
+            fail("Should have thrown InvalidEntityException");
+        } catch(InvalidEntityException e) {
+            assertEquals(
+                    "scheduleType set to once with no time of day start, but expires in under a day (this can cause confusing behavior)",
+                    e.getErrors().get("scheduleType").get(0));
+        }
+    }
```
DwayneJengSage, line 46: Would it be worthwhile to have two tests here, one with an expires < 1 day, one with expires > 1 day, so we can verify both cases?

test/org/sagebionetworks/bridge/models/schedules/ScheduleTestUtils.java
```diff
@@ -35,7 +36,8 @@ public static long asLong(String string) {
     public static void assertDates(List<ScheduledActivity> activities, String... output) {
         assertEquals(output.length, activities.size());
         for (int i=0; i < activities.size(); i++) {
-            assertEquals(asLong(output[i]), activities.get(i).getScheduledOn().getMillis());
+            assertEquals(new DateTime(asLong(output[i]), DateTimeZone.UTC), activities.get(i).getScheduledOn());
+            //assertEquals(asLong(output[i]), activities.get(i).getScheduledOn().getMillis());
```
DwayneJengSage: Nitpick: Remove commented assert.

DwayneJengSage (changes requested):
The general approach is good. I think we can tighten up on addScheduledActivity...() to make it easier to reason with.

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
alxdarksage: I've refactored this so the method takes LocalDate and LocalTime. it's clearer now that in all cases where times have not been set, we're using MIDNIGHT.

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
alxdarksage: Yes, refactored

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
alxdarksage: Yes, because the events come from the DAO getActivityEventMap() method and we set it to UTC there in case the server time zone isn't UTC.

test/org/sagebionetworks/bridge/models/schedules/ScheduleTest.java
alxdarksage: I've added a case the next day in Moscow, it verifies that the time is normalized to a common day so if someone moves, their one-time/persistent task won't double up.

test/org/sagebionetworks/bridge/validators/ScheduleValidatorTest.java
alxdarksage: Added, uncovered a defect, fixed ScheduleContext.

test/org/sagebionetworks/bridge/dynamodb/DynamoActivityEventDaoTest.java
alxdarksage: I changed the initial times to Moscow time, and these assertions still pass (times converted to UTC).

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
alxdarksage: This is now tested by DynamoActivityEventDaoTest, I believe.

alxdarksage:
Okay, I've updated.

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -127,8 +128,12 @@ protected DateTime getFirstEventDateTime(ScheduleContext context, String eventId
      * If scheduling hasn't reached the end time, or hasn't accumulated the minimum number of tasks, returns true, or 
      * false otherwise. 
      */
-    protected boolean shouldContinueScheduling(ScheduleContext context, DateTime scheduledTime, List<ScheduledActivity> scheduledActivities) {
-        boolean boundaryNotMet = scheduledTime.isBefore(context.getEndsOn()) || hasNotMetMinimumCount(context, scheduledActivities.size());
+    protected boolean shouldContinueScheduling(ScheduleContext context, DateTime scheduledTime,
+            List<ScheduledActivity> scheduledActivities) {
+        
+        boolean boundaryNotMet = scheduledTime.withZoneRetainFields(context.getZone()).isBefore(context.getEndsOn()) ||
```
DwayneJengSage: Why are we shifting the timezone of scheduledTime? withZoneRetainFields() sets the timezone without changing the time, so this shifts the timezone by 5-8 hours (assuming continental US). This leads to weird off-by-one errors in up to 1/3rd of cases.

If the idea is that interval and cron scheduling should only be used for dates but not times, we could force both DateTimes to the same arbitrary timezone, convert to LocalDates, and compare dates. If this is the case, it might be worthwhile to push this change up to IntervalScheduler and CronScheduler (the only places that use them) to make this method easier to reason about. LocalDate supports plus(Period), so this works trivially in IntervalScheduler. CronScheduler needs converting anyway.

test/org/sagebionetworks/bridge/json/DateUtilsTest.java
```diff
@@ -212,4 +212,20 @@ public void getDateTimeOrDefaultThrowsException() {
...
+    // or maybe the day before... this won't matter for one-time task or persistent 
+    // tasks unless they expire, but we prevent a short expiration window). Then if 
+    // the user moves across the datetime boundary, they won't get another task.
+    @Test
+    public void eventToPriorUTCMidnight() {
+        DateTime dateTime = DateTime.parse("2016-11-06T16:32.123-07:00");
+        DateTime midnight = DateUtils.dateTimeToMidnightUTC(dateTime);
+        assertEquals("2016-11-06T00:00:00.000Z", midnight.toString());
+        
+        // Time zone not an issue here, it will normalize to UTC
```
DwayneJengSage: Nitpick: This comment seems to apply to the _other_ case.

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -88,10 +86,12 @@ protected void addScheduledActivityAtTime(List<ScheduledActivity> scheduledActiv
         return scheduledActivities.subList(0, Math.min(scheduledActivities.size(), count));
     }
     
-    private boolean isInWindow(DateTime scheduledTime) {
+    private boolean isInWindow(LocalDate localDate, LocalTime localTime) {
+        DateTime scheduledTime = localDate.toDateTime(localTime).withZoneRetainFields(DateTimeZone.UTC);
```
DwayneJengSage: I think localDate.toDateTime(localTime, DateTimeZone.UTC) is what you want here.

Also, is UTC the right timezone to use here? This represents the time the task will be scheduled for the user, so it should use the context's local time zone.

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -102,10 +102,11 @@ private boolean isBeforeWindowEnd(DateTime scheduledTime) {
         return (endsOn == null || scheduledTime.isEqual(endsOn) || scheduledTime.isBefore(endsOn));
     }
     
-    private DateTime getExpiresOn(DateTime scheduledTime) {
+    private DateTime getExpiresOn(LocalDate localDate, LocalTime localTime) {
         if (schedule.getExpires() == null) {
             return null;
         }
+        DateTime scheduledTime = localDate.toDateTime(localTime).withZoneRetainFields(DateTimeZone.UTC);
```
DwayneJengSage: Same here.

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -39,39 +44,32 @@ protected DateTime getScheduledTimeBasedOnEvent(ScheduleContext context) {
...
             if (expiresOn == null || expiresOn.isAfter(context.getNow())) {
                 for (Activity activity : schedule.getActivities()) {
                     ScheduledActivity schActivity = ScheduledActivity.create();
                     schActivity.setSchedulePlanGuid(plan.getGuid());
                     schActivity.setTimeZone(context.getZone());
                     schActivity.setHealthCode(context.getCriteriaContext().getHealthCode());
                     schActivity.setActivity(activity);
-                    schActivity.setScheduledOn(scheduledTime);
-                    schActivity.setGuid(activity.getGuid() + ":" + scheduledTime.toLocalDateTime().toString());
+                    schActivity.setScheduledOn(localDate.toDateTime(localTime, context.getZone()));
```
DwayneJengSage: Instead of converting localDate and localTime to a dateTime and then immediately throwing away the timezone, it would make more sense for ScheduledActivity.setScheduledOn() to take in a LocalDateTime. If it's too much work to refactor all uses of setScheduledOn(), we could create an overload that takes LocalDateTime.

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -39,39 +44,32 @@ protected DateTime getScheduledTimeBasedOnEvent(ScheduleContext context) {
...
-            "Scheduled DateTime does not have requested time zone: " + scheduledTime.getZone());
-
-        // If this time point is outside of the schedule's active window, skip it.
-        if (isInWindow(scheduledTime)) {
+            ScheduleContext context, LocalDate localDate, LocalTime localTime) {
+        
+        if (isInWindow(localDate, localTime)) {
             // As long at the activities are not already expired, add them.
-            DateTime expiresOn = getExpiresOn(scheduledTime);
+            DateTime expiresOn = getExpiresOn(localDate, localTime);
```
DwayneJengSage: Similarly, expiresOn is stored in ScheduledActivity as a LocalDateTime. Does it make sense to force this to a DateTime with an arbitrary timezone, only to immediately discard it?

app/org/sagebionetworks/bridge/models/schedules/CronActivityScheduler.java
```diff
@@ -25,10 +26,10 @@
             
             while (shouldContinueScheduling(context, scheduledTime, scheduledActivities)) {
                 Date next = trigger.getFireTimeAfter(scheduledTime.toDate());
-                scheduledTime = new DateTime(next, context.getZone());
+                scheduledTime = new DateTime(next, DateTimeZone.UTC);
                 
                 if (shouldContinueScheduling(context, scheduledTime, scheduledActivities)) {
-                    addScheduledActivityForAllTimes(scheduledActivities, plan, context, scheduledTime);    
+                    addScheduledActivityAtTime(scheduledActivities, plan, context, scheduledTime.toLocalDate(), scheduledTime.toLocalTime());
```
DwayneJengSage: Do cron schedules not have times? If they don't, wouldn't it make sense to schedule it at midnight to be consistent with the other schedulers?

DwayneJengSage (changes requested):

DwayneJengSage:
Also, in SchedulerValidatorTest, would it be worthwhile to have two tests here, one with an expires < 1 day, one with expires > 1 day, so we can verify both cases?

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
alxdarksage: This is a good observation and as a refactor, fairly straightforward to do, and the results are better. But it's going to add a lot of changes on top of what's already here. I'm inclined to file a JIRA and deal with it after this is rolled out.

alxdarksage: Let's discuss because you're the one who ultimately has to review it.

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
alxdarksage: Yes, this works.

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
alxdarksage: Also changed, also works

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
alxdarksage: Changed this. One test broke, but it was added as part of this check in and the results look correct to me (the test was wrong).

test/org/sagebionetworks/bridge/json/DateUtilsTest.java
alxdarksage: removed

app/org/sagebionetworks/bridge/models/schedules/CronActivityScheduler.java
alxdarksage: They do have times, so the time portion is carried over (only one per calculation of the cron expression). Cron is in UTC which is why it's fundamentally useless for scheduling, but when it was used by YMedia, the time portion was treated as a local time, for example a "10" for "ten in the morning" though obviously not in UTC time. So we calculate in UTC, and then switch to local time zone, preserving the time portion. This works well enough and is the existing behavior.

alxdarksage:
Added positive case validator test as well.

I also switched over to just setting localScheduledOn and localExpiresOn.

DwayneJengSage:
Overall, this looks good.

Just one small comment about ActivityScheduler.isInWindow(): Is UTC the right timezone for this? This represents the time the task will be scheduled for the user, so it seems like it should use the context's local time zone. Otherwise, we may have tasks ending prematurely and/or tasks going past the scheduled end time.

ALX: I've changed this so it is now doing the correct thing. In fixing this, it breaks three tests that existed before the refactoring. I've gone through those tests and believe they are adjusted and correct (and passing). In the course of doing this, I rewrote ActivityScheduler because it was getting confusing (too many time-testing methods). There are now two: shouldContinueScheduling and isValidForScheduling.

app/org/sagebionetworks/bridge/json/DateUtils.java
```diff
@@ -214,4 +215,12 @@ public static DateTime getDateTimeOrDefault(String value, DateTime defaultValue)
...
     }
     
+    /**
+     * Convert DateTime instance to midnight UTC.
+     * @param dateTime
+     * @return dateTime at midnight in UTC
+     */
+    public static DateTime dateTimeToMidnightUTC(DateTime dateTime) {
+        return new DateTime(dateTime, DateTimeZone.UTC).withTime(LocalTime.MIDNIGHT);
+    }
```
alxdarksage: This is not used anymore and could be removed, though not harmful it's in this PR.

app/org/sagebionetworks/bridge/models/schedules/IntervalActivityScheduler.java
```diff
@@ -21,14 +22,17 @@
...
-                addScheduledActivityForAllTimes(scheduledActivities, plan, context, datetime);
+                LocalTime localTime = addScheduledActivityForAllTimes(scheduledActivities, plan, context, datetime.toLocalDate());
                 // A one-time activity with no interval (for example); don't loop
                 if (schedule.getInterval() == null) {
                     return trimScheduledActivities(scheduledActivities);
                 }
-                datetime = datetime.plus(schedule.getInterval());
+                // We reset the time portion of the timestamp to the last time that was used to schedule.
+                // Otherwise to-the-minute resolution of the start/end time boundaries won't be correct.
+                datetime = datetime.withTime(localTime).plus(schedule.getInterval());
```
DwayneJengSage, line 35: Is there a way to test this change?

DwayneJengSage (changes requested):
Also, for ActivityScheduler.isInWindow(): Is UTC the right timezone for this? This represents the time the task will be scheduled for the user, so it seems like it should use the context's local time zone. Otherwise, we may have tasks ending prematurely and/or tasks going past the scheduled end time.

ALX: Yes. We explain this window as a window in UTC that applies to everyone, everywhere. It would be used to implement something like the end of a study, or Eric has used it I think in order to create weekly schedules. It's one time for everyone regardless of where they are.

ALX: Can't find your comment but yes, we're comparing UTC time to a local time that's shifted in UTC and that's an error. I could fix this but it will break 3 existing tests. I don't think it'll break anything in production, so I'll change it.

app/org/sagebionetworks/bridge/models/schedules/IntervalActivityScheduler.java
alxdarksage: Yes, there are specific tests that fail if you don't do this (the ScheduledActivityServiceRecurringTest returns the wrong number of tasks). I can show why it occurs though I'm not sure why it was masked until these other changes were made.

DwayneJengSage:
As per offline conversation, it seems like we're finding and pulling in a bunch of unrelated bug fixes into this change. This is bloating the change and causing this code review to take much longer than expected. As per conversation, we should roll back to an earlier change and work from there. https://github.com/Sage-Bionetworks/BridgePF/pull/1277/commits/65e1ce569f180074f2fdf5e1564f1d3f64bab1b1 is the change that came closest to passing code review, with the only outstanding code review comment:

Just one small comment about ActivityScheduler.isInWindow(): Is UTC the right timezone for this? This represents the time the task will be scheduled for the user, so it seems like it should use the context's local time zone. Otherwise, we may have tasks ending prematurely and/or tasks going past the scheduled end time.

My gut says there are also other functional changes at that point in the pull request that should be removed and factored into a separate JIRA/bugfix, to prevent further bloat.

DwayneJengSage:
Replaced by https://github.com/Sage-Bionetworks/BridgePF/pull/1290

PR #1286, opened 2016-11-15
deal with dedupe use case 2: log duplicate uploads for same healthcod…

zaclinsage:
…e, createdOn and schemaId

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDao.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDao.java
@@ -2,6 +2,8 @@ package org.sagebionetworks.bridge.dynamodb;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Resource;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -13,6 +15,7 @@ import com.amazonaws.services.dynamodbv2.model.Condition;
 import org.joda.time.DateTime;
 import org.sagebionetworks.bridge.BridgeUtils;
 import org.sagebionetworks.bridge.dao.HealthDataDao;
+import org.sagebionetworks.bridge.exceptions.BridgeServiceException;
 import org.sagebionetworks.bridge.models.healthdata.HealthDataRecord;
 import org.sagebionetworks.bridge.models.healthdata.HealthDataRecordBuilder;
 
@@ -23,9 +26,12 @@ import org.springframework.stereotype.Component;
 /** DynamoDB implementation of {@link org.sagebionetworks.bridge.dao.HealthDataDao}. */
 @Component
 public class DynamoHealthDataDao implements HealthDataDao {
+    private static final long CREATED_ON_OFFSET_MILIS = 100;
+
     private DynamoDBMapper mapper;
     private DynamoIndexHelper healthCodeIndex;
     private DynamoIndexHelper uploadDateIndex;
+    private DynamoIndexHelper healthCodeCreatedOnIndex;
 
     /** DynamoDB mapper for the HealthDataRecord table. This is configured by Spring. */
     @Resource(name = "healthDataDdbMapper")
@@ -50,6 +56,16 @@ public class DynamoHealthDataDao implements HealthDataDao {
     public void setUploadDateIndex(DynamoIndexHelper uploadDateIndex) {
         this.uploadDateIndex = uploadDateIndex;
     }
+
+    /**
+     * DynamoDB Index reference for the healthCodeCreatedOnIndex index. This is needed because the DynamoDB mapper does allow queries
+     * using global secondary indices. This is configured by Spring
+     */
+    @Resource(name = "healthDataHealthCodeCreatedOnIndex")
+    public void setHealthCodeCreatedOnIndex(DynamoIndexHelper healthCodeCreatedOnIndex) {
+        this.healthCodeCreatedOnIndex = healthCodeCreatedOnIndex;
+    }
+
 
     /** {@inheritDoc} */
     @Override
@@ -102,11 +118,33 @@ public class DynamoHealthDataDao implements HealthDataDao {
     /** {@inheritDoc} */
     @Override
     public List<HealthDataRecord> getRecordsByHealthCodeCreatedOnSchemaId(@Nonnull String healthCode, @Nonnull Long createdOn, @Nonnull String schemaId) {
-        List<HealthDataRecord> recordsByHealthCode = healthCodeIndex.queryKeys(HealthDataRecord.class, "healthCode",
-                healthCode, null);
+        Condition rangeKeyCondition = new Condition()
+                .withComparisonOperator(ComparisonOperator.BETWEEN)
+                .withAttributeValueList(new AttributeValue().withN(String.valueOf((createdOn - CREATED_ON_OFFSET_MILIS)))
+                        , new AttributeValue().withN(String.valueOf((createdOn + CREATED_ON_OFFSET_MILIS))));
 
-        return recordsByHealthCode.stream()
-                .filter(record -> record.getCreatedOn().equals(createdOn) && record.getSchemaId().equals(schemaId))
-                .collect(Collectors.toList());
+        DynamoHealthDataRecord queryRecord = new DynamoHealthDataRecord();
+        queryRecord.setHealthCode(healthCode);
+
+        DynamoDBQueryExpression<DynamoHealthDataRecord> expression = new DynamoDBQueryExpression<DynamoHealthDataRecord>()
+                .withConsistentRead(false)
+                .withHashKeyValues(queryRecord)
+                .withRangeKeyCondition("createdOn", rangeKeyCondition);
+
+        List<DynamoHealthDataRecord> recordList = mapper.query(DynamoHealthDataRecord.class, expression);
+
+        List<HealthDataRecord> finalRecords = new ArrayList<>();
+        for (Object oneResult : recordList) {
+            if (!DynamoHealthDataRecord.class.isInstance(oneResult)) {
+                // This should never happen, but just in case.
+                throw new BridgeServiceException(String.format(
+                        "DynamoDB returned objects of type %s instead of %s",
+                        oneResult.getClass().getName(), DynamoHealthDataRecord.class.getName()));
+            }
+
+            finalRecords.add((HealthDataRecord) oneResult);
+        }
+
+        return finalRecords;
     }
 }
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDao.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDao.java
@@ -3,7 +3,14 @@ package org.sagebionetworks.bridge.dynamodb;
 import javax.annotation.Nonnull;
 import javax.annotation.Resource;
 import java.util.List;
+import java.util.stream.Collectors;
 
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBQueryExpression;
+import com.amazonaws.services.dynamodbv2.document.RangeKeyCondition;
+import com.amazonaws.services.dynamodbv2.model.AttributeValue;
+import com.amazonaws.services.dynamodbv2.model.ComparisonOperator;
+import com.amazonaws.services.dynamodbv2.model.Condition;
+import org.joda.time.DateTime;
 import org.sagebionetworks.bridge.BridgeUtils;
 import org.sagebionetworks.bridge.dao.HealthDataDao;
 import org.sagebionetworks.bridge.models.healthdata.HealthDataRecord;
@@ -90,5 +97,16 @@ public class DynamoHealthDataDao implements HealthDataDao {
     @Override
     public HealthDataRecordBuilder getRecordBuilder() {
         return new DynamoHealthDataRecord.Builder();
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public List<HealthDataRecord> getRecordsByHealthCodeCreatedOnSchemaId(@Nonnull String healthCode, @Nonnull Long createdOn, @Nonnull String schemaId) {
+        List<HealthDataRecord> recordsByHealthCode = healthCodeIndex.queryKeys(HealthDataRecord.class, "healthCode",
+                healthCode, null);
+
+        return recordsByHealthCode.stream()
+                .filter(record -> record.getCreatedOn().equals(createdOn) && record.getSchemaId().equals(schemaId))
+                .collect(Collectors.toList());
     }
 }
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDao.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDao.java
@@ -3,16 +3,14 @@ package org.sagebionetworks.bridge.dynamodb;
 import javax.annotation.Nonnull;
 import javax.annotation.Resource;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBQueryExpression;
-import com.amazonaws.services.dynamodbv2.document.RangeKeyCondition;
 import com.amazonaws.services.dynamodbv2.model.AttributeValue;
 import com.amazonaws.services.dynamodbv2.model.ComparisonOperator;
 import com.amazonaws.services.dynamodbv2.model.Condition;
-import org.joda.time.DateTime;
 import org.sagebionetworks.bridge.BridgeUtils;
 import org.sagebionetworks.bridge.dao.HealthDataDao;
 import org.sagebionetworks.bridge.exceptions.BridgeServiceException;
@@ -26,7 +24,7 @@ import org.springframework.stereotype.Component;
 /** DynamoDB implementation of {@link org.sagebionetworks.bridge.dao.HealthDataDao}. */
 @Component
 public class DynamoHealthDataDao implements HealthDataDao {
-    private static final long CREATED_ON_OFFSET_MILIS = 100;
+    private static final long CREATED_ON_OFFSET_MILLIS = TimeUnit.MILLISECONDS.convert(1, TimeUnit.HOURS);
 
     private DynamoDBMapper mapper;
     private DynamoIndexHelper healthCodeIndex;
@@ -62,7 +60,7 @@ public class DynamoHealthDataDao implements HealthDataDao {
      * using global secondary indices. This is configured by Spring
      */
     @Resource(name = "healthDataHealthCodeCreatedOnIndex")
-    public void setHealthCodeCreatedOnIndex(DynamoIndexHelper healthCodeCreatedOnIndex) {
+    public final void setHealthCodeCreatedOnIndex(DynamoIndexHelper healthCodeCreatedOnIndex) {
         this.healthCodeCreatedOnIndex = healthCodeCreatedOnIndex;
     }
 
@@ -120,8 +118,8 @@ public class DynamoHealthDataDao implements HealthDataDao {
     public List<HealthDataRecord> getRecordsByHealthCodeCreatedOnSchemaId(@Nonnull String healthCode, @Nonnull Long createdOn, @Nonnull String schemaId) {
         Condition rangeKeyCondition = new Condition()
                 .withComparisonOperator(ComparisonOperator.BETWEEN)
-                .withAttributeValueList(new AttributeValue().withN(String.valueOf((createdOn - CREATED_ON_OFFSET_MILIS)))
-                        , new AttributeValue().withN(String.valueOf((createdOn + CREATED_ON_OFFSET_MILIS))));
+                .withAttributeValueList(new AttributeValue().withN(String.valueOf((createdOn - CREATED_ON_OFFSET_MILLIS)))
+                        , new AttributeValue().withN(String.valueOf((createdOn + CREATED_ON_OFFSET_MILLIS))));
 
         DynamoHealthDataRecord queryRecord = new DynamoHealthDataRecord();
         queryRecord.setHealthCode(healthCode);
@@ -145,6 +143,8 @@ public class DynamoHealthDataDao implements HealthDataDao {
             finalRecords.add((HealthDataRecord) oneResult);
         }
 
-        return finalRecords;
+        return finalRecords.stream()
+                .filter(record -> record.getSchemaId().equals(schemaId))
+                .collect(Collectors.toList());
     }
 }
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
@@ -46,14 +46,12 @@ public class DynamoHealthDataRecord implements HealthDataRecord {
     private ExporterStatus synapseExporterStatus;
 
     /** {@inheritDoc} */
-    @JsonSerialize(using = DateTimeToLongSerializer.class)
     @Override
     public Long getCreatedOn() {
         return createdOn;
     }
 
     /** @see #getCreatedOn */
-    @JsonDeserialize(using = DateTimeToLongDeserializer.class)
     public void setCreatedOn(Long createdOn) {
         this.createdOn = createdOn;
     }
@@ -168,14 +166,12 @@ public class DynamoHealthDataRecord implements HealthDataRecord {
 
     /** {@inheritDoc} */
     @DynamoDBIndexRangeKey(attributeName = "uploadedOn", globalSecondaryIndexName = "study-uploadedOn-index")
-    @JsonSerialize(using = DateTimeToLongSerializer.class)
     @Override
     public Long getUploadedOn() {
         return uploadedOn;
     }
 
     /** @see #getUploadedOn */
-    @JsonDeserialize(using = DateTimeToLongDeserializer.class)
     public void setUploadedOn(Long uploadedOn) {
         this.uploadedOn = uploadedOn;
     }
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
@@ -46,12 +46,15 @@ public class DynamoHealthDataRecord implements HealthDataRecord {
     private ExporterStatus synapseExporterStatus;
 
     /** {@inheritDoc} */
+    @DynamoDBIndexRangeKey(attributeName = "createdOn", globalSecondaryIndexName = "healthCode-createdOn-index")
+    @JsonSerialize(using = DateTimeToLongSerializer.class)
     @Override
     public Long getCreatedOn() {
         return createdOn;
     }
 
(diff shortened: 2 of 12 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDao.java
```diff
@@ -91,4 +98,15 @@ public HealthDataRecord getRecordById(@Nonnull String id) {
...
+    /** {@inheritDoc} */
+    @Override
+    public List<HealthDataRecord> getRecordsByHealthCodeCreatedOnSchemaId(@Nonnull String healthCode, @Nonnull Long createdOn, @Nonnull String schemaId) {
+        List<HealthDataRecord> recordsByHealthCode = healthCodeIndex.queryKeys(HealthDataRecord.class, "healthCode",
+                healthCode, null);
+
+        return recordsByHealthCode.stream()
+                .filter(record -> record.getCreatedOn().equals(createdOn) && record.getSchemaId().equals(schemaId))
+                .collect(Collectors.toList());
+    }
```
DwayneJengSage, line 149: I think this needs to be a bit broader than that. Some duplicate uploads might not come in at the exact same time, but rather spaced a few seconds or a few minutes apart. You'll want a start and end time and use a range query.

Also, this query is very poor performing. Your query gets all the records for a particular healthcode, which could be hundreds if not thousands of records.

You also call queryKeys(), which only gets they keys (healthcode, recordId), which means schemaId and createdOn will be blank. You'll want to just call query().

Lastly, you'll want a unit test that calls through to the real DynamoDB, so you can make sure your code works against the actual DDB, and not a mock DDB that works the way you think it does (but might not actually). See for example DynamoUploadSchemaDaoDdbTest vs DynamoUploadSchemaDaoMockTest.

app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
```diff
@@ -46,14 +46,12 @@
     private ExporterStatus synapseExporterStatus;
 
     /** {@inheritDoc} */
-    @JsonSerialize(using = DateTimeToLongSerializer.class)
```
DwayneJengSage, line 49: Why is are the JSON serializers and deserializers being removed from this class (here and in the other 3 cases in this file)? This will break the JSON format of the health data record.

app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
```diff
@@ -22,6 +26,15 @@
 
     private List<UploadValidationHandler> handlerList;
     private UploadDao uploadDao;
+    private HealthDataService healthDataService;
+
+    public void setHealthDataService(HealthDataService healthDataService) {
+        this.healthDataService = healthDataService;
+    }
```
DwayneJengSage, line 33: Make this final. Otherwise, subclasses my accidentally override this and lead to mysterious NullPointerExceptions.

app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
```diff
@@ -111,6 +124,30 @@ public void run() {
         }
 
         // TODO: if validation fails, wipe the files from S3
+
+        // dedupe logic over here:
+        dedupeHelper(context.getRecordId());
```
DwayneJengSage: Play doesn't log exceptions from async threads. You need a try-catch here.

app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
```diff
@@ -111,6 +124,30 @@ public void run() {
...
+    }
+
+    /**
+     * helper method: query healthdatarecord by healthcode, schema and createdOn from the upload just finished above,
+     * if there are more than 1 such records in ddb, there are duplicates. log that information but not stop or delete anything right now
+     * @param healthRecordId
+     */
+    private void dedupeHelper(String healthRecordId) {
+        // get necessary information for query first
+        HealthDataRecord record = healthDataService.getRecordById(healthRecordId);
```
DwayneJengSage, line 152: There might not be a record for this upload. (For example, if the upload failed validation.) What does your code do in this case?

app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
```diff
@@ -111,6 +124,30 @@ public void run() {
...
+        Long createdOn = record.getCreatedOn();
+        String healthCode = record.getHealthCode();
+        String schemaId = record.getSchemaId();
+
+        List<HealthDataRecord> retList = healthDataService.getRecordsByHealthcodeCreatedOnSchemaId(healthCode, createdOn, schemaId);
+
+        if (retList.size() > 1) {
+            logger.warn(String.format("Duplicate health data records for record id: %s, health code: %s, created on:  %s, schema id: %s, duplicate size: %s",
+                    healthRecordId, healthCode, createdOn, schemaId, retList.size()));
+        }
```
DwayneJengSage, line 184: It's not enough to simply log that a duplicate exists. If we want to analyze, we need to actually identify the duplicates.

Also, this should be info, not warning. Warnings mean there's an immediately actionable if there are lots of them. This doesn't qualify.

Lastly, you'll want a unit test to make sure this works. It might be worth factoring the warning message to a "handleDuplicate()" method that you can spy and verify that it was called. You'll also want to test the cases where there is no record for this upload, where the record is not a duplicate, where there is a duplicate, and where there's more than one duplicate.

app/org/sagebionetworks/bridge/upload/UploadValidationTaskFactory.java
```diff
@@ -33,6 +35,11 @@ public void setUploadDao(UploadDao uploadDao) {
         this.uploadDao = uploadDao;
     }
 
+    @Autowired
+    public void setHealthDataService(HealthDataService healthDataService) {
+        this.healthDataService = healthDataService;
+    }
```
DwayneJengSage, line 41: Similarly, make this final.

DwayneJengSage (changes requested):
Also, how did you test this? Unit tests? Integration tests? Manual tests? Testing a few cases manually and verifying the log message(s) would definitely help make sure your code works as intended.

app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDao.java
```diff
@@ -16,9 +26,12 @@
 /** DynamoDB implementation of {@link org.sagebionetworks.bridge.dao.HealthDataDao}. */
 @Component
 public class DynamoHealthDataDao implements HealthDataDao {
+    private static final long CREATED_ON_OFFSET_MILIS = 100;
```
DwayneJengSage: I think this is too narrow. We can have dupes that are submitted seconds or even minutes apart. I'd suggest setting this to an hour or so.

app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDao.java
```diff
@@ -44,6 +57,16 @@ public void setUploadDateIndex(DynamoIndexHelper uploadDateIndex) {
         this.uploadDateIndex = uploadDateIndex;
     }
 
+    /**
+     * DynamoDB Index reference for the healthCodeCreatedOnIndex index. This is needed because the DynamoDB mapper does allow queries
+     * using global secondary indices. This is configured by Spring
+     */
+    @Resource(name = "healthDataHealthCodeCreatedOnIndex")
+    public void setHealthCodeCreatedOnIndex(DynamoIndexHelper healthCodeCreatedOnIndex) {
```
DwayneJengSage: Make this final.

app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDao.java
```diff
@@ -91,4 +114,37 @@ public HealthDataRecord getRecordById(@Nonnull String id) {
...
+        Condition rangeKeyCondition = new Condition()
+                .withComparisonOperator(ComparisonOperator.BETWEEN)
+                .withAttributeValueList(new AttributeValue().withN(String.valueOf((createdOn - CREATED_ON_OFFSET_MILIS)))
+                        , new AttributeValue().withN(String.valueOf((createdOn + CREATED_ON_OFFSET_MILIS))));
+
+        DynamoHealthDataRecord queryRecord = new DynamoHealthDataRecord();
+        queryRecord.setHealthCode(healthCode);
+
+        DynamoDBQueryExpression<DynamoHealthDataRecord> expression = new DynamoDBQueryExpression<DynamoHealthDataRecord>()
+                .withConsistentRead(false)
```
DwayneJengSage, line 128: Why are we turning off consistent reads? If two duplicates come in at near the same time, we want to be be able to catch it.

test/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDaoDdbTest.java
```diff
@@ -1,30 +1,40 @@
...
+import java.util.List;
+
 @ContextConfiguration("classpath:test-context.xml")
 @RunWith(SpringJUnit4ClassRunner.class)
 public class DynamoHealthDataDaoDdbTest {
     private static final long CREATED_ON = 1424136378727L;
-    private static final String HEALTH_CODE = "health-code";
-    private static final String RECORD_ID = "record-id";
+    private static final String HEALTH_CODE = TestUtils.randomName(DynamoHealthDataDaoDdbTest.class);
+    private static final String RECORD_ID = TestUtils.randomName(DynamoHealthDataDaoDdbTest.class);
```
DwayneJengSage: This almost certainly doesn't do what you want it to do. This creates the random name once during the initialization of the JVM and uses the same ID for all tests. This means there's a chance that tests might interfere with each other.

You're better off making this non-static and adding the initialization to an @Before method.

app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
```diff
@@ -111,6 +122,61 @@ public void run() {
         }
 
         // TODO: if validation fails, wipe the files from S3
+
+        // dedupe logic over here:
+        try {
+            dedupeHelper(context.getRecordId());
+        } catch (ClassNotFoundException e) {
+            logClassNotFoundException(e);
+        }
```
DwayneJengSage, line 133: I think there are more exceptions than just ClassNotFoundException. In theory, you need to be able to handle any and all exception types, since the ExceptionInterceptor doesn't work on async threads.

app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
```diff
@@ -111,6 +122,61 @@ public void run() {
...
+     * @param healthRecordId
+     */
+    private void dedupeHelper(String healthRecordId) throws ClassNotFoundException {
+        // get necessary information for query first
+        HealthDataRecord record = healthDataService.getRecordById(healthRecordId);
+
+        // if there is no record in ddb,
+        if (record == null) {
+            throw new ClassNotFoundException("Record not found in ddb");
+        }
```
DwayneJengSage, line 157: It's a perfectly normal scenario for an upload to not have a record, such as when validation fails. We shouldn't throw an exception here.

Also, healthDataService.getRecordById() will throw a Bad Request Exception if there is no record ID, so we need to handle that case as well.

app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
```diff
@@ -111,6 +122,61 @@ public void run() {
...
+        List<HealthDataRecord> retList = healthDataService.getRecordsByHealthcodeCreatedOnSchemaId(healthCode, createdOn, schemaId);
+
+        if (retList.size() > 1) {
+            logger.info(String.format("Duplicate health data records for record id: %s, health code: %s, created on:  %s, schema id: %s, duplicate size: %s",
+                    healthRecordId, healthCode, createdOn, schemaId, retList.size()));
+
+            logDuplicateUploadRecords(retList);
+        } else if (retList.size() == 0) {
+            // should not be reached since getRecordById() already test if there is a record exists
+            throw new ClassNotFoundException("Record not found in ddb");
```
DwayneJengSage: Do we need this else if check here? Is ClassNotFoundException the right exception to throw here? Is there any particular reason we're throwing an exception instead of logging?

app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
```diff
@@ -111,6 +122,61 @@ public void run() {
...
+
+        Long createdOn = record.getCreatedOn();
+        String healthCode = record.getHealthCode();
+        String schemaId = record.getSchemaId();
+
+        List<HealthDataRecord> retList = healthDataService.getRecordsByHealthcodeCreatedOnSchemaId(healthCode, createdOn, schemaId);
+
+        if (retList.size() > 1) {
+            logger.info(String.format("Duplicate health data records for record id: %s, health code: %s, created on:  %s, schema id: %s, duplicate size: %s",
+                    healthRecordId, healthCode, createdOn, schemaId, retList.size()));
```
DwayneJengSage: Don't log the health code. This is a potential security/privacy risk.

app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
```diff
@@ -111,6 +122,61 @@ public void run() {
...
+        // else there is only one such record exists in ddb, means no duplicate -- do nothing
+    }
+
+    void logDuplicateUploadRecords(List<HealthDataRecord> dupeRecords) {
+        for (HealthDataRecord dupeRecord: dupeRecords) {
+            logger.info("Duplicate HealthDataRecord: " + "record id: " + dupeRecord.getId()
+                    + ", createdOn: " + dupeRecord.getCreatedOn().toString()
+                    + ", healthCode: " + dupeRecord.getHealthCode() + ", schemaId: " + dupeRecord.getSchemaId()
+                    + ", studyId: " + dupeRecord.getStudyId() + ", uploadDate: " + dupeRecord.getUploadDate().toString()
+                    + ", uploadId: " + dupeRecord.getUploadId());
```
DwayneJengSage: Similarly, don't log the healthCode here either.

Also, the original record ID and duplicate record ID should be on the same line, to make log analysis easier.

Also, we need to make sure that we aren't logging records as duplicates of themselves.

test/org/sagebionetworks/bridge/upload/UploadValidationTaskTest.java
```diff
@@ -3,37 +3,129 @@
...
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
-import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.notNull;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyZeroInteractions;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.*;
```
DwayneJengSage: Code convention is to list out all imports we're using instead of using *

test/org/sagebionetworks/bridge/upload/UploadValidationTaskTest.java
```diff
@@ -3,37 +3,129 @@
...
 import org.sagebionetworks.bridge.dynamodb.DynamoUpload2;
+import org.sagebionetworks.bridge.json.BridgeObjectMapper;
+import org.sagebionetworks.bridge.models.healthdata.HealthDataRecord;
+import org.sagebionetworks.bridge.models.healthdata.HealthDataRecordBuilder;
 import org.sagebionetworks.bridge.models.upload.UploadStatus;
+import org.sagebionetworks.bridge.services.HealthDataService;
 
 public class UploadValidationTaskTest {
+    public UploadValidationTaskTest() throws IOException {
+    }
```
DwayneJengSage: Tests shouldn't have constructors.

app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDao.java
```diff
@@ -91,4 +114,37 @@ public HealthDataRecord getRecordById(@Nonnull String id) {
...
+                // This should never happen, but just in case.
+                throw new BridgeServiceException(String.format(
+                        "DynamoDB returned objects of type %s instead of %s",
+                        oneResult.getClass().getName(), DynamoHealthDataRecord.class.getName()));
+            }
+
+            finalRecords.add((HealthDataRecord) oneResult);
+        }
+
+        return finalRecords;
```
DwayneJengSage: Where do we filter on schema ID?

test/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDaoDdbTest.java
```diff
@@ -86,4 +99,62 @@ public void ddbRoundtrip() throws Exception {
             mapper.delete(savedRecord);
         }
     }
+
+    @Test
+    public void getRecordsByHealthCodeCreatedOnSchemaId() throws IOException {
```
DwayneJengSage: We should also add tests for:
* different health code
* createdOn out of range
* different schema ID

test/org/sagebionetworks/bridge/upload/UploadValidationTaskTest.java
```diff
@@ -150,6 +244,130 @@ public void writeValidationStatusException() {
...
+
+        // execute
+        task.run();
+
+        // verify log helper was NOT called
+        verify(task, times(0)).logDuplicateUploadRecords(any());
+    }
+
+    @Test
+    public void dedupeNullRecord() {
```
DwayneJengSage, line 341: Should we also have a test case where upload.getRecordId() is null?

test/org/sagebionetworks/bridge/upload/UploadValidationTaskTest.java
```diff
@@ -150,6 +244,130 @@ public void writeValidationStatusException() {
...
+
+        // set up validation task
+        UploadValidationTask task = spy(new UploadValidationTask(ctx));
+        task.setHandlerList(handlerList);
+        // return a null record
+        healthDataService = mock(HealthDataService.class);
+        when(healthDataService.getRecordById(any())).thenReturn(null);
+        task.setHealthDataService(healthDataService);
+        task.run();
+        verify(task).logClassNotFoundException(any());
```
DwayneJengSage: Would it be more useful to have a verify that logDuplicateUploadRecords *isn't* called?

test/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDaoDdbTest.java
```diff
@@ -1,30 +1,40 @@
...
 public class DynamoHealthDataDaoDdbTest {
     private static final long CREATED_ON = 1424136378727L;
-    private static final String HEALTH_CODE = "health-code";
-    private static final String RECORD_ID = "record-id";
+    private static final String HEALTH_CODE = TestUtils.randomName(DynamoHealthDataDaoDdbTest.class);
+    private static final String RECORD_ID = TestUtils.randomName(DynamoHealthDataDaoDdbTest.class);
     private static final String DATA_TEXT = "{\"data\":\"dummy value\"}";
     private static final String METADATA_TEXT = "{\"metadata\":\"dummy meta value\"}";
-    private static final String SCHEMA_ID = "schema-id";
+    private static final String SCHEMA_ID = TestUtils.randomName(DynamoHealthDataDaoDdbTest.class);
```
DwayneJengSage: Randomizing the schema ID is probably not necessary, since nothing here uses it as a key.

test/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDaoDdbTest.java
```diff
@@ -86,4 +99,62 @@ public void ddbRoundtrip() throws Exception {
...
+        testRecord.setUserExternalId(USER_EXTERNAL_ID);
+        testRecord.setUserDataGroups(TestConstants.USER_DATA_GROUPS);
+        testRecord.setData(BridgeObjectMapper.get().readTree(DATA_TEXT));
+        testRecord.setMetadata(BridgeObjectMapper.get().readTree(METADATA_TEXT));
+        testRecord.setStudyId(TestConstants.TEST_STUDY_IDENTIFIER);
+
+        mapper.save(testRecord);
+
+        // query that record
+        List<HealthDataRecord> retList = dao.getRecordsByHealthCodeCreatedOnSchemaId(HEALTH_CODE, CREATED_ON, SCHEMA_ID);
```
DwayneJengSage: getRecordsByHealthCodeCreatedOnSchemaId() should be moved inside the try-finally. This way, if it throws, the test still cleans up after itself.

test/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDaoDdbTest.java
```diff
@@ -86,4 +99,62 @@ public void ddbRoundtrip() throws Exception {
...
+        testRecord.setUploadedOn(UPLOADED_ON);
+        testRecord.setUploadId(UPLOAD_ID);
+        testRecord.setUserSharingScope(ParticipantOption.SharingScope.SPONSORS_AND_PARTNERS);
+        testRecord.setUserExternalId(USER_EXTERNAL_ID);
+        testRecord.setUserDataGroups(TestConstants.USER_DATA_GROUPS);
+        testRecord.setData(BridgeObjectMapper.get().readTree(DATA_TEXT));
+        testRecord.setMetadata(BridgeObjectMapper.get().readTree(METADATA_TEXT));
+        testRecord.setStudyId(TestConstants.TEST_STUDY_IDENTIFIER);
+
+        mapper.save(testRecord);
```
DwayneJengSage, line 80: It might be worthwhile to move record creation to an @Before and record deleting to an @After. This removes duplicate code, and if the test fails and then the record deletion also fails, we get a more useful error message.

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDaoDdbTest.java
```diff
@@ -1,89 +1,193 @@
...
     private static final LocalDate UPLOAD_DATE = LocalDate.parse("2016-05-06");
     private static final String UPLOAD_ID = "upload-id";
     private static final long UPLOADED_ON = 1462575525894L;
     private static final String USER_EXTERNAL_ID = "external-id";
 
+    private static final String HEALTH_CODE_2 = "test-healthcode-2";
+    private static final String SCHEMA_ID_2 = "test-schema-id-2";
+
+    private String HEALTH_CODE;
+    private String RECORD_ID;
```
DwayneJengSage: Style note: Non-static members should be camel-case: healthCode, recordId

test/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDaoDdbTest.java
```diff
@@ -1,89 +1,193 @@
...
+
+//    private String SCHEMA_ID;
```
DwayneJengSage: We can remove this unnecessary comment.

test/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDaoDdbTest.java
```diff
@@ -1,89 +1,193 @@
...
     private DynamoDBMapper mapper;
 
+    @Autowired
+    private DynamoHealthDataDao dao;
+
+    @Before
+    public void setup() throws IOException {
+        HEALTH_CODE = TestUtils.randomName(DynamoHealthDataDaoDdbTest.class);
+        RECORD_ID = TestUtils.randomName(DynamoHealthDataDaoDdbTest.class);
+//        SCHEMA_ID = TestUtils.randomName(DynamoHealthDataDaoDdbTest.class);
```
DwayneJengSage: Also, this comment.

app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
```diff
@@ -111,6 +124,65 @@ public void run() {
...
+        logger.info("Origin Record: " + "record id: " + originRecord.getId());
+
+        for (HealthDataRecord dupeRecord: dupeRecords) {
+            if (!dupeRecord.getId().equals(originRecord.getId())) {
+                logger.info("Duplicate HealthDataRecord: " + "record id: " + dupeRecord.getId()
+                        + ", createdOn: " + dupeRecord.getCreatedOn().toString()
+                        + ", schemaId: " + dupeRecord.getSchemaId()
+                        + ", studyId: " + dupeRecord.getStudyId() + ", uploadDate: " + dupeRecord.getUploadDate().toString()
+                        + ", uploadId: " + dupeRecord.getUploadId());
+            }
```
DwayneJengSage, line 183: We want to log the original record ID and the dupe record ID in the same log line. Otherwise, it'll be difficult to match up what record is a duplicate of what record.

test/org/sagebionetworks/bridge/upload/UploadValidationTaskTest.java
```diff
@@ -3,37 +3,191 @@
...
+
+    private final List<UploadValidationHandler> handlerList = ImmutableList.of(
+            new MessageHandler("foo was here"), new MessageHandler("bar was here"),
+            new MessageHandler("kilroy was here"), new RecordIdHandler(RECORD_ID));
+
+
+    private HealthDataService healthDataService;
+
+    public UploadValidationTaskTest() throws IOException {
+    }
```
DwayneJengSage: Test classes shouldn't have constructors.

Also, we need a unit test for when context.getRecordId() is null.

test/org/sagebionetworks/bridge/upload/UploadValidationTaskTest.java
```diff
@@ -3,37 +3,191 @@
...
+//                .withUploadId(UPLOAD_ID)
+//                .withUserSharingScope(ParticipantOption.SharingScope.SPONSORS_AND_PARTNERS)
+//                .withUserExternalId(USER_EXTERNAL_ID)
+//                .withUserDataGroups(TestConstants.USER_DATA_GROUPS)
+//                .withData(BridgeObjectMapper.get().readTree(DATA_TEXT))
+//                .withMetadata(BridgeObjectMapper.get().readTree(METADATA_TEXT))
+//                .build();
+//
+//        TEST_RECORD_DUPE_LIST_NORMAL = Arrays.asList(TEST_RECORD, TEST_RECORD_DUPE);
+//        TEST_RECORD_DUPE_LIST_MULTI = Arrays.asList(TEST_RECORD, TEST_RECORD_DUPE, TEST_RECORD_DUPE_2);
```
DwayneJengSage: Why are these commented out?

test/org/sagebionetworks/bridge/upload/UploadValidationTaskTest.java
```diff
@@ -150,6 +306,131 @@ public void writeValidationStatusException() {
...
+
+        // set up validation task
+        UploadValidationTask task = spy(new UploadValidationTask(ctx));
+        task.setHandlerList(handlerList);
+        // return a null record
+        healthDataService = mock(HealthDataService.class);
+        when(healthDataService.getRecordById(any())).thenReturn(null);
+        task.setHealthDataService(healthDataService);
+        task.run();
+//        verify(task).logClassNotFoundException(any());
```
DwayneJengSage: Remove commented out code.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #1290, opened 2016-11-19
Onetime task fix

alxdarksage:


app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -88,10 +88,12 @@ protected void addScheduledActivityAtTime(List<ScheduledActivity> scheduledActiv
         return scheduledActivities.subList(0, Math.min(scheduledActivities.size(), count));
     }
     
-    private boolean isInWindow(DateTime scheduledTime) {
+    private boolean isInWindow(LocalDate localDate, LocalTime localTime) {
+        DateTime scheduledTime = localDate.toDateTime(localTime, DateTimeZone.UTC);
```
DwayneJengSage: Is UTC the right timezone for this? This represents the time the task will be scheduled for the user, so it seems like it should use the context's local time zone. Otherwise, we may have tasks ending prematurely and/or tasks going past the scheduled end time.

app/org/sagebionetworks/bridge/models/schedules/CronActivityScheduler.java
```diff
@@ -25,10 +26,10 @@
             
             while (shouldContinueScheduling(context, scheduledTime, scheduledActivities)) {
                 Date next = trigger.getFireTimeAfter(scheduledTime.toDate());
-                scheduledTime = new DateTime(next, context.getZone());
+                scheduledTime = new DateTime(next, DateTimeZone.UTC);
                 
                 if (shouldContinueScheduling(context, scheduledTime, scheduledActivities)) {
-                    addScheduledActivityForAllTimes(scheduledActivities, plan, context, scheduledTime);    
+                    addScheduledActivityAtTime(scheduledActivities, plan, context, scheduledTime.toLocalDate(), scheduledTime.toLocalTime());
```
DwayneJengSage, line 36: I think the best way to reason behind this logic is to say that the CronScheduler is timezone-agnostic, and that the schedules produced by the cron expression correspond to a Joda LocalDateTime.

However, since the QuartzScheduler uses the old Java Date libraries, which doesn't have the concept, we're forced to use a full-fledged date time. By always scheduling using a consistent, fixed timezone (UTC), we can simulate a LocalDateTime, and we pass a LocalDate and LocalTime to addScheduleActivityAtTime(), so this confusion is isolated to this one method.

A comment clarifying this to be the case would be useful for future us.

DwayneJengSage (changes requested):
Two comments. Otherwise looks good.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceRecurringTest.java
```diff
@@ -91,6 +91,9 @@ public void after() {
         }
     }
     
+    // WARNING: The notes here are not correct due to changes we've made in the way we test the time zone.
+    // The last day MSK seems to fall outside the scope of the tests. This needs to be fixed in later
+    // updates but we're currently addressing the duplication of tasks.
     @Test
     public void retrievalActivitiesAcrossTimeAndTimeZones() throws Exception {
```
DwayneJengSage, line 97: I think we established that this test succeeds before 4pm but fails after 5pm. Is it worth using DateTimeUtils to set a fixed "now" time so that we have a deterministic build?

DwayneJengSage (changes requested):
Looks good. Just one comment about the unit tests. I think we established these tests can randomly fail depending on time of day.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceRecurringTest.java
alxdarksage: This test does fix the time at the top of the test. I ran them yesterday evening, I'll run them now, but this test should not vary.

alxdarksage: This test does call DateTimeUtils.setCurrentMillisFixed but still varies, and I know from prior exploration, that's not holding for all time values being created by the test. Looking at it.

alxdarksage: This test does set a fixed time but the enrollment timestamp of the user is changing, and this causes the tests to generate different numbers of tasks at different times (I'm pretty sure this is the cause). This should be fixed probably. But testing this at 03:00, 10:00 and 22:00 on my machine, it's possible to get all the tests to pass with stable numbers if I add one final check that the scheduledOn time is correct against the localized endsOn, or the scheduledOnAfterEndsOnNoMinimumCount method. 

(Otherwise the number of tasks returned changes in a couple of tests, and we would have to weaken or remove the tests which I don't want to do).

So I would deploy this BUT I want to run it this evening because I'm worried that setting the time on my machine is not proving that this time shift issue is addressed.

DwayneJengSage (approved):

DwayneJengSage:
Looks good, but it looks like there's a merge conflict now.

DwayneJengSage:
Hold on. I misunderstood the thing about the tests.

DwayneJengSage:
So it looks like we set a fixed "now" at the start of the test, but *after* we create the test account. So the test account's enrollment time depends on the time of day, and this randomly causes tests to fail depending on the time of day. This easy fix (for the test) would be to fix the time *before* we create the test account.

As already discussed, we should file a separate JIRA for this bug and fix in a later change, to prevent further complicating this pull request.

DwayneJengSage (changes requested):
Need to fix "now" time in ScheduledActivityServiceRecurringTest.retrievalActivitiesAcrossTimeAndTimeZones() *before* creating the test account.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceRecurringTest.java
alxdarksage: Moved enrollment so it is at a fixed time (the time frozen in the test -2 days). Added back some tasks that will appear when event is at that time.

DwayneJengSage (approved):

PR #1294, opened 2016-11-28
Apple has changed the name of their platform, we need to map this to existing platform name so that filtering continues as expected.

alxdarksage:


test/org/sagebionetworks/bridge/models/ClientInfoTest.java
```diff
@@ -226,4 +227,17 @@ public void deprecatedOperatingSystemParsed() {
...
     }
+    
+    @Test
+    public void synonymOsNameConvertedToCanonicalValue() {
+        ClientInfo info = ClientInfo.parseUserAgentString(VALID_LONG_WITH_SYNONYM);
+        
+        assertEquals("Cardio Health", info.getAppName());
+        assertEquals(1, info.getAppVersion().intValue());
+        assertEquals("Unknown iPhone", info.getDeviceName());
+        assertEquals(IOS, info.getOsName());
```
zaclinsage: you test it with 'iOS' but assert it against 'IOS'. Will parseUserAgentString() automatically convert to upper case?

zaclinsage (approved):
good for me

PR #1295, opened 2016-11-29
notifyByEmail not being set to default value of true under some use cases

alxdarksage:
Change this value to a Boolean object so that it can be left null, rather than the default value of false, if it is not set through calls that update all fields of StudyParticipant. This will allow it to preserve its default value.


zaclinsage (approved):
looks good to me

PR #1298, opened 2016-11-30
Consolidating user session retrieval methods

alxdarksage:
Which ensures 410/412s occur together, because logic isn't different. All the methods now use the same code. Addresses BRIDGE-1550, BRIDGE-1411.

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/SubpopulationController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/SubpopulationController.java
@@ -5,12 +5,10 @@ import static org.sagebionetworks.bridge.Roles.DEVELOPER;
 import static org.sagebionetworks.bridge.Roles.RESEARCHER;
 
 import java.util.List;
-import java.util.Set;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Controller;
 
-import org.sagebionetworks.bridge.Roles;
 import org.sagebionetworks.bridge.exceptions.UnauthorizedException;
 import org.sagebionetworks.bridge.models.GuidVersionHolder;
 import org.sagebionetworks.bridge.models.accounts.UserSession;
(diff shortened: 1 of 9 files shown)


app/org/sagebionetworks/bridge/play/controllers/SubpopulationController.java
```diff
@@ -69,10 +63,8 @@ public Result getSubpopulation(String guid) {
         return okResult(subpop);
     }
     public Result deleteSubpopulation(String guid, String physicalDeleteString) {
-        UserSession session = getAuthenticatedSession();
-        if (!session.isInRole(DELETE_ROLES)) {
-            throw new UnauthorizedException();
-        }
+        UserSession session = getAuthenticatedSession(ADMIN, DEVELOPER);
```
zaclinsage, line 66: will this just throws an exception if user does not have admin or dev permission?

test/org/sagebionetworks/bridge/play/controllers/SurveyControllerTest.java
```diff
@@ -132,11 +144,18 @@ private void setUserSession(String studyIdentifier) {
...
+        
+        Map<SubpopulationGuid,ConsentStatus> consentStatuses = Maps.newHashMap();
+        consentStatuses.put(guid, status);
+        
         session = new UserSession(participant);
         session.setStudyIdentifier(new StudyIdentifierImpl(studyIdentifier));
-        doReturn(session).when(controller).getAuthenticatedSession();
-        doReturn(session).when(controller).getAuthenticatedAndConsentedSession();
-        doReturn(session).when(controller).getAuthenticatedSession(any(Roles.class));
+        session.setConsentStatuses(consentStatuses);
```
zaclinsage: why we need to set consent statuses up while we didn't use it in previous test codes?

app/org/sagebionetworks/bridge/play/controllers/SubpopulationController.java
alxdarksage: Yes

zaclinsage (approved):
looks good to me

PR #1301, opened 2016-12-05
Udd change in bridgepf and warning header

zaclinsage:


app/org/sagebionetworks/bridge/play/controllers/BaseController.java
```diff
@@ -258,13 +260,25 @@ void verifySupportedVersionOrThrowException(Study study) throws UnsupportedVersi
                 LOG.debug("Malformed Accept-Language header sent: " + acceptLanguageHeader);
             }
         }
+
+        // if no Accept-Language header detected, we shall add an extra warning header
+        Http.Response response = Http.Context.current().response();
+        response.setHeader(BridgeConstants.BRIDGE_API_STATUS_HEADER, BridgeConstants.BRIDGE_WARNING_STATUS);
```
alxdarksage, line 266: What happens if they send a junk accept language header, we should also include a warning;
I think the header needs to be a message: "Warning: Accept-Language header is missing or invalid; filtering based on language not available" and "Warning: User-Agent header is missing or invalid; filtering based on application version not available"

app/org/sagebionetworks/bridge/play/controllers/BaseController.java
```diff
@@ -258,13 +260,25 @@ void verifySupportedVersionOrThrowException(Study study) throws UnsupportedVersi
...
     ClientInfo getClientInfoFromUserAgentHeader() {
         String userAgentHeader = request().getHeader(USER_AGENT);
         ClientInfo info = ClientInfo.fromUserAgentCache(userAgentHeader);
-        
+
+        // if the user agent cannot be parsed (probably due to missing user agent string or unrecognizable user agent),
+        // should set an extra header to http response as warning - we should have an user agent info for filtering to work
+        if (info.equals(ClientInfo.UNKNOWN_CLIENT)) {
+            Http.Response response = Http.Context.current().response();
+            response.setHeader(BridgeConstants.BRIDGE_API_STATUS_HEADER, BridgeConstants.BRIDGE_WARNING_STATUS);
```
alxdarksage, line 279: Yes, but need to include an explanation of what is wrong, something we can log from the client. Otherwise, we don't know what the exact problem is as an end user.

test/org/sagebionetworks/bridge/play/controllers/BaseControllerIntegTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @After
+    public void after() {
+        helper.deleteUser(testUser);
+    }
+
+    @Test
+    public void setWarningHeaderWhenNoUserAgentOrNoAcceptLanguage() {
+        // This is expensive so test them all after starting the server, in one test
+        TestUtils.runningTestServerWithSpring(() -> {
```
alxdarksage, line 50: We want to avoid writing any more integration tests in BridgePF. Please write a test that verifies the controller sets the headers, that does not start a Play server. If we write a true integration test, we can do it in the integration tests project. This kind of test is extremely slow to run.

alxdarksage (changes requested):

alxdarksage:
Zach, has this been split out into two other PRs? If so, you can close.

PR #1302, opened 2016-12-05
add null check to getUploadsForStudy APIs

DwayneJengSage:
For whatever reason, some uploads have record IDs, but no records. Under normal conditions, this shouldn't be possible, but it comes up occasionally in test environments. Either way, we shouldn't be throwing NullPointerExceptions for data integrity issues.

Testing done:
* manual tests with a test set that includes (a) failed uploads (b) successful uploads (c) successful uploads with no record
* updated unit tests

### DIFF
--- a/test/org/sagebionetworks/bridge/services/UploadServiceMockTest.java
+++ b/test/org/sagebionetworks/bridge/services/UploadServiceMockTest.java
@@ -7,6 +7,7 @@ import static org.junit.Assert.fail;
 import static org.mockito.Mockito.any;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.eq;
+import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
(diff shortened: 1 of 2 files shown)


test/org/sagebionetworks/bridge/services/UploadServiceMockTest.java
```diff
@@ -236,20 +231,38 @@ public void canGetStudyUploads() throws Exception {
                 START_TIME, END_TIME);
         
         verify(mockDao).getStudyUploads(TestConstants.TEST_STUDY, START_TIME, END_TIME);
+        validateUploadMocks(returned);
+    }
+
+    private void validateUploadMocks(DateTimeRangeResourceList<? extends UploadView> returned) {
         verify(mockHealthDataService).getRecordById("record-id");
+        verify(mockHealthDataService).getRecordById("missing-record-id");
```
zaclinsage, line 239: is it helpful to declare those constants at the top of the test class?

PR #1305, opened 2016-12-06
Httpheader

zaclinsage:


app/org/sagebionetworks/bridge/BridgeConstants.java
```diff
@@ -20,7 +20,14 @@
     
     public static final String BRIDGE_API_STATUS_HEADER = "Bridge-Api-Status";
 
-    public static final String BRIDGE_DEPRECATED_STATUS = "deprecated";
+    public static final String BRIDGE_DEPRECATED_STATUS = "you're calling a deprecated endpoint";
+
+    public static final String BRIDGE_WARNING_STATUS = "Bridge-Warning: you're calling a deprecated endpoint; you haven't included an Accept-Language header; we can't parse your User-Agent header";
```
alxdarksage: Can be removed

app/org/sagebionetworks/bridge/BridgeConstants.java
```diff
@@ -20,7 +20,14 @@
...
+
+    public static final String WARN_NO_USER_AGENT =  "we can't parse your User-Agent header";
+
+    public static final String WARN_NO_ACCEPT_LANGUAGE = "you haven't included an Accept-Language header";
```
alxdarksage: messages used to include why we were warning people, we should add back (e.g. here we won't filter based on the caller's language)

app/org/sagebionetworks/bridge/play/controllers/BaseController.java
```diff
@@ -258,13 +260,34 @@ void verifySupportedVersionOrThrowException(Study study) throws UnsupportedVersi
...
+        // should set an extra header to http response as warning - we should have an user agent info for filtering to work
+        if (info.equals(ClientInfo.UNKNOWN_CLIENT)) {
+            Http.Response response = Http.Context.current().response();
+            if (response.getHeaders().containsKey(BridgeConstants.BRIDGE_API_STATUS_HEADER)) {
+                String previousWarning = response.getHeaders().get(BridgeConstants.BRIDGE_API_STATUS_HEADER);
+                response.setHeader(BridgeConstants.BRIDGE_API_STATUS_HEADER, previousWarning + "; " + BridgeConstants.WARN_NO_USER_AGENT);
+            } else {
+                response.setHeader(BridgeConstants.BRIDGE_API_STATUS_HEADER, BridgeConstants.WARN_NO_USER_AGENT);
+            }
+        }
```
alxdarksage, line 277: Can you make a method, addWarningToHeader(String warning), to consolidate this logic?

test/org/sagebionetworks/bridge/play/controllers/BaseControllerTest.java
```diff
@@ -138,7 +131,60 @@ public void doesNotThrowErrorWhenUserAgentStringInvalid() throws Exception {
...
+        Http.Response mockResponse = BaseController.response();
+        verify(mockResponse).setHeader(BRIDGE_API_STATUS_HEADER, WARN_NO_USER_AGENT);
+
+        assertNull(info.getAppName());
+        assertNull(info.getAppVersion());
+        assertNull(info.getOsName());
+        assertNull(info.getOsVersion());
+        assertNull(info.getSdkName());
+        assertNull(info.getSdkVersion());
+    }
```
alxdarksage, line 225: Add a verification that we don't set the header if the User-Agent string parses correctly

test/org/sagebionetworks/bridge/play/controllers/BaseControllerTest.java
```diff
@@ -350,7 +396,74 @@ public void canGetLanguagesWhenNotInSessionOrHeader() throws Exception {
...
+        // with no accept language header at all, things don't break;
+        LinkedHashSet<String> langs = controller.getLanguagesFromAcceptLanguageHeader();
+        // testing this because the rest of these tests will use ImmutableSet.of()
+        assertTrue(langs instanceof LinkedHashSet);
+        assertEquals(ImmutableSet.of(), langs);
+
+        // verify if it set warning header
+        Http.Response mockResponse = BaseController.response();
+        verify(mockResponse).setHeader(BRIDGE_API_STATUS_HEADER, WARN_NO_ACCEPT_LANGUAGE);
+    }
```
alxdarksage, line 517: Add happy path case. In one of these cases as well, you can verify that the message is appended rather than added, to verify that works.

test/org/sagebionetworks/bridge/play/controllers/BaseControllerTest.java
```diff
@@ -350,7 +396,74 @@ public void canGetLanguagesWhenNotInSessionOrHeader() throws Exception {
...
+
     @Test(expected = NotAuthenticatedException.class)
     public void getSessionAuthenticatedFail() {
         UserSession session = new UserSession(new StudyParticipant.Builder().build());
```
alxdarksage, line 521: Does the deprecated header test have to be changed? I don't think so if 1) you consolidate the code adding the header and 2) verify that messages are concatenated correctly.

alxdarksage (changes requested):

app/org/sagebionetworks/bridge/BridgeConstants.java
zaclinsage: what's the meaning of this comment? why current warning msg is not explaining why it warns people? the msg used to have just "warning'

test/org/sagebionetworks/bridge/play/controllers/BaseControllerTest.java
zaclinsage: moved the concatenated test case specifically to addWarningMsg() method

test/org/sagebionetworks/bridge/play/controllers/BaseControllerTest.java
zaclinsage: moved the concatenated test case specifically to addWarningMsg() method

test/org/sagebionetworks/bridge/play/controllers/BaseControllerTest.java
zaclinsage: done

app/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -68,6 +69,16 @@ public static String resolveTemplate(String template, Map<String,String> values)
         }
         return template;
     }
+
+    public static void addWarningMessage(String msg) {
+        Http.Response response = Http.Context.current().response();
```
alxdarksage: I really dislike accessing this thread local variable in a static utility class like this... I would make this method a base method in BaseController, which is part of the Play infrastructure. That's because this code will have to be rewritten if we switch off of the Play Framework.

alxdarksage (changes requested):
This is great, I have only one comment about using BridgeUtils for the header functionality, then this can be merged.

PR #1307, opened 2016-12-07
Revert task fix

alxdarksage:
Times are now based off the timestamp once again.

The plan instead is to record the time zone in the event table, in order to derive a local time of day that will work regardless of time zone changes. We'll backfill where we're able or probably use a central time zone (like central) instead.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
+++ b/app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
@@ -4,23 +4,18 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static java.util.Comparator.comparing;
 import static java.util.stream.Collectors.toList;
-import static java.util.stream.Collectors.toSet;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static org.sagebionetworks.bridge.util.BridgeCollectors.toImmutableList;
 
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 
 import org.joda.time.DateTime;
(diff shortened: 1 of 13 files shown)


app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
```diff
@@ -164,26 +118,11 @@ public void deleteActivitiesForUser(String healthCode) {
     }
     
     protected List<ScheduledActivity> updateActivitiesAndCollectSaves(List<ScheduledActivity> scheduledActivities, List<ScheduledActivity> dbActivities) {
-        // BRIDGE-1589: After adjusting activity GUIDs' time portion, we will have generated db activities with duplicate 
-        // GUIDs in dbActivities, and Maps.uniqueIndex will throw an error because the GUIDs are not all unique. 
-        // Build the map manually while de-duplicating the db activities, particularly saving any copy that has 
-        // client-persisted state in it (startedOn is good enough here). 
-        // Map<String, ScheduledActivity> dbMap = Maps.uniqueIndex(dbActivities, ScheduledActivity::getGuid);
+        Map<String, ScheduledActivity> dbMap = Maps.uniqueIndex(dbActivities, ScheduledActivity::getGuid);
```
liujoshua, line 121: Do we need a migration to ensure duplicate Guid is no longer an issue? (e.g. did any DB Activities in prod get changed with the old code?

liujoshua (approved):

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
alxdarksage: Once this fix is rolled back, yes, there could be duplicated tasks that are visible, though it's been a month now and I'm wondering how many of these are left.

PR #1308, opened 2016-12-08
Fix the time zone after the first time you retrieve it

alxdarksage:
The behavior now:

- we retrieve and save the first time zone we get from the user as a participant option, and use that to schedule a sequence of events from an event timestamp;
- once we get to calculating the time of day of a task, we use the local times supplied in the schedule. The final time is alway returned with the time zone of the request, in effect making these timestamps LocalDateTime objects.

This fixes a couple of issues:

- fixing the time zone of the sequence calculation will prevent duplicates when daylight savings occurs;
- activities will be scheduled from the current day of the request, always, preventing situations where a shift to UTC occurs and so, at some points of the day in some time zones, the date of the schedule is wrong (e.g. late in the day on the west coast, scheduler starts on the next day);


app/org/sagebionetworks/bridge/models/accounts/StudyParticipant.java
```diff
@@ -172,7 +178,8 @@ public boolean equals(Object obj) {
                 && Objects.equals(id, other.id) && Objects.equals(languages, other.languages)
                 && Objects.equals(lastName, other.lastName) && Objects.equals(notifyByEmail, other.notifyByEmail)
                 && Objects.equals(password, other.password) && Objects.equals(roles, other.roles)
-                && Objects.equals(sharingScope, other.sharingScope) && Objects.equals(status, other.status);
+                && Objects.equals(sharingScope, other.sharingScope) && Objects.equals(status, other.status)
+                && Objects.equals(timeZone, other.timeZone);
```
Erin-Mounts, line 182: so this won't consider a UTC time and a local-time-with-timezone to be equal even if they represent the same instant (ignoring relativistic effects)? if that's the intended behavior, then cool

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -338,6 +340,10 @@ private IdentifierHolder saveParticipant(Study study, Set<Roles> callerRoles, St
         if (study.isExternalIdValidationEnabled()) {
             options.remove(EXTERNAL_IDENTIFIER);
         }
+        // You can't update time zone after it is set.
```
Erin-Mounts: So if the participant goes on a trip, or moves to a different time zone, or heaven forbid their local jurisdiction switches between standard and daylight time, is the time zone now wrong? Do we care?

Erin-Mounts (approved):
Couple of questions, can't tell if you have test coverage that answers them, otherwise looks fine as far as I can tell.

liujoshua (approved):

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: When calculating the series of times for repeating events, we proceed to the current day using the first time zone seen from the user, so that these tasks show up on that first day they are requested. This will prevent the tasks from appearing the next day, as we see late in the day here in PST. However, all repeating schedules require times-of-day to be specified, and these are always expressed in local time. So once the scheduler gets to the current day, the user is given those tasks in the time zone of the request, and should be correct.

app/org/sagebionetworks/bridge/models/accounts/StudyParticipant.java
alxdarksage: This is a comparison test for the study participant and is mostly used for tests, so we want to know if the participant object has a different time zone associated with it. Doesn't compare two time stamps.

app/org/sagebionetworks/bridge/dao/ParticipantOption.java
```diff
@@ -5,17 +5,27 @@
...
 
+    TIME_ZONE(null, "timeZone") {
+        public String fromParticipant(StudyParticipant participant) {
+            return (participant.getTimeZone() == null) ? null : participant.getTimeZone().toString();
+        }
+        public String deserialize(JsonNode node) {
+            checkNotNull(node);
+            return DateUtils.parseZoneFromOffsetString(node.asText()).toString();
+        }        
+    },
```
DwayneJengSage: Should we add unit tests in ParticipantOptionTest?

app/org/sagebionetworks/bridge/models/accounts/ParticipantOptionsLookup.java
```diff
@@ -21,6 +24,12 @@ public ParticipantOptionsLookup(Map<String,String> options) {
         this.options = options;
     }
     
+    public DateTimeZone getTimeZone(ParticipantOption option) {
+        String value = options.get(option.name());
+        // This method is null-safe (returns null)
+        return DateUtils.parseZoneFromOffsetString(value);
+    }
```
DwayneJengSage, line 31: What happens if the the value isn't a properly formatted timezone? Does it throw or does it return null? Is this consistent with the other getters? Is it worth adding a unit test for this?

app/org/sagebionetworks/bridge/play/controllers/ScheduledActivityController.java
```diff
@@ -82,14 +84,23 @@ public Result updateScheduledActivities() throws Exception {
...
+        // This zone is the zone of the request and the endsOn is scheduled relative to the user's
+        // time zone at the time of a request.
+        DateTimeZone zone = addEndsOnWithZone(builder, untilString, offset, daysAhead);
+
+        // This zone is the first zone used to schedule for the user, ever. If we don't have that:
+        // this request is the first contact from the user with a time zone, so save that.
+        DateTimeZone timeZone = session.getParticipant().getTimeZone();
+        if (timeZone == null) {
+            timeZone = persistTimeZone(session, zone);
+        }
```
DwayneJengSage: It seems like there are two different concepts of "time zone" in play here. The first is the time zone that the user wants to get events for. The second is the timezone of which we interpret event epoch millis into local times. The second one seems to exist solely because we were recording event times as epoch millis before we realized we also needed to know what timezone they're in, and now it's untenable to go back and change that.

A detailed comment explaining all this would be super useful for future archeology.

Also, is it worth adding two separate time zone fields to ScheduleContext that reflect this understanding instead of relying on getTimeZone() and getEndsOn().getTimeZone()? We could call them "request time zone" and "event time zone" or something similar. This will make the code much more readable and much easier to reason about.

Also, in terms of naming in this function: "zone" vs "timeZone" is really ambiguous. Maybe "requestTimeZone" vs "persistedTimeZone"? "requestTimeZone" vs "eventTimeZone"?

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -64,7 +64,9 @@ protected void addScheduledActivityAtTime(List<ScheduledActivity> scheduledActiv
                 for (Activity activity : schedule.getActivities()) {
                     ScheduledActivity schActivity = ScheduledActivity.create();
                     schActivity.setSchedulePlanGuid(plan.getGuid());
-                    schActivity.setTimeZone(context.getZone());
+                    //schActivity.setTimeZone(context.getZone());
+                    // Use the time zone of the request, not the time zone we've been sequencing against. 
+                    schActivity.setTimeZone(context.getEndsOn().getZone());
```
DwayneJengSage: Minor: Remove commented out code.

Also "the time zone we've been sequencing against" is a little unclear. Maybe it should say something like "the time zone used to parse events"? See other comment about renaming ScheduleContext timezones.

test/org/sagebionetworks/bridge/models/schedules/ActivitySchedulerTest.java
```diff
@@ -324,7 +324,7 @@ public void scheduleIsTranslatedToTimeZoneWhenCreatedAndPersisted() {
         
         // User is in Moscow, however.
         DateTimeZone zone = DateTimeZone.forOffsetHours(3);
-        List<ScheduledActivity> activities = schedule.getScheduler().getScheduledActivities(plan, getContext(zone, DateTime.now().plusDays(1)));
+        List<ScheduledActivity> activities = schedule.getScheduler().getScheduledActivities(plan, getContext(zone, DateTime.now(zone).plusDays(1)));
```
DwayneJengSage: scheduleIsTranslatedToTimeZoneWhenCreatedAndPersisted() also uses DateTime.now() using the default timezone. This test will probably fail when we run this on other machines.

Also, instead of using DateTime.now(), should we be using NOW?

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
DwayneJengSage: Also, a few test cases

1. Scheduler is interpreting the event to the correct date. Examples:
* Enrollment=1487552400000 (2017-02-20T01:00Z), timeZone=-08:00, endsOnTimeZone=-08:00, schedule=daily at 11pm, expected=2017-02-19T23:00-0800
* Enrollment=1487552400000 (2017-02-20T01:00Z), timeZone=+09:00, endsOnTimeZone=+09:00, schedule=daily at 11pm, expected=2017-02-20T23:00+0900

2. Scheduler is interpreting the event to the right local time. Examples:
* Enrollment=1487552400000 (2017-02-20T01:00Z), timeZone=-08:00, endsOnTimeZone=-08:00, schedule=one hour after enrollment, expected=2017-02-19T18:00-0800
* Enrollment=1487552400000 (2017-02-20T01:00Z), timeZone=+09:00, endsOnTimeZone=+09:00, schedule=one hour after enrollment, expected=2017-02-20T11:00+0900

3. Local times having timezone correctly applied. Examples:
* Enrollment=1487552400000 (2017-02-20T01:00Z), timeZone=-08:00, endsOnTimeZone=+09:00, schedule=one hour after enrollment, expected=2017-02-19T18:00+0900 (Should this be the 19th or the 20th?)
* Enrollment=1487552400000 (2017-02-20T01:00Z), timeZone=+09:00, endsOnTimeZone=-08:00, schedule=one hour after enrollment, expected=2017-02-20T11:00-0800

Are these test cases represented in the unit tests?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -457,7 +464,8 @@ public void updateParticipantWithExternalIdValidationAddingId() {
         assertTrue(options.get(DATA_GROUPS).contains("group2"));
         assertTrue(options.get(LANGUAGES).contains("de"));
         assertTrue(options.get(LANGUAGES).contains("fr"));
-        assertNull(options.get(EXTERNAL_IDENTIFIER));
+        assertNull(options.get(EXTERNAL_IDENTIFIER)); // can't set this
+        assertNull(options.get(TIME_ZONE)); // can't set this
```
DwayneJengSage, line 464: Should time zone be a separate test instead of bundling it with external ID tests? Having modular tests makes things easier to test and reason about.

Also, should we have separate tests for adding time zone (no time zone to new time zone) vs changing time zone?

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -395,14 +397,19 @@ private IdentifierHolder saveParticipant(Study study, Set<Roles> callerRoles, St
...
+    private Map<ParticipantOption, String> getAllOptionsButExternalId(Study study, StudyParticipant participant,
+            boolean isNew) {
         Map<ParticipantOption, String> options = Maps.newHashMap();
         for (ParticipantOption option : ParticipantOption.values()) {
             options.put(option, option.fromParticipant(participant));
         }
         if (study.isExternalIdValidationEnabled()) {
             options.remove(EXTERNAL_IDENTIFIER);
         }
+        // You can't update time zone after it is set.
```
DwayneJengSage: Maybe changing the name from TIME_ZONE to something like ORIGINAL_EVENT_TIME_ZONE would be less confusing. Also, updating the comment to explain that we need to store the time zone in the participant, because we didn't store it in the event, and now we can't go back and change it. This means that someone can figure out what's going on by looking at the code instead of relying on tribal knowledge.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceDuplicateTest.java
```diff
@@ -270,24 +268,22 @@ public void testExistingDuplicationScenario() throws Exception {
         verify(schedulePlanService).getSchedulePlans(any(), any());
         
         allWithinQueryWindow(activities, context);
-        //assertEquals(16, activities.size());
-        assertEquals(1, filterByGuid(activities, "bea8fd5d-7622-451f-a727-f9e37f00e1be:2016-06-30T19:43:07.951").size());
-        assertEquals(1, filterByGuid(activities, "6966c3d7-0949-43a8-804e-efc25d0f83e2:2016-06-30T19:43:07.951").size());
-        assertEquals(1, filterByGuid(activities, "79cf1788-a087-4fa3-92e4-92e43d9699a7:2016-06-30T19:43:07.951").size());
-        // NOTE: this task has reset because the timestamp is different
-        // assertNotNull(filterByLabel(activities, "Training Session 1").get(0).getStartedOn());
+        assertEquals(0, filterByGuid(activities, "bea8fd5d-7622-451f-a727-f9e37f00e1be:2016-06-30T12:43:07.951").size());
```
DwayneJengSage: Why does this event disappear, but the others don't?

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceDuplicateTest.java
```diff
@@ -45,7 +45,6 @@
  * a standard time and midnight seems to work regardless of when the timestamp is).
  */
 @RunWith(MockitoJUnitRunner.class)
-@Ignore // The current code does not prevent duplicates.
 public class ScheduledActivityServiceDuplicateTest {
```
DwayneJengSage, line 48: None of these tests seem to test the scenario where the user creates a bunch of events in one timezone, then asks for activities in another timezone. Are we sure this test actually tests what we're testing for?

Also, we test for the presence of tasks we expect. What about the absence of tasks we *don't* expect? iirc, the duplicated tasks showed up because they had different timestamps that were off by an hour (equal to the change in timezone). Should our tests verify that that the task with the wrong timestamp isn't there?

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceDuplicateTest.java
```diff
@@ -300,13 +296,12 @@ public void settingEnrollmetToNextDayUTCWorks() throws Exception {
         verify(schedulePlanService).getSchedulePlans(any(), any());
         
         allWithinQueryWindow(activities, context);
-        assertEquals(1, filterByGuid(activities, "bea8fd5d-7622-451f-a727-f9e37f00e1be:2016-06-30T19:43:07.951").size());
-        assertEquals(1, filterByGuid(activities, "6966c3d7-0949-43a8-804e-efc25d0f83e2:2016-06-30T19:43:07.951").size());
-        assertEquals(1, filterByGuid(activities, "79cf1788-a087-4fa3-92e4-92e43d9699a7:2016-06-30T19:43:07.951").size());
-        // NOTE: this task has reset because the timestamp is different
-        //assertNotNull(filterByLabel(activities, "Training Session 1").get(0).getStartedOn());
+        assertEquals(0, filterByGuid(activities, "bea8fd5d-7622-451f-a727-f9e37f00e1be:2016-06-30T12:43:07.951").size());
```
DwayneJengSage: Similarly, why did this task disappear?

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -408,140 +406,155 @@ public void orderActivitieFiltersAndSorts() {
...
-                                        "'type':'TaskReference'"+
-                                    "},"+
-                                    "'activityType':'task',"+
-                                    "'type':'Activity'"+
+        // As long as time zone is consistent, the right number of tasks will be generated on 
+        // the day of the request, regardless of the hour of the day.
+        executeComplexTestInTimeZone("06", DateTimeZone.forOffsetHours(-7));
+        executeComplexTestInTimeZone("20", DateTimeZone.forOffsetHours(-7));
+        executeComplexTestInTimeZone("06", DateTimeZone.forOffsetHours(3));
+        executeComplexTestInTimeZone("20", DateTimeZone.forOffsetHours(3));
```
DwayneJengSage: Should the offsets be big enough to shift the day to the previous and next days?

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -408,140 +406,155 @@ public void orderActivitieFiltersAndSorts() {
...
+    }
+    
+    private void executeComplexTestInTimeZone(String hourOfDay, DateTimeZone timeZone) throws Exception {
+        DateTime now = DateTime.parse("2017-02-"+hourOfDay+"T22:00:00.000").withZone(timeZone);
```
DwayneJengSage: The variable is called "hourOfDay", but you use it as the day, not the hour.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceRecurringTest.java
```diff
@@ -130,7 +127,7 @@ public void retrievalActivitiesAcrossTimeAndTimeZones() throws Exception {
         String pst1 = DateTime.now(PST).toLocalDate().toString();
         String pst2 = DateTime.now(PST).plusDays(1).toLocalDate().toString();
         String pst3 = DateTime.now(PST).plusDays(2).toLocalDate().toString();
-        String pst4 = DateTime.now(PST).plusDays(3).toLocalDate().toString();
+        //String pst4 = DateTime.now(PST).plusDays(3).toLocalDate().toString();
```
DwayneJengSage: Why is this changing?

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceRecurringTest.java
```diff
@@ -149,11 +146,11 @@ public void retrievalActivitiesAcrossTimeAndTimeZones() throws Exception {
         // the window, over in Moscow... that is not returned because although it exists, we 
         // filter it out from the persisted activities retrieved from the db.
         activities = service.getScheduledActivities(getContextWith2DayWindow(PST));
-        assertEquals(4, activities.size());
+        assertEquals(3, activities.size());
         assertEquals(pst1+"T10:00:00.000-07:00", activities.get(0).getScheduledOn().toString());
         assertEquals(pst2+"T10:00:00.000-07:00", activities.get(1).getScheduledOn().toString());
         assertEquals(pst3+"T10:00:00.000-07:00", activities.get(2).getScheduledOn().toString());
-        assertEquals(pst4+"T10:00:00.000-07:00", activities.get(3).getScheduledOn().toString());
+        //assertEquals(pst4+"T10:00:00.000-07:00", activities.get(3).getScheduledOn().toString());
```
DwayneJengSage: Similarly here?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/dao/ParticipantOption.java
alxdarksage: Added tests

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
alxdarksage: These are all good tests so I've added them to IntervalActivitySchedulerTest as they're variations on that scheduler. They do pass. There are similar test cases but if feel like the more tests checking these issues, the better.

app/org/sagebionetworks/bridge/play/controllers/ScheduledActivityController.java
alxdarksage: Yes, we could split these out and now probably should. I would probably call them PersistedTimeZone (usually captured the first time the user asks for tasks, and intended to insure scheduling is correct on that first day) and requestTimeZone. However, this will lead to a lot of code change. Should I do that separately, after the other changes are reviewed and merged?

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Yes see comment above, I think we should split out the two time zones entirely, but maybe in a next step. It's not exactly the "original event" time zone because we don't capture that when they enroll (create an account), it's the first time zone we get when they are asking for the first time for activities... I would just call this the persisted time zone, but that seems redundant because we're persisting it. Maybe INITIAL_TIME_ZONE.

test/org/sagebionetworks/bridge/models/schedules/ActivitySchedulerTest.java
alxdarksage: Yes, used NOW everywhere, and aligned DateTimeUtils.setCurrentMillisFixed(); with now for consistency (I do not know why these were different but that seems like a mistake to me).

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: We ended up removing the ability to set the time zone when creating/updating a participant.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceDuplicateTest.java
alxdarksage: I added changing endsOn to the MSK time zone in the @before method, and the right tasks are still returned (without duplicates).

ScheduledActivityServiceRecurringTest manipulates the time zone extensively to verify that works correctly, so I also changed the initial persisted time zone to be different from the Moscow/PST time zones used in the test, to verify those tests still work.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceDuplicateTest.java
alxdarksage: There's no task for this because there's two of these tasks with that guid, one finished and one not (the duplicate). The scheduler correctly keeps the finished task and thus, sees it is done and does not return it from the scheduling service. (This test is probably not as useful as it was since these duplicates in production have all long since probably expired.)

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceDuplicateTest.java
alxdarksage: Same reason as above, scheduler determines (despite presence of duplicates in the table) that the task is done.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceRecurringTest.java
alxdarksage: It shouldn't, this is being reverted back to have pst4, I wonder if this wouldn't break this test, it only passes with all four PST dates right now and I don't believe I did anything today that would have caused a failing test to succeed.

test/org/sagebionetworks/bridge/models/schedules/ActivitySchedulerTest.java
```diff
@@ -53,7 +53,7 @@ public void before() {
         plan.setGuid("BBB");
         
         // Day of tests is 2015-04-06T10:10:10.000-07:00 for purpose of calculating expiration
-        DateTimeUtils.setCurrentMillisFixed(1428340210000L);
+        DateTimeUtils.setCurrentMillisFixed(NOW.getMillis());
```
alxdarksage: These were different, which is really confusing. Now freezing system time to NOW which is what you'd expect, this did change the dates in the test but it's all consistent with expectations

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceRecurringTest.java
```diff
@@ -146,11 +144,11 @@ public void retrievalActivitiesAcrossTimeAndTimeZones() throws Exception {
         // the window, over in Moscow... that is not returned because although it exists, we 
         // filter it out from the persisted activities retrieved from the db.
         activities = service.getScheduledActivities(getContextWith2DayWindow(PST));
-        assertEquals(3, activities.size());
+        assertEquals(4, activities.size());
```
alxdarksage: I'm still trying to figure out how this passed with 3 activities, I did introduce EST into this test

app/org/sagebionetworks/bridge/models/accounts/ParticipantOptionsLookup.java
alxdarksage: Sorry, I changed a path where we're parsing JSON to throw a BadRequestException... here in the lookup we're assuming the values have been written correctly to DDB and they should come back correctly... if they don't and it's an IllegalArgumentException, that is translated to a 500 response and that's what we want. We want to catch it if it's wrong when we're reading the values from the user.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: Changed

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: Changed to 06 and 23, which will switch to prior day/next day for one of the two for each time zone.

app/org/sagebionetworks/bridge/play/controllers/ScheduledActivityController.java
DwayneJengSage: I know we agreed to save the big refactor for a later date, but this code change should still at least make sense as is. As is, it's a big confusing mess of poorly named variables and requires a lot of code archeology and tribal knowledge to understand what's going on.

What we definitely need:

Comments describing how there are two different concepts of "time zone" in play here. The first is the time zone that the user wants to get events for. The second is the timezone of which we interpret event epoch millis into local times. The second one seems to exist solely because we were recording event times as epoch millis before we realized we also needed to know what timezone they're in, and now it's untenable to go back and change that.

The local variables in this method don't make sense. "zone" and "timeZone" are incredibly ambiguous and confusing, and without the above clarifying comment, it's impossible to understand what they represent.

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -64,7 +64,8 @@ protected void addScheduledActivityAtTime(List<ScheduledActivity> scheduledActiv
                 for (Activity activity : schedule.getActivities()) {
                     ScheduledActivity schActivity = ScheduledActivity.create();
                     schActivity.setSchedulePlanGuid(plan.getGuid());
-                    schActivity.setTimeZone(context.getZone());
+                    // Use the time zone of the request, not the time zone we've been sequencing against.
```
DwayneJengSage: Similarly, this comment "the time zone we've been sequencing against" is incredibly unclear and confusing. Maybe it should say something like "the time zone used to parse events"?

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -154,6 +155,7 @@ public StudyParticipant getParticipant(Study study, Account account, boolean inc
         builder.withExternalId(lookup.getString(EXTERNAL_IDENTIFIER));
         builder.withDataGroups(lookup.getStringSet(DATA_GROUPS));
         builder.withLanguages(lookup.getOrderedStringSet(LANGUAGES));
+        builder.withTimeZone(lookup.getTimeZone(TIME_ZONE));
```
DwayneJengSage, line 158: There used to be code in this class to prevent TIME_ZONE from being saved on update. I can't seem to find that code anymore.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceDuplicateTest.java
```diff
@@ -270,24 +269,22 @@ public void testExistingDuplicationScenario() throws Exception {
...
         allWithinQueryWindow(activities, context);
-        //assertEquals(16, activities.size());
-        assertEquals(1, filterByGuid(activities, "bea8fd5d-7622-451f-a727-f9e37f00e1be:2016-06-30T19:43:07.951").size());
-        assertEquals(1, filterByGuid(activities, "6966c3d7-0949-43a8-804e-efc25d0f83e2:2016-06-30T19:43:07.951").size());
-        assertEquals(1, filterByGuid(activities, "79cf1788-a087-4fa3-92e4-92e43d9699a7:2016-06-30T19:43:07.951").size());
-        // NOTE: this task has reset because the timestamp is different
-        // assertNotNull(filterByLabel(activities, "Training Session 1").get(0).getStartedOn());
+        assertEquals(0, filterByGuid(activities, "bea8fd5d-7622-451f-a727-f9e37f00e1be:2016-06-30T12:43:07.951").size());
+        assertEquals(1, filterByGuid(activities, "6966c3d7-0949-43a8-804e-efc25d0f83e2:2016-06-30T12:43:07.951").size());
+        assertEquals(1, filterByGuid(activities, "79cf1788-a087-4fa3-92e4-92e43d9699a7:2016-06-30T12:43:07.951").size());
```
DwayneJengSage: I don't believe these tests are sufficient. You test that the activity with the correct timestamp (12:43) is present, but we don't test that the activity with the wrong timestampp (11:43) is absent. So we're not really testing for duplicate. Every test in this class seems to have the same problem.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceDuplicateTest.java
```diff
@@ -270,24 +269,22 @@ public void testExistingDuplicationScenario() throws Exception {
         verify(schedulePlanService).getSchedulePlans(any(), any());
         
         allWithinQueryWindow(activities, context);
-        //assertEquals(16, activities.size());
-        assertEquals(1, filterByGuid(activities, "bea8fd5d-7622-451f-a727-f9e37f00e1be:2016-06-30T19:43:07.951").size());
-        assertEquals(1, filterByGuid(activities, "6966c3d7-0949-43a8-804e-efc25d0f83e2:2016-06-30T19:43:07.951").size());
-        assertEquals(1, filterByGuid(activities, "79cf1788-a087-4fa3-92e4-92e43d9699a7:2016-06-30T19:43:07.951").size());
-        // NOTE: this task has reset because the timestamp is different
-        // assertNotNull(filterByLabel(activities, "Training Session 1").get(0).getStartedOn());
+        assertEquals(0, filterByGuid(activities, "bea8fd5d-7622-451f-a727-f9e37f00e1be:2016-06-30T12:43:07.951").size());
```
DwayneJengSage: Why is this activity absent for this test, but present for the other tests?

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -408,140 +406,155 @@ public void orderActivitieFiltersAndSorts() {
...
+        // As long as time zone is consistent, the right number of tasks will be generated on 
+        // the day of the request, regardless of the hour of the day.
+        executeComplexTestInTimeZone("06", DateTimeZone.forOffsetHours(-7));
+        executeComplexTestInTimeZone("23", DateTimeZone.forOffsetHours(-7));
+        executeComplexTestInTimeZone("06", DateTimeZone.forOffsetHours(3));
+        executeComplexTestInTimeZone("23", DateTimeZone.forOffsetHours(3));
+    }
+    
+    private void executeComplexTestInTimeZone(String dayOfMonth, DateTimeZone timeZone) throws Exception {
+        DateTime now = DateTime.parse("2017-02-"+dayOfMonth+"T22:00:00.000").withZone(timeZone);
```
DwayneJengSage: I think the change is wrong. Semantically, you clearly meant hour, since that's the only time the timezones make sense.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -408,140 +406,155 @@ public void orderActivitieFiltersAndSorts() {
...
-                                        "'type':'TaskReference'"+
-                                    "},"+
-                                    "'activityType':'task',"+
-                                    "'type':'Activity'"+
+        // As long as time zone is consistent, the right number of tasks will be generated on 
+        // the day of the request, regardless of the hour of the day.
+        executeComplexTestInTimeZone("06", DateTimeZone.forOffsetHours(-7));
+        executeComplexTestInTimeZone("23", DateTimeZone.forOffsetHours(-7));
+        executeComplexTestInTimeZone("06", DateTimeZone.forOffsetHours(3));
+        executeComplexTestInTimeZone("23", DateTimeZone.forOffsetHours(3));
```
DwayneJengSage: This change is also wrong. 06:00-0700 converts to 13:00Z and 06:00+0300 converts to 03:00Z, so you're still not breaking the day boundaries.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -408,140 +406,155 @@ public void orderActivitieFiltersAndSorts() {
...
+                
+                ScheduleContext context = new ScheduleContext.Builder()
+                    .withClientInfo(info)
+                    .withStudyIdentifier("test-study")
+                    .withUserDataGroups(Sets.newHashSet("parkinson","test_user"))
+                        .withEndsOn(DateTime.now().plusDays(1).withHourOfDay(23).withMinuteOfHour(59).withSecondOfMinute(59))
+                        .withTimeZone(timeZone)
+                    .withHealthCode("AAA")
+                    .withUserId(USER_ID)
+                    .withAccountCreatedOn(DateTime.now().minusDays(4))
```
DwayneJengSage: endsOn() and accountCreatedOn() uses DateTime.now() with the default timezone, which means this test is almost certainly not using the timezone you expect it to.

In fact, given how timezones is pretty critical in the whole scheduling thing, we should probably change all instances of DateTime.now() to a fixed time that we know the timezone of.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceRecurringTest.java
```diff
@@ -199,16 +195,16 @@ public void persistedActivitiesAreFilteredByEndsOn() throws Exception {
...
+    private ScheduleContext getContext(DateTimeZone persistedZone, DateTime endsOn) {
         // Setting the endsOn value to the end of the day, as we do in the controller.
         return new ScheduleContext.Builder()
             .withStudyIdentifier(study.getStudyIdentifier())
             .withClientInfo(ClientInfo.UNKNOWN_CLIENT)
-            .withTimeZone(zone)
+            .withTimeZone(persistedZone)
             .withAccountCreatedOn(DateTime.now())
             // Setting the endsOn value to the end of the day, as we do in the controller.
             .withEndsOn(endsOn.withHourOfDay(23).withMinuteOfHour(59).withSecondOfMinute(59))
```
DwayneJengSage: Similarly, this test uses endsOn() with the default timezone. Could that explain why we see inconsistent results with this test?

test/org/sagebionetworks/bridge/models/schedules/ActivitySchedulerTest.java
DwayneJengSage: Minor: Update comment to reflect reality.

test/org/sagebionetworks/bridge/models/schedules/IntervalActivitySchedulerTest.java
```diff
@@ -608,10 +610,91 @@ public void eventIsLateUTC() {
...
+    }
+    
+    private String firstTimeStampFor(int initialTZOffset, int requestTZOffset, Schedule schedule) {
+        DateTime enrollment = new DateTime(1487552400000L); // 2017-02-20T01:00Z
+        DateTimeZone initialTimeZone = DateTimeZone.forOffsetHours(initialTZOffset);
+        DateTimeZone requestTimeZone = DateTimeZone.forOffsetHours(requestTZOffset);
+        
+        // NOTE: The events map is adjusted to initial time zone in the service, we must do that
+        // here for this test to pass. Just calling scheduler without a "localized" event map will fail.
+        events.put("enrollment", enrollment.withZone(initialTimeZone));
```
DwayneJengSage: Is this because the logic for parsing the enrollment time lives in ScheduledActivityService, but this test is in ActivityScheduler? If so, I would argue that this test doesn't fully exercise our code. Rather, we make an assumption about how callers will behave, which may not be true if there's a bug in the caller.

We should move this test up to ScheduledActivityService to verify our fix works end-to-end.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceRecurringTest.java
```diff
@@ -97,48 +100,40 @@ public void after() {
...
-        // We start this test in the early morning in Russia, in the future so the new user's
-        // enrollment doesn't screw up the test.
-        DateTimeUtils.setCurrentMillisFixed(targetDateTime.getMillis());
-
-        // These time zones are far apart and for our chosen time, Dave will be teleporting to the 
-        // previous day. Our scheduler must do something rational.
-        DateTimeZone MSK = DateTimeZone.forOffsetHours(3);
-        DateTimeZone PST = DateTimeZone.forOffsetHours(-7);
+        // We start this test in the early morning in Russia
+        DateTimeUtils.setCurrentMillisFixed(targetDateTime.withZone(MSK).getMillis());
```
DwayneJengSage: Minor: Epoch milliseconds doesn't depend on time zone.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/play/controllers/ScheduledActivityController.java
alxdarksage: Rewrote these comments and renamed the context time zone to initialTimeZone (in contract to requestTimeZone).

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
alxdarksage: Rewrote comment

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Yes thanks for spotting this, in merging a conflict yesterday with the refactor of ParticipantService, I missed that. I've added it back.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceDuplicateTest.java
alxdarksage: The persisted tasks include a task with this GUID that has been finished, so it should be removed from the results. Another test executes without any of the persisted tasks, to verify that it is returned in that case. There's also a duplication of this task so when it is returned, there should only be 1.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceDuplicateTest.java
alxdarksage: These should not have the time portion in the string, I intended to match the GUIDs only of all the tasks returned. We're verifying there are no duplicates (in one recurring case, I search by Label because the plan produces three activities, I just want to verify that "Training Session 1" occurs only one time. Removed all the timestamps.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: Yes, sorry. I eliminated all the string parsing nonsense, and and also introduced a NOW variable and used it everywhere in addition to DateTimeUtils.setCurrentMillisFixed(). Verified these now are on different days.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: Yes, totally wrong. Changed using Joda's API.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: Rewrote this test to use a fixed NOW value (this was also fixed in @before with DateTimeUtils.setCurrentMillisFixed() but this is not very easy to follow; actually forcing now to be supplied to ScheduleContext, rather than using DateTime.now() in the build method, might also help reduce time errors in these tests).

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceRecurringTest.java
alxdarksage: Removed

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceRecurringTest.java
alxdarksage: I've rewritten this test to use NOW instead of DateTimeUtils functions, and I'm now setting the NOW value in the schedule context, and removed all uses of DateTime.now(). I believe this should make the tests stable whenever they are run (we shouldn't rely on system time at all in the scheduler).

test/org/sagebionetworks/bridge/models/schedules/ActivitySchedulerTest.java
alxdarksage: Updated

test/org/sagebionetworks/bridge/models/schedules/IntervalActivitySchedulerTest.java
alxdarksage: Moved this to the ScheduledEventService, removed application of the time zone to the map.

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -256,12 +258,15 @@ public void updateParticipant(Study study, Set<Roles> callerRoles, StudyParticip
...
-            Account account, StudyParticipant participant) {
+            Account account, StudyParticipant participant, boolean isUpdate) {
         for (ParticipantOption option : ParticipantOption.values()) {
             options.put(option, option.fromParticipant(participant));
         }
         // External identifier is handled by the ExternalIdService
         options.remove(EXTERNAL_IDENTIFIER);
+        if (isUpdate) {
+            options.remove(TIME_ZONE);
+        }
```
DwayneJengSage: As per IM discussion, time zone should always come from the getActivities, and not from createParticipant. With that in mind, we can remove the isUpdate flag and always remove TIME_ZONE.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -408,140 +409,155 @@ public void orderActivitieFiltersAndSorts() {
...
-                                        "'type':'TaskReference'"+
-                                    "},"+
-                                    "'activityType':'task',"+
-                                    "'type':'Activity'"+
+        // As long as time zone is consistent, the right number of tasks will be generated on 
+        // the day of the request, regardless of the hour of the day.
+        executeComplexTestInTimeZone(3, DateTimeZone.forOffsetHours(-7));
+        executeComplexTestInTimeZone(3, DateTimeZone.forOffsetHours(-7));
+        executeComplexTestInTimeZone(23, DateTimeZone.forOffsetHours(8));
+        executeComplexTestInTimeZone(23, DateTimeZone.forOffsetHours(8));
```
DwayneJengSage: I think you got these backwards again, I think you meant hour 3 offset -7, hour 23 offset -7, hour 3 offset +8, hour 23 offset +8. Instead, you did hour 3 offset -7 (which converts to 10:00Z) and hour 23 offset +8 (which converts to 15:00Z), and you did each test twice.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceRecurringTest.java
```diff
@@ -97,48 +100,37 @@ public void after() {
...
         
-        // We start this test in the early morning in Russia, in the future so the new user's
-        // enrollment doesn't screw up the test.
-        DateTimeUtils.setCurrentMillisFixed(targetDateTime.getMillis());
-
-        // These time zones are far apart and for our chosen time, Dave will be teleporting to the 
-        // previous day. Our scheduler must do something rational.
-        DateTimeZone MSK = DateTimeZone.forOffsetHours(3);
-        DateTimeZone PST = DateTimeZone.forOffsetHours(-7);
+        DateTimeUtils.setCurrentMillisSystem();
```
DwayneJengSage, line 110: Minor: Did you mean current system time or NOW? Can current time of day change the veracity of the test? Or do we believe at this point that the scheduler is stable regardless of time of day?

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -408,140 +409,155 @@ public void orderActivitieFiltersAndSorts() {
...
+    }
+    
+    private void executeComplexTestInTimeZone(int hourOfDay, DateTimeZone timeZone) throws Exception {
+        DateTime now = NOW.withHourOfDay(hourOfDay).withZone(timeZone);
+        DateTimeUtils.setCurrentMillisFixed(now.getMillis());
```
DwayneJengSage: I think we over-corrected. This test now doesn't depend on the inputs at all, which means we're probably not testing what we think we're testing.

DwayneJengSage (changes requested):
I think we're close. Just a few more small comments.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: DateTimeUtils set the time to create the user, but otherwise the now value (which is adjusted) is used through the rest of the test (consistently, e.g. I am setting now in the schedule context which I do not do in enough of the tests). So it should use the "now" value in the test.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceRecurringTest.java
alxdarksage: Yes I changed the test to remove reliance on DateTime.now(), except that this one call to create a user, will not take a timestamp. So I set the time using DateTimeUtils while creating the user, then change it back to system time. Which should not be used in the scheduler.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: I can remove the setCurrentMillisSystem() call... it should not make any difference because system time/DateTime.now() shouldn't be used by the scheduler.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -408,140 +409,155 @@ public void orderActivitieFiltersAndSorts() {
...
+        // the day of the request, regardless of the hour of the day.
+        executeComplexTestInTimeZone(3, DateTimeZone.forOffsetHours(-7));
+        executeComplexTestInTimeZone(23, DateTimeZone.forOffsetHours(-7));
+        executeComplexTestInTimeZone(3, DateTimeZone.forOffsetHours(8));
+        executeComplexTestInTimeZone(23, DateTimeZone.forOffsetHours(8));
+    }
+    
+    private void executeComplexTestInTimeZone(int hourOfDay, DateTimeZone timeZone) throws Exception {
+        DateTime now = NOW.withZone(timeZone).withHourOfDay(hourOfDay);
+        DateTimeUtils.setCurrentMillisFixed(now.getMillis());
```
DwayneJengSage: This test still doesn't depend on "now", which means we're not actually testing what we think we're testing.

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -256,12 +258,13 @@ public void updateParticipant(Study study, Set<Roles> callerRoles, StudyParticip
     }
 
     private void updateAccountOptionsAndRoles(Study study, Set<Roles> callerRoles, Map<ParticipantOption, String> options,
-            Account account, StudyParticipant participant) {
+            Account account, StudyParticipant participant, boolean isUpdate) {
```
DwayneJengSage: isUpdate no longer in use?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Yes, removed from method signature.

DwayneJengSage (approved):

PR #1312, opened 2016-12-08
Fix for one-time task without times having a time portion in local time zone

alxdarksage:
Some tests are disabled that were added around the duplicate task issue... that's definitely not fixed by this. That fix is in another PR.


Erin-Mounts:
looks good to me, modulo my utter lack of context for the code this is embedded in

liujoshua (approved):

PR #1316, opened 2016-12-13
Update libs

alxdarksage:
- updating to AWS SDK 1.10.20;
- removing a few marshallers that we'd have to convert but that weren't used anywhere
- tests pass; also ran the integration tests against these changes, and all pass


DwayneJengSage (commented):
Changes look good. As per offline discussion, it may or may not be worth adding unit tests for:

* JsonNodeMarshaller
* StringSetMarshaller

DwayneJengSage (approved):
Looks good. Feel free to merge whenever you're ready.

PR #1321, opened 2016-12-20
Adding public to the study index record, and create a public endpoint for retrieving these

alxdarksage:
Needed for the Resilience Project.

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoReportIndexDao.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoReportIndexDao.java
@@ -2,6 +2,8 @@ package org.sagebionetworks.bridge.dynamodb;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import java.util.Map;
+
 import javax.annotation.Resource;
 
 import org.springframework.stereotype.Component;
@@ -16,9 +18,19 @@ import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
 
 import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper;
 import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBQueryExpression;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBSaveExpression;
+import com.amazonaws.services.dynamodbv2.model.AttributeValue;
+import com.amazonaws.services.dynamodbv2.model.ConditionalCheckFailedException;
+import com.amazonaws.services.dynamodbv2.model.ExpectedAttributeValue;
+import com.google.common.collect.ImmutableMap;
 
 @Component
 public class DynamoReportIndexDao implements ReportIndexDao {
+
+    private static final DynamoDBSaveExpression DOES_NOT_EXIST_EXPRESSION = new DynamoDBSaveExpression()
+            .withExpected(new ImmutableMap.Builder<String,ExpectedAttributeValue>()
+                    .put("key", new ExpectedAttributeValue(false))
+                    .put("identifier", new ExpectedAttributeValue(false)).build());
     
     private DynamoDBMapper mapper;
 
@@ -28,6 +40,17 @@ public class DynamoReportIndexDao implements ReportIndexDao {
     }
     
     @Override
+    public ReportIndex getIndex(ReportDataKey key) {
+        checkNotNull(key);
+        
+        DynamoReportIndex hashKey = new DynamoReportIndex();
+        hashKey.setKey(key.getIndexKeyString());
+        hashKey.setIdentifier(key.getIdentifier());
+        
+        return mapper.load(hashKey);
+    }
+
+    @Override
     public void addIndex(ReportDataKey key) {
         checkNotNull(key);
         
@@ -35,11 +58,11 @@ public class DynamoReportIndexDao implements ReportIndexDao {
         index.setKey(key.getIndexKeyString());
         index.setIdentifier(key.getIdentifier());
 
-        // Don't recreate if it exists. There isn't a non-key attribute to use for a 
-        // conditional save operation, so load, test, then save.
-        DynamoReportIndex dbIndex = mapper.load(index);
-        if (dbIndex == null) {
-            mapper.save(index);
+        try {
+            mapper.save(index, DOES_NOT_EXIST_EXPRESSION);    
+        } catch(ConditionalCheckFailedException e) {
+            // Do not throw an exception if the index already exists. This is called as a side
+            // effect of saving a report, and can be called multiple times for a report.
         }
     }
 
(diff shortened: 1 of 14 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoReportIndexDao.java
```diff
@@ -33,8 +34,13 @@ public void addIndex(ReportDataKey key) {
...
         index.setIdentifier(key.getIdentifier());
-        
-        mapper.save(index);
+
+        // Don't recreate if it exists. There isn't a non-key attribute to use for a 
+        // conditional save operation, so load, test, then save.
+        DynamoReportIndex dbIndex = mapper.load(index);
+        if (dbIndex == null) {
+            mapper.save(index);
+        }
```
DwayneJengSage, line 66: There is a way to do a conditional save on the key. See for example DynamoUploadSchema https://github.com/Sage-Bionetworks/BridgePF/blob/develop/app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java#L56 Might take a little massaging to get this to work with hash-and-range key.

Upsides are that this will make the save operation atomic (as well as save a tiny amount of Read Cap). Downsides are this throws a ConditionalCheckFailedException.

app/org/sagebionetworks/bridge/dynamodb/DynamoReportIndexDao.java
```diff
@@ -52,6 +58,21 @@ public void removeIndex(ReportDataKey key) {
...
+        DynamoReportIndex hashKey = new DynamoReportIndex();
+        hashKey.setKey(index.getKey());
+        hashKey.setIdentifier(index.getIdentifier());
+        
+        DynamoReportIndex dbIndex = mapper.load(hashKey);
+        if (dbIndex == null) {
+            throw new EntityNotFoundException(ReportIndex.class);
+        }
+        mapper.save(index);
+    }
```
DwayneJengSage, line 102: Similarly, I think you can do a conditional save here.

app/org/sagebionetworks/bridge/dynamodb/DynamoReportIndexDao.java
```diff
@@ -52,6 +58,21 @@ public void removeIndex(ReportDataKey key) {
...
+        checkNotNull(index);
+        
+        DynamoReportIndex hashKey = new DynamoReportIndex();
+        hashKey.setKey(index.getKey());
+        hashKey.setIdentifier(index.getIdentifier());
+        
+        DynamoReportIndex dbIndex = mapper.load(hashKey);
+        if (dbIndex == null) {
+            throw new EntityNotFoundException(ReportIndex.class);
+        }
```
DwayneJengSage, line 101: Is it worth adding a unit test to test updating a non-existent index?

app/org/sagebionetworks/bridge/models/DateRangeResourceList.java
```diff
@@ -4,13 +4,18 @@
...
 public class DateRangeResourceList<T> {
     
     private final List<T> items;
     private final LocalDate startDate;
     private final LocalDate endDate;
 
-    public DateRangeResourceList(List<T> items, LocalDate startDate, LocalDate endDate) {
+    @JsonCreator
+    public DateRangeResourceList(@JsonProperty("items") List<T> items, @JsonProperty("startDate") LocalDate startDate,
+            @JsonProperty("endDate") LocalDate endDate) {
```
DwayneJengSage, line 18: Is it worth adding a deserialization test to DateRangeResourceListTest?

app/org/sagebionetworks/bridge/play/controllers/ReportController.java
```diff
@@ -254,6 +276,53 @@ public Result deleteStudyReportRecord(String identifier, String dateString) {
...
+        
+        return okResult("Report index updated.");
+    }
+    
+    private void verifyIndex(final StudyIdentifier studyId, final String identifier) {
+        reportService.getReportIndices(studyId, ReportType.STUDY).getItems().stream()
+            .filter(index -> index.isPublic() && index.getIdentifier().equals(identifier))
+            .findFirst()
+            .orElseThrow(() -> new EntityNotFoundException(ReportIndex.class));
+    }
```
DwayneJengSage, line 329: This seems incredibly inefficient, especially given that the identifier is already a range key on the ReportIndex table. We can get the ReportIndex using a simple hash-and-range key query.

Also, the findFirst() is redundant, since there will only ever be at most one ReportIndex with that specific study, report type, and identifier combination.

test/org/sagebionetworks/bridge/play/controllers/ReportControllerTest.java
```diff
@@ -410,8 +415,76 @@ public void adminCanDeleteParticipantIndex() {
     }
     
     @Test(expected = UnauthorizedException.class)
-    public void nonAdminCannotDeleteParticipantIndex() {
-        controller.deleteParticipantReportIndex(REPORT_ID);
+    public void nonAdminCannotDeleteParticipantIndex() throws Exception {
+        Result result = controller.deleteParticipantReportIndex(REPORT_ID);
+        assertResult(result);
```
DwayneJengSage: Will this assert ever fire? From a reading of the code and the test, it looks like the delete() call will always through UnauthorizedException.

test/org/sagebionetworks/bridge/play/controllers/ReportControllerTest.java
```diff
@@ -410,8 +415,76 @@ public void adminCanDeleteParticipantIndex() {
...
+    
+    @Test(expected = EntityNotFoundException.class)
+    public void missingPublicStudyReturns404() throws Exception {
+        controller.getPublicStudyReport(TEST_STUDY.getIdentifier(), "does-not-exist", "2016-05-02", "2016-05-09");
+    }
+    
+    @Test(expected = EntityNotFoundException.class)
+    public void privatePublicStudyReturns404() throws Exception {
+        controller.getPublicStudyReport(TEST_STUDY.getIdentifier(), REPORT_ID, "2016-05-02", "2016-05-09");
     }
```
DwayneJengSage, line 504: I don't see anywhere in the test where we create this report. Are we sure this test adequately covers the "private report" case?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/dynamodb/DynamoReportIndexDao.java
alxdarksage: Found this and added. For addIndex we do want to silently return if index exists, so this is now swallowing the exception.

app/org/sagebionetworks/bridge/dynamodb/DynamoReportIndexDao.java
alxdarksage: Added

app/org/sagebionetworks/bridge/models/DateRangeResourceList.java
alxdarksage: Added

app/org/sagebionetworks/bridge/play/controllers/ReportController.java
alxdarksage: Added a getIndex() method to the ReportIndexDao in order to retrieve this, changed code.

test/org/sagebionetworks/bridge/play/controllers/ReportControllerTest.java
alxdarksage: No it sure won't, removed it

test/org/sagebionetworks/bridge/play/controllers/ReportControllerTest.java
alxdarksage: Yes, thanks for catching this. I added the report and index with public=false.

DwayneJengSage:
Looks good except for one comment.

You made some changes to DateRangeResourceList, and we decided to add deserialization test to DateRangeResourceListTest. However, I'm not seeing that test. Did you forget to git add?

app/org/sagebionetworks/bridge/models/DateRangeResourceList.java
alxdarksage: Sorry, now it's really added

alxdarksage:
Added deserialization test to the end of DateRangeResourceListTest.

DwayneJengSage (approved):

PR #1322, opened 2016-12-22
Upload Validation ignore empty uploads

DwayneJengSage:
Right now, it's possible to submit an upload with an empty attachment. This leads to wonky things down the line. This change adds logic to ignore empty uploads, treating them as if they didn't exist.

Also fixed a non-failing NullPointerException in the unit tests.

Testing done:
* updated unit tests
* ran upload integ tests
* manual tested by uploading an upload with an empty attachment


alxdarksage:
Looks good to merge. Do you want to log when this happens, or anything like that, even if it's not an error?

PR #1323, opened 2017-01-03
deactivate study

zaclinsage:


app/org/sagebionetworks/bridge/dynamodb/DynamoStudyDao.java
```diff
@@ -59,14 +60,39 @@ public Study getStudy(String identifier) {
...
     @Override
     public List<Study> getStudies() {
         DynamoDBScanExpression scan = new DynamoDBScanExpression();
-        
-        List<DynamoStudy> mappings = mapper.scan(DynamoStudy.class, scan);
+
+        // get all active studies
+        List<DynamoStudy> mappings = mapper.scan(DynamoStudy.class, scan).stream()
+                .filter(dynamoStudy -> dynamoStudy.isActive())
+                .collect(Collectors.toList());
```
alxdarksage: You can add this as a criteria to the scan itself, which would be slightly better for performance once we have lots of these deactivated studies down the road. It then becomes easier later to add a flag to include or not include deactivated studies... you just include this filtering criteria in the scan, or not.

app/org/sagebionetworks/bridge/dynamodb/DynamoStudyDao.java
```diff
@@ -105,4 +131,18 @@ public void deleteStudy(Study study) {
 
         mapper.delete(study);
     }
+
+    @Override
+    public void deactivateStudy(Study study) {
+        checkNotNull(study, Validate.CANNOT_BE_BLANK, "study");
```
alxdarksage: If you have a deactivate method, I would have it only take the study ID: then load the study, changing this one property only, before updating the study. This way, there can be no side effects to deactivation (no other fields or properties changed). Deactivated studies need to be capable of being re-enabled; I think this can be done as part of the normal update call for admins.

app/org/sagebionetworks/bridge/play/controllers/StudyController.java
```diff
@@ -140,12 +140,17 @@ public Result createSynapse(String synapseUserId) throws Exception {
...
         getAuthenticatedSession(ADMIN);
         if (studyWhitelist.contains(identifier)) {
             return forbidden(Json.toJson(identifier + " is protected by whitelist."));
         }
-        studyService.deleteStudy(identifier);
+        if ("true".equals(physical)) {
+            studyService.deleteStudy(identifier);
+        } else {
+            studyService.deactivateStudy(identifier);
+        }
```
alxdarksage: I would prefer this branching logic be moved to the service... as a general rule, controllers should only be concerned with converting HTTP requests into an equivalent Java call, and any logic based on that can go in the service. So here, just converting the physical flag to a boolean before calling studyService.deleteStudy(identifier, physical).

app/org/sagebionetworks/bridge/dynamodb/DynamoStudyDao.java
```diff
@@ -59,14 +60,39 @@ public Study getStudy(String identifier) {
...
+        checkArgument(isNotBlank(identifier), Validate.CANNOT_BE_BLANK, "identifier");
+
+        DynamoStudy study = new DynamoStudy();
+        study.setIdentifier(identifier);
+        study = mapper.load(study);
+        if (study == null) {
+            throw new EntityNotFoundException(Study.class, "Study '"+identifier+"' not found.");
+        }
+
         return study;
```
alxdarksage, line 64: As in getStudies() below, you could add an includeDeleted boolean flag to the existing getStudy() call, and locally use that method with true rather than false. This is less important since I don't believe you can get a study until you re-enable it, outside of the DAO itself.

(As discussed, the service can provide overloaded methods that always set this to false, so everything continues to compile.)

For the use case of an admin restoring a deactivated study, the admin needs to be able to retrieve all studies, including deactivated studies, and should be able to update a deactivated study with the "active" flag set to false (I don't think this needs to be exposed as a separate API endpoint, but note that setting the active flag should be protected by the "isAdminUpdate" flag in the StudyService). Then getStudy() and everything else will work again.

test/org/sagebionetworks/bridge/dynamodb/DynamoStudyDaoTest.java
```diff
@@ -52,7 +52,7 @@ public void before() {
     public void after() {
         for (String oneStudyId : studyIdsToDelete) {
             try {
-                Study study = studyDao.getStudy(oneStudyId);
+                Study study = studyDao.getStudyOnlyForDao(oneStudyId);
                 studyDao.deleteStudy(study);
```
alxdarksage, line 56: Can a user delete a deactivated study? They should be able to do this. I don't think they should need to activate the study before they can delete it, either. This may require changing the delete to only take a studyId.

app/org/sagebionetworks/bridge/dynamodb/DynamoStudyDao.java
```diff
@@ -59,14 +60,39 @@ public Study getStudy(String identifier) {
         if (study == null) {
             throw new EntityNotFoundException(Study.class, "Study '"+identifier+"' not found.");
         }
+        if (!study.isActive()) {
+            throw new EntityNotFoundException(Study.class, "Study '"+identifier+"' is de-activated.");
+        }
```
alxdarksage: From the outside world, the error message should be same: "Study 'foo' not found." For all intents and purposes, it has been "deleted" for everyone but admins. You can combine these two tests.

alxdarksage (changes requested):
In this checkin, you should probably also include deactivated studies for admins in the getStudies() call. If everything is done right, all deletes will continue to be physical, so the resulting studies should be exactly the same as before. But then you can get feedback in integration tests/researcher UI to verify the studies have only been deactivated... otherwise I don't see how you could tell vs. the studies being permanently deleted.

Also important because admins should be able to re-activate a study, or delete it from the de-activated state, and these can be tested in this PR as well.

app/org/sagebionetworks/bridge/play/controllers/StudyController.java
```diff
@@ -140,12 +141,18 @@ public Result createSynapse(String synapseUserId) throws Exception {
...
         getAuthenticatedSession(ADMIN);
         if (studyWhitelist.contains(identifier)) {
             return forbidden(Json.toJson(identifier + " is protected by whitelist."));
         }
-        studyService.deleteStudy(identifier);
+        if ("true".equals(physical)) {
+            studyService.deleteStudy(identifier, true);
+        } else {
+            studyService.deleteStudy(identifier, false);
+        }
```
alxdarksage: studyService.deleteStudy(identifier, Boolean.valueOf(physical)); // would be terser

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -121,24 +122,37 @@ public final void setSynapseClient(SynapseClient synapseClient) {
         this.synapseClient = synapseClient;
     }
 
-    public Study getStudy(String identifier) {
+    private Study getStudy(String identifier, boolean containDeactivated) {
```
alxdarksage: boolean includeDeleted more consistent with surveys

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -121,24 +122,37 @@ public final void setSynapseClient(SynapseClient synapseClient) {
...
+
         Study study = cacheProvider.getStudy(identifier);
         if (study == null) {
             study = studyDao.getStudy(identifier);
             cacheProvider.setStudy(study);
         }
+
+        if (study != null && !study.isActive() && !containDeactivated) {
+            throw new EntityNotFoundException(Study.class, "Study '"+identifier+"' is de-activated.");
+        }
```
alxdarksage, line 137: I think the logic is wrong... if study is null, we DO thrown ENFE, don't we? study.isActive/containDeactivated are paired relative to that:

(study == null || (!study.isActive() && !containDeactivated))

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -225,19 +239,30 @@ public Study createSynapseProjectTeam(Long userId, Study study) throws SynapseEx
...
+
         study.setStormpathHref(originalStudy.getStormpathHref());
-        study.setActive(true);
-        // And this cannot be set unless you're an administrator. Regardless of what the 
+        // And this cannot be set unless you're an administrator. Regardless of what the
         // developer set, set these back to the original study.
         if (!isAdminUpdate) {
+            // prevent non-admins update a deactivated study
+            if (!originalStudy.isActive()) {
+                throw new EntityNotFoundException(Study.class, "Study '"+ study.getIdentifier() +"' is deactivated.");
```
alxdarksage: Study not found (don't use word deactivated in API). Also if you're going to throw a 404, you might as well do it before doing any further work, so I would get the original study and check it at the top of the method. Same with the exception you throw below.

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -225,19 +239,30 @@ public Study createSynapseProjectTeam(Long userId, Study study) throws SynapseEx
...
+            }
+
             study.setHealthCodeExportEnabled(originalStudy.isHealthCodeExportEnabled());
             study.setEmailVerificationEnabled(originalStudy.isEmailVerificationEnabled());
         }
+
+        // prevent anyone changing active to false -- it should be done by deactivateStudy() method
+        if (originalStudy.isActive() && !study.isActive()) {
+            throw new IllegalAccessError("Nobody can set field of active to false in update method.");
+        }
```
alxdarksage, line 262: throw BadRequestException, "Study cannot be deleted on update." Otherwise this is returned as a 500 error I believe, and really it's a 400.

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -259,23 +284,32 @@ public Study updateStudy(Study study, boolean isAdminUpdate) {
...
             throw new EntityNotFoundException(Study.class, "Study '"+identifier+"' not found");
         }
-        studyDao.deleteStudy(existing);
-        directoryDao.deleteDirectoryForStudy(existing);
-        subpopService.deleteAllSubpopulations(existing.getStudyIdentifier());
-        cacheProvider.removeStudy(identifier);
+
+        if (!physical) {
+            // deactivate
+            if (!existing.isActive()) throw new EntityNotFoundException(Study.class, "Study '"+identifier+"' is deactivated before.");
```
alxdarksage: Use curly braces please

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -259,23 +284,32 @@ public Study updateStudy(Study study, boolean isAdminUpdate) {
...
+            studyDao.deactivateStudy(existing.getIdentifier());
```
alxdarksage, line 308: I think you need to remove it from the cache as well. It might show up in certain calls if you don't, not sure

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
```diff
@@ -121,6 +122,59 @@ private void assertDirectoryUpdated(Consumer<Study> consumer) {
...
+        when(studyDao.getStudy(study.getIdentifier())).thenReturn(study);
+
+        service.deleteStudy(study.getIdentifier(), false);
+    }
+
+    @Test(expected = EntityNotFoundException.class)
+    public void deactivateStudyNotFound() {
+        Study study = getTestStudy();
+        study.setActive(false);
+        when(studyDao.getStudy(study.getIdentifier())).thenReturn(null);
```
alxdarksage, line 141: Does this method return null, rather than throwing EntityNotFoundException? Just checking. Also I would verify that the dao is not called to delete the study.

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
```diff
@@ -121,6 +122,59 @@ private void assertDirectoryUpdated(Consumer<Study> consumer) {
...
+        service.deleteStudy(study.getIdentifier(), false);
+    }
+
+    @Test(expected = EntityNotFoundException.class)
+    public void nonAdminsCannotUpdateDeactivatedStudy() {
+        Study study = getTestStudy();
+        study.setActive(false);
+        when(studyDao.getStudy(study.getIdentifier())).thenReturn(study);
+
+        service.updateStudy(study, false);
```
alxdarksage, line 156: verify that the dao was never called to update study?

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
```diff
@@ -121,6 +122,59 @@ private void assertDirectoryUpdated(Consumer<Study> consumer) {
...
+    public void nonAdminsCannotSetActiveToFalse() {
+        Study originalStudy = getTestStudy();
+        originalStudy.setActive(true);
+        when(studyDao.getStudy(originalStudy.getIdentifier())).thenReturn(originalStudy);
+
+        Study study = getTestStudy();
+        study.setIdentifier(originalStudy.getIdentifier());
+        study.setActive(false);
+
+        service.updateStudy(study, false);
```
alxdarksage, line 171: Verify dao was not called

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
```diff
@@ -121,6 +122,59 @@ private void assertDirectoryUpdated(Consumer<Study> consumer) {
...
+    public void adminCannotSetActiveToFalse() {
+        Study originalStudy = getTestStudy();
+        originalStudy.setActive(true);
+        when(studyDao.getStudy(originalStudy.getIdentifier())).thenReturn(originalStudy);
+
+        Study study = getTestStudy();
+        study.setIdentifier(originalStudy.getIdentifier());
+        study.setActive(false);
+
+        service.updateStudy(study, true);
```
alxdarksage, line 186: same

alxdarksage (changes requested):

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -259,22 +285,34 @@ public Study updateStudy(Study study, boolean isAdminUpdate) {
...
             throw new EntityNotFoundException(Study.class, "Study '"+identifier+"' not found");
         }
-        studyDao.deleteStudy(existing);
-        directoryDao.deleteDirectoryForStudy(existing);
-        subpopService.deleteAllSubpopulations(existing.getStudyIdentifier());
+
+        if (!physical) {
+            // deactivate
+            if (!existing.isActive()) {
+                throw new EntityNotFoundException(Study.class, "Study '"+identifier+"' is deactivated before.");
```
alxdarksage: Change message to "Study not found."

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -121,24 +123,37 @@ public final void setSynapseClient(SynapseClient synapseClient) {
...
-        
+
         Study study = cacheProvider.getStudy(identifier);
         if (study == null) {
             study = studyDao.getStudy(identifier);
             cacheProvider.setStudy(study);
         }
+
+        if (study != null && !study.isActive() && !includeDeleted) {
+            throw new EntityNotFoundException(Study.class, "Study '"+identifier+"' is de-activated.");
```
alxdarksage: Change message to "Study not found" (this is actually a bit of a security issue, you should not be able to deduce that a study once existed in the system when it has been "deleted").

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -225,19 +240,30 @@ public Study createSynapseProjectTeam(Long userId, Study study) throws SynapseEx
...
+                throw new EntityNotFoundException(Study.class, "Study '"+ study.getIdentifier() +"' not found.");
+            }
+
             study.setHealthCodeExportEnabled(originalStudy.isHealthCodeExportEnabled());
             study.setEmailVerificationEnabled(originalStudy.isEmailVerificationEnabled());
         }
+
+        // prevent anyone changing active to false -- it should be done by deactivateStudy() method
+        if (originalStudy.isActive() && !study.isActive()) {
+            throw new BadRequestException("Nobody can set field of active to false in update method.");
```
alxdarksage: Message: "Study cannot be deleted through an update."

alxdarksage (changes requested):
This is fine but some messages still need to be changed.

PR #1324, opened 2017-01-06
limit dupe upload logging to 10 dupes

DwayneJengSage:
Right now, we log every duplicate if an upload is a duplicate this is bad because (a) it consumes a lot of DDB capacity (b) it logs a quadratic amount of logs.

This change caps the max dupes to 10, in both of those places.

Testing done:
* updated unit tests
* ran UploadTest integ tests
* manual test: Set max dupes to 2, then tested with 3 dupes and verified logs.

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDao.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDao.java
@@ -5,15 +5,15 @@ import javax.annotation.Resource;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
 
 import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBQueryExpression;
 import com.amazonaws.services.dynamodbv2.model.AttributeValue;
 import com.amazonaws.services.dynamodbv2.model.ComparisonOperator;
 import com.amazonaws.services.dynamodbv2.model.Condition;
+
+import org.sagebionetworks.bridge.BridgeConstants;
 import org.sagebionetworks.bridge.BridgeUtils;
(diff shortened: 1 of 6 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDao.java
```diff
@@ -121,19 +121,19 @@ public HealthDataRecordBuilder getRecordBuilder() {
...
+                    break;
+                }
             }
-
-            finalRecords.add((HealthDataRecord) oneResult);
         }
 
-        return finalRecords.stream()
-                .filter(record -> record.getSchemaId().equals(schemaId))
-                .collect(Collectors.toList());
```
alxdarksage, line 137: I think stream.limit(BridgeConstants.DUPE_RECORDS_MAX_COUNT) should also work here. The stream is supposed to be smart enough to stop after the limit number without processing the rest, and this is all happening in memory.

app/org/sagebionetworks/bridge/upload/UploadValidationTask.java
```diff
@@ -160,27 +163,34 @@ private void dedupeHelper(String healthRecordId) throws BadRequestException {
...
+                dupeRecordIdList.add(dupeRecordId);
+                numDupes++;
+
+                if (numDupes >= BridgeConstants.DUPE_RECORDS_MAX_COUNT) {
+                    // HealthDataService shouldn't return more than MAX_COUNT, but in case it does, have an extra check
+                    // here to prevent quadratic logging.
+                    break;
+                }
+            }
+        }
```
alxdarksage, line 180: retList.stream().filter().limit(10) would be terser

zaclinsage (commented):
Can merge after alx comments

alxdarksage:
We've merged it. ;-)

PR #1325, opened 2017-01-09
Adding get study report index, and changing to a more useful default date range.

alxdarksage:
Some tweaks to the public reporting stuff based on implementing it in the researcher UI:

- get the study index to get publication state, for both developers and researchers
- getting date range from yesterday to yesterday doesn't return anything obvious, get from yesterday until today.


liujoshua (approved):
Looks good

PR #1327, opened 2017-01-10
Adding study configuration for platform ARNs needed to register devices for notifications.

alxdarksage:
It's just a place to hold the metadata. We are not going to try and configure this when we create a study (not clear to me we'll even have notifications in each study). It must be done manually because you need a lot of stuff from the external notification systems like certificates, and this is more easily done through SNS.

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
@@ -50,12 +50,14 @@ public final class DynamoStudy implements Study {
     private boolean emailVerificationEnabled;
     private boolean externalIdValidationEnabled;
     private Map<String, Integer> minSupportedAppVersions;
+    private Map<String, String> pushNotificationARNs;
 
     public DynamoStudy() {
         profileAttributes = new HashSet<>();
         taskIdentifiers = new HashSet<>();
         dataGroups = new HashSet<>();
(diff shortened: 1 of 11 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
```diff
@@ -343,6 +345,19 @@ public void setExternalIdValidationEnabled(boolean externalIdValidationEnabled)
     public void setMinSupportedAppVersions(Map<String,Integer> map) {
         this.minSupportedAppVersions = (map == null) ? new HashMap<>() : map;
     }
+    
+    /** {@inheritDoc} */
+    @Override
+    public Map<String,String> getPushNotificationARNs() {
+        return pushNotificationARNs;
+    }
```
DwayneJengSage, line 353: Is it worth adding to DynamoStudyDaoTest.crudStudy() to make sure we can write these values to / read from DDB?

DwayneJengSage (changes requested):
Looks good. One comment about a possible test.

app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
alxdarksage: Yes, definitely. Added explicit test they are persisted in the DAO test.

PR #1329, opened 2017-01-12
Adding NotificationRegistration DDB table and DAO.

alxdarksage:


app/org/sagebionetworks/bridge/models/notifications/NotificationRegistration.java
```diff
@@ -0,0 +1,76 @@
...
+     * device identifier in subsequent requests. Can also be used to unregister 
+     * for notifications. 
+     */
+    public String getGuid();
+    
+    /** @see getGuid */
+    public void setGuid(String guid);
+    
+    /**
+     * The platformEndpoint ARN generated when the user's device is registered
```
liujoshua: What's the difference between platformARN, endpointARN, and platformEndpoint ARN?

app/org/sagebionetworks/bridge/dynamodb/DynamoNotificationRegistrationDao.java
```diff
@@ -0,0 +1,192 @@
...
+        return registration;
+    }
+    
+    /**
+     * Update an endpoint using the GUID that was supplied on creation. The only thing you can actually update 
+     * is the device token, but this is important to be able to update as it changes on some platforms.
+     */
+    @Override
+    public NotificationRegistration updateRegistration(String platformARN, NotificationRegistration registration) {
+        checkNotNull(platformARN);
```
liujoshua: platformARN is unused. Not sure if that's what the method comment is talking about

app/org/sagebionetworks/bridge/dynamodb/DynamoNotificationRegistrationDao.java
```diff
@@ -0,0 +1,192 @@
...
+        return registration;
+    }
+
+    @Override
+    public NotificationRegistration createRegistration(String platformARN, NotificationRegistration registration) {
+        checkNotNull(platformARN);
+        checkNotNull(registration);
+        checkNotNull(registration.getHealthCode());
+        checkNotNull(registration.getDeviceId());
+        checkNotNull(registration.getOsName());
```
liujoshua, line 92: what will osName be used for?

liujoshua (approved):
Looks good. Couple questions around what certain fields mean

app/org/sagebionetworks/bridge/models/notifications/NotificationRegistration.java
alxdarksage: platformARN is an identifier representing the set-up of a notifications service like APMS for a specific application. What SNS calls either the platformEndpointARN or the endpointARN, and which I decided I would internally call the endpointARN, is a registration between a specific device, and the platformARN. That's the thing we need to send a push notification to an individual device. It's what gets registered with a topic as well. So I will change the above "platformEndpoint ARN" to match this terminology.

app/org/sagebionetworks/bridge/dynamodb/DynamoNotificationRegistrationDao.java
alxdarksage: As I'm building this, it doesn't seem that necessary, but I am just recording locally the device identifier and platform in case we need it for something. Otherwise we can't easily migrate later. The osName tells me whether the device identifier is really a device token (iOS) or a registrationId (Android).

app/org/sagebionetworks/bridge/dynamodb/DynamoNotificationRegistrationDao.java
alxdarksage: Okay, I'll remove. Makes sense that it's not used if we've already created this record.

alxdarksage:
Okay, I've updated based on your comments.

liujoshua (approved):
Looks good to me

PR #1330, opened 2017-01-12
Notifications service

alxdarksage:
Adding API (controller, service) to CRUD push notification registrations, and to retrieve them through the participants API for an individual user (I imagine making these available to support personnel along with topic registrations).

### DIFF
--- a/app/org/sagebionetworks/bridge/exceptions/NotImplementedException.java
+++ b/app/org/sagebionetworks/bridge/exceptions/NotImplementedException.java
@@ -2,6 +2,11 @@ package org.sagebionetworks.bridge.exceptions;
 
 import org.apache.http.HttpStatus;
 
+/**
+ * This endpoint has not been implemented for this study. It has been disabled, or 
+ * not completely configured. The caller cannot fix the call to succeed, but it is 
+ * not an unexpected server error, it is intentional.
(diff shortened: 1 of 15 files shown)


app/org/sagebionetworks/bridge/exceptions/NotImplementedException.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.apache.http.HttpStatus;
+
+/**
+ * This endpoint has not been implemented for this study. It has been disabled, or 
+ * not completely configured. The caller cannot fix the call to succeed, but it is 
+ * not an unexpected server error, it is intentional.
+ */
+@SuppressWarnings("serial")
+public class NotImplementedException extends BridgeServiceException {
```
liujoshua, line 11: Do you think the SDK should get a new exception as well? Or roll it up into BridgeSDKException?

liujoshua (approved):
Looks good to me

app/org/sagebionetworks/bridge/exceptions/NotImplementedException.java
alxdarksage: Good point. I should add that to the SDK changes.

PR #1332, opened 2017-01-13
Add ability to send a message to an individual participant.

alxdarksage:
This finishes the basic functionality to register for push notifications and to receive them at an individual level. Topics will allow us to send to groups of participants (TBD).

### DIFF
--- a/app/org/sagebionetworks/bridge/models/notifications/NotificationMessage.java
+++ b/app/org/sagebionetworks/bridge/models/notifications/NotificationMessage.java
@@ -1,0 +1,79 @@
+package org.sagebionetworks.bridge.models.notifications;
+
+import java.util.Objects;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+/**
+ * For information on the push notification JSON structures that are sent to SNS, see:
+ * http://docs.aws.amazon.com/sns/latest/dg/mobile-push-send-custommessage.html
+ * 
+ * Note: There are many more fields that we can and will provide once we have an end-to-end system working. So we have
+ * introduced a builder in anticipation of expanding this model.
+ */
+@JsonDeserialize(builder=NotificationMessage.Builder.class)
+public final class NotificationMessage {
+
+    /**
+     * Very short rendition of the notification. For example, this value will be displayed on an Apple iWatch, where the
+     * message is shown for a short time only.
+     */
+    private final String subject;
+    
+    /**
+     * Full notification message. 
+     */
+    private final String message;
+    
+    NotificationMessage(String subject, String message) {
+        this.subject = subject;
+        this.message = message;
+    }
+
+    public String getSubject() {
+        return subject;
+    }
+
+    public String getMessage() {
+        return message;
+    }
+    
+    @Override
+    public int hashCode() {
+        return Objects.hash(subject, message);
+    }
+    
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        } else if (obj == null || getClass() != obj.getClass()) {
+            return false;
+        }
+        NotificationMessage other = (NotificationMessage)obj;
+        return Objects.equals(subject, other.getSubject()) &&
+               Objects.equals(message, other.getMessage());
+    }
+    
+    @Override
+    public String toString() {
+        return String.format("NotificationMessage[subject=%s, message%s]", subject, message);
+    }
+    
+    public static class Builder {
+        private String subject;
+        private String message;
+        
+        public Builder withSubject(String subject) {
+            this.subject = subject;
+            return this;
+        }
+        public Builder withMessage(String message) {
+            this.message = message;
+            return this;
+        }
+        public NotificationMessage build() {
+            return new NotificationMessage(subject, message);
+        }
+    }
+}
--- a/app/org/sagebionetworks/bridge/models/notifications/NotificationMessage.java
+++ b/app/org/sagebionetworks/bridge/models/notifications/NotificationMessage.java
@@ -2,6 +2,8 @@ package org.sagebionetworks.bridge.models.notifications;
 
 import java.util.Objects;
 
+import org.sagebionetworks.bridge.models.BridgeEntity;
+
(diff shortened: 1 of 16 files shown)


app/org/sagebionetworks/bridge/models/notifications/NotificationMessage.java
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * Very short rendition of the notification. For example, this value will be displayed on an Apple iWatch, where the
+     * message is shown for a short time only.
+     */
+    private final String subject;
+    
+    /**
+     * Full notification message. 
+     */
+    private final String message;
```
liujoshua, line 28: In cases where full message can be displayed, (e.g. non iWatch) is `message` used in lieu of `subject`, or should they both be displayed when possible?

app/org/sagebionetworks/bridge/models/notifications/NotificationMessage.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.util.Objects;
+
+import org.sagebionetworks.bridge.models.BridgeEntity;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+/**
+ * For information on the push notification JSON structures that are sent to SNS, see:
+ * http://docs.aws.amazon.com/sns/latest/dg/mobile-push-send-custommessage.html
```
liujoshua, line 11: Awesome! I didn't realize SNS could deliver to GCM, APNS, ADM 👍

app/org/sagebionetworks/bridge/services/NotificationsService.java
```diff
@@ -1,33 +1,50 @@
...
 import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
+import org.sagebionetworks.bridge.validators.NotificationMessageValidator;
 import org.sagebionetworks.bridge.validators.NotificationRegistrationValidator;
 import org.sagebionetworks.bridge.validators.Validate;
 
+import com.amazonaws.AmazonServiceException;
+import com.amazonaws.services.sns.AmazonSNSClient;
+import com.amazonaws.services.sns.model.PublishRequest;
+import com.amazonaws.services.sns.model.PublishResult;
+import com.newrelic.agent.deps.com.google.common.collect.Lists;
```
liujoshua: Internal Guava :(

app/org/sagebionetworks/bridge/services/NotificationsService.java
```diff
@@ -106,6 +128,45 @@ public void deleteRegistration(String healthCode, String guid) {
...
+    public void sendNotification(StudyIdentifier studyId, String healthCode, NotificationMessage message) {
+        checkNotNull(studyId);
+        checkNotNull(healthCode);
+        checkNotNull(message);
+        
+        Validate.entityThrowingException(NotificationMessageValidator.INSTANCE, message);
+        
+        List<NotificationRegistration> registrations = notificationRegistrationDao.listRegistrations(healthCode);
+        if (registrations.isEmpty()) {
+            throw new BadRequestException("Participant has not registered to receive push notifications.");
```
liujoshua, line 147: From what I understand, it's expected for participants not to receive notifications since they can opt out. Would it be better not to throw an exception since non-delivery is an expected use case, maybe return a 204? Would it be useful to provide info in the response along the lines of how many devices it was delivered to, which devices, etc?

app/org/sagebionetworks/bridge/services/NotificationsService.java
```diff
@@ -106,6 +128,45 @@ public void deleteRegistration(String healthCode, String guid) {
...
+        List<String> errorMessages = Lists.newArrayListWithCapacity(registrations.size());
+        for (NotificationRegistration registration : registrations) {
+            String endpointARN = registration.getEndpointARN();
+            
+            PublishRequest request = new PublishRequest().withTargetArn(endpointARN)
+                    .withSubject(message.getSubject()).withMessage(message.getMessage());
+            
+            try {
+                PublishResult result = snsClient.publish(request);
+                LOG.debug("Sent message to participant, study="+studyId.getIdentifier()+", endpointARN="+endpointARN+", message ID="+result.getMessageId());
```
liujoshua: Let's put some spaces between + signs

liujoshua (approved):
Looks good to merge, couple questions/suggestions

app/org/sagebionetworks/bridge/models/notifications/NotificationMessage.java
alxdarksage: This really depends on the platform, and I noticed that using the simple interface for sending messages, SNS did not add the subject to the JSON it produced. But there's a field that maps to this in both APNS and GCM so when I move to constructing the JSON myself, it should be used (but again I'm not 100% sure how. I'll update the docs as we learn more).

app/org/sagebionetworks/bridge/services/NotificationsService.java
alxdarksage: Yikes. Fixed.

app/org/sagebionetworks/bridge/services/NotificationsService.java
alxdarksage: In the case where there are no registrations, we know for a fact it will fail, so right now the UI disables the button to send a notification to the user. I think we should probably return 400 here to tell the caller since we know it's pointless. In the case where it succeeds though, we should return 202 "Accepted" since actually, we don't know what the ultimate disposition of the message will be. Even without errors, it may fail to be delivered for any number of reasons. So I'll change that.

app/org/sagebionetworks/bridge/services/NotificationsService.java
alxdarksage: Changed

app/org/sagebionetworks/bridge/services/NotificationsService.java
liujoshua: That all makes sense

liujoshua:
Looks good, merge when ready

PR #1333, opened 2017-01-14
DynamoDBMapper.query() ignores LIMIT; move to queryPage()

DwayneJengSage:
See https://github.com/aws/aws-sdk-java/issues/736

This is causing throttling errors for us, since our paginated APIs and other uses of query() are consuming a lot more DDB capacity than expected, in some cases unbounded capacity. This change moves most usages of query() with LIMIT to queryPage(), which does respect LIMIT. (Note that some cases can't be fixed, like ExternalId. Since LIMIT and filters interact in weird ways, assignmentFilter doesn't work anymore. See https://sagebionetworks.jira.com/browse/BRIDGE-1652)

I also removed the redundant dupe max records from UploadValidationTask. Because we were enforcing the limit in two places, this masked the bug in DynamoHealthDataDao that was browning out DDB.

Testing done:
* manually tested upload dedupe detection - This is the case where we discovered the issue.
* integ tests and unit tests


liujoshua (approved):
Looks good. This should prevent unbounded capacity consumption

PR #1334, opened 2017-01-18
Adding NotificationTopic table (mapping record and the DAO).

alxdarksage:
Developers create topics for their study, and participants will subscribe to these topics in order to receive notifications en masse. This is the beginning of creating the Topic CRUD APIs.

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoNotificationRegistration.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoNotificationRegistration.java
@@ -1,6 +1,5 @@
 package org.sagebionetworks.bridge.dynamodb;
 
-import org.sagebionetworks.bridge.json.BridgeTypeName;
 import org.sagebionetworks.bridge.json.DateTimeToLongDeserializer;
 import org.sagebionetworks.bridge.json.DateTimeToLongSerializer;
 import org.sagebionetworks.bridge.models.notifications.NotificationRegistration;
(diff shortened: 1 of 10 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoNotificationRegistration.java
```diff
@@ -13,7 +13,6 @@
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 
 @DynamoDBTable(tableName = "NotificationRegistration")
-@BridgeTypeName("NotificationRegistration")
```
alxdarksage, line 16: It makes more sense to me to put these on the interface; this is part of the public interface of the object.

app/org/sagebionetworks/bridge/dynamodb/DynamoNotificationTopicDao.java
```diff
@@ -0,0 +1,146 @@
...
+        CreateTopicRequest request = new CreateTopicRequest().withName(snsTopicName);
+        CreateTopicResult result = snsClient.createTopic(request);
+        topic.setTopicARN(result.getTopicArn());
+        
+        mapper.save(topic);
+        return topic;
+    }
+
+    @Override
+    public NotificationTopic updateTopic(StudyIdentifier studyId, NotificationTopic topic) {
```
liujoshua, line 103: Can we get studyId from the NotificationTopic object instead of a parameter?

liujoshua (approved):
Looks good and ready to merge

app/org/sagebionetworks/bridge/dynamodb/DynamoNotificationTopicDao.java
alxdarksage: That might make sense. The studyId should ultimately come from the user's session so it can't be set to another study, either by carrying it over, or ensuring it is set in the service.

PR #1340, opened 2017-01-20
Compound Activity Definitions

DwayneJengSage:
The next step in adding schema IDs to schedules. Previously, we had created CompoundActivities in schedules. However, we don't want researchers to have to re-type the same list of schemas multiple times for each schedule. So we want them to be able to define Compound Activity "Definitions" separately, which can be referenced in schedules.

This includes DDB definitions, DAO, service, validator, and controller. This doesn't include scheduler logic to automatically transcribe definitions into schedules. That's the next change list.

Testing done:
* added unit tests
* manually tested CRUD methods

### DIFF
--- a/app/org/sagebionetworks/bridge/dao/CompoundActivityDefinitionDao.java
+++ b/app/org/sagebionetworks/bridge/dao/CompoundActivityDefinitionDao.java
@@ -8,14 +8,13 @@ import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
 /** DAO for basic CRUD and list operations for compound activity definitions. */
 public interface CompoundActivityDefinitionDao {
     /** Creates a compound activity definition. */
-    CompoundActivityDefinition createCompoundActivityDefinition(StudyIdentifier studyId,
-            CompoundActivityDefinition compoundActivityDefinition);
+    CompoundActivityDefinition createCompoundActivityDefinition(CompoundActivityDefinition compoundActivityDefinition);
 
-    /**
-     * Deletes a compound activity definition. This is intended to be used by admin accounts to clean up after tests or
-     * other administrative tasks.
-     */
+    /** Deletes a compound activity definition. */
     void deleteCompoundActivityDefinition(StudyIdentifier studyId, String taskId);
+
+    /** Deletes all compound activity definitions in the specified study. Used when we physically delete a study. */
+    void deleteAllCompoundActivityDefinitionsInStudy(StudyIdentifier studyId);
 
     /** List all compound activity definitions in a study. */
     List<CompoundActivityDefinition> getAllCompoundActivityDefinitionsInStudy(StudyIdentifier studyId);
@@ -24,6 +23,5 @@ public interface CompoundActivityDefinitionDao {
     CompoundActivityDefinition getCompoundActivityDefinition(StudyIdentifier studyId, String taskId);
 
     /** Update a compound activity definition. */
-    CompoundActivityDefinition updateCompoundActivityDefinition(StudyIdentifier studyId, String taskId,
-            CompoundActivityDefinition compoundActivityDefinition);
+    CompoundActivityDefinition updateCompoundActivityDefinition(CompoundActivityDefinition compoundActivityDefinition);
 }
--- a/app/org/sagebionetworks/bridge/dao/CompoundActivityDefinitionDao.java
+++ b/app/org/sagebionetworks/bridge/dao/CompoundActivityDefinitionDao.java
@@ -1,0 +1,29 @@
+package org.sagebionetworks.bridge.dao;
+
+import java.util.List;
+
+import org.sagebionetworks.bridge.models.schedules.CompoundActivityDefinition;
+import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
+
+/** DAO for basic CRUD and list operations for compound activity definitions. */
+public interface CompoundActivityDefinitionDao {
+    /** Creates a compound activity definition. */
+    CompoundActivityDefinition createCompoundActivityDefinition(StudyIdentifier studyId,
+            CompoundActivityDefinition compoundActivityDefinition);
+
+    /**
+     * Deletes a compound activity definition. This is intended to be used by admin accounts to clean up after tests or
+     * other administrative tasks.
+     */
+    void deleteCompoundActivityDefinition(StudyIdentifier studyId, String taskId);
+
+    /** List all compound activity definitions in a study. */
+    List<CompoundActivityDefinition> getAllCompoundActivityDefinitionsInStudy(StudyIdentifier studyId);
+
+    /** Get a compound activity definition by ID. */
+    CompoundActivityDefinition getCompoundActivityDefinition(StudyIdentifier studyId, String taskId);
+
+    /** Update a compound activity definition. */
+    CompoundActivityDefinition updateCompoundActivityDefinition(StudyIdentifier studyId, String taskId,
+            CompoundActivityDefinition compoundActivityDefinition);
+}
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoCompoundActivityDefinition.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoCompoundActivityDefinition.java
@@ -1,0 +1,116 @@
+package org.sagebionetworks.bridge.dynamodb;
+
+import java.util.List;
+
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBHashKey;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBRangeKey;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBTable;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBTypeConverted;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBVersionAttribute;
+import com.fasterxml.jackson.annotation.JsonFilter;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.google.common.collect.ImmutableList;
+
+import org.sagebionetworks.bridge.models.schedules.CompoundActivityDefinition;
+import org.sagebionetworks.bridge.models.schedules.SchemaReference;
+import org.sagebionetworks.bridge.models.schedules.SurveyReference;
+
+/**
+ * DynamoDB implementation of CompoundActivityDefinition. The hash key is study ID and the range key is the task ID.
+ * This allows us to namespace compound activity definitions to studies and to list all definitions in a study.
+ */
+@DynamoDBTable(tableName = "CompoundActivityDefinition")
+@JsonFilter("filter")
+public class DynamoCompoundActivityDefinition implements CompoundActivityDefinition {
+    private List<SchemaReference> schemaList = ImmutableList.of();
+    private String studyId;
+    private List<SurveyReference> surveyList = ImmutableList.of();
+    private String taskId;
+    private Long version;
+
+    /** {@inheritDoc} */
+    @DynamoDBTypeConverted(converter = SchemaReferenceListMarshaller.class)
+    @Override
+    public List<SchemaReference> getSchemaList() {
+        return schemaList;
+    }
+
+    /** @see #getSchemaList */
+    @Override
+    public void setSchemaList(List<SchemaReference> schemaList) {
+        this.schemaList = schemaList != null ? ImmutableList.copyOf(schemaList) : ImmutableList.of();
+    }
+
+    /** {@inheritDoc} */
+    @DynamoDBHashKey
+    @Override
+    public String getStudyId() {
+        return studyId;
+    }
+
+    /** @see #getStudyId */
+    @Override
+    public void setStudyId(String studyId) {
+        this.studyId = studyId;
+    }
+
+    /** {@inheritDoc} */
+    @DynamoDBTypeConverted(converter = SurveyReferenceListMarshaller.class)
+    @Override
+    public List<SurveyReference> getSurveyList() {
+        return surveyList;
+    }
+
+    /** @see #getSurveyList */
+    @Override
+    public void setSurveyList(List<SurveyReference> surveyList) {
+        this.surveyList = surveyList != null ? ImmutableList.copyOf(surveyList) : ImmutableList.of();
+    }
+
+    /** {@inheritDoc} */
+    @DynamoDBRangeKey
+    public String getTaskId() {
+        return taskId;
+    }
+
+    /** @see #getTaskId */
+    @Override
+    public void setTaskId(String taskId) {
+        this.taskId = taskId;
+    }
+
+    /** For use with DynamoDB versioning and concurrency. */
+    @DynamoDBVersionAttribute
+    public Long getVersion() {
+        return version;
+    }
+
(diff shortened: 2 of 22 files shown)


app/org/sagebionetworks/bridge/dao/CompoundActivityDefinitionDao.java
```diff
@@ -0,0 +1,29 @@
...
+    /** List all compound activity definitions in a study. */
+    List<CompoundActivityDefinition> getAllCompoundActivityDefinitionsInStudy(StudyIdentifier studyId);
+
+    /** Get a compound activity definition by ID. */
+    CompoundActivityDefinition getCompoundActivityDefinition(StudyIdentifier studyId, String taskId);
+
+    /** Update a compound activity definition. */
+    CompoundActivityDefinition updateCompoundActivityDefinition(StudyIdentifier studyId, String taskId,
+            CompoundActivityDefinition compoundActivityDefinition);
+}
```
alxdarksage, line 27: deleteAllCompoundActivityDefinition to call when physically deleting a study?

app/org/sagebionetworks/bridge/dynamodb/DynamoCompoundActivityDefinitionDao.java
```diff
@@ -0,0 +1,133 @@
...
+
+    /** {@inheritDoc} */
+    @Override
+    public CompoundActivityDefinition createCompoundActivityDefinition(StudyIdentifier studyId,
+            CompoundActivityDefinition compoundActivityDefinition) {
+        // Currently, all CompoundActivityDefinitions are DynamoCompoundActivityDefinitions.
+        DynamoCompoundActivityDefinition ddbDef = (DynamoCompoundActivityDefinition) compoundActivityDefinition;
+
+        // Set study to prevent people from creating defs in other studies.
+        ddbDef.setStudyId(studyId.getIdentifier());
```
alxdarksage: I would do this in the service, so if we move off of DDB, all this logic is preserved in the re-usable service. If I could do it over, I'd put the DAO in a package with the service, package visible, so only the service can use it, and no one can side-step the logic in the service.

app/org/sagebionetworks/bridge/dynamodb/DynamoCompoundActivityDefinitionDao.java
```diff
@@ -0,0 +1,133 @@
...
+        ddbHashKey.setStudyId(studyId.getIdentifier());
+        DynamoDBQueryExpression<DynamoCompoundActivityDefinition> ddbQueryExpr =
+                new DynamoDBQueryExpression<DynamoCompoundActivityDefinition>().withHashKeyValues(ddbHashKey);
+
+        // execute query
+        List<DynamoCompoundActivityDefinition> defList = mapper.query(DynamoCompoundActivityDefinition.class,
+                ddbQueryExpr);
+
+        // because of generics wonkiness, we need to convert this to a list of parent class CompoundActivityDefinition
+        return defList.stream().collect(Collectors.toList());
```
alxdarksage: I found that ImmutableList.copyOf(defList) also works, I can't find a way to do this that doesn't iterate over the results

app/org/sagebionetworks/bridge/dynamodb/DynamoCompoundActivityDefinitionDao.java
```diff
@@ -0,0 +1,133 @@
...
+        ddbDef.setTaskId(taskId);
+
+        // Call DDB mapper to save.
+        try {
+            mapper.save(ddbDef);
+        } catch (ConditionalCheckFailedException ex) {
+            throw new ConcurrentModificationException(ddbDef);
+        }
+
+        return ddbDef;
```
alxdarksage: Do you think we should be adding the version attribute to all model objects? I ask because I didn't add it to NotificationRegistration or NotificationTopic, it didn't seem super important to prevent overwriting changes.

app/org/sagebionetworks/bridge/models/schedules/CompoundActivityDefinition.java
```diff
@@ -0,0 +1,56 @@
...
+import org.sagebionetworks.bridge.dynamodb.DynamoCompoundActivityDefinition;
+import org.sagebionetworks.bridge.json.BridgeObjectMapper;
+import org.sagebionetworks.bridge.json.BridgeTypeName;
+import org.sagebionetworks.bridge.models.BridgeEntity;
+
+/**
+ * Compound activity definition, which is keyed off taskID and stored in persistent storage. This allows study
+ * developers to define a task once and use that task in multiple schedules.
+ */
+@BridgeTypeName("CompoundActivityDefinition")
```
alxdarksage, line 19: Can we call this CompoundActivity in the API? It's shorter and consistent with Activity.

app/org/sagebionetworks/bridge/models/studies/Study.java
```diff
@@ -29,6 +29,11 @@
         new SimpleFilterProvider().addFilter("filter",
         SimpleBeanPropertyFilter.filterOutAllExcept("name", "identifier")));
 
+    /** Convenience method for creating a Study using a concrete implementation. */
+    static Study create() {
```
alxdarksage: Why not public?

app/org/sagebionetworks/bridge/play/controllers/CompoundActivityDefinitionController.java
```diff
@@ -0,0 +1,82 @@
...
+                requestDef);
+        return created(CompoundActivityDefinition.PUBLIC_DEFINITION_WRITER.writeValueAsString(createdDef));
+    }
+
+    /**
+     * Deletes a compound activity definition. This is intended to be used by admin accounts to clean up after tests or
+     * other administrative tasks.
+     */
+    public Result deleteCompoundActivityDefinition(String studyId, String taskId) {
+        getAuthenticatedSession(ADMIN);
```
alxdarksage: In the course of designing, developers are going to want to delete these. That raises issues... regardless of who does the deletion, we probably shouldn't do it if it's in a schedule. But it's not too difficult to re-create one of these if they accidentally delete it.

conf/routes
```diff
@@ -13,6 +13,12 @@ POST   /v3/auth/signUp                  @org.sagebionetworks.bridge.play.control
 POST   /v3/auth/verifyEmail             @org.sagebionetworks.bridge.play.controllers.AuthenticationController.verifyEmail
 POST   /v3/auth/resendEmailVerification @org.sagebionetworks.bridge.play.controllers.AuthenticationController.resendEmailVerification
 
+# Compound Activity Definitions
+GET    /v3/compoundactivitydefinitions @org.sagebionetworks.bridge.play.controllers.CompoundActivityDefinitionController.getAllCompoundActivityDefinitionsInStudy
+POST   /v3/compoundactivitydefinitions @org.sagebionetworks.bridge.play.controllers.CompoundActivityDefinitionController.createCompoundActivityDefinition
+GET    /v3/compoundactivitydefinitions/:taskId @org.sagebionetworks.bridge.play.controllers.CompoundActivityDefinitionController.getCompoundActivityDefinition(taskId: String)
+POST   /v3/compoundactivitydefinitions/:taskId @org.sagebionetworks.bridge.play.controllers.CompoundActivityDefinitionController.updateCompoundActivityDefinition(taskId: String)
```
alxdarksage, line 21: Can we call /v3/compoundactivities ?

app/org/sagebionetworks/bridge/dynamodb/DynamoCompoundActivityDefinition.java
```diff
@@ -0,0 +1,116 @@
...
+ * This allows us to namespace compound activity definitions to studies and to list all definitions in a study.
+ */
+@DynamoDBTable(tableName = "CompoundActivityDefinition")
+@JsonFilter("filter")
+public class DynamoCompoundActivityDefinition implements CompoundActivityDefinition {
+    private List<SchemaReference> schemaList = ImmutableList.of();
+    private String studyId;
+    private List<SurveyReference> surveyList = ImmutableList.of();
+    private String taskId;
+    private Long version;
```
alxdarksage, line 29: These are going to need a name or something to display them in a selector (a list or pulldown or something). 

I'm not sure what taskId is... is this the compoundActivityDefinitionId or Guid? If these are only created when we schedule tasks (and that's the task's ID), how will people define these beforehand to include in schedules?

alxdarksage (commented):
Maybe it would be easier to discuss how these are created and then included in schedules when you get in, I'm a little fuzzy on that. In addition my biggish comments were 1) I would call the CompoundActivity in the API because it's shorter and definition doesn't add to the meaning (unless these are instantiated for tasks as CompoundActivity instances or something), and 2) I would allow developers to delete these. They aren't so complicated they can't be recreated if deleted, and it's super annoying not to be able to delete things. I do have BRIDGE-1420 to prevent these kinds of deletes from breaking a schedule.

DwayneJengSage:
> Maybe it would be easier to discuss how these are created and then included in schedules when you get in, I'm a little fuzzy on that.

Forwarded Alx the original email from the previous discussion and discussed offline. In summary, this is to associate schemas and surveys to a task ID. We don't want researchers to have to enter the same list of schemas and surveys multiple times, so we allow them to define a "definition", which can be references in schedules. There's a later code change to autopopulate these in the scheduler.

> I would call the CompoundActivity in the API because it's shorter and definition doesn't add to the meaning (unless these are instantiated for tasks as CompoundActivity instances or something)

As per our previous email, the names we had previously decided on wer CompoundActivity (for references in the the scheduler output) and CompoundActivityDefinition (for the definition that the researchers can create once and refer to multiple times). I'm open to name changes, but I don't believe the names will actually get any shorter.

I'm also open to the idea of combining CompoundActivityDefinition and CompoundActivity, but then we'd need to keep track of separate implementations for the Dynamo implementation and for the implementation used in the scheduler. Plus the validator would get messy. At this point, I'm not entirely sure which way would be simpler and cleaner.

> I would allow developers to delete these.

Sure, I can do that.

app/org/sagebionetworks/bridge/dao/CompoundActivityDefinitionDao.java
DwayneJengSage: Added.

app/org/sagebionetworks/bridge/dynamodb/DynamoCompoundActivityDefinitionDao.java
DwayneJengSage: Added, to both create() and to update().

app/org/sagebionetworks/bridge/dynamodb/DynamoCompoundActivityDefinitionDao.java
DwayneJengSage: Changed to ImmutableList.copyOf(). I find this more elegant. Thanks!

app/org/sagebionetworks/bridge/dynamodb/DynamoCompoundActivityDefinitionDao.java
DwayneJengSage: The version attribute helps us catch race conditions by error'ing them out at the DDB level.

On the other hand, if we take out the version attribute entirely, we can combine the create() and update() methods in the DAO, since at that point, they'd be identical. Let me know if you have a preference either way.

app/org/sagebionetworks/bridge/models/schedules/CompoundActivityDefinition.java
DwayneJengSage: As mentioned in my other comment, we currently have CompoundActivity as the reference/instance object in the schedule, and CompoundActivityDefinition as the definition. We can change the names, but I don't think it would be any shorter.

app/org/sagebionetworks/bridge/play/controllers/CompoundActivityDefinitionController.java
DwayneJengSage: Changed to allow developers to delete these.

conf/routes
DwayneJengSage: That depends on whether (a) we want to call this compoundactivities in the URL regardless of what the underlying object is called or (b) we want to remain consistent with the actual objects we're CRUD'ing.

app/org/sagebionetworks/bridge/dynamodb/DynamoCompoundActivityDefinition.java
DwayneJengSage: The task ID is the same task ID as in TaskReference and in the Study.getTaskIdentifiers(). My assumption was that we didn't need a new set of identifiers for Compound Activities since the idea was to associate a list of schemas and surveys to tasks, and that there would always be a one-to-one mapping between task IDs and those lists. We _could_ add a separate set of CompoundActivityDefinitionIds, but I feel like this would only add confusion, and I'm not sure what the benefits are.

The task IDs themselves are already defined by researchers/developers in the Bridge Study Manager, so I assume they're already human readable. The assumption is whatever we currently use for TaskReferences could also be used for CompoundActivities. If for whatever that user experience is still confusing, I could trivially add name to CompoundActivity and CompoundActivityDefinition.

DwayneJengSage:
Pushed an update just to move this forward. This is still pending a few tests, and I need to update a few unit tests.

app/org/sagebionetworks/bridge/dynamodb/DynamoCompoundActivityDefinition.java
alxdarksage: Yes we discussed this, you can ignore, they'll reference these by taskId

app/org/sagebionetworks/bridge/dynamodb/DynamoCompoundActivityDefinitionDao.java
alxdarksage: I should probably add version just in case someone is confused and saves over their updates in a new browser window or something. I'll do that today for notifications. I think I'd keep create/update because in create, you'd ignore primary keys and replace them (creating a copy), whereas in update, you're expecting to update an existing record.

app/org/sagebionetworks/bridge/models/schedules/CompoundActivityDefinition.java
alxdarksage: I see. In the UI I'm probably going to continue showing these as Task Identifiers or even better, Activities, where there's no naming conflict issue. Then tell people to schedule the activities.

alxdarksage (approved):
Looks good, merge when ready.

test/org/sagebionetworks/bridge/validators/CompoundActivityDefinitionValidatorTest.java
```diff
@@ -0,0 +1,163 @@
...
+        CompoundActivityDefinition def = makeValidDef();
+        def.setStudyId(studyId);
+
+        try {
+            Validate.entityThrowingException(VALIDATOR, def);
+            fail("expected exception");
+        } catch (InvalidEntityException ex) {
+            assertTrue(ex.getMessage().contains("studyId must be specified"));
+        }
+    }
```
alxdarksage, line 97: FWIW I created a TestUtils.assertValidatorMessage(Validator validator, Object object, String fieldName, String error) method to use on these kinds of tests

alxdarksage (approved):
Looks good.

PR #1344, opened 2017-01-24
Added fields to describe and report on pattern-based string constraints

alxdarksage:
Also did the following clean-up:

- Switched survey validation tests over to use a more stringent utility method and found many minor errors in the formatting of error messages for surveys, which are cleaned up;

- Removed the equals/hashCode methods from constraints. These had no tests and failed equalVerifier testing, and are not used except for convenience in a couple of tests, so there was no justification to keep them;

- Removed a number of unused import statements and variables left by prior check-ins;


DwayneJengSage:
Changes look good. For future reference, can we separate the test / error message refactors into a separate changelist from the actual functional change? This way, we don't have to scroll through pages of refactors to find the relevant functional changes. Thanks!

DwayneJengSage (approved):

PR #1345, opened 2017-01-25
add createdOn timezone to health data record

DwayneJengSage:
Currently, createdOn is being exported as epoch milliseconds. We want to export the timezone as well. This change adds timezone as a separate field to the health data record. Note that we made it a separate field instead of changing createdOn to a DateTime because (a) createdOn is used as a range key index and (b) backwards compatibility.

Added unit tests. Manually tested cases where createdOn is present and where createdOn were absent.

This will also need a corresponding change in BridgeEX, and probably a minor update in Integ Tests. (Coming soon.)

See also https://sagebionetworks.jira.com/browse/BRIDGE-1658

### DIFF
--- a/app/org/sagebionetworks/bridge/models/healthdata/HealthDataRecord.java
+++ b/app/org/sagebionetworks/bridge/models/healthdata/HealthDataRecord.java
@@ -15,17 +15,20 @@ import org.sagebionetworks.bridge.models.BridgeEntity;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import org.joda.time.format.DateTimeFormat;
+import org.joda.time.format.DateTimeFormatter;
 
 /** This class represents health data and associated metadata. */
 @BridgeTypeName("HealthData")
 @JsonDeserialize(as = DynamoHealthDataRecord.class)
 public interface HealthDataRecord extends BridgeEntity {
+    DateTimeFormatter TIME_ZONE_FORMATTER = DateTimeFormat.forPattern("Z");
     ObjectWriter PUBLIC_RECORD_WRITER = new BridgeObjectMapper().writer(
             new SimpleFilterProvider().addFilter("filter",
                     SimpleBeanPropertyFilter.serializeAllExcept("healthCode")));
 
-    public enum ExporterStatus {
-        NOT_EXPORTED, SUCCEEDED;
+    enum ExporterStatus {
+        NOT_EXPORTED, SUCCEEDED
     }
 
     /**
@@ -33,6 +36,9 @@ public interface HealthDataRecord extends BridgeEntity {
      * (start of epoch).
      */
     Long getCreatedOn();
+
+    /** Time zone of the createdOn timestamp, expressed as a 4-digit string with sign. Examples: "-0800", "+0900" */
+    String getCreatedOnTimeZone();
 
     /** Health data, in JSON format. */
     JsonNode getData();
--- a/app/org/sagebionetworks/bridge/upload/IosSchemaValidationHandler2.java
+++ b/app/org/sagebionetworks/bridge/upload/IosSchemaValidationHandler2.java
@@ -29,6 +29,7 @@ import org.sagebionetworks.bridge.json.BridgeObjectMapper;
 import org.sagebionetworks.bridge.json.DateUtils;
 import org.sagebionetworks.bridge.json.JsonUtils;
 import org.sagebionetworks.bridge.models.GuidCreatedOnVersionHolderImpl;
+import org.sagebionetworks.bridge.models.healthdata.HealthDataRecord;
 import org.sagebionetworks.bridge.models.healthdata.HealthDataRecordBuilder;
 import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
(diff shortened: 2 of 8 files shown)


app/org/sagebionetworks/bridge/upload/IosSchemaValidationHandler2.java
```diff
@@ -332,11 +333,14 @@ private static void validateInfoJsonFileList(UploadValidationContext context, St
         }
 
         if (createdOn == null) {
-            // Recover by using current time.
+            // Recover by using current time. Don't set a timezone, since it's indeterminate.
             context.addMessage(String.format("upload ID %s has no timestamps, using current time", uploadId));
-            createdOn = DateUtils.getCurrentDateTime();
+            recordBuilder.withCreatedOn(DateUtils.getCurrentMillisFromEpoch());
```
alxdarksage, line 338: Can we get the time zone from the system, e.g. DateTimeZone.getDefault()?

app/org/sagebionetworks/bridge/models/healthdata/HealthDataRecord.java
```diff
@@ -15,17 +15,20 @@
...
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import org.joda.time.format.DateTimeFormat;
+import org.joda.time.format.DateTimeFormatter;
 
 /** This class represents health data and associated metadata. */
 @BridgeTypeName("HealthData")
 @JsonDeserialize(as = DynamoHealthDataRecord.class)
 public interface HealthDataRecord extends BridgeEntity {
+    DateTimeFormatter TIME_ZONE_FORMATTER = DateTimeFormat.forPattern("Z");
```
alxdarksage, line 25: Just checking that this will parse all 4 of the acceptable formats:
<time>Z
<time>±hh:mm
<time>±hhmm
<time>±hh

test/org/sagebionetworks/bridge/models/healthdata/HealthDataRecordTest.java
```diff
@@ -299,13 +305,15 @@ public void testSerialization() throws Exception {
         // start with JSON
         String jsonText = "{\n" +
                 "   \"createdOn\":\"2014-02-12T13:45-0800\",\n" +
+                "   \"createdOnTimeZone\":\"-0800\",\n" +
```
alxdarksage, line 309: What happens if the time zone in createdOn is different from createdOnTimeZone (what do you think should happen, and we should probably document). Can the time zone be parsed directly from createdOn, or parsed out of it if createdOnTimeZone is not provided?

alxdarksage (commented):
I leave it to you to decide if you want to do anything based on my comments. Other than that, looks good.

app/org/sagebionetworks/bridge/upload/IosSchemaValidationHandler2.java
DwayneJengSage: Discussed offline. If the users don't submit a timezone, instead of submitting a default timezone, we should submit no timezone.

app/org/sagebionetworks/bridge/models/healthdata/HealthDataRecord.java
DwayneJengSage: We're using this to print timezones, not parse them. I've added tests to verify that we're printing what we expect.

test/org/sagebionetworks/bridge/models/healthdata/HealthDataRecordTest.java
DwayneJengSage: Discussed offline. This is an artifact of how in DDB this is represented as epoch milliseconds, but we convert it to DateTime when returning it as JSON. We never read this from an API, and in the uploads, we always get it from createdOn, so there will never be a data discrepancy.

It may be confusing to show a timestamp in a timezone that's not the user-submitted one, but it's not a problem unique to Bridge, and I'm not sure how to fix it.

test/org/sagebionetworks/bridge/models/healthdata/HealthDataRecordTest.java
```diff
@@ -387,4 +388,21 @@ public void testSerialization() throws Exception {
...
     }
+
+    @Test
+    public void testTimeZoneFormatter() {
+        testTimeZoneFormatter("+0000", DateTimeZone.UTC);
+        testTimeZoneFormatter("+0000", DateTimeZone.forOffsetHours(0));
+        testTimeZoneFormatter("+0900", DateTimeZone.forOffsetHours(+9));
+        testTimeZoneFormatter("-0800", DateTimeZone.forOffsetHours(-8));
+        testTimeZoneFormatter("+1345", DateTimeZone.forOffsetHoursMinutes(+13, +45));
+        testTimeZoneFormatter("-0330", DateTimeZone.forOffsetHoursMinutes(-3, -30));
```
alxdarksage, line 399: "Z" is valid ISO 8601 time zone, will that work? same with something like "+13:45", will that work? I'm getting this from https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC

DwayneJengSage: Added more tests.

PR #1349, opened 2017-01-26
Subscriptions

alxdarksage:
Adding the ability to subscribe to topics once you have a device registration.

- also consolidated the creation of dummy test objects to use static methods on TestUtils.


app/org/sagebionetworks/bridge/dynamodb/DynamoTopicSubscriptionDao.java
```diff
@@ -0,0 +1,116 @@
...
+        mapper.delete(subscription);
+    }
+
+    @Override
+    public void delete(TopicSubscription subscription) {
+        checkNotNull(subscription);
+        // No exception is caught here, so user will see errors, but nevertheless we will
+        // try and clean up both data stores.
+        try {
+            snsClient.unsubscribe(subscription.getSubscriptionARN());
```
liujoshua: If we fail to unsubscribe, don't we want to keep the DDB record for integrity so we can attempt deletion again? Or is there a reason this is different from how deletion is done in unsubscribe?

liujoshua (approved):

app/org/sagebionetworks/bridge/dynamodb/DynamoTopicSubscriptionDao.java
alxdarksage: The only place this is called is if we catch a NotFoundException when loading SNS topics. So we kind of expect that it will fail again and what we want is to get rid of the mapping record.

Nothing in this method indicates this how it is being used... I'll try and rename for clarity.

alxdarksage:
I've tried to rename that method to be clearer what its unusual purpose is.

liujoshua:
That's a good, descriptive name

PR #1354, opened 2017-01-27
Enable pattern error message validation.

alxdarksage:
I've updated the integration tests in all environments to pass once this is enabled, so this is now ok to push out.


DwayneJengSage:
This looks like the exact inverse of https://github.com/Sage-Bionetworks/BridgePF/pull/1348 Looks good to me!

DwayneJengSage (approved):

PR #1359, opened 2017-01-27
Remove the use of DynamoTestUtil.clearTable() on critical health code tables.

alxdarksage:
While testing notifications, I discovered my account's healthCode was changing in the local environment. After verifying this has not happened in production, and tracking down why it occurs, it comes down to the fact we're deleting records from the key mapping table, HealthCode, during the tests.

This would be catastrophic if it happened in any other environment. I don't know how these tests would run anywhere else, but I'm removing the use of this method so it doesn't happen locally.


DwayneJengSage:
It looks like the only remaining place this is being used is DynamoBackfillDaoTest. How difficult would it be to remove it from there as well, and then delete clearTables() so we can be sure no one is using it?

DwayneJengSage:
Other than that, change looks good.

alxdarksage:
Yeah let me remove the last use of clearTables() today, sounds good.

alxdarksage:
Okay, I changed that last test and removed the utility class that implements clearTable().

test/org/sagebionetworks/bridge/dynamodb/DynamoBackfillDaoTest.java
```diff
@@ -13,34 +13,58 @@
...
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
 
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapperConfig;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapperConfig.ConsistentReads;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapperConfig.SaveBehavior;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.newrelic.agent.deps.com.google.common.collect.Lists;
```
DwayneJengSage: Remove NewRelic from our imports. Is there any way we can permanently disable this library from showing up in our code?

DwayneJengSage (changes requested):
One minor change. Otherwise, looks good.

alxdarksage:
I removed that import. I also found a type filter feature in Eclipse where I can add the new relic package that keeps getting imported (haven't used before but it worked when fixing this class).

PR #1361, opened 2017-02-01
add logging when we fail to create ReportIndex; clean up ReportIndex unit tests

DwayneJengSage:
We're seeing strange failures where the ReportIndex is simply never created. Added some logging to one case where we're silently squelching errors to see if we can track this down.

Also, rewrote the DynamoReportIndexDaoTest, since many of these tests are nonsensical.

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoReportIndexDao.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoReportIndexDao.java
@@ -6,6 +6,8 @@ import java.util.Map;
 
 import javax.annotation.Resource;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
 import org.sagebionetworks.bridge.dao.ReportIndexDao;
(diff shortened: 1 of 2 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoReportIndexDao.java
```diff
@@ -63,6 +66,7 @@ public void addIndex(ReportDataKey key) {
         } catch(ConditionalCheckFailedException e) {
             // Do not throw an exception if the index already exists. This is called as a side
             // effect of saving a report, and can be called multiple times for a report.
+            LOG.debug("Error saving index in DynamoReportIndexDao.addIndex(): " + e.getMessage());
```
liujoshua, line 69: Is debug high enough log level?

liujoshua (approved):
Looks good. One question about the log level because I don't know all the scenarios we're interested in investigating

app/org/sagebionetworks/bridge/dynamodb/DynamoReportIndexDao.java
DwayneJengSage: I've confirmed that debug does make it into our logs.

liujoshua (approved):

PR #1363, opened 2017-02-02
min/maxAppVersion in schemas

DwayneJengSage:
This adds min/maxAppVersion in schemas. In a future changes, we will be using this information to allow the Scheduler to automatically populate schedules with schema revisions.


app/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -242,5 +242,25 @@ public static int getIntOrDefault(String value, int defaultValue) {
             throw new BadRequestException(value + " is not an integer");
         }
     }
-    
+
+    /**
+     * Creates a new copy of the map, removing any entries that have a null value (particularly easy to do this in
+     * JSON).
+     */
+    public static <K,V> Map<K,V> withoutNullEntries(Map<K,V> map) {
```
liujoshua, line 250: Would it be desirable to use this setting across the board?

https://fasterxml.github.io/jackson-databind/javadoc/2.6/com/fasterxml/jackson/databind/SerializationFeature.html#WRITE_NULL_MAP_VALUES

liujoshua (approved):

app/org/sagebionetworks/bridge/BridgeUtils.java
DwayneJengSage: Probably, but I don't think it'll help in this specific case. We want to remove null entries from the map from the data itself, not just when it's being serialized to JSON.

PR #1364, opened 2017-02-03
remove mapper.count to save ddb throughput

zaclinsage:
manual test done;
unit test done;
integ test done;

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
@@ -148,9 +148,7 @@ public class DynamoUploadDao implements UploadDao {
 
         String nextPageOffsetKey = (page.getLastEvaluatedKey() != null) ? page.getLastEvaluatedKey().get(UPLOAD_ID).getS() : null;
 
-        int total = mapper.count(DynamoUpload2.class, createCountQuery(studyId.getIdentifier(), startTime, endTime));
-
-        return new PagedResourceList<>(uploadList, null, pageSize, total)
+        return new PagedResourceList<>(uploadList, null, pageSize, uploadList.size())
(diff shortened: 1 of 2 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
```diff
@@ -148,9 +148,7 @@ public Upload getUpload(@Nonnull String uploadId) {
 
         String nextPageOffsetKey = (page.getLastEvaluatedKey() != null) ? page.getLastEvaluatedKey().get(UPLOAD_ID).getS() : null;
 
-        int total = mapper.count(DynamoUpload2.class, createCountQuery(studyId.getIdentifier(), startTime, endTime));
-
-        return new PagedResourceList<>(uploadList, null, pageSize, total)
+        return new PagedResourceList<>(uploadList, null, pageSize, uploadList.size())
```
alxdarksage, line 151: If there's more records than are returned in one page, total will be wrong. Total is so that the client can correctly render the total number of pages that are present... in paged resource lists, it's not the same as the number of records in a page.

alxdarksage (changes requested):
"total" in any paged list that is returned from the UI is the total number of records that fit the query... not the number of records in that page. You can get that from the length of the array itself.

DwayneJengSage (approved):

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
alxdarksage: We discussed offline and this code does not lie in the path of the call that is used by the UI and that currently implements pagination off a count of all the records in the query... OTOW this is fine to merge from my perspective.

alxdarksage (approved):

PR #1366, opened 2017-02-08
Improvements to exception-based responses

alxdarksage:
In creating a ConstraintViolationException I wanted to have some extra information in the JSON generated by the exception. which caused me to recognize that our handling of exception serialization was pretty bad (I don't believe we were serializing anything unless we specified it in a handler class that was doing instanceof checks... not very maintainable).

(Addresses BRIDGE-1697, BRIDGE-1372).

Changed the following:
- changed ExceptionInterceptor to serialize exceptions directly and then remove the fields that are specific to Throwables and not desirable to return to clients (this allows extra fields in our business exceptions to be exposed);
- added a couple of fields to the standard exception JSON: statusCode and type
- changed the BridgeErrorHandler which was returning everything as a 400 with an HTML response body, rather than the real response code at that point and a JSON payload that will work with client-side expectations about the content being returned;
- the BridgeErrorHandler now returns a different payload for 404s generated by the fact that an URL on the server is wrong, vs. a 404 because an entity doesn't exist in Bridge (type EndpointNotFoundException vs. type EntityNotFoundException)
- changed EntityAlreadyExistsException so it can handle any keys for any entity more generically;


test/org/sagebionetworks/bridge/exceptions/ConstraintViolationExceptionTest.java
```diff
@@ -0,0 +1,105 @@
...
+        assertEquals(409, result.status());
+        JsonNode node = BridgeObjectMapper.get().readTree(Helpers.contentAsString(result));
+        
+        assertEquals("Referenced in survey", node.get("message").asText());
+        assertEquals("ConstraintViolationException", node.get("type").asText());
+        assertEquals(409, node.get("statusCode").asInt());
+        
+        JsonNode entityKeys = node.get("entityKeys");
+        assertEquals("schemaIdValue", entityKeys.get("schemaId").asText());
+        assertEquals("10", entityKeys.get("schemaRevision").asText());
```
alxdarksage, line 58: This should probably be a Map<String,Object> so it can store revision as a number

app/org/sagebionetworks/bridge/play/interceptors/ExceptionInterceptor.java
```diff
@@ -60,8 +68,12 @@ private Result getResult(final Throwable throwable) throws JsonProcessingExcepti
...
         }
         String message = getMessage(throwable, status);
-        final ExceptionMessage exceptionMessage = new ExceptionMessage(throwable, message);
-        return Results.status(status, Json.toJson(exceptionMessage));
+        
+        ObjectNode node = (ObjectNode)BridgeObjectMapper.get().valueToTree(throwable);
+        node.put("message", message);
+        node.remove(UNEXPOSED_FIELD_NAMES);
+        
+        return Results.status(status, node);
```
DwayneJengSage, line 76: Should this filter out things like healthCode, studyId, etc? Or is that the responsibility of the code throwing the exception?

DwayneJengSage (approved):
Looks good. One open question, non-bocking.

zaclinsage (approved):
LGTM

app/org/sagebionetworks/bridge/play/interceptors/ExceptionInterceptor.java
alxdarksage: None of the exceptions should have compromising fields like this, and I've removed just serializing an entire entity as part of an exception (though those should prevent serializing health code too).

test/org/sagebionetworks/bridge/exceptions/ConstraintViolationExceptionTest.java
alxdarksage: I'll change this in the next check in when I start to use it.

PR #1367, opened 2017-02-08
Referential integrity checks for survey deletes

alxdarksage:
Survey deletion rules. Note that I have rewritten the DAO and moved these more complicated rules into the service. If there's concern about accessing the DAO directly, I can reorganize packages so the DAO is only visible to the service.

Logical deletes (hiding survey but keeping in db): once published, you cannot delete a survey. This means that once the survey might be used in a scheduled activity, you can always find it in the user interface, look up its schema, etc. This is more strict than it was before. This is how developers delete.

Physical deletes (remove from db): these are only used during tests or by admins, but we now verify that the survey is not in use in a schedule before allowing it. This is because we ignore the publication flag when doing these deletes. Tests must delete schedules first, the integration tests pass as is (I'll add a test for the constraint itself). There are two rules:

- If a schedule references to a specific survey version, don't allow it to be deleted. You can't make these through the Bridge Study Manager, but they're allowable in the API;

- If a schedule references the most-recently published version of a survey, verify this delete is not removing the last published instance of the survey. This is the more common case right now.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/SurveyService.java
+++ b/app/org/sagebionetworks/bridge/services/SurveyService.java
@@ -34,57 +34,22 @@ import org.springframework.validation.Validator;
 @Component
 public class SurveyService {
 
-    /**
-     * The survey is referenced in an activity by exact guid/createdOn version
-     */
-    private static final BiPredicate<Activity, GuidCreatedOnVersionHolder> EXACT_MATCH = (activity, keys) -> {
-        if (activity.getSurvey() != null) {
-            return activity.getSurvey().equalsSurvey(keys);
-        } else if (activity.getCompoundActivity() != null) {
-            CompoundActivity compoundActivity = activity.getCompoundActivity();
-            for (SurveyReference aSurveyRef : compoundActivity.getSurveyList()) {
-                if (aSurveyRef.equalsSurvey(keys)) {
-                    return true;
-                }
-            }
-        }
-        return false;
-    };
-
-    /**
-     * The survey is referenced in an activity that wants the most recently published version. If there is a match, we
-     * verify that there is more than one published version to fall back on.
-     */
-    private static final BiPredicate<Activity, GuidCreatedOnVersionHolder> PUBLISHED_MATCH = (activity, keys) -> {
-        if (activity.getSurvey() != null) {
-            return activity.getSurvey().getGuid().equals(keys.getGuid());
-        } else if (activity.getCompoundActivity() != null) {
-            CompoundActivity compoundActivity = activity.getCompoundActivity();
-            for (SurveyReference aSurveyRef : compoundActivity.getSurveyList()) {
-                if (aSurveyRef.getGuid().equals(keys.getGuid())) {
-                    return true;
-                }
-            }
-        }
-        return false;
-    };
-
     private Validator validator;
     private SurveyDao surveyDao;
     private SchedulePlanService schedulePlanService;
 
     @Autowired
-    public void setSurveyDao(SurveyDao surveyDao) {
+    final void setSurveyDao(SurveyDao surveyDao) {
         this.surveyDao = surveyDao;
     }
 
     @Autowired
-    public void setValidator(SurveyValidator validator) {
+    final void setValidator(SurveyValidator validator) {
         this.validator = validator;
     }
 
     @Autowired
-    public void setSchedulePlanService(SchedulePlanService schedulePlanService) {
+    final void setSchedulePlanService(SchedulePlanService schedulePlanService) {
         this.schedulePlanService = schedulePlanService;
     }
 
@@ -174,7 +139,7 @@ public class SurveyService {
      * survey version that could have been sent to users will remain in the API so you can look at its 
      * schema, etc. This is how study developers should delete surveys. 
      */
-    public void deleteSurvey(StudyIdentifier studyId, GuidCreatedOnVersionHolder keys) {
+    public void deleteSurvey(GuidCreatedOnVersionHolder keys) {
         checkArgument(StringUtils.isNotBlank(keys.getGuid()), "Survey GUID cannot be null/blank");
         checkArgument(keys.getCreatedOn() != 0L, "Survey createdOn timestamp cannot be 0");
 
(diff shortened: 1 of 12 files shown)


app/org/sagebionetworks/bridge/services/SurveyService.java
```diff
@@ -113,45 +163,54 @@ public Survey publishSurvey(StudyIdentifier study, GuidCreatedOnVersionHolder ke
...
      */
-    public void deleteSurvey(GuidCreatedOnVersionHolder keys) {
+    public void deleteSurvey(StudyIdentifier studyId, GuidCreatedOnVersionHolder keys) {
         checkArgument(StringUtils.isNotBlank(keys.getGuid()), "Survey GUID cannot be null/blank");
         checkArgument(keys.getCreatedOn() != 0L, "Survey createdOn timestamp cannot be 0");
 
+        Survey existing = surveyDao.getSurvey(keys);
+        if (existing.isDeleted()) {
+            throw new EntityNotFoundException(Survey.class);
+        }
```
DwayneJengSage: getSurvey() and isDeleted() are duplicated in both the Service and in the DAO. I feel like either this is business logic and should live in the Service, or it's implementation logic and should live in the DAO, but not both.

The isDeleted() check should be easy to get rid of.

The getSurvey() call would require you to pass the survey down through deleteSurvey(), which is kind of wonky. I'll leave the second part up to your judgement call.

test/org/sagebionetworks/bridge/models/schedules/SurveyReferenceTest.java
```diff
@@ -0,0 +1,42 @@
...
+        assertEquals(IDENTIFIER, ref.getIdentifier());
+        assertEquals(GUID, ref.getGuid());
+        assertEquals(CREATED_ON, ref.getCreatedOn());
+        assertTrue(ref.getHref().contains("/v3/surveys/abc/revisions/2017-02-09T20:15:59.558Z"));
+        
+        GuidCreatedOnVersionHolder keys1 = new GuidCreatedOnVersionHolderImpl(GUID, CREATED_ON.getMillis());
+        assertTrue(ref.equalsSurvey(keys1));
+        
+        GuidCreatedOnVersionHolder keys2 = new GuidCreatedOnVersionHolderImpl("def", CREATED_ON.getMillis());
+        assertFalse(ref.equalsSurvey(keys2));
```
DwayneJengSage, line 32: Branch coverage: same guid, different createdOn?

app/org/sagebionetworks/bridge/services/SurveyService.java
```diff
@@ -22,33 +34,72 @@
...
+        } else if (activity.getCompoundActivity() != null) {
+            CompoundActivity compoundActivity = activity.getCompoundActivity();
+            for (SurveyReference aSurveyRef : compoundActivity.getSurveyList()) {
+                if (aSurveyRef.getGuid().equals(keys.getGuid())) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    };
```
DwayneJengSage: EXACT_MATCH and PUBLISHED_MATCH are identical except for the use of SurveyReference.equalsSurvey(keys) vs SurveyReference.getGuid().equals(keys.getGuid()). Would it be worth moving the bulk of this code to findFirstMatchingPlan() and just lambda-izing the equality functions?

app/org/sagebionetworks/bridge/services/SurveyService.java
```diff
@@ -22,33 +34,72 @@
...
+
     @Autowired
     public void setValidator(SurveyValidator validator) {
         this.validator = validator;
     }
 
+    @Autowired
+    public void setSchedulePlanService(SchedulePlanService schedulePlanService) {
+        this.schedulePlanService = schedulePlanService;
+    }
```
DwayneJengSage: Minor: Make this final, to prevent accidental overriding.

app/org/sagebionetworks/bridge/services/SurveyService.java
```diff
@@ -113,45 +163,54 @@ public Survey publishSurvey(StudyIdentifier study, GuidCreatedOnVersionHolder ke
...
-     * 
-     * @param keys
+     * Logically delete this survey (mark it deleted and do not return it from any list-based APIs; continue 
+     * to provide it when the version is specifically referenced). Once a survey is published, you cannot 
+     * delete it, because we do not know if it has already been dereferenced in scheduled activities. Any 
+     * survey version that could have been sent to users will remain in the API so you can look at its 
+     * schema, etc. This is how study developers should delete surveys. 
      */
-    public void deleteSurvey(GuidCreatedOnVersionHolder keys) {
+    public void deleteSurvey(StudyIdentifier studyId, GuidCreatedOnVersionHolder keys) {
```
DwayneJengSage: Minor: Do we need to take in the studyId here? Is it just for consistency with deleteSurveyPermanently?

DwayneJengSage (changes requested):
Overall looks good. A few small comments. I like the thorough tests in SurveyServiceMockTest.

test/org/sagebionetworks/bridge/models/schedules/SurveyReferenceTest.java
```diff
@@ -0,0 +1,47 @@
...
+        // Doesn't match published links, only specific ones.
+        ref = new SurveyReference(IDENTIFIER, GUID, null);
+        assertFalse(ref.equalsSurvey(keys1));
+        assertFalse(ref.equalsSurvey(keys2));
+        
+        assertFalse(ref.equalsSurvey(null));
+        
+        // Same guid, different createdOn
+        GuidCreatedOnVersionHolder keys3 = new GuidCreatedOnVersionHolderImpl("def", DateTime.now().getMillis());
+        assertFalse(ref.equalsSurvey(keys3));
```
DwayneJengSage: This test case is actually for different guid _and_ different createdOn. Which is probably also worth testing, but we're still missing the test case for same guid different createdOn.

DwayneJengSage (approved):
Looks good. One comment about tests. Feel free to merge after adding the test. Thanks!

PR #1368, opened 2017-02-09
Create study and user

zaclinsage:
unit test done;
manual test done;
java client should be changed accordingly;


test/org/sagebionetworks/bridge/models/studies/StudyAndUserHolderTest.java
```diff
@@ -0,0 +1,92 @@
...
+    private static final String TEST_STUDY_ID = "test-study-id";
+    private static final String TEST_STUDY_NAME = "test=study-name";
+    private static final String TEST_USER_EMAIL = "test+user@email.com";
+    private static final String TEST_USER_EMAIL_2 = "test+user+2@email.com";
+    private static final String TEST_USER_FIRST_NAME = "test_user_first_name";
+    private static final String TEST_USER_LAST_NAME = "test_user_last_name";
+    private static final String TEST_USER_PASSWORD = "test_user_password";
+
+    @Test
+    public void deserializeCorrectly() throws Exception {
```
DwayneJengSage: We need a test for the other way around, because the user will send us JSON and we need to parse it into a POJO.

Also, terminology note, deserialization is when you take a string and turn it into a POJO. Serialization is when you take a POJO and turn it into a string.

test/org/sagebionetworks/bridge/models/studies/StudyAndUserHolderTest.java
```diff
@@ -0,0 +1,92 @@
...
+        String studyAndUserString = BridgeObjectMapper.get().writeValueAsString(holder);
+        JsonNode studyAndUserNode = BridgeObjectMapper.get().readTree(studyAndUserString);
+        JsonNode studyNode = studyAndUserNode.get("study");
+        ArrayNode users = (ArrayNode) studyAndUserNode.get("users");
+
+        // verify
+        assertEquals(TEST_STUDY_ID, studyNode.get("identifier").asText());
+        assertEquals(TEST_STUDY_NAME, studyNode.get("name").asText());
+        assertEquals(TEST_USER_EMAIL, studyNode.get("supportEmail").asText());
+        assertEquals(true, studyNode.get("active").asBoolean());
```
DwayneJengSage: You should use textValue() and boolValue() instead of asText() and asBooloean(). For textValue(), this might not make a difference, because if it's a non-string value, it almost certainly won't match your test.

For boolValue(), this matters. We want to verify that we're serializing into the boolean true and not the string "true".

app/org/sagebionetworks/bridge/play/controllers/StudyController.java
```diff
@@ -136,6 +138,15 @@ public Result createStudy() throws Exception {
         return okResult(new VersionHolder(study.getVersion()));
     }
 
+    public Result createStudyAndUser() throws Exception {
+        getAuthenticatedSession(ADMIN);
+
+        StudyAndUserHolder studyAndUserHolder = parseJson(request(), StudyAndUserHolder.class);
+        Study study = studyService.createStudyAndUser(studyAndUserHolder);
+
+        return okResult(new VersionHolder(study.getVersion()));
```
DwayneJengSage: Minor: This should be createdResult()

test/org/sagebionetworks/bridge/play/controllers/StudyControllerTest.java
```diff
@@ -201,6 +213,50 @@ public void deactivateStudyThrowsGoodException() throws Exception {
...
+
+        // execute
+        controller.createStudyAndUser();
+
+        // verify
+        verify(mockStudyService, times(1)).createStudyAndUser(any());
+        StudyAndUserHolder capObj = argumentCaptor.getValue();
+        assertEquals(study, capObj.getStudy());
+        assertEquals(mockUsers, capObj.getUsers());
+    }
```
DwayneJengSage: Minor: We should validate that result.status() includes the expected status code and that the result contains something reasonable.

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -166,6 +176,40 @@ public Study getStudy(StudyIdentifier studyId) {
...
     }
 
+    public Study createStudyAndUser(StudyAndUserHolder studyAndUserHolder) throws SynapseException {
+        checkNotNull(studyAndUserHolder, Validate.CANNOT_BE_NULL, "study and users");
+        if (studyAndUserHolder.getUsers() == null) {
+            throw new BadRequestException("User list cannot be null.");
+        }
+        if (studyAndUserHolder.getStudy() == null) {
+            throw new BadRequestException("Study cannot be null.");
+        }
```
DwayneJengSage: Is it worth creating a validator for StudyAndUserHolder? This allows us to check for missing (or empty) user list and missing study, as well as validate the parameters in the study and users. This means callers won't have to keep going back and fixing their errors one by one. Rather, they get a big list of all the errors that they can fix at once.

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -166,6 +176,40 @@ public Study getStudy(StudyIdentifier studyId) {
...
+
+        Study study = studyAndUserHolder.getStudy();
+
+        // first create study
+        study = createStudy(study);
+
+        // then create users for that study
+        // send verification email as well
+        for (StudyParticipant user: users) {
+            IdentifierHolder identifierHolder = participantService.createParticipant(study, user.getRoles(), user,true);
```
DwayneJengSage: We should check that the roles only contain Developer or Researcher or both. We don't want people to be able to create more admins or worker accounts with this API. Also, a user with no roles is probably not appropriate here.

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
```diff
@@ -216,6 +234,115 @@ public void adminCannotSetActiveToFalse() {
     }
 
     @Test
+    public void createStudyAndUser() throws SynapseException {
```
DwayneJengSage, line 242: createStudy() and createSynapseProjectTeam() are already tested elswhere. It might be easiest to just spy StudyService, create mocks around createStudy() and createSynapse(), and just verify that these are called with the expected arguments.

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
```diff
@@ -248,7 +375,7 @@ public void createSynapseProjectTeam() throws SynapseException {
         AccessControlList capturedProjectAcl = argumentProjectAcl.getValue();
         Set<ResourceAccess> capturedProjectAclSet = capturedProjectAcl.getResourceAccess();
         assertNotNull(capturedProjectAclSet);
-        assertEquals(capturedProjectAclSet.size(), 2); // only has target user and exporter
+        assertEquals(capturedProjectAclSet.size(), 3); // only has target user, exporter and team
```
DwayneJengSage, line 482: Do we need to validate the 3rd ACL?

conf/bridge-server.conf
```diff
@@ -3,6 +3,7 @@ synapse.api.key = your-api-key-here
 
 exporter.synapse.id = 3325672
 test.synapse.user.id = 3348228
+admin.synapse.user.id = 3351155
```
DwayneJengSage: I thought we decided on having a list of individual accounts rather than having a shared account?

test/org/sagebionetworks/bridge/dynamodb/DynamoUploadDaoTest.java
```diff
@@ -230,6 +230,7 @@ public void canPaginateUploads() throws Exception {
         List<? extends Upload> items = paginatedList.getItems();
         assertEquals(1, items.size()); // only return 1 upload in each page
         assertEquals(1, paginatedList.getPageSize());
+        assertEquals(1, paginatedList.getTotal());
```
DwayneJengSage: I think we decided to remove getTotal() from the test, since it wasn't a meaningful value. (Here and two other places in this file.)

DwayneJengSage (changes requested):
Overall design considerations: Is it better for this logic to live here or in the Admin App?

app/org/sagebionetworks/bridge/models/VersionHolder.java
```diff
@@ -1,10 +1,12 @@
 package org.sagebionetworks.bridge.models;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 public class VersionHolder {
 
     private final Long version;
     
-    public VersionHolder(Long version) {
+    public VersionHolder(@JsonProperty("version")Long version) {
```
alxdarksage: This is usually a readonly property (we create and return it but don't deserialize it), so if you aren't deserializing it anywhere, you shouldn't need this annotation.

app/org/sagebionetworks/bridge/models/studies/StudyAndUserHolder.java
```diff
@@ -0,0 +1,31 @@
+package org.sagebionetworks.bridge.models.studies;
+
+import java.util.List;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import org.sagebionetworks.bridge.models.BridgeEntity;
+import org.sagebionetworks.bridge.models.accounts.StudyParticipant;
+
+public final class StudyAndUserHolder implements BridgeEntity {
```
alxdarksage: I think you can just call this StudyWithUsers or StudyAndUsers. "Holder" as a word is reserved for key-holding objects that replace full entities, which this is not.

app/org/sagebionetworks/bridge/play/controllers/StudyController.java
```diff
@@ -136,13 +140,23 @@ public Result createStudy() throws Exception {
         return okResult(new VersionHolder(study.getVersion()));
     }
 
-    public Result createSynapse(String synapseUserId) throws Exception {
+    public Result createStudyAndUser() throws Exception {
+        getAuthenticatedSession(ADMIN);
+
+        StudyAndUserHolder studyAndUserHolder = parseJson(request(), StudyAndUserHolder.class);
+        Study study = studyService.createStudyAndUser(studyAndUserHolder);
```
alxdarksage: Minor: use plural..
StudyAndUser**s** studyAndUser**s** = parseJson(....)
createStudyAndUser**s**

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -166,6 +176,75 @@ public Study getStudy(StudyIdentifier studyId) {
         return studyDao.getStudies();
     }
 
+    public Study createStudyAndUser(StudyAndUserHolder studyAndUserHolder) throws SynapseException {
+        checkNotNull(studyAndUserHolder, Validate.CANNOT_BE_NULL, "study and users");
+        if (studyAndUserHolder.getAdminIds() == null) {
+            throw new BadRequestException("Admin Ids cannot be null.");
+        }
```
alxdarksage: And should have at least one value?

app/org/sagebionetworks/bridge/services/StudyService.java
alxdarksage: I agree with this, also it can help prevent partially creating the synapse entities only to receive an error, which will leave things broken.

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -204,10 +283,15 @@ public Study createStudy(Study study) {
...
-        if (StringUtils.isBlank(synapseUserId)) {
-            throw new BadRequestException("Synapse User ID is invalid.");
+    public Study createSynapseProjectTeam(List<String> synapseUserIds, Study study) throws SynapseException {
+        if (synapseUserIds == null) {
+            throw new BadRequestException("Synapse User IDs cannot be null.");
+        }
+
+        if (synapseUserIds.isEmpty()) {
+            throw new BadRequestException("Synapse User IDs cannot be empty.");
         }
```
alxdarksage, line 290: You could combine these tests

conf/bridge-server.conf
```diff
@@ -1,5 +1,5 @@
-synapse.user = your-username-here
-synapse.api.key = your-api-key-here
+synapse.user = Bridge.Study.Creator
+synapse.api.key = SW6jpgyt8ot4Hv2yAyIbEn5WTWi9jSOgbL/DPaM3/eEo//P8Cmy+w/dfN6NCblpRs0R2yWdwTzHR4qyHy1dsvg==
```
alxdarksage: You'll have to remove this from the PR, and then regenerate this key, de-activating this in Synapse. This shouldn't be checked into source control, it should be handled like other keys

conf/routes
```diff
@@ -168,14 +168,15 @@ POST /v3/recordExportStatuses      @org.sagebionetworks.bridge.play.controllers.
 
 # Studies
 GET    /v3/studies                  @org.sagebionetworks.bridge.play.controllers.StudyController.getAllStudies(format: String ?= null, summary: String ?= null)
-POST   /v3/studies                  @org.sagebionetworks.bridge.play.controllers.StudyController.createStudy 
+POST   /v3/studies                  @org.sagebionetworks.bridge.play.controllers.StudyController.createStudy
+POST   /v3/studies/createStudyAndUser   @org.sagebionetworks.bridge.play.controllers.StudyController.createStudyAndUser
```
alxdarksage: I dislike the use of "create" in a POST method which is always already a create. /v3/studies/init or something like that would be preferable (still not great but better).

test/org/sagebionetworks/bridge/models/studies/StudyAndUserHolderTest.java
```diff
@@ -0,0 +1,97 @@
...
+                "\t  \"name\": \"test=study-name\",\n" +
+                "\t  \"active\": \"true\"\n" +
+                "\t},\n" +
+                "\t\"users\": [\n" +
+                "\t\t{\n" +
+                "\t\t\t\"firstName\": \"test_user_first_name\",\n" +
+                "\t\t\t\"lastName\": \"test_user_last_name\",\n" +
+                "\t\t\t\"email\": \"test+user@email.com\",\n" +
+                "\t\t\t\"password\": \"test_user_password\",\n" +
+                "\t\t\t\"roles\": [\"DEVELOPER\",\"RESEARCHER\"]\n" +
```
alxdarksage: These should be lowercase... I would verify that works in your test, kind of surprised that upper-case works tbh

alxdarksage (changes requested):
Some changes, minor and major

app/org/sagebionetworks/bridge/services/StudyService.java
zaclinsage: already verified it

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -221,10 +302,12 @@ public Study createSynapseProjectTeam(String synapseUserId, Study study) throws
...
         // then check if the user id exists
-        try {
-            synapseClient.getUserProfile(synapseUserId);
-        } catch (SynapseNotFoundException e) {
-            throw new BadRequestException("Synapse user Id is invalid.");
+        for (String userId : synapseUserIds) {
+            try {
+                synapseClient.getUserProfile(userId);
+            } catch (SynapseNotFoundException e) {
+                throw new BadRequestException("Synapse user Id is invalid.");
```
alxdarksage: So, we could add the ID to these exceptions so the user knows which one has the error.

alxdarksage (changes requested):

PR #1370, opened 2017-02-13
Make it possible to validate external IDs without also requiring them when an account is created.

alxdarksage:
- Used the new study flag, isExternalIdRequiredOnSignUp, to make it possible to validate IDs without requiring them at sign up. Or to require an external ID that is not validated on sign up.

- took a check of the study configuration out of the ExternalIdService because this just made it harder to test that the correct behavior is occurring when called from ParticipantService (which is where it is called from).

- fixed an error in the test where we were passing the constant HEALTH_CODE instead of EXTERNAL_ID to the external ID service, which was confusing

- refactored ParticipantServiceTest, redoing the existing tests around external ID validation, so I could add many more tests of every combination of validation / required on create for both creation and update calls;


DwayneJengSage:
For future reference, can we separate the non-functional whitespace changes from the functional changes? It looks like a huge chunk of this changelist is just whitespace changes, which really bogs down the review process.

app/org/sagebionetworks/bridge/services/ExternalIdService.java
```diff
@@ -73,19 +73,15 @@ public void reserveExternalId(Study study, String externalIdentifier) {
...
     
     public void assignExternalId(Study study, String externalIdentifier, String healthCode) {
         checkNotNull(study);
         checkArgument(isNotBlank(externalIdentifier));
         checkArgument(isNotBlank(healthCode));
         
-        if (study.isExternalIdValidationEnabled()) {
-            externalIdDao.assignExternalId(study.getStudyIdentifier(), externalIdentifier, healthCode);
-        }
+        externalIdDao.assignExternalId(study.getStudyIdentifier(), externalIdentifier, healthCode);
```
DwayneJengSage: There's a use in UserProfileController that doesn't check study.isExternalIdValidationEnabled(). This might be a functional change. Should we patch it there as well?

Given that removing the check from here forces us to add the check in a whole bunch of other places, I would argue that here is probably the correct place to add this check. It enforces one of our invariants, keeps all the logic in once place, and reduces the overall complexity in calling code.

If we _do_ decide to move checks out of here, this method (as well as reserveExternalId()) no longer use the study. How much work is it to change this to take StudyIdentifier instead and have callers call with a StudyIdentifier instead of a Study?

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -307,102 +309,113 @@ public void resendConsentAgreement(Study study, SubpopulationGuid subpopGuid, St
...
-        // If we're validating the ID, we do this through the externalIdService, which writes to the participant options
-        // table when its appropriate to do so
-        if (study.isExternalIdValidationEnabled()) {
-            options.remove(EXTERNAL_IDENTIFIER);
+
+            updateValidatedExternalId(study, participant, account.getHealthCode());
         }
+
+        // Save all participant options except for the external ID, if we are going to use the registration service
+        Map<ParticipantOption, String> options = getAllOptionsButExternalId(study, participant);
```
DwayneJengSage: This comment used to say

        // If we're validating the ID, we do this through the externalIdService, which writes to the participant options
        // table when its appropriate to do so

The new comment is much more confusing. Case in point, I spent about 15 minutes tracing through the new code, old code, new tests, old tests, until I found this comment and figured out what was going on.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -896,6 +814,268 @@ public void sendNotification() {
...
+    // Creating an account and supplying an externalId
+    
+    @Test
+    public void createExternalIdValidatedRequiredWithValue() {
+        setupExternalIdTest(true, true);
+        
+        participantService.createParticipant(STUDY, CALLER_ROLES, PARTICIPANT, false);
+        
+        verifyIdReservation(EXTERNAL_ID);
+        verifyNotSetAsOption();
```
DwayneJengSage: A quick comment here explaining why when externalIdValidationEnabled=true means we don't set participant options. Otherwise, this leads to further confusion down the line.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -896,6 +814,268 @@ public void sendNotification() {
...
+    
+    @Test
+    public void updateExternalIdNotValidatedNotRequiredNewValue() {
+        setupExternalIdTest(false, false);
+        
+        participantService.updateParticipant(STUDY, CALLER_ROLES, NEW_ID_PARTICIPANT);
+        
+        verifyNotSetAsReservation();
+        verifySetAsOption("newExternalId");
+    }
```
DwayneJengSage: I noticed that you've thoroughly exhaustively tested every possibility. Generally speaking, this is a very commendable thing.

However, this also tells me that saveParticipant() is growing too complex. This is not sustainable, as its unit tests will continue to grow exponentially, and eventually we won't be able change it at all.

I don't know if now is the right time to do this, but at some point before the next update to saveParticipants(), we should refactor the code so that the complexity and test requirements of this class grow linearly rather than exponentially. At the very least, if we can isolate some of the variables from one another, we can reduce some of the complexity and avoid having to write 2^N tests.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/ExternalIdService.java
alxdarksage: It would be safest to add this back, and check redundantly in the ParticipantService to verify the functionality is not executed in the tests.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: The introduction of these behavior flags in particular has caused exponential growth in the tests, but I will admit I don't have an answer about how to reduce the complexity right now (other than to split the create/update code paths, which would make it easier to reason about but wouldn't reduce the number of tests).

alxdarksage:
I added back the configuration tests to ExternalIdService because it is called from that one legacy controller method to set an external ID, and that legacy endpoint which may be in use. I also added comments to the tests about what each test expects. 

Regarding complexity of ParticipantService: yes the flags are creating combinatorial increases in the method's complexity. I don't have a fix for this or I would attempt it... I think the create/update methods should be split, even though they'd be partly redundant, just to make them easier to reason about. But this wouldn't make the tests simpler or fewer in number, I don't think.

test/org/sagebionetworks/bridge/services/ExternalIdServiceTest.java
```diff
@@ -81,43 +83,21 @@ public void addExternalIds() {
...
     }
     
     @Test
-    public void reserveExternalIdWithoutVerification() {
-        STUDY.setExternalIdValidationEnabled(false);
-        externalIdService.reserveExternalId(STUDY, EXT_ID);
-        
-        verify(externalIdDao, never()).reserveExternalId(STUDY.getStudyIdentifier(), EXT_ID);
-        verifyNoMoreInteractions(optionsService);
-    }
```
DwayneJengSage: We should add these tests back in.

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -307,102 +309,115 @@ public void resendConsentAgreement(Study study, SubpopulationGuid subpopGuid, St
...
+        boolean newAccountAssigningId = isNew && study.isExternalIdValidationEnabled()
+                && isNotBlank(participant.getExternalId());
         Account account = null;
         if (isNew) {
             // Don't set it yet. Create the user first, and only assign it if that's successful.
             // Allows us to assure that credentials and ID will be related or not created at all.
-            if (isNotBlank(participant.getExternalId())) {
-                externalIdService.reserveExternalId(study, participant.getExternalId());    
+            if (newAccountAssigningId) {
+                externalIdService.reserveExternalId(study, participant.getExternalId());
```
DwayneJengSage, line 356: This check is now redundant with ExternalIdService. I would argue that having this check in ParticipantService is the cause of the complexity explosion we're seeing. We're making the code less modular and trying to make ParticipantService do too much.

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/services/ExternalIdServiceTest.java
```diff
@@ -99,25 +100,22 @@ public void reserveExternalIdWithoutVerification() {
     
     @Test
     public void assignExternalIdWithVerification() {
-        STUDY.setExternalIdValidationEnabled(true);
         externalIdService.assignExternalId(STUDY, EXT_ID, HEALTH_CODE);
         
         verify(externalIdDao).assignExternalId(STUDY.getStudyIdentifier(), EXT_ID, HEALTH_CODE);
-        verify(optionsService).setString(STUDY.getStudyIdentifier(), HEALTH_CODE, EXTERNAL_IDENTIFIER, EXT_ID);
```
DwayneJengSage: Was this verify(optionsService) mistakenly deleted?

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #1371, opened 2017-02-14
pull updated bridge-base to allow multiple DDB indices to use the same range key

DwayneJengSage:
Pulls in the change from https://github.com/Sage-Bionetworks/bridge-base/pull/30

Testing done: Created new tables under a clean namespace and verified a sample of the indices.


alxdarksage (approved):
Approved to merge after bridge-base

PR #1375, opened 2017-02-16
External id on signup refactor

alxdarksage:
https://sagebionetworks.jira.com/browse/BRIDGE-1715

### DIFF
--- a/app/org/sagebionetworks/bridge/services/ExternalIdService.java
+++ b/app/org/sagebionetworks/bridge/services/ExternalIdService.java
@@ -75,13 +75,8 @@ public class ExternalIdService {
         checkNotNull(study);
         checkNotNull(healthCode);
         
-        if (study.isExternalIdValidationEnabled() && externalIdentifier != null) {
-            ParticipantOptionsLookup lookup = optionsService.getOptions(healthCode);
-            String existingId = lookup.getString(EXTERNAL_IDENTIFIER);
-
-            if (!externalIdentifier.equals(existingId)) {
-                externalIdDao.reserveExternalId(study.getStudyIdentifier(), externalIdentifier);    
-            }
+        if (study.isExternalIdValidationEnabled() && isNotBlank(externalIdentifier)) {
+            externalIdDao.reserveExternalId(study.getStudyIdentifier(), externalIdentifier);    
         }
     }
     
--- a/app/org/sagebionetworks/bridge/services/ExternalIdService.java
+++ b/app/org/sagebionetworks/bridge/services/ExternalIdService.java
@@ -2,6 +2,7 @@ package org.sagebionetworks.bridge.services;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
+import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static org.sagebionetworks.bridge.dao.ParticipantOption.EXTERNAL_IDENTIFIER;
 
@@ -17,6 +18,7 @@ import org.sagebionetworks.bridge.dao.ExternalIdDao;
 import org.sagebionetworks.bridge.exceptions.BadRequestException;
 import org.sagebionetworks.bridge.models.PagedResourceList;
 import org.sagebionetworks.bridge.models.accounts.ExternalIdentifierInfo;
+import org.sagebionetworks.bridge.models.accounts.ParticipantOptionsLookup;
 import org.sagebionetworks.bridge.models.studies.Study;
 import org.sagebionetworks.bridge.validators.ExternalIdsValidator;
 import org.sagebionetworks.bridge.validators.Validate;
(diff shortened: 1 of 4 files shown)


test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -151,8 +144,8 @@
     @Mock
     private CacheProvider cacheProvider;
     
-    @Mock
-    private ExternalIdService externalIdService;
+    @Mock 
+    private ExternalIdDao externalIdDao;
```
DwayneJengSage: DAO is not in use?

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -214,11 +213,68 @@ public void signUserOut(Study study, String email) {
...
+
+    private void updateAccountOptionsAndRoles(Study study, Set<Roles> callerRoles, Map<ParticipantOption, String> options,
+            Account account, StudyParticipant participant) {
+        for (ParticipantOption option : ParticipantOption.values()) {
+            options.put(option, option.fromParticipant(participant));
+        }
+        // If validation is enabled, don't just save this like any other option, regardless of what it is.
+        if (study.isExternalIdValidationEnabled()) {
+            options.remove(EXTERNAL_IDENTIFIER);
+        }
```
DwayneJengSage: Why not just remove External ID all the time? ExternalIdService already saves the ID to Participant Options. This way, this code doesn't depend on the validation flag, and we can simplify our business logic and our tests.

app/org/sagebionetworks/bridge/services/ExternalIdService.java
```diff
@@ -69,22 +71,36 @@ public void addExternalIds(Study study, List<String> externalIdentifiers) {
...
+        
+        if (study.isExternalIdValidationEnabled() && externalIdentifier != null) {
+            ParticipantOptionsLookup lookup = optionsService.getOptions(healthCode);
+            String existingId = lookup.getString(EXTERNAL_IDENTIFIER);
 
-        if (study.isExternalIdValidationEnabled()) {
-            externalIdDao.reserveExternalId(study.getStudyIdentifier(), externalIdentifier);
+            if (!externalIdentifier.equals(existingId)) {
+                externalIdDao.reserveExternalId(study.getStudyIdentifier(), externalIdentifier);    
+            }
```
DwayneJengSage: What's the use case for allowing the user to reserve an external ID that's different from their existing one? It will fail when they try to assign it anyway.

test/org/sagebionetworks/bridge/services/ExternalIdServiceTest.java
```diff
@@ -156,4 +148,150 @@ public void deleteExternalIdsWithValidationEnabled() {
...
+    @Test(expected = BadRequestException.class)
+    public void updateExternalIdValidatedNewValue() {
+        setupExternalIdTest(true, EXT_ID);
+        
+        externalIdService.reserveExternalId(STUDY, "newExternalId", HEALTH_CODE);
+        externalIdService.assignExternalId(STUDY, "newExternalId", HEALTH_CODE);
+        
+        // Updating validated value with new value throws exception
+        verifySetAsReservation("newExternalId");
+    }
```
DwayneJengSage: updateExternalIdValidatedNewValue() and updateExternalIdNotValidatedNewValue() appear to be identical to updateExternalIdValidatedWithChangedValue() and updateExternalIdNotValidatedWithChangedValue().

test/org/sagebionetworks/bridge/services/ExternalIdServiceTest.java
```diff
@@ -156,4 +148,150 @@ public void deleteExternalIdsWithValidationEnabled() {
...
+    @Test
+    public void createExternalIdNotValidatedWithValue() {
+        setupExternalIdTest(false, null);
+        
+        externalIdService.reserveExternalId(STUDY, EXT_ID, HEALTH_CODE);
+        externalIdService.assignExternalId(STUDY, EXT_ID, HEALTH_CODE);
+        
+        // Adding unvalidated ID, set as an option
+        verifySetAsOption(EXT_ID);
+    }
```
DwayneJengSage, line 153: reserve() and assign() don't actually do anything when validation=false. Is it really necessary to have half a dozen tests that all test for the exact same thing when validation=false?

app/org/sagebionetworks/bridge/services/ExternalIdService.java
```diff
@@ -69,22 +71,36 @@ public void addExternalIds(Study study, List<String> externalIdentifiers) {
...
+            String existingExternalId = lookup.getString(EXTERNAL_IDENTIFIER);
+
+            if (isBlank(existingExternalId)) {
+                if (isNotBlank(externalIdentifier)) {
+                    externalIdDao.assignExternalId(study.getStudyIdentifier(), externalIdentifier, healthCode);
+                }
+            } else if (!existingExternalId.equals(externalIdentifier)) {
+                throw new BadRequestException(
+                        "External ID cannot be changed or removed after assignment.");
+            }
```
DwayneJengSage, line 98: This logic is duplicated between reserve() and assign(). If we change the semantics so you always have to call reserve() before assign(), we could simplify this logic. This means, by the time we get to assign(), we know that we aren't trying to change or remove an existing external ID.

test/org/sagebionetworks/bridge/services/ExternalIdServiceTest.java
```diff
@@ -80,25 +85,9 @@ public void addExternalIds() {
...
-        externalIdService.reserveExternalId(STUDY, EXT_ID);
-        
-        verify(externalIdDao, never()).reserveExternalId(STUDY.getStudyIdentifier(), EXT_ID);
-        verifyNoMoreInteractions(optionsService);
-    }
-    
-    @Test
     public void assignExternalIdWithVerification() {
+        doReturn(lookup).when(optionsService).getOptions(HEALTH_CODE);
+
```
DwayneJengSage: This test appears to be redundant with createExternalIdValidatedWithValue().

test/org/sagebionetworks/bridge/services/ExternalIdServiceTest.java
```diff
@@ -108,6 +97,8 @@ public void assignExternalIdWithVerification() {
     
     @Test
     public void assignExternalIdWithoutVerification() {
+        doReturn(lookup).when(optionsService).getOptions(HEALTH_CODE);
+
```
DwayneJengSage: This test appears to be redundant with createExternalIdNotValidatedWithValue().

test/org/sagebionetworks/bridge/services/ExternalIdServiceTest.java
DwayneJengSage: The verifySetAsReservation() here will never be called, since assign() throws.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -1200,19 +925,18 @@ private void setupExternalIdTest(boolean withValidation, boolean requiredOnSignU
...
     }
     
-    private void verifyNotSetAsReservation() {
-        verify(externalIdService, never()).reserveExternalId(STUDY, EXTERNAL_ID);
-        verify(externalIdService, never()).assignExternalId(STUDY, EXTERNAL_ID, HEALTH_CODE);
+    private void verifyNotSetAsReservation(String externalId, boolean isUpdate) {
+        // This is now always called, so it's not so easy to verify.
+        if (!isUpdate) {
+            verify(externalIdService).reserveExternalId(STUDY, externalId, HEALTH_CODE);    
+        }
```
DwayneJengSage: Similarly, if we changed the semantics so reserve is always called, we could remove the if statement and simplify the test.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: yes, removed if clause

app/org/sagebionetworks/bridge/services/ExternalIdService.java
alxdarksage: removed

app/org/sagebionetworks/bridge/services/ExternalIdService.java
alxdarksage: Once you remove the check in reserve the logic isn't duplicated. It just does a blind reserve whether assigned or not, or changed or not.

test/org/sagebionetworks/bridge/services/ExternalIdServiceTest.java
alxdarksage: Removed the four redundant tests. I would like to leave the remaining 10 tests for the various combinations of calls with existing states.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Cleaned up tests (some of these helper methods aren't needed once you simplify).

test/org/sagebionetworks/bridge/services/ExternalIdServiceTest.java
alxdarksage: Okay, I've removed all but one test for not validated path. This does depend on knowing these service methods do nothing if that flag is set to false, were that to change, the tests might not catch change in behavior.

DwayneJengSage (approved):

PR #1376, opened 2017-02-17
ScheduledActivityService resolves schema revs

DwayneJengSage:
This change refactors ScheduledActivityService to resolve compound activities and schema revs (in addition to surveys). This change is fairly big, largely in part because I had to make a bunch of changes to enable testing ScheduledActivityService.

I also fixed an issue where, once the activities were created in DDB, they would never be updated. This is bad because, if we publish a new survey version, or bump a new schema rev, it could potentially not show up to the end user (or to app developers) for days. I added a fix so that if the user hasn't done anything with the activity (not started, not finished, not expired), we can update the activity in DDB.

In addition to integ tests and unit tests, I also performed the manual tests with the following definitions:
* 3 published surveys
* 3 schemas, one of which has rev 1 w/ min app version 2 and rev 2 with min app version 4
* Compound Activity Definition as follows:
`
{
  "taskId":"compound-from-def",
  "schemaList":[
    {"id":"simple-test-schema"}
  ],
  "surveyList":[
    {"guid":"2a93e632-c760-4b52-b567-ec36b26acfa4"}
  ]
}
`

* Schedule PLan as follows:
`
{
  "label":"Schedule w/ Compound Activity",
  "strategy":{
    "type":"SimpleScheduleStrategy",
    "schedule":{
      "scheduleType":"ONCE",
      "activities":[
        {
          "label":"Compound Activity From Definition",
          "compoundActivity":{
            "taskIdentifier":"compound-from-def"
          }
        },
        {
          "label":"Compound Activity Without Definition",
          "compoundActivity":{
            "taskIdentifier":"compound-without-def",
            "schemaList":[
              {"id":"simple-attachment"}
            ],
            "surveyList":[
              {"guid":"b974e46d-e230-4367-b2ed-60524a2b94d7"}
            ]
          }
        },
        {
          "label":"Survey Activity",
          "survey":{"guid":"b953673d-0e94-4b75-ad66-53ba025a98a0"}
        },
        {
          "label":"Task Activity with Schema",
          "task":{
            "identifier":"back-compat-task",
            "schema":{"id":"upload-v2-manual-test"}
          }
        }
      ]
    }
  }
}
`

First I tested with app version 3 and verified that the schema was resolved with revision 1. Then, I tested with app version 5 and verified the schema was resolved with revision 2. In both cases, I verified that the compound activity, all 3 schemas, and all 3 surveys were being resolved.

Integ test changes at https://github.com/Sage-Bionetworks/BridgeIntegrationTests/pull/134 makes this change easier to test, but it's not strictly required.


alxdarksage:
"I added a fix so that if the user hasn't done anything with the activity (not started, not finished, not expired), we can update the activity in DDB."

-- This is probably okay, these tasks could have been retrieved and then cached for days on the client. But worst case scenario, the task submits with the schema value even if it would have been updated if they had asked again.

alxdarksage (approved):

PR #1379, opened 2017-02-24
Fixing special case of UTC time zone parsing.

alxdarksage:
When you parse "+00:00" into a Joda DateTimeZone object, and call toString() on it, you get back the string "UTC". Changed ParticipantOption to preserve this as "+00:00" and improved our utility for parsing time zones so it returns DateTimeZone.UTC if it encounters "UTC" as the string.

This is the only case I could find of this in the documentation, the other offsets are recorded as we'd expect, so far as I can tell.

### DIFF
--- a/app/org/sagebionetworks/bridge/dao/ParticipantOption.java
+++ b/app/org/sagebionetworks/bridge/dao/ParticipantOption.java
@@ -21,23 +21,13 @@ public enum ParticipantOption {
 
     TIME_ZONE(null, "timeZone") {
         public String fromParticipant(StudyParticipant participant) {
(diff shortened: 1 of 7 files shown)


DwayneJengSage (changes requested):
Looks good. I think we also want to change ParticipantOptionsService.setDateTimeZone() to use DateUtils, for consistency.

Also, we should run integ tests ConsentTest.canWithdrawFromAllConsentsInStudy() and ScheduleTest.criteriaBasedScheduleIsFilteredForUse() locally as a final verification that everything is fixed.

Thanks!

DwayneJengSage (approved):

PR #1381, opened 2017-02-28
Factor out request time zone

alxdarksage:
Separated out the request time zone into a separate field (rather than reading endsOn).

### DIFF
--- a/app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
+++ b/app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
@@ -229,18 +229,24 @@ public final class ScheduleContext {
         
         public ScheduleContext build() {
             // ScheduleContext is validated to have the correct values:
-            // - requestTimeZone is required
+            // - requestTimeZone is required (when validated)
             // - now is advisable, but still not required
             // - endsOn OR daysAhead needs to be supplied to calculate endsOn from now
             if (initialTimeZone == null) {
                 initialTimeZone = requestTimeZone;
             }
-            if (now == null && requestTimeZone != null) {
-                now = DateTime.now(requestTimeZone);
-            }
-            if (endsOn == null && now != null) {
-                endsOn = now.plusDays(daysAhead).withHourOfDay(23).withMinuteOfHour(59).withSecondOfMinute(59)
-                        .withMillisOfSecond(999);
+            if (requestTimeZone != null) {
+                if (now == null) {
+                    now = DateTime.now(requestTimeZone);
+                } else {
+                    now = now.withZone(requestTimeZone);    
+                }
+                if (endsOn == null) {
+                    endsOn = now.plusDays(daysAhead).withHourOfDay(23).withMinuteOfHour(59).withSecondOfMinute(59)
+                            .withMillisOfSecond(999);
+                } else {
+                    endsOn = endsOn.withZone(requestTimeZone);
+                }
             }
             return new ScheduleContext(initialTimeZone, requestTimeZone, endsOn, events, now, minimumPerSchedule,
                     accountCreatedOn, contextBuilder.build());
--- a/app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
+++ b/app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
@@ -24,6 +24,7 @@ import com.google.common.collect.ImmutableMap;
 public final class ScheduleContext {
     
     private final DateTimeZone initialTimeZone;
+    private final DateTimeZone requestTimeZone;
     private final DateTime endsOn;
     private final Map<String,DateTime> events;
     private final DateTime now;
@@ -31,9 +32,11 @@ public final class ScheduleContext {
     private final int minimumPerSchedule;
     private final CriteriaContext criteriaContext;
     
-    private ScheduleContext(DateTimeZone initialTimeZone, DateTime endsOn, Map<String, DateTime> events, DateTime now,
-            int minimumPerSchedule, DateTime accountCreatedOn, CriteriaContext criteriaContext) {
+    private ScheduleContext(DateTimeZone initialTimeZone, DateTimeZone requestTimeZone, DateTime endsOn,
+            Map<String, DateTime> events, DateTime now, int minimumPerSchedule, DateTime accountCreatedOn,
+            CriteriaContext criteriaContext) {
         this.initialTimeZone = initialTimeZone;
+        this.requestTimeZone = requestTimeZone;
         this.endsOn = endsOn;
         this.events = events;
         this.now = now;
@@ -49,6 +52,16 @@ public final class ScheduleContext {
      */
     public DateTimeZone getInitialTimeZone() {
         return initialTimeZone;
+    }
+
+    /**
+     * The time zone of the current request, sent to the server as a UTC offset. This is the time zone that local 
+     * date and time schedules should use when returning timestamps. For example, a recurring activity at 02:00 PM,
+     * if requested for the time zone -05:00, will return a timestamp of "YYYY-MM-DDT14:00:00.000-05:00". In 
+     * effect this makes the timestamp a local date and time value.
+     */
+    public DateTimeZone getRequestTimeZone() {
+        return requestTimeZone;
     }
     
     /**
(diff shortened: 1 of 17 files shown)


test/org/sagebionetworks/bridge/models/schedules/ScheduleContextTest.java
```diff
@@ -67,6 +72,7 @@ public void builderWorks() {
                 .withClientInfo(clientInfo)
                 .withStudyIdentifier(studyId)
                 .withInitialTimeZone(PST)
+                .withRequestTimeZone(PST)
```
DwayneJengSage: We set the requestTimeZone, but we don't verify it anywhere?

Also, is it worth making requestTimeZone and initialTimeZone different to make sure we're not crossing our wires?

app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
```diff
@@ -200,13 +228,22 @@ public Builder withContext(ScheduleContext context) {
         }
         
         public ScheduleContext build() {
-            // pretty much everything else is optional. I would like healthCode to be required, but it's not:
-            // we use these selection criteria to select subpopulations on sign up.
-            if (now == null) {
-                now = (initialTimeZone == null) ? DateTime.now() : DateTime.now(initialTimeZone);
+            // ScheduleContext is validated to have the correct values:
+            // - requestTimeZone is required
```
DwayneJengSage: Clarify in comment that requestTimeZone is required, but is enforced in the validator, not here.

Also, is it worth having a unit test where requestTimeZone is null, since nothing in this class actually prevents that.

app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
```diff
@@ -200,13 +228,22 @@ public Builder withContext(ScheduleContext context) {
...
+            // - now is advisable, but still not required
+            // - endsOn OR daysAhead needs to be supplied to calculate endsOn from now
+            if (initialTimeZone == null) {
+                initialTimeZone = requestTimeZone;
+            }
```
DwayneJengSage, line 237: Unit test for setting initialTimeZone based off requestTimeZone?

app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
```diff
@@ -200,13 +228,22 @@ public Builder withContext(ScheduleContext context) {
...
+            if (now == null && requestTimeZone != null) {
+                now = DateTime.now(requestTimeZone);
+            }
```
DwayneJengSage: Before, not specifying timezone would result in now using the default timezone. Now, now is null. Is that intentional?

test/org/sagebionetworks/bridge/models/schedules/IntervalActivitySchedulerTest.java
```diff
@@ -70,18 +70,19 @@ public void onceWithoutTimesUsesLocalTime() throws Exception {
...
         schedule.setScheduleType(ScheduleType.ONCE);
         schedule.setExpires("P2M");
+        DateTimeZone zone = DateTimeZone.forOffsetHours(-7);
         
         ScheduleContext context= new ScheduleContext.Builder()
                 .withStudyIdentifier(TEST_STUDY)
-                .withInitialTimeZone(DateTimeZone.forOffsetHours(-7))
-                .withEndsOn(ENROLLMENT.plusDays(2)) // in UTC
+                .withRequestTimeZone(zone)
+                .withDaysAhead(2)
```
DwayneJengSage: In this test, the initial and the requested timezone were different. Now they are the same. This changes the semantics of the test. Are you sure we're not losing test coverage on this?

Semantically, we should be replacing endsOn.getTimeZone() with requestTimeZone, not initialTimeZone. This seems to be happening a lot in a lot of the other tests as well.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceRecurringTest.java
```diff
@@ -179,32 +179,25 @@ public void persistedActivitiesAreFilteredByEndsOn() throws Exception {
...
-            .withEndsOn(endsOn.withHourOfDay(23).withMinuteOfHour(59).withSecondOfMinute(59))
-            .withHealthCode(testUser.getHealthCode())
-            .withUserId(testUser.getId()).build();
-    }
-}
+                .withStudyIdentifier(study.getStudyIdentifier())
+                .withClientInfo(ClientInfo.UNKNOWN_CLIENT)
+                .withInitialTimeZone(initialTimeZone)
+                .withRequestTimeZone(requestTimeZone)
+                .withNow(now.withZone(requestTimeZone))
```
DwayneJengSage, line 198: Does any of our functionality depend on the timezone of now? This change changes the timezone of now in some tests, which may change the functionality and test coverage.

app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
```diff
@@ -200,13 +228,22 @@ public Builder withContext(ScheduleContext context) {
...
+            if (endsOn == null && now != null) {
+                endsOn = now.plusDays(daysAhead).withHourOfDay(23).withMinuteOfHour(59).withSecondOfMinute(59)
+                        .withMillisOfSecond(999);
```
DwayneJengSage: now and requestedTimeZone may be different. This may lead you into weird states where endsOn isn't in the requestedTimeZone.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
alxdarksage: Added (also verifying null works).

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceRecurringTest.java
alxdarksage: Nothing breaks...  logically I'd expect this since we're trying to return tasks in the request time zone, so starting calculations from that time zone for scheduledOn times should work. However, I could go back to generating this in the default time zone (probably UTC in prod) if you think I should.

app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
alxdarksage: Added (and test for requestTimeZone == null).

alxdarksage (commented):
I've updated. I didn't change the conversion of "now" to the user's request time zone. I believe this works, but I could revert it to system time zone. I was trying to set the stage to provide now through the API at a later date, if desired for testing or any other purpose.

app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
```diff
@@ -200,13 +228,28 @@ public Builder withContext(ScheduleContext context) {
...
+            // - endsOn OR daysAhead needs to be supplied to calculate endsOn from now
+            if (initialTimeZone == null) {
+                initialTimeZone = requestTimeZone;
+            }
+            if (requestTimeZone != null) {
+                if (now == null) {
+                    now = DateTime.now(requestTimeZone);
+                } else {
+                    now = now.withZone(requestTimeZone);    
+                }
```
DwayneJengSage, line 243: There are a few places in the code where we assume ScheduleContext.getNow() is not null. Under the old code, the builder would always ensure now is not null. Under the new code, we cannot guarantee that to be the case. Do we need to fix this?

test/org/sagebionetworks/bridge/dynamodb/DynamoScheduledActivityDaoMockTest.java
```diff
@@ -131,7 +131,7 @@ public void testGetActivities() throws Exception {
         ScheduleContext context = new ScheduleContext.Builder()
             .withStudyIdentifier(TEST_STUDY)
             .withClientInfo(ClientInfo.UNKNOWN_CLIENT)
-            .withInitialTimeZone(PACIFIC_TIME_ZONE)
+            .withRequestTimeZone(PACIFIC_TIME_ZONE)
```
DwayneJengSage, line 134: There are a lot of tests where we're replacing initialTimeZone with requestTimeZone instead of endsOn.getTimeZone() with requestTimeZone. This changes the semantics of a lot of tests, and means we could be losing test coverage. Are we sure we want to do this?

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceRecurringTest.java
DwayneJengSage: Tests not failing isn't the same as nothing breaking, especially if we changed the tests, as is the case here. Instead of trying to logically deduce why these tests don't matter, I'd rather keep these tests intact and prove that everything is still working.

PR #1385, opened 2017-03-04
Enforce manual creation of activity GUID on schedule plan create

alxdarksage:
We're going to index scheduled activities by the activity guid, and the current design has led to cases where, if a client does not provide GUIDs for the activities, they get recreated and break things. So we really need for that to not happen.

I may explore creating a key based on the content of an Activity, such that even if the user deletes and recreates an activity in JSON, it has a stable GUID. I think this could prevent a class of problems where people are screwing around with the schedule in the editor.

Also, in doing this:
- refactor testing of validation errors; 
- added tests of schedule plan controller
- fix survey validator which is different in behavior from all other validators

Moving toward greater consistency in how errors are reported.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/SchedulePlanService.java
+++ b/app/org/sagebionetworks/bridge/services/SchedulePlanService.java
@@ -60,14 +60,17 @@ public class SchedulePlanService {
         plan.setVersion(null);
         plan.setGuid(BridgeUtils.generateGuid());
         
-        List<Schedule> schedules = plan.getStrategy().getAllPossibleSchedules();
-        for (Schedule schedule : schedules) {
-            for (int i=0; i < schedule.getActivities().size(); i++) {
-                Activity activity = schedule.getActivities().get(i);
-                
-                Activity activityWithGuid = new Activity.Builder().withActivity(activity)
-                        .withGuid(BridgeUtils.generateGuid()).build();
-                schedule.getActivities().set(i, activityWithGuid);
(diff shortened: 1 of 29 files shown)


app/org/sagebionetworks/bridge/services/SchedulePlanService.java
```diff
@@ -69,6 +83,23 @@ public SchedulePlan updateSchedulePlan(Study study, SchedulePlan plan) {
...
+        // Verify that all GUIDs that are supplied already exist in the plan. If they don't (or GUID is null),
+        // then add a new GUID.
+        SchedulePlan existing = schedulePlanDao.getSchedulePlan(study, plan.getGuid());
+        Set<String> existingGuids = getAllActivityGuids(existing);
+        
+        List<Schedule> schedules = plan.getStrategy().getAllPossibleSchedules();
+        for (Schedule schedule : schedules) {
+            for (int i=0; i < schedule.getActivities().size(); i++) {
+                Activity activity = schedule.getActivities().get(i);
+                if (activity.getGuid() == null || !existingGuids.contains(activity.getGuid())) {
```
liujoshua: What is the reason for creating an activity with a new guid if that activity/guid is not in the plan? should an activity only belong to one schedule?

alxdarksage: The case where in editing, a user is adding a new activity to a schedule plan. That activity should not have a guid, and we should assign one.

liujoshua (approved):

PR #1387, opened 2017-03-08
Update scheduled activity table

alxdarksage:
As a prelude to being able to save arbitrary client data with a task, a new method for retrieving tasks by the scheduled on timestamp. This is per activity in a schedule plan.

This was branched from another PR that must be merged first (that does some work to ensure that activity GUIDs are stable and don't change when clients make certain mistakes with the JSON they send to the server).

### DIFF
--- a/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
+++ b/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
@@ -88,15 +88,14 @@ import org.sagebionetworks.bridge.upload.UploadValidationHandler;
  * production-only Spring configs, see {@link BridgeProductionSpringConfig}. For test-only Spring configs, see
  * BridgeTestSpringConfig in tests.
  */
-@ComponentScan(basePackages = "org.sagebionetworks.bridge", excludeFilters = @ComponentScan.Filter(
-        type = FilterType.ANNOTATION, value = Configuration.class))
+@ComponentScan(basePackages = "org.sagebionetworks.bridge", excludeFilters = @ComponentScan.Filter(type = FilterType.ANNOTATION, value = Configuration.class))
 @Configuration
 public class BridgeSpringConfig {
     @Bean(name = "redisProviders")
     public List<String> redisProviders() {
         return Lists.newArrayList("REDISCLOUD_URL", "REDISTOGO_URL");
     }
-    
+
     @Bean(name = "bridgeObjectMapper")
     public BridgeObjectMapper bridgeObjectMapper() {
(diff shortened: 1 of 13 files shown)


app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
```diff
@@ -367,6 +367,13 @@ public DynamoIndexHelper uploadHealthCodeRequestedOnIndex(AmazonDynamoDBClient d
         return DynamoIndexHelper.create(DynamoUpload2.class, "healthCode-requestedOn-index", dynamoDBClient, dynamoNamingHelper, dynamoUtils);
     }
     
+    @Bean(name = "healthCodeActivityGuidIndex")
+    @Autowired
+    public DynamoIndexHelper healthCodeActivityGuidIndex(AmazonDynamoDBClient dynamoDBClient, DynamoUtils dynamoUtils,
+            DynamoNamingHelper dynamoNamingHelper) {
+        return DynamoIndexHelper.create(DynamoUpload2.class, "healthCodeActivityGuid-localScheduledOn-index", dynamoDBClient, dynamoNamingHelper, dynamoUtils);
```
liujoshua: Is this supposed to be ScheduleActivity.class or really DynamoUpload2.class?

app/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceList.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+public class ForwardCursorPagedResourceList<T> {
+
+    private final List<T> items;
+    private final int pageSize;
+    private final @Nullable Long offsetBy;
+    private final Map<String,String> filters = Maps.newHashMap();
+
+    @JsonCreator
+    public ForwardCursorPagedResourceList(
```
liujoshua, line 31: should we return hasNext to the client?

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
alxdarksage: You're right, I replaced that last thing and didn't change that. This would fail tests. Fixing

app/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceList.java
alxdarksage: Sure, adding

alxdarksage (commented):
Okay, changed that config method (thanks for spotting that), and added a "hasNext" property to the forward cursor paged list.

PR #1391, opened 2017-03-10
add Shared Module Library study to bootstrapper; prevent encryption keys from being recreated

DwayneJengSage:
Added Shared Module Library "shared" study to the default bootstrapper. This allows other devs to have this study in their local envs, if they need to test Shared Module Library stuff. (For dev, staging, and prod, we still need to create these studies manually, which I've already done.) See https://sagebionetworks.jira.com/browse/BRIDGE-1542

As a side effect, this also prevents encryption certs from being recreated when you recreate a study https://sagebionetworks.jira.com/browse/BRIDGE-593


alxdarksage (approved):

alxdarksage:
Merge when ready

PR #1394, opened 2017-03-14
Mods to code for AWS deployment

consolecowboy:
Per request, small changes to code to allow for the following

- Changes to BridgeProductionSpringConfig.java to allow for no password for connecting to Elasticache is lieu of passworded Redis server
- Addition of dist directory and deployment scripts to pushing to Elastic Beanstalk

### DIFF
--- a/dist/.ebextensions/01_bridgecf.config
+++ b/dist/.ebextensions/01_bridgecf.config
@@ -1,0 +1,126 @@
+files:
+  "/root/build-bridge-server-conf.sh":
+    mode: "000755"
+    owner: root
+    group: root
+    content: |
+        #!/bin/bash
+        # Script to dynamically build bridge server config from env variables
+        # Brian Holt <beholt@gmail.com> 02.09.2017
+        mkdir /home/webapp/.bridge
+        chown ${EXECUSER} /home/webapp/.bridge
+
+        cat <<- EOF > /home/webapp/.bridge/bridge-server.conf
+        bridge.env=${bridge_env:-dummy-value}
+        bridge.user=${bridge_user:-dummy-value}
+
+        admin.email = ${admin_email:-dummy-value}
+        admin.password = ${admin_password:-dummy-value}
+
+        async.worker.thread.count = ${async_worker_thread_count:-20}
+
+        aws.key = ${default_aws_key:-AWS}
+        aws.secret.key = ${default_aws_secret_key:-AWS}
+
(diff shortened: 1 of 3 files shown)


dist/.ebextensions/01_bridgecf.config
```diff
@@ -0,0 +1,126 @@
...
+        # Brian Holt <beholt@gmail.com> 02.09.2017
+        mkdir /home/webapp/.bridge
+        chown ${EXECUSER} /home/webapp/.bridge
+
+        cat <<- EOF > /home/webapp/.bridge/bridge-server.conf
+        bridge.env=${bridge_env:-dummy-value}
+        bridge.user=${bridge_user:-dummy-value}
+
+        admin.email = ${admin_email:-dummy-value}
+        admin.password = ${admin_password:-dummy-value}
```
alxdarksage, line 18: Is :-dummy-value a syntax for a default value?

dist/.ebextensions/01_bridgecf.config
```diff
@@ -0,0 +1,126 @@
...
+        aws.secret.key = ${default_aws_secret_key:-AWS}
+
+        aws.key.upload = ${aws_key_upload:-AWS}
+        aws.secret.key.upload = ${aws_secret_key_upload:-AWS}
+
+        aws.key.upload.cms = ${aws_key_upload_cms:-AWS}
+        aws.secret.key.upload.cms = ${aws_secret_key_upload_cms:-AWS}
+
+        bridge.healthcode.key = ${bridge_healthcode_key:-KST6Md7/phHLZg+1FBgbmngKi53K/e7gLptQOEDii0M=}
+        bridge.healthcode.redis.key = ${bridge_healthcode_redis_key:-zEjhUL/FVsN8vti6HO27XgrM32i1a3huEuXWD4Hq06I=}
```
alxdarksage, line 32: These are not the values we actually use, I assume. These were probably copied from the default bridge config file?

consolecowboy:
Yes, they are the default values from the file in the conf file, to prevent
any blank values from being passed through.

These values are all set via cloud formation, which then passes them onto
elastic beanstalk via environmental variables. This script is run to
produce the final running .bridge/bridge-server.conf file.  Ideally they'd
all be set to proper values, which will allow for tweaking directly from
the elastic beanstalk control web interface (or the cli, as we see fit.)

-- BEH --

PR #1395, opened 2017-03-15
Custom client data for activities

alxdarksage:


test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -266,24 +270,39 @@ public void updateActivitiesWorks() {
...
+        assertEquals(scheduledActivities.get(3).getClientData(), dbActivities.get(3).getClientData());
         
         // Correct published activities
         ScheduledActivity publishedActivity1 = publishCapture.getAllValues().get(0);
         assertEquals(scheduledActivities.get(1).getGuid(), publishedActivity1.getGuid());
         ScheduledActivity publishedActivity2 = publishCapture.getAllValues().get(1);
         assertEquals(scheduledActivities.get(2).getGuid(), publishedActivity2.getGuid());
+    }
+    
+    @Test(expected = BridgeServiceException.class)
```
DwayneJengSage: Should this be BadRequestException?

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
```diff
@@ -157,16 +160,28 @@ public void updateScheduledActivities(String healthCode, List<ScheduledActivity>
...
+                throw new BadRequestException("Client data too large ("+CLIENT_DATA_MAX_BYTES+" bytes limit)");
+            }
+            ScheduledActivity dbActivity = activityDao.getActivity(healthCode, schActivity.getGuid());
+            boolean addToSaves = false;
+            if (hasUpdatedClientData(schActivity, dbActivity)) {
+                dbActivity.setClientData(schActivity.getClientData());
+                addToSaves = true;
+            }
             if (schActivity.getStartedOn() != null || schActivity.getFinishedOn() != null) {
                 // We do not need to add the time zone here. Not returning these to the user.
```
DwayneJengSage: Is it worth simplifying the logic and removing the top-level if statement?

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
```diff
@@ -214,6 +229,29 @@ public void deleteActivitiesForUser(String healthCode) {
...
     
+    /**
+     * If the client data is being added or removed, or if it is different, then the activity is being 
+     * updated.
+     */
+    private boolean hasUpdatedClientData(ScheduledActivity schActivity, ScheduledActivity dbActivity) {
+        JsonNode schNode = (schActivity == null || schActivity.getClientData() == null) ? 
+                null : schActivity.getClientData();
+        JsonNode dbNode = (dbActivity == null || dbActivity.getClientData() == null) ? 
+                null : dbActivity.getClientData();
```
DwayneJengSage: The logic below can handle nulls just fine, so I think you can just do dbActivity == null instead of (dbActivity == null || dbActivity.getClientData() == null).

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
```diff
@@ -214,6 +229,29 @@ public void deleteActivitiesForUser(String healthCode) {
...
+        
+        return (schNode == null && dbNode != null) ||
+               (schNode != null && dbNode == null) ||
+               (schNode != null && dbNode != null && !schNode.equals(dbNode));
```
DwayneJengSage: Would using Objects.equals() simplify this logic?

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
```diff
@@ -214,6 +229,29 @@ public void deleteActivitiesForUser(String healthCode) {
             .collect(toImmutableList());
     }
     
+    /**
+     * If the client data is being added or removed, or if it is different, then the activity is being 
+     * updated.
+     */
+    private boolean hasUpdatedClientData(ScheduledActivity schActivity, ScheduledActivity dbActivity) {
```
DwayneJengSage: Is it worth adding unit test cases for

- both null
- added
- removed
- changed
- same

DwayneJengSage (changes requested):
Mostly minor changes. Overall looks good.

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
alxdarksage: Yes, now that we are tracking when to save, outer if statement isn't needed. Removed.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: Yes, changed

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
alxdarksage: Added

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
alxdarksage: Yes, removed

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
alxdarksage: Can be simplified, though I've added a comment because then it's not as obvious to me what it's testing for.

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
```diff
@@ -214,6 +226,28 @@ public void deleteActivitiesForUser(String healthCode) {
...
+     * updated.
+     */
+    protected boolean hasUpdatedClientData(ScheduledActivity schActivity, ScheduledActivity dbActivity) {
+        JsonNode schNode = (schActivity == null) ? null : schActivity.getClientData();
+        JsonNode dbNode = (dbActivity == null) ? null : dbActivity.getClientData();
+        
+        // One is null and the other is not, or they are not equal (JsonNode equals() accepts 
+        // null parameter, so it is testing both).
+        return (schNode == null && dbNode != null) ||
+               (schNode != null && !schNode.equals(dbNode));
```
DwayneJengSage: To clarify, I meant https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#equals(java.lang.Object,%20java.lang.Object), which I'm 99% sure would both be simpler and more readable.

DwayneJengSage (changes requested):
Also, the merge conflicts.

alxdarksage:
Updated to use Objects.equals() and resolved conflict

DwayneJengSage (approved):
Looks good. Merge when ready.

PR #1397, opened 2017-03-17
Return user session

alxdarksage:
- adds SessionUpdateService to consolidate updates to session (later, we'll use this to update the user's notification topic subscriptions based on changing criteria. It doesn't include all updates to the participant object, I'm just consolidating session manipulation this PR);

- refactored consent service tests to verify the contents of the returned session;

- removed setting a cookie in HTTP response (no longer used);

- marking some UserProfile endpoints as deprecated at this point. They are now also returning the session, but no one should be using them (they were never carried over into the Swagger version of our Java SDK, for example).

In general, the session is returned when signing in, when changing consent (either consenting or withdrawing), and when the user updates their own participant record, even through some of the earlier field-by-field update endpoints.


app/org/sagebionetworks/bridge/services/SessionUpdateService.java
```diff
@@ -0,0 +1,147 @@
...
+    public void updateParticipant(UserSession session, CriteriaContext context, StudyParticipant participant) {
+        session.setParticipant(participant);
+        
+        Map<SubpopulationGuid,ConsentStatus> statuses = consentService.getConsentStatuses(context);
+        session.setConsentStatuses(statuses);
+        
+        cacheProvider.setUserSession(session);
+    }
+    
+    public void updateDataGroups(UserSession session, CriteriaContext context, Set<String> dataGroups) {
```
liujoshua: I'm confused by this method signature (and updateParticipant above) and how it should be called.

CriteriaContext also contains the participant's data groups. Does the dataGroups param need to match the data groups in the context param? 

Looks to me like the steps are
- get new dataGroups
- set new dataGroups on context
- call updateDataGroups with params: old session, and updated context and dataGroup
- the old session passed in as param gets updated with data groups and consents as a side effect

Is that right?

liujoshua (approved):

PR #1398, opened 2017-03-21
Use range with existing GUID key, remove GSI and related fields.

alxdarksage:
You can use a between query on the GUID (which is a sortable string that sorts by date because it ends with an ISO timestamp). Removing GSI and related fields, switching to a query of the primary hash/range keys.


zaclinsage (approved):
LGTM

PR #1400, opened 2017-03-27
refactor UploadSchemaDao into UploadSchemaService

DwayneJengSage:
There's a lot of business logic in DynamoUploadSchemaDao that should be in UploadSchemaService, most notably logic for creating schemas from surveys and logic for managing. This change moves a lot of that logic from the DAO to the Service, and likewise for the tests. It's mostly a mechanical refactor, but also includes a few minor changes such as:
* Using Apache common lang StringUtil.isBlank() instead of Guava Strings.isNullOrEmpty()
* Fix for https://sagebionetworks.jira.com/browse/BRIDGE-1698, which prevents multiple surveys from pointing at the same schema
* Validator now checks for study ID and schema rev (because validator is now called after study ID and schema rev are set)
* DynamoUploadSchema.getKey() now returns null when studyId or schemaId are not specified instead of throwing InvalidEntityException. This is to enable querying using the mapper API instead of the now defunct DynamoIndexHelper.

Sorry in advance for the massive change. There was a lot of code and test that had to be moved.


alxdarksage:
This is great. I'd like to do this to other DAOs too, it'll make it easier to switch to a RDBMS if we want. I have also thought about organizing packages so that the DAO is only visible to the service, so there's no way you can skip execution of the service logic.

PR #1407, opened 2017-03-29
Fix amazon message

alxdarksage:
And, remove the original history API. It's not used anywhere (was only used in the researcher UI which now uses the new endpoint).

This has been removed from SDK and integration tests have been moved to new API, so this removal should not break anything.


app/org/sagebionetworks/bridge/dynamodb/DynamoScheduledActivityDao.java
```diff
@@ -117,7 +94,12 @@ final void setDdbMapper(DynamoDBMapper mapper) {
             query.withExclusiveStartKey(map);
         }
         
-        QueryResultPage<DynamoScheduledActivity> queryResult = mapper.queryPage(DynamoScheduledActivity.class, query);
+        QueryResultPage<DynamoScheduledActivity> queryResult = null;
+        try {
+            queryResult = mapper.queryPage(DynamoScheduledActivity.class, query);
+        } catch(AmazonServiceException e) {
+            throw new BadRequestException(e.getErrorMessage());
+        }
```
DwayneJengSage: AmazonServiceException includes things like internal server error and throttling exceptions. We shouldn't munge them all into a 400 Bad Request exception, as that gives the client the wrong impression of why the server call failed. Additionally, throwing Bad Request exceptions hides the exception from our ExceptionInterceptor, which means we lose visibility into when our dependent services are behaving badly.

AmazonServiceException has a getStatusCode() field. If it's in the 4XX range, we can return a BadRequestException. Otherwise, we should throw 500.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/dynamodb/DynamoScheduledActivityDao.java
alxdarksage: I agree, actually I'm wondering if we should do this in ExceptionInterceptor because there are many places where an unusual input to AWS could cause an exception... any of the 400s should probably not be returned as generic 500 errors, as ExceptionInterceptor currently does.

Right now ExceptionInterceptor looks to see if it has a BridgeServerException, and if so handles it differently. We could do the same with AmazonServiceException for the status and the error message.

alxdarksage:
Updated to process Amazon runtime exceptions in the ExceptionHandler, and there we now look at the status code returned from Amazon to determine if it's a BadRequestException or a BridgeServiceException.

DwayneJengSage (approved):

PR #1408, opened 2017-03-30
Add a route for researcher to withdraw a user from a single consent.

alxdarksage:
BRIDGE-1656. Right now researchers can withdraw a user from all consents, we've had requests to withdraw from individual consents, so, here's an API for that.

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
@@ -246,6 +246,19 @@ public class ParticipantController extends BaseController {
         return okResult("User has been withdrawn from the study.");
     }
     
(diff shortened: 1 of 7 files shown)


app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
```diff
@@ -246,6 +246,19 @@ public Result withdrawFromAllConsents(String userId) {
         return okResult("User has been withdrawn from the study.");
     }
     
+    public Result withdrawConsent(String userId, String guid) {
```
liujoshua: Could we call the param subpopulationGuid?

liujoshua (approved):
LGTM. One minor comment

PR #1409, opened 2017-03-30
Create study and user

zaclinsage:
Also fix no name in Stormpath issue;
unit tests done;
manual tests done:
  create study w/wt existed user email in Synapse;


app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -244,10 +248,14 @@ public Study createStudyAndUsers(StudyAndUsers studyAndUsers) throws SynapseExce
...
             IdentifierHolder identifierHolder = participantService.createParticipant(study, user.getRoles(), user,true);
+
             NewUser synapseUser = new NewUser();
             synapseUser.setEmail(user.getEmail());
-            synapseClient.newAccountEmailValidation(synapseUser, SYNAPSE_REGISTER_END_POINT);
-
+            try {
+                synapseClient.newAccountEmailValidation(synapseUser, SYNAPSE_REGISTER_END_POINT);
+            } catch (SynapseServerException e) {
+                LOG.error("Email: " + user.getEmail() + " already exists in Synapse", e);
```
alxdarksage: This doesn't test to see if the exception is specific to the case where the user already has an account... I thought on Slack you told me there is a specific message in that case you can use to filter this down further, rethrowing anything that's not that specific error?

alxdarksage (changes requested):

PR #1410, opened 2017-03-31
create new admin api to create user with given study id

zaclinsage:
unit tests done;

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/UserManagementController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/UserManagementController.java
@@ -1,6 +1,11 @@
 package org.sagebionetworks.bridge.play.controllers;
 
+import static org.sagebionetworks.bridge.Roles.ADMIN;
+
 import com.fasterxml.jackson.databind.JsonNode;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Controller;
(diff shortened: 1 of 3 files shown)


app/org/sagebionetworks/bridge/play/controllers/UserManagementController.java
```diff
@@ -41,6 +40,27 @@ public Result createUser() throws Exception {
...
+    public Result createUserWithStudyId(String studyId) throws Exception {
+        getAuthenticatedSession(ADMIN);
+        Study study = studyService.getStudy(studyId);
+
+        JsonNode node = requestToJSON(request());
+        StudyParticipant participant = parseJson(request(), StudyParticipant.class);
+
+        boolean consent = JsonUtils.asBoolean(node, CONSENT_FIELD);
+
+        UserSession userSession = userAdminService.createUser(study, participant, null, false, consent);
```
alxdarksage: Because you are creating a different user from the caller, you probably shouldn't return the session, unless you anticipate using this in tests.

alxdarksage (changes requested):

app/org/sagebionetworks/bridge/play/controllers/UserManagementController.java
zaclinsage: looks like integ test does not need this return session. will remove it

PR #1413, opened 2017-04-03
Adding metadata to send out sign in links.

alxdarksage:
For BRIDGE-1714 and BRIDGE-1778

Adding the email template for an email that would contain a link to sign the user into an app on the phone. Also adding a boolean flag (false by default) to turn this feature on and off. Only an admin can enable the flag, we didn't seem to want it enabled for all studies.

This is just the data added to the Study, which makes it available through the API. Tested here, and by updating the BSM to edit the template and the enablement flag.


app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -268,6 +287,7 @@ public Study createStudy(Study study) {
         study.setActive(true);
         study.setStrictUploadValidationEnabled(true);
         study.setEmailVerificationEnabled(true);
+        study.setEmailSignInEnabled(false);
```
DwayneJengSage, line 302: Is it worth adding a unit test to verify isEmailSignInEnabled cannot be set to true when created?

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -392,13 +412,17 @@ public Study updateStudy(Study study, boolean isAdminUpdate) {
             study.setEmailVerificationEnabled(originalStudy.isEmailVerificationEnabled());
             study.setExternalIdValidationEnabled(originalStudy.isExternalIdValidationEnabled());
             study.setExternalIdRequiredOnSignup(originalStudy.isExternalIdRequiredOnSignup());
+            study.setEmailSignInEnabled(originalStudy.isEmailSignInEnabled());
```
DwayneJengSage, line 427: Similarly, a unit test to verify isEmailSignInEnabled cannot be changed on update?

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -170,9 +180,18 @@ public Study getStudy(String identifier, boolean includeDeleted) {
...
+            // studies, treat it as if it doesn't exist.
+            if (!study.isActive() && !includeDeleted) {
+                throw new EntityNotFoundException(Study.class, "Study not found.");
+            }
+            // Because this template does not currently exist in studies, add the default if it is null.
+            if (study.getEmailSignInTemplate() == null) {
+                EmailTemplate template = new EmailTemplate(defaultEmailSignInTemplateSubject,
+                        defaultEmailSignInTemplate, MimeType.HTML);
+                study.setEmailSignInTemplate(template);
+            }
```
DwayneJengSage, line 198: Is there any way to test this? Right now, all the existing tests go through create() or update(), which means this code path is not being hit in any of our tests.

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -538,42 +562,31 @@ private boolean studyDirectoryHasChanged(Study originalStudy, Study study) {
...
-    private EmailTemplate fillOutTemplate(EmailTemplate template, String defaultSubject, String defaultBody) {
-        if (template == null) {
-            template = new EmailTemplate(defaultSubject, defaultBody, MimeType.HTML);
+        if (study.getVerifyEmailTemplate() == null) {
+            EmailTemplate template = new EmailTemplate(defaultEmailVerificationTemplateSubject,
+                    defaultEmailVerificationTemplate, MimeType.HTML);
+            study.setVerifyEmailTemplate(template);
         }
-        if (StringUtils.isBlank(template.getSubject())) {
-            template = new EmailTemplate(defaultSubject, template.getBody(), template.getMimeType());
```
DwayneJengSage, line 581: This change means if we have a template with a subject and no body (or a body and no subject), we replace the whole template instead of filling in the missing attributes. Is this expected?

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -586,6 +599,9 @@ private void sanitizeHTML(Study study) {
         
         template = study.getResetPasswordTemplate();
         study.setResetPasswordTemplate(sanitizeEmailTemplate(template));
+        
+        template = study.getEmailSignInTemplate();
+        study.setEmailSignInTemplate(sanitizeEmailTemplate(template));
```
DwayneJengSage, line 616: It looks like we don't have any tests for sanitizing email templates. Is this something we should be testing?

app/org/sagebionetworks/bridge/services/email/EmailSignInEmailProvider.java
```diff
@@ -0,0 +1,59 @@
...
+    
+    @Override
+    public MimeTypeEmail getMimeTypeEmail() throws MessagingException {
+        Map<String,String> map = Maps.newHashMap();
+        map.put("studyName", study.getName());
+        map.put("supportEmail", study.getSupportEmail());
+        map.put("technicalEmail", study.getTechnicalEmail());
+        map.put("sponsorName", study.getSponsorName());
+        map.put("token", token);
+        map.put("host", BridgeConfigFactory.getConfig().getHostnameWithPostfix("webservices"));
```
DwayneJengSage, line 39: Not necessary for this change, but I've noticed that every call to BridgeUtils.resolveTemplate() has a similar but different map. It might be worthwhile filing a JIRA to refactor BridgeUtils.resolveTemplate() to take in a study and fill in a basic common set of template vars, so we don't have to re-define them with every call.

app/org/sagebionetworks/bridge/services/email/EmailSignInEmailProvider.java
```diff
@@ -0,0 +1,59 @@
...
+        
+        MimeTypeEmailBuilder builder = new MimeTypeEmailBuilder();
+
+        String formattedSubject = BridgeUtils.resolveTemplate(subject, map);
+        builder.withSubject(formattedSubject);
+
+        final String sendFromEmail = String.format("%s <%s>", study.getName(), study.getSupportEmail());
+        builder.withSender(sendFromEmail);
```
DwayneJengSage, line 47: I think we previously had a bug where if the study name contained a comma, things would break in really funky ways. How did we fix that bug, and will the bug fix get picked up here?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/StudyService.java
alxdarksage: I've added a test for this in the mock test.

app/org/sagebionetworks/bridge/services/StudyService.java
alxdarksage: Added to StudyServiceTest alongside tests for the other flags through create/update with different roles. (crudStudy test).

app/org/sagebionetworks/bridge/services/StudyService.java
alxdarksage: Added all these flags to the mock study service test that compares admin and developer ability to update (developer can't, admin can).

app/org/sagebionetworks/bridge/services/StudyService.java
alxdarksage: Yes. I changed this on purpose. This is caught by validation and that seems to make more sense than silently replacing it with a default template when someone submits incomplete data.

app/org/sagebionetworks/bridge/services/StudyService.java
alxdarksage: I added a few tests of this to StudyServiceMockTest.

app/org/sagebionetworks/bridge/services/email/EmailSignInEmailProvider.java
alxdarksage: Filed: BRIDGE-1782

app/org/sagebionetworks/bridge/services/email/EmailSignInEmailProvider.java
alxdarksage: Hmm... it's not trapped in validation, and I see no fix in the email provider classes. BRIDGE-591 is about quoting the study name, which we do here. Is there another issue?

alxdarksage:
Updated based on review.

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -550,57 +574,49 @@ private boolean studyDirectoryHasChanged(Study originalStudy, Study study) {
     }
     
     /**
-     * When the password policy or templates are not included, they are set to some sensible default 
+     * When the password policy or templates are not included, they are set to some sensible defaults.  
      * values. 
      * @param study
      */
-    private void setDefaultsIfAbsent(Study study) {
+    protected final void setDefaultsIfAbsent(Study study) {
```
DwayneJengSage, line 581: Minor: Does this need to be marked protected final?

DwayneJengSage (changes requested):
Overall looks good. One minor comment.

Also, is there any way we can verify that the fix for https://sagebionetworks.jira.com/browse/BRIDGE-591 is applied here? A simple reading of the code suggests we're *not* quoting the sender name. Is that happening somewhere else? Is there any way we can test it and know for sure?

app/org/sagebionetworks/bridge/services/StudyService.java
alxdarksage: No... that is not needed for testing. Reverting.

DwayneJengSage (approved):

PR #1418, opened 2017-04-06
Request sign in via email.

alxdarksage:
Pretty much as described here, and now complete: https://sagebionetworks.jira.com/wiki/pages/viewpage.action?pageId=141923170

One thing to note is that I adapted SignIn to capture all the stuff we look at for authentication, rather than create new objects for each call with slightly different parameters. These are not entity objects, they are just JSON deserialization value objects. Updated the SignInValidator to validate different stuff based on the purpose of the signIn object. The Email object could also be replaced, but I'll leave that for another day.


app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -91,7 +98,49 @@ final void setEmailValidator(EmailValidator validator) {
...
+    
+    public UserSession emailSignIn(Study study, CriteriaContext context, String email, String token) {
+        String cacheKey = getEmailSignInCacheKey(study, email);
+        
+        String storedToken = cacheProvider.getString(cacheKey);
+        if (storedToken == null || !storedToken.equals(token)) {
+            throw new AuthenticationFailedException();
+        }
+        
+        Account account = accountDao.getAccountWithEmail(study, email);
```
alxdarksage: Verify the session is set in cache

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -211,9 +260,9 @@ public void resetPassword(PasswordReset passwordReset) throws BridgeServiceExcep
         accountDao.resetPassword(passwordReset);
     }
     
-    private UserSession getSessionFromAccount(Study study, CriteriaContext context, Account account) {
+    protected UserSession getSessionFromAccount(Study study, CriteriaContext context, Account account) {
```
alxdarksage: Did tests ultimately require this be more accessible?

app/org/sagebionetworks/bridge/services/email/EmailSignInEmailProvider.java
```diff
@@ -16,31 +20,38 @@
...
     }
     
     @Override
     public MimeTypeEmail getMimeTypeEmail() throws MessagingException {
+        String encodedRecipientEmail = null;
+        try {
+            encodedRecipientEmail = URLEncoder.encode(recipientEmail, "UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            throw new BadRequestException("Email was malformed: " + recipientEmail + ".");
+        }
```
alxdarksage: UnsupportedEncodingException = this never happens if you use UTF-8, error is wrong here as well.

app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
```diff
@@ -492,12 +494,24 @@ private Account constructAccount(StudyIdentifier studyId, com.stormpath.sdk.acco
...
+    public String getAccountIdFromEmail(Study study, String email) {
+        Directory directory = client.getResource(study.getStormpathHref(), Directory.class);
+        
+        AccountList accounts = directory.getAccounts(Accounts.where(Accounts.email().eqIgnoreCase(email)));
+        if (accounts.getSize() > 0) {
+            com.stormpath.sdk.account.Account acct = accounts.iterator().next();
+            return BridgeUtils.getIdFromStormpathHref(acct.getHref());
+        }
+        return null;
+    }
```
alxdarksage: Did they ever add a getAccount() method that's less wonky?

app/org/sagebionetworks/bridge/models/accounts/SignIn.java
```diff
@@ -26,7 +32,7 @@ public String getPassword() {
         return password;
     }
     
-    public boolean isBlank() {
```
zaclinsage, line 29: why we remove this method?

app/org/sagebionetworks/bridge/play/controllers/ApplicationController.java
```diff
@@ -18,15 +27,34 @@ public Result loadApp() throws Exception {
...
     }
 
-    public Result verifyEmail(String study) {
-        Study studyObj = studyService.getStudy(study);
+    public Result verifyEmail(String studyId) {
+        if (isBlank(studyId)) {
+            throw new BadRequestException("study parameter is required");
+        }
+        Study study = studyService.getStudy(studyId);
         return ok(views.html.verifyEmail.render(ASSETS_HOST, ASSETS_BUILD,
```
zaclinsage: what's the view.html for? Is it used just for rendering html page?

zaclinsage (commented):
LGTM

app/org/sagebionetworks/bridge/play/controllers/ApplicationController.java
```diff
@@ -18,15 +27,34 @@ public Result loadApp() throws Exception {
         return ok(views.html.index.render());
     }
 
-    public Result verifyEmail(String study) {
-        Study studyObj = studyService.getStudy(study);
+    public Result verifyEmail(String studyId) {
+        if (isBlank(studyId)) {
+            throw new BadRequestException("study parameter is required");
+        }
```
DwayneJengSage: We generally put input validation in the service, not in the controller. Would it be better to have the null check in StudyService and have it throw a BadRequest instead?

Also, is it worth adding a unit test for this case? Another reason fro putting the input validation in the Service is that it would be much easier to test.

app/org/sagebionetworks/bridge/play/controllers/ApplicationController.java
```diff
@@ -18,15 +27,34 @@ public Result loadApp() throws Exception {
...
-                StringEscapeUtils.escapeHtml4(studyObj.getName()), studyObj.getSupportEmail()));
+                StringEscapeUtils.escapeHtml4(study.getName()), study.getSupportEmail()));
     }
 
-    public Result resetPassword(String study) {
-        Study studyObj = studyService.getStudy(study);
+    public Result resetPassword(String studyId) {
+        if (isBlank(studyId)) {
+            throw new BadRequestException("studyId parameter is required");
+        }
```
DwayneJengSage: Similar concern here.

app/org/sagebionetworks/bridge/play/controllers/ApplicationController.java
```diff
@@ -18,15 +27,34 @@ public Result loadApp() throws Exception {
...
         return ok(views.html.resetPassword.render(ASSETS_HOST, ASSETS_BUILD,
-                StringEscapeUtils.escapeHtml4(studyObj.getName()), studyObj.getSupportEmail()));
+                StringEscapeUtils.escapeHtml4(study.getName()), study.getSupportEmail()));
+    }
+    
+    public Result startSession(String studyId, String email, String token) {
+        if (isBlank(studyId)) {
+            throw new BadRequestException("study ID is required");
+        }
+        SignIn signIn = new SignIn(studyId, email, null, token);
```
DwayneJengSage: This method has a bit of business logic in it and it has no unit tests. It also doesn't use views.html rendering, returns JSON instead of HTML, calls through to AuthenticationService, and is related to Auth. It looks entirely redundant with AuthenticationController.emailSignIn().

Would it be better to move this to AuthenticationController and combine it with emailSignIn?

conf/routes
```diff
@@ -3,6 +3,7 @@ GET    /                           @org.sagebionetworks.bridge.play.controllers.
 GET    /index.html                 @org.sagebionetworks.bridge.play.controllers.ApplicationController.loadApp
 GET    /mobile/verifyEmail.html    @org.sagebionetworks.bridge.play.controllers.ApplicationController.verifyEmail(study: String ?= "api")
 GET    /mobile/resetPassword.html  @org.sagebionetworks.bridge.play.controllers.ApplicationController.resetPassword(study: String ?= "api")
+GET    /mobile/startSession.html   @org.sagebionetworks.bridge.play.controllers.ApplicationController.startSession(study: String ?= null, email: String ?= null, token: String ?= null)
```
DwayneJengSage, line 6: Similarly, should we combine this with /v3/auth/email/signIn?

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
```diff
@@ -93,9 +117,8 @@ public Result resetPassword() throws Exception {
     private Result signInWithRetry(int retryCounter) throws Exception {
         UserSession session = getSessionIfItExists();
         if (session == null) {
-            JsonNode json = requestToJSON(request());
             SignIn signIn = parseJson(request(), SignIn.class);
-            Study study = getStudyOrThrowException(json);
+            Study study = studyService.getStudy(signIn.getStudyId());
```
DwayneJengSage, line 123: I believe this has the side effect of disabling the shut-off switch for sign-in. Do we have a test (unit or integration) that verifies the shut-off switch works at sign-in?

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -36,20 +47,27 @@
...
 public class AuthenticationService {
-    
+
     private final Logger logger = LoggerFactory.getLogger(AuthenticationService.class);
     
+    private static final EmailVerificationValidator EMAIL_VERIFICATION_VALIDATOR = new EmailVerificationValidator();
+    private static final SignInValidator EMAIL_SIGNIN_REQUEST_VALIDATOR = new SignInValidator(EMAIL_REQUEST);
+    private static final SignInValidator PASSWORD_SIGNIN_VALIDATOR = new SignInValidator(PASSWORD);
+    private static final SignInValidator EMAIL_SIGNIN_VALIDATOR = new SignInValidator(EMAIL);
+    private static final EmailValidator EMAIL_VALIDATOR = new EmailValidator();
```
DwayneJengSage: Minor: Instead of creating these as constants in AuthenticationService, would be better to create a singleton in the class itself? For example, EmailVerificationValidator.INSTANCE or SignInValidator.EMAIL_REQUEST_VALIDATOR.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -0,0 +1,284 @@
...
+        service.setSendMailService(sendMailService);
+        service.setStudyService(studyService);
+
+        doReturn(study).when(studyService).getStudy(STUDY_ID);
+        doReturn(STUDY_ID).when(study).getIdentifier();
+        doReturn(true).when(study).isEmailSignInEnabled();
+        doReturn(new StudyIdentifierImpl(STUDY_ID)).when(study).getStudyIdentifier();
+        doReturn(new EmailTemplate("subject","body",MimeType.TEXT)).when(study).getEmailSignInTemplate();
+        doReturn("sender@sender.com").when(study).getSupportEmail();
+        doReturn("Sender").when(study).getName();
```
DwayneJengSage: Study is a data type with no behaviors that need mocking, and we shouldn't be changing this anytime soon. Would it be better to create a concrete Study (using Study.create()) and just filling in the actual values instead of using a mock?

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -0,0 +1,284 @@
...
+        verify(cacheProvider).setString(eq(CACHE_KEY), stringCaptor.capture(), eq(60));
+        assertNotNull(stringCaptor.getValue());
+
+        verify(sendMailService).sendEmail(providerCaptor.capture());
+        
+        EmailSignInEmailProvider provider = providerCaptor.getValue();
+        MimeTypeEmail email = provider.getMimeTypeEmail();
+        assertEquals(RECIPIENT_EMAIL, email.getRecipientAddresses().get(0));
+        assertEquals("\"Sender\" <sender@sender.com>", email.getSenderAddress());
+        assertEquals("body", email.getMessageParts().get(0).getContent());
```
DwayneJengSage: This test seems to test the implementation of EmailSignInEmailProvider, not AuthenticationService.requestEmailSignIn(). It would be better to verify that the email provider contains the values we passed in (study, email, token). If necessary, we may need to expose some getters for tests.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -0,0 +1,284 @@
...
+        doReturn(TOKEN).when(cacheProvider).getString(CACHE_KEY);
+        doReturn(account).when(accountDao).getAccountWithEmail(study, RECIPIENT_EMAIL);
+        doReturn(PARTICIPANT).when(participantService).getParticipant(study, account, false);
+        doReturn(CONSENTED_STATUS_MAP).when(consentService).getConsentStatuses(any());
+        
+        UserSession retSession = service.emailSignIn(CONTEXT, SIGN_IN);
+        
+        assertNotNull(retSession);
+        verify(accountDao).getAccountWithEmail(study, RECIPIENT_EMAIL);
+        verify(cacheProvider).removeString(CACHE_KEY);
```
DwayneJengSage: We have logic for optionally setting the password. However, that logic appears to be untested.

app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
```diff
@@ -258,6 +260,20 @@ public void resetPassword(PasswordReset passwordReset) {
     }
     
     @Override
+    public void changePassword(Account account, String newPassword) {
```
DwayneJengSage, line 262: Unit tests?

app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
```diff
@@ -492,12 +508,24 @@ private Account constructAccount(StudyIdentifier studyId, com.stormpath.sdk.acco
         return account;
     }
     
-    private Account getAccountWithEmail(Study study, String email) {
+    public String getAccountIdFromEmail(Study study, String email) {
+        Directory directory = client.getResource(study.getStormpathHref(), Directory.class);
```
DwayneJengSage: This is not used? It's also untested and has a lot of redundancy with getAccountFromEmail(). One of these should call the other, or they should both call the same base method.

test/org/sagebionetworks/bridge/stormpath/StormpathAccountDaoMockTest.java
```diff
@@ -263,6 +273,19 @@ public void authenticate() {
     }
 
     @Test
+    public void getAccountWithEmail() {
+        mockAccountWithoutHealthCode();
+        doReturn(accountList).when(directory).getAccounts(any(AccountCriteria.class));
+        
+        dao.getAccountWithEmail(study, "email@email.com");
```
DwayneJengSage: Is it worth putting data in accountList and verify that it's correctly passed back? Also, do we need to verify that we're passing in study correctly and getting the right directory?

app/org/sagebionetworks/bridge/validators/SignInValidator.java
```diff
@@ -18,13 +32,21 @@ public boolean supports(Class<?> clazz) {
...
+        if (isBlank(signIn.getEmail())) {
+            errors.rejectValue("email", "is required");
+        }            
+        if (type == Type.PASSWORD) {
+            if (isBlank(signIn.getPassword())) {
+                errors.rejectValue("password", "required");
+            }
+        } else if (type == Type.EMAIL) {
+            if (isBlank(signIn.getToken())) {
+                errors.rejectValue("token", "is required");
```
DwayneJengSage: Unit tests?

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerMockTest.java
```diff
@@ -79,12 +83,18 @@
     @Mock
     CacheProvider cacheProvider;
     
+    @Mock
+    UserSession userSession;
```
DwayneJengSage: Similarly, UserSession is also just getters and setters. We should use a concrete one instead of mocking it.

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerMockTest.java
```diff
@@ -96,7 +106,48 @@ public void before() {
...
+        doReturn(participant).when(userSession).getParticipant();
+        study.setIdentifier("study-test");
+        doReturn(study).when(studyService).getStudy("study-test");
+        doReturn(true).when(userSession).isAuthenticated();
+        doReturn(userSession).when(authenticationService).emailSignIn(any(CriteriaContext.class), any(SignIn.class));
+        
+        Result result = controller.emailSignIn();
+        assertEquals(200, result.status());
+        JsonNode node = BridgeObjectMapper.get().readTree(Helpers.contentAsString(result));
+        assertTrue(node.get("authenticated").asBoolean());
```
DwayneJengSage: asBoolean() automatically converts ints (zero vs non-zero) and strings ("true" vs "false") into booleans. If you want to verify both value _and_ type, use booleanValue().

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -0,0 +1,284 @@
...
+    @Mock
+    private StudyService studyService;
+    @Mock
+    private PasswordResetValidator passwordResetValidator;
+    @Mock
+    private Study study;
+    @Mock
+    private Account account;
+    @Mock
+    private UserSession session;
```
DwayneJengSage: Similarly, study and session don't need to be mocked, and account can just use SimpleAccount.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -0,0 +1,284 @@
...
+    
+    @Captor
+    private ArgumentCaptor<String> stringCaptor;
+    @Captor
+    private ArgumentCaptor<EmailSignInEmailProvider> providerCaptor;
+    
+    @Spy
+    private AuthenticationService service;
```
DwayneJengSage: It seems like we don't actually spy this anywhere. Is this necessary?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/models/accounts/SignIn.java
alxdarksage: Not used anywhere in the code. The standard way we validate data from these payload objects is by using a Validator, which allows us to return all the errors keyed in a map a client can use to display errors to a user.

app/org/sagebionetworks/bridge/play/controllers/ApplicationController.java
alxdarksage: Yes, it's for testing. We have a link and it needs to go somewhere, and it might be useful for developers to get a session key. Once installed, an app will intercept this link and process it in the app (using our API to complete the call and capture the session).

app/org/sagebionetworks/bridge/play/controllers/ApplicationController.java
alxdarksage: Moved this check into StudyService.

app/org/sagebionetworks/bridge/play/controllers/ApplicationController.java
alxdarksage: This is receiving the link directly from the email where the values are parameters in a GET request, not a JSON post. It cannot be combined with the other method. It is here so that if an email is sent and the context is such that the app does not intercept the link, you will get back a session. It might be useful for testing, it's not useful in production because there's nothing the user could do with the session outside of the app.

conf/routes
alxdarksage: No, it's here for testing purposes. The other API is the one the app should/will use when it intercepts the link.

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
alxdarksage: I added a couple of tests for this because it doesn't seem to be verified working anywhere (at the controller level). Then fixed it so this is thrown correctly on sign in and email sign in.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage: Moved all these to be constants on the respective validator classes.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: Switched all of these, and removed session (was not actually used in tests).

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: Not it isn't used. Removing.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: Switched to using study object.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: Added getters to test the values being directly passed into the provider

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: Yes, you're right. added a test that it's not called in a sign in without a password, and that it is called when a password is provided.

app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
alxdarksage: Added test.

app/org/sagebionetworks/bridge/stormpath/StormpathAccountDao.java
alxdarksage: Yes you're right, this must have been part of an earlier attempt to implement. Removed.

test/org/sagebionetworks/bridge/stormpath/StormpathAccountDaoMockTest.java
alxdarksage: Added both to this test.

app/org/sagebionetworks/bridge/validators/SignInValidator.java
alxdarksage: Added these.

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerMockTest.java
alxdarksage: Changed to object.

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerMockTest.java
alxdarksage: I didn't now that. I always mean booleanValue(), not asBoolean()... changed this.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -0,0 +1,300 @@
...
+
+    private Study study;
+
+    private Account account;
+    
+    private AuthenticationService service;
+
+    @Before
+    public void before() {
+        study = new DynamoStudy();
```
DwayneJengSage: Minor: We should use Study.create() to avoid leaking the Dynamo implementation.

DwayneJengSage (approved):
One minor change in the tests, non-blocking. Everything else looks good.

PR #1423, opened 2017-04-13
Publish workflow schemas

zaclinsage:
unit tests done;
manual tests done;


DwayneJengSage:
Looks good. Can we also add this to BridgeDocs and JavaSDK?

We'll also want to add this to integration tests UploadSchemaTest.optionalFields(), to make sure the field is being written and read end-to-end: https://github.com/Sage-Bionetworks/BridgeIntegrationTests/blob/develop/src/test/java/org/sagebionetworks/bridge/sdk/integration/UploadSchemaTest.java#L204

Also, I forgot one requirement. When we import the schema, we need to set the published flag. See https://github.com/Sage-Bionetworks/BridgePF/blob/develop/app/org/sagebionetworks/bridge/services/SharedModuleService.java#L100

We'll also want to test for this in integ test SharedModuleTest: https://github.com/Sage-Bionetworks/BridgeIntegrationTests/blob/develop/src/test/java/org/sagebionetworks/bridge/sdk/integration/SharedModuleTest.java#L187

test/org/sagebionetworks/bridge/services/SharedModuleServiceTest.java
```diff
@@ -90,6 +90,7 @@ public void byIdAndVersionSchemaSuccess() {
 
         // mock schema service
         UploadSchema sharedSchema = UploadSchema.create();
+        sharedSchema.setPublished(true);
```
DwayneJengSage: This is not the way to test this. We need to test that we set the schema to published before we create it in the local study. What you're doing here is setting the shared schema to published. This actually makes it impossible to test that we're setting the published flag, because we can't tell whether we set the schema to published before creating it, or if it was already published to begin with.

Better to set the published flag to false, and then when verifying the call to createUploadSchema(), we verify that the passed in schema is published.

test/org/sagebionetworks/bridge/services/SharedModuleServiceTest.java
```diff
@@ -168,6 +169,7 @@ public void latestPublishedSuccess() {
...
```
DwayneJengSage: Same here.

DwayneJengSage (changes requested):

DwayneJengSage:
See also https://github.com/Sage-Bionetworks/BridgeIntegrationTests/pull/149

test/org/sagebionetworks/bridge/services/SharedModuleServiceTest.java
```diff
@@ -99,6 +102,7 @@ public void byIdAndVersionSchemaSuccess() {
         assertEquals(SharedModuleType.SCHEMA, status.getModuleType());
         assertEquals(SCHEMA_ID, status.getSchemaId());
         assertEquals(SCHEMA_REV, status.getSchemaRevision().intValue());
+        assertTrue(sharedSchema.getPublished());
```
DwayneJengSage: This is not how we want to test this either. This tests that the schema returned by the getUploadSchema call has its published value set to true. This tests the implementation, not the behavior. We want to test that the schema passed into the createSchema() call has its published value set to true. (Here and in the other place.)

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/services/SharedModuleServiceTest.java
Zealing: Yes you are correct!

zaclinsage: yes you are correct!

PR #1424, opened 2017-04-13
Non auth view module

zaclinsage:


alxdarksage (approved):

test/org/sagebionetworks/bridge/play/controllers/SharedModuleMetadataControllerTest.java
```diff
@@ -120,9 +120,6 @@ public void byIdAndVersion() throws Exception {
         Result result = controller.getMetadataByIdAndVersion(MODULE_ID, MODULE_VERSION);
         assertEquals(200, result.status());
         assertMetadataInResult(result);
-
-        // validate permissions
-        verify(controller).getAuthenticatedSession(Roles.DEVELOPER);
```
DwayneJengSage: It might be worth changing this to verify that we never ask for an authenticated session. This way, tests can verify that we don't need auth. (Here and the other 3 places.)

Also, in integ tests, is there a way to call Bridge without an account? We might want to add integ tests to verify this.

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/play/controllers/SharedModuleMetadataControllerTest.java
alxdarksage: Regarding non-authenticated calls... yes, through ApiClientProvider, not ClientManager (the class I wrote to make working with sessions easier).

zaclinsage: good point. I will add non-auth user in integ test.

DwayneJengSage (approved):

DwayneJengSage:
See also https://github.com/Sage-Bionetworks/BridgeIntegrationTests/pull/148

PR #1427, opened 2017-04-19
mock test for surveyservice

zaclinsage:


alxdarksage:
This is mixing Spring and mock... we usually try and avoid this...

DwayneJengSage:
It's unavoidable. This is a legacy unit test that should be moved to integ tests. In the meantime, we shouldn't be adding MySQL dependencies to unit tests.

PR #1429, opened 2017-04-26
Starting to port over to Okta.

alxdarksage:
DO NOT MERGE

Status: OktaAccountDao reimplements the functionality that StormpathAccountDao + Stormpath provided. I am able to get the BSM up and running, using the Okta evaluation account. However, this only currently works for one study.

- Some fields like status and roles I have mapped to UserProfile custom fields (Okta suggestions for how to implement these were less direct, more complex, and offered no advantages);
- Implemented sending emails and managing workflows and verifications on our side, ignoring the Okta assumptions which are slightly different than Stormpath;
- there is no API that I can find to adjust the custom fields on a UserProfile. This is the logical equivalent of customData in Stormpath, but while customData is dynamic (new fields can be added), that's not the case for Okta's UserProfile. This makes subpopulation consent records difficult to store, we'd have to come up with a workaround.

Finally, Okta has backtracked on allowing us to create/delete our own organizations, which means we'd have to have all accounts in one organization. To deal with the reality that a user could have the same email address with different passwords in two different apps, we'd have to namespace the email addresses.

While this is all probably doable, it's worth stopping to discuss if this is really worth it to us. It's seems pretty hacky.


DwayneJengSage:
Closing this, since we're going with a different approach.

PR #1431, opened 2017-04-26
Do not change sharing status on consent

alxdarksage:
Only on withdrawal when user is ejected from the study (when withdrawing means the user is no longer consented to the study, we change the status to NO_SHARING). Note for posterity, the user can still then turn their sharing back on through the "self participant" API and currently, that'll cause their data to be exported again. See BRIDGE-1293.

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
@@ -5,6 +5,7 @@ import static org.sagebionetworks.bridge.dao.ParticipantOption.SHARING_SCOPE;
 import java.util.Map;
 
 import org.joda.time.DateTime;
+
 import org.sagebionetworks.bridge.dao.ParticipantOption.SharingScope;
 import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
 import org.sagebionetworks.bridge.models.CriteriaContext;
@@ -18,7 +19,6 @@ import org.sagebionetworks.bridge.models.subpopulations.ConsentSignature;
 import org.sagebionetworks.bridge.models.subpopulations.Subpopulation;
 import org.sagebionetworks.bridge.models.subpopulations.SubpopulationGuid;
 import org.sagebionetworks.bridge.services.ConsentService;
-import org.sagebionetworks.bridge.services.ParticipantOptionsService;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Controller;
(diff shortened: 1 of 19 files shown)


app/org/sagebionetworks/bridge/services/SessionUpdateService.java
```diff
@@ -76,14 +83,16 @@ public void updateSharingScope(UserSession session, SharingScope sharingScope) {
...
         session.setConsentStatuses(statuses);
         
-        StudyParticipant.Builder builder = builder(session);
-        builder.withSharingScope(sharingScope);
-        if (!session.doesConsent()) {
+        StudyParticipant.Builder builder = builder(session).withSharingScope(sharingScope);
+        
+        if (isWithdrawing && !session.doesConsent()) {
             builder.withSharingScope(SharingScope.NO_SHARING);
+            optionsService.setEnum(session.getStudyIdentifier(), session.getHealthCode(),
```
liujoshua: SharingScope is being set here during withdrawal

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
```diff
@@ -176,7 +170,7 @@ private Result giveConsentForVersion(int version, SubpopulationGuid subpopGuid)
```
liujoshua, line 171: when giving consent, SharingScope gets set in line 174 above, consentService.consentToResearch

liujoshua (approved):
Looks good. I pointed out the difference in ways/places consent is set in giving consent vs withdrawing consent, which made it a bit confusing. Consent has always been complicated and not sure whether this can be avoided due to dependence on current session to tell if user is consented in another subpop before setting to NO_SHARING during withdrawal.

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
alxdarksage: Going to fix this issue, then merge. Thanks for pointing out this could be done in the consent service.

alxdarksage: Okay I've cleaned this up so that:
- controller creates the criteriacontext (or the service for the participantService where the context is not based on the request, it's based on looking up information about the user being manipulated);
- the service updates the user's state if sharing should be turned off on a withdrawal
- the session update service will reflect this change when updating the session (again, if withdrawing)

liujoshua (approved):

PR #1432, opened 2017-04-28
Add UIHint.SELECT to NumericaContraint's supported hints

liujoshua:


app/org/sagebionetworks/bridge/models/surveys/NumericalConstraints.java
```diff
@@ -10,7 +10,7 @@
     private Double step;
     
     public NumericalConstraints() {
-        setSupportedHints(EnumSet.of(UIHint.NUMBERFIELD, UIHint.SLIDER));    
+        setSupportedHints(EnumSet.of(UIHint.NUMBERFIELD, UIHint.SLIDER, UIHint.SELECT));
```
alxdarksage, line 13: I would add validation that minValue and maxValue are set for the slider and select uiHints, because without these, I don't know how the client can render them. The step value has a default value of 1 or 1.0, so it remains optional.

alxdarksage (changes requested):

PR #1437, opened 2017-05-04
set copyright notice when updating a survey

liujoshua:
Fixes issue where edits to existing survey's copyright notice in study manager weren't persisted.


alxdarksage (commented):
Merged.

PR #1440, opened 2017-05-10
Rename ScheduleContext.now to ScheduleContext.startsOn, which is more accurate.

alxdarksage:
We're about to introduce a new endpoint (/v4/activities) that will allow the client to retrieve activities for an arbitrary date range. When we do this, the "now" value in the ScheduleContext will really be the "startsOn" value of the query, and only "now" by default in the v3 of the API. Just doing a rename here to separate this change from the functional changes that are coming.

### DIFF
--- a/app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
+++ b/app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
@@ -180,7 +180,9 @@ public final class ScheduleContext {
             return this;
         }
         public Builder withAccountCreatedOn(DateTime accountCreatedOn) {
-            this.accountCreatedOn = new DateTime(accountCreatedOn, DateTimeZone.UTC);
+            if (accountCreatedOn != null) {
+                this.accountCreatedOn = new DateTime(accountCreatedOn, DateTimeZone.UTC);    
+            }
             return this;
         }
         public Builder withLanguages(LinkedHashSet<String> languages) {
--- a/app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
+++ b/app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
@@ -24,19 +24,19 @@ import com.google.common.collect.ImmutableMap;
 public final class ScheduleContext {
     
     private final DateTimeZone initialTimeZone;
+    private final DateTime startsOn;
     private final DateTime endsOn;
     private final Map<String,DateTime> events;
-    private final DateTime now;
     private final DateTime accountCreatedOn;
     private final int minimumPerSchedule;
     private final CriteriaContext criteriaContext;
     
-    private ScheduleContext(DateTimeZone initialTimeZone, DateTime endsOn, Map<String, DateTime> events, DateTime now,
+    private ScheduleContext(DateTimeZone initialTimeZone, DateTime startsOn, DateTime endsOn, Map<String, DateTime> events,
             int minimumPerSchedule, DateTime accountCreatedOn, CriteriaContext criteriaContext) {
         this.initialTimeZone = initialTimeZone;
         this.endsOn = endsOn;
         this.events = events;
-        this.now = now;
+        this.startsOn = startsOn;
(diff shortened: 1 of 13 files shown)


test/org/sagebionetworks/bridge/models/schedules/ScheduleContextTest.java
```diff
@@ -81,7 +81,7 @@ public void builderWorks() {
         assertEquals(3, context.getMinimumPerSchedule());
         assertEquals("healthCode", context.getCriteriaContext().getHealthCode());
         assertEquals(TestConstants.USER_DATA_GROUPS, context.getCriteriaContext().getUserDataGroups());
-        assertEquals(now, context.getNow());
+        assertEquals(now, context.getStartsOn());
```
DwayneJengSage: Minor: This isn't a problem with the code per se, but your change did shine a spotlight on the fact that we didn't have any unit tests on withContext(). We should add that here. It's only 2 extra lines to this test.

Also, endsOn and startsOn are both set to potentially the same value. We should change the test so they use different values to make sure we aren't getting our wires crossed.

app/org/sagebionetworks/bridge/play/controllers/ScheduledActivityController.java
```diff
@@ -132,7 +132,7 @@ public Result updateScheduledActivities() throws Exception {
                 .withUserAgent(request().getHeader(USER_AGENT))
                 .withLanguages(context.getCriteriaContext().getLanguages())
                 .withUserDataGroups(context.getCriteriaContext().getUserDataGroups())
-                .withActivitiesAccessedOn(context.getNow())
+                .withActivitiesAccessedOn(DateUtils.getCurrentDateTime().withZone(requestTimeZone))
```
DwayneJengSage, line 135: We should update the unit tests to make sure we're saving the right time zone.

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/models/schedules/ScheduleContextTest.java
alxdarksage: Updated to test the withContext() method. (This found an error when copying a context where accountCreatedOn is null).

app/org/sagebionetworks/bridge/play/controllers/ScheduledActivityController.java
alxdarksage: Added

app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
```diff
@@ -176,12 +175,14 @@ public Builder withUserDataGroups(Set<String> userDataGroups) {
...
-            this.now = now;
+        public Builder withStartsOn(DateTime startsOn) {
+            this.startsOn = startsOn;
             return this;
         }
         public Builder withAccountCreatedOn(DateTime accountCreatedOn) {
-            this.accountCreatedOn = new DateTime(accountCreatedOn, DateTimeZone.UTC);
+            if (accountCreatedOn != null) {
+                this.accountCreatedOn = new DateTime(accountCreatedOn, DateTimeZone.UTC);    
+            }
```
DwayneJengSage, line 185: This change has uncovered 2 more test gaps. We don't have tests for:

* withContext() doesn't test accountCreatedOn. (I mean, we found out indirectly that it does, but if withContext() suddenly stopped copying accountCreatedOn, we'd have no way of detecting that.)
* withAccountCreatedOn() null case.

DwayneJengSage (commented):
2 minor test comments. I don't want to bloat this change request, but at the same time, we should have tests for all our code.

app/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
alxdarksage: Okay, I created a method to test all values being set, and then set through the withContext method, and another method to test none of the values being set, and then that again being copied through the withContext method.

liujoshua (approved):

PR #1441, opened 2017-05-11
Return ScheduledActivity in same TimeZone as in the request

liujoshua:


app/org/sagebionetworks/bridge/dao/ScheduledActivityDao.java
```diff
@@ -2,6 +2,7 @@
 
 import java.util.List;
 
+import org.joda.time.Chronology;
```
alxdarksage: I don't think you need Chronology, at least elsewhere I've called dateTime.getZone() without issue.

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
```diff
@@ -125,8 +129,13 @@ final void setSurveyService(SurveyService surveyService) {
             throw new BadRequestException(EITHER_BOTH_DATES_OR_NEITHER);
         }
 
+        DateTimeZone timezone = scheduledOnStart.getChronology().getZone();
```
alxdarksage: Can you just call scheduledOnStart.getZone() ?

alxdarksage (changes requested):
Only one minor comment about use of getChronology() most places...

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
liujoshua: Turns out, yes

PR #1442, opened 2017-05-11
Date based scheduling

alxdarksage:
- Introduces new endpoint to retrieve activities within a date range of up to two weeks.
- All tasks for that time period are returned with the exception of deleted tasks (even if they are expired or finished)

### DIFF
--- a/app/org/sagebionetworks/bridge/models/schedules/ScheduledActivity.java
+++ b/app/org/sagebionetworks/bridge/models/schedules/ScheduledActivity.java
@@ -10,6 +10,7 @@ import org.sagebionetworks.bridge.dynamodb.DynamoScheduledActivity;
 import org.sagebionetworks.bridge.json.BridgeObjectMapper;
 import org.sagebionetworks.bridge.models.BridgeEntity;
 
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBIgnore;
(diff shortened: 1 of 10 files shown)


app/org/sagebionetworks/bridge/models/schedules/ScheduledActivity.java
```diff
@@ -61,8 +62,28 @@ public int compare(ScheduledActivity scheduledActivity1, ScheduledActivity sched
             return result;
         }
     };
-
-    ScheduledActivityStatus getStatus();
+    
+    @DynamoDBIgnore
+    default ScheduledActivityStatus getStatus() {
```
liujoshua, line 67: default?

app/org/sagebionetworks/bridge/models/schedules/ScheduledActivity.java
```diff
@@ -61,8 +62,28 @@ public int compare(ScheduledActivity scheduledActivity1, ScheduledActivity sched
             return result;
         }
     };
-
-    ScheduledActivityStatus getStatus();
+    
+    @DynamoDBIgnore
```
liujoshua, line 66: This should still be in DynamoScheduledActivity, shouldn't it?

liujoshua (changes requested):

PR #1446, opened 2017-05-16
Account workflow service

alxdarksage:
Moves the email verification and password reset workflows out of AccountDao to a separate service, that can be used between AccountDao implementations.

I did remove the implementation from StormpathAccountDao, which now calls this service. We could deploy and verify this works separate from the rest of the refactor (at that point we're not using Stormpath's workflow support at all). If you'd rather leave StormpathAccountDao alone, and I can undo that so it continues to use Stormpath's workflow support. Just let me know.


app/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -280,4 +282,18 @@ public static DateTime getDateTimeOrDefault(String value, DateTime defaultValue)
...
     }
+    
+    public static String encodeURIComponent(String component) {
+        String encoded = null;
+        if (component != null) {
+            try {
+                encoded = URLEncoder.encode(component, "UTF-8");
+            } catch (UnsupportedEncodingException e) {
+                // UTF-8 is always supported, so this should never happen. 
+                throw new BadRequestException(e.getMessage());
```
liujoshua: Would a BridgeServiceException make more sense? BadRequestException suggests the client did something wrong

app/org/sagebionetworks/bridge/services/email/BasicEmailProvider.java
```diff
@@ -0,0 +1,106 @@
...
+    @Override
+    public MimeTypeEmail getMimeTypeEmail() throws MessagingException {
+        tokenMap.put("studyName", study.getName());
+        tokenMap.put("studyId", study.getIdentifier());
+        tokenMap.put("supportEmail", study.getSupportEmail());
+        tokenMap.put("technicalEmail", study.getTechnicalEmail());
+        tokenMap.put("sponsorName", study.getSponsorName());
+        tokenMap.put("host", BridgeConfigFactory.getConfig().getHostnameWithPostfix("webservices"));
+        
+        final MimeTypeEmailBuilder builder = new MimeTypeEmailBuilder();
```
liujoshua: Nit-pick: Could we name this emailBuilder? To distinguish from the (Provider) Builder inner class?

app/org/sagebionetworks/bridge/services/AccountWorkflowService.java
```diff
@@ -0,0 +1,233 @@
...
+    
+    /**
+     * Using the verification token that was sent to the user, verify the email address. 
+     * If the token is invalid, it fails quietly. If the token exists but the account 
+     * does not, it throws an exception (this would be unexpected).
+     */
+    public void verifyEmail(EmailVerification verification) {
+        checkNotNull(verification);
+
+        VerificationData data = restoreVerification(verification.getSptoken());
```
liujoshua, line 138: Could we have it fail on a non-existent SpToken? My thought is that failing silently could make debugging for devs and reporting errors for customers frustrating since there is no sign at the point of failure.

Guessing the correct token should be on the same order of difficulty as guessing the correct email plus password, and I don't think getting an error leaks any info since the token itself shouldn't have any semantic meaning that could identify the user, the probability of getting this error is approximately 100% without a real token.

liujoshua (changes requested):

app/org/sagebionetworks/bridge/services/email/BasicEmailProvider.java
```diff
@@ -0,0 +1,106 @@
...
+
+        final String formattedSubject = BridgeUtils.resolveTemplate(template.getSubject(), tokenMap);
+        builder.withSubject(formattedSubject);
+
+        Set<String> senderEmails = BridgeUtils.commaListToOrderedSet(study.getSupportEmail());
```
DwayneJengSage: We have a recurring problem where sometimes the support email doesn't verify their email in SES. Now that we're using this email for email verification and password reset, do you think that might be a problem?

app/org/sagebionetworks/bridge/BridgeUtils.java
alxdarksage: Changed

app/org/sagebionetworks/bridge/services/email/BasicEmailProvider.java
alxdarksage: Changed

app/org/sagebionetworks/bridge/services/AccountWorkflowService.java
alxdarksage: This is not the current behavior but I agree with you that it's best to return a 400 here. Changed it.

app/org/sagebionetworks/bridge/services/email/BasicEmailProvider.java
alxdarksage: Filed a couple of JIRAs to improve this:
https://sagebionetworks.jira.com/browse/BRIDGE-1847
https://sagebionetworks.jira.com/browse/BRIDGE-1846

I can work on this sooner rather than later.

alxdarksage (commented):
Updated based on comments.

liujoshua (approved):
Look great

PR #1452, opened 2017-05-19
Move requirement for MultiValue question to have choices from save va…

liujoshua:
…lidation to publish validation


alxdarksage (approved):
Good to merge. Build is running right now though for a bug fix I merged.

PR #1457, opened 2017-05-22
Disable email until verified

alxdarksage:
- SendMailService no longer has a default email address that it uses to send email, this is always provided by the MimeTypeEmailProvider;
- if the email is not verified by SES, throw an error;
- cache whether or not support email is verified by SES in Redis
- default system email is still in our configuration and we can still use it to send certain emails if we want them to be from us, but we must now explicitly provide it in the MimeTypeEmailProvider we use

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
@@ -98,9 +98,8 @@ public class AuthenticationController extends BaseController {
     }
 
     public Result requestResetPassword() throws Exception {
-        JsonNode json = requestToJSON(request());
         Email email = parseJson(request(), Email.class);
-        Study study = getStudyOrThrowException(json);
+        Study study = studyService.getStudy(email.getStudyIdentifier());
         authenticationService.requestResetPassword(study, email);
         return okResult("If registered with the study, we'll email you instructions on how to change your password.");
     }
--- a/app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
@@ -22,8 +22,6 @@ import org.sagebionetworks.bridge.models.accounts.StudyParticipant;
 import org.sagebionetworks.bridge.models.accounts.UserSession;
 import org.sagebionetworks.bridge.models.accounts.UserSessionInfo;
 import org.sagebionetworks.bridge.models.studies.Study;
-import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
-import org.sagebionetworks.bridge.models.studies.StudyIdentifierImpl;
 import org.springframework.stereotype.Controller;
 
 import play.mvc.BodyParser;
@@ -92,20 +90,24 @@ public class AuthenticationController extends BaseController {
     public Result resendEmailVerification() throws Exception {
         JsonNode json = requestToJSON(request());
         Email email = parseJson(request(), Email.class);
-        StudyIdentifier studyIdentifier = getStudyIdentifierOrThrowException(json);
-        authenticationService.resendEmailVerification(studyIdentifier, email);
+        Study study = getStudyOrThrowException(json);
+        authenticationService.resendEmailVerification(study.getStudyIdentifier(), email);
         return okResult("If registered with the study, we'll email you instructions on how to verify your account.");
     }
 
     public Result requestResetPassword() throws Exception {
         Email email = parseJson(request(), Email.class);
         Study study = studyService.getStudy(email.getStudyIdentifier());
+        verifySupportedVersionOrThrowException(study);
         authenticationService.requestResetPassword(study, email);
+
         return okResult("If registered with the study, we'll email you instructions on how to change your password.");
     }
 
     public Result resetPassword() throws Exception {
         PasswordReset passwordReset = parseJson(request(), PasswordReset.class);
+        Study study = studyService.getStudy(passwordReset.getStudyIdentifier());
+        verifySupportedVersionOrThrowException(study);
         authenticationService.resetPassword(passwordReset);
         return okResult("Password has been changed.");
     }
@@ -173,11 +175,6 @@ public class AuthenticationController extends BaseController {
         Study study = studyService.getStudy(studyId);
         verifySupportedVersionOrThrowException(study);
         return study;
-    }
-
-    private StudyIdentifier getStudyIdentifierOrThrowException(JsonNode node) {
-        String studyId = getStudyStringOrThrowException(node);
-        return new StudyIdentifierImpl(studyId);
     }
 
     private String getStudyStringOrThrowException(JsonNode node) {
--- a/app/org/sagebionetworks/bridge/services/EmailVerificationService.java
+++ b/app/org/sagebionetworks/bridge/services/EmailVerificationService.java
@@ -47,7 +47,7 @@ public class EmailVerificationService {
         this.cacheProvider = cacheProvider;
     }
     
-    protected String getVerifiedAddressKey(String emailAddress) {
+    private String getVerifiedAddressKey(String emailAddress) {
         return emailAddress + KEY_POSTFIX;
     }
     
--- a/app/org/sagebionetworks/bridge/services/EmailVerificationService.java
+++ b/app/org/sagebionetworks/bridge/services/EmailVerificationService.java
@@ -9,7 +9,10 @@ import javax.annotation.Resource;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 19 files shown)


app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
```diff
@@ -98,9 +98,8 @@ public Result resendEmailVerification() throws Exception {
     }
 
     public Result requestResetPassword() throws Exception {
-        JsonNode json = requestToJSON(request());
         Email email = parseJson(request(), Email.class);
-        Study study = getStudyOrThrowException(json);
+        Study study = studyService.getStudy(email.getStudyIdentifier());
```
DwayneJengSage, line 100: This change means the shut-off switch won't kick in for request reset password. Is this intentional?

Also, there seem to be no unit tests for this at all.

test/org/sagebionetworks/bridge/services/EmailVerificationServiceTest.java
```diff
@@ -27,130 +35,187 @@
...
+    @Test
+    public void isUnverifiedAndCached() throws Exception {
+        when(cacheProvider.getString(EMAIL_ADDRESS_KEY)).thenReturn("UNVERIFIED");
+        assertFalse(service.isVerified(EMAIL_ADDRESS));
+    }
+    
+    @Test
+    public void isVerifiedUncached() {
+        doReturn(EmailVerificationStatus.VERIFIED).when(service).getEmailStatus(EMAIL_ADDRESS);
+        assertTrue(service.isVerified(EMAIL_ADDRESS));
```
DwayneJengSage: Should we also be verifying that we're caching the answer?

app/org/sagebionetworks/bridge/services/EmailVerificationService.java
```diff
@@ -25,14 +28,42 @@
...
+        this.cacheProvider = cacheProvider;
+    }
+    
+    private String getVerifiedAddressKey(String emailAddress) {
+        return emailAddress + KEY_POSTFIX;
+    }
+    
+    private EmailVerificationStatus cacheAndReturn(String emailAddress, EmailVerificationStatus status) {
+        String key = getVerifiedAddressKey(emailAddress);
+        cacheProvider.setString(key, status.name(), 60*60*24);
```
DwayneJengSage: 24 hours seems like a long time. IIRC, there's no callback when someone verifies their email in SES, so someone could verify their email, but the Study won't be able to send any email for potentially 24 hours.

Maybe 5 minutes is more reasonable? Or maybe we only cache verified emails, since that's the common use case and is very unlikely to change?

app/org/sagebionetworks/bridge/services/email/MimeTypeEmailProvider.java
```diff
@@ -4,6 +4,13 @@
 
 public interface MimeTypeEmailProvider {
 
+    /**
+     * Get the sender email address without any further formatting. So for example, if the provider formats the sender
+     * email as <code>"Study Name" &lt;email@email.com&gtl</code>, This method should return only
+     * <code>email@email.com</code>.
+     */
+    String getPlainSenderEmail();
```
DwayneJengSage: It looks like the sender email will always be study.getSupportEmail(). (Which it probably should be, since verifying emails in SES seems to be a huge bottleneck, from a herding cats point of view.) Given this, how much of a refactor would it be so MimeTypeEmailProvider is an abstract class instead of an interface, and always knows to get the sender email from study.getSupportEmail()?

Similarly, MimeTypeEmail.getSenderAddress() seems to be the same in all providers, and the logic for generating and escaping them seems to be the same. Is it worth consolidating this to the provider, so we don't need to write new sender formatting code with each provider?

test/org/sagebionetworks/bridge/services/EmailVerificationServiceTest.java
```diff
@@ -27,130 +35,187 @@
...
-
+    @Mock
+    private CacheProvider cacheProvider;
+    @Spy
     private EmailVerificationService service;
-    
+
     private ArgumentCaptor<GetIdentityVerificationAttributesRequest> getCaptor;
-    
+
```
DwayneJengSage: There seems to be a lot of whitespace and formatting changes in this file. Can you turn off auto-format? All the formatting changes is making it difficult to find the actual changes in this file.

test/org/sagebionetworks/bridge/services/SendMailViaAmazonServiceTest.java
```diff
@@ -0,0 +1,80 @@
...
+        service.setEmailVerificationService(emailVerificationService);
+    }
+    
+    @Test
+    public void unverifiedEmailThrowsException() {
+        when(emailVerificationService.isVerified(SUPPORT_EMAIL)).thenReturn(false);
+        
+        BasicEmailProvider provider = new BasicEmailProvider.Builder()
+                .withStudy(Study.create())
+                .withRecipientEmail(SUPPORT_EMAIL)
```
DwayneJengSage: This test doesn't do what you think it does. You created a blank study, then tried to send the email to SUPPORT_EMAIL. EmailVerificationService tries to verify that that a blank email address is verified. This throws, but for the wrong reason.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
alxdarksage: Yes there's no test. Added a couple for requestResetPassword, added back the app version kill switch check.

test/org/sagebionetworks/bridge/services/EmailVerificationServiceTest.java
alxdarksage: Added

app/org/sagebionetworks/bridge/services/EmailVerificationService.java
alxdarksage: Hmm, yes. This happens outside of our system. Okay, I've reduced it to 5 minutes. From testing I can tell you when you trigger it from the UI, the UI calls getEmailStatus() and this always updates the cache. So you notice no delay when setting up a study to work with it. However outside the UI you would have to wait.

app/org/sagebionetworks/bridge/services/email/MimeTypeEmailProvider.java
alxdarksage: Sounds good. Refactored to use an abstract class, added tests.

test/org/sagebionetworks/bridge/services/SendMailViaAmazonServiceTest.java
alxdarksage: Yikes. Fixed this test.

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
alxdarksage: So, there weren't tests for other methods in AuthenticationController either, so I've added them as well, and they all check the kill switch now if the User-Agent string is present.

test/org/sagebionetworks/bridge/services/EmailVerificationServiceTest.java
```diff
@@ -59,98 +71,163 @@ public void verifiedEmailTakesNoAction() {
...
+    }
+    
+    @Test
+    public void isVerifiedUncached() {
+        doReturn(EmailVerificationStatus.VERIFIED).when(service).getEmailStatus(EMAIL_ADDRESS);
+        
+        assertTrue(service.isVerified(EMAIL_ADDRESS));
+        
+        verify(cacheProvider).setString(eq(service.getVerifiedAddressKey(EMAIL_ADDRESS)),
+                eq("VERIFIED"), anyInt());
```
DwayneJengSage: Minor: Should we be using EMAIL_ADDRESS_KEY instead? (Here and in the two other tests below.)

app/org/sagebionetworks/bridge/services/email/MimeTypeEmailProvider.java
```diff
@@ -1,9 +1,39 @@
...
+import com.google.common.collect.Iterables;
+
+public abstract class MimeTypeEmailProvider {
 
-    MimeTypeEmail getMimeTypeEmail() throws MessagingException;
+    private Study study;
+    
+    MimeTypeEmailProvider(Study study) {
+        this.study = study;
+    }
```
DwayneJengSage: Minor: Should this be protected? If for whatever reason, the subclasses are in a different package.

DwayneJengSage (approved):
Two minor comments. Otherwise, looks good.

PR #1458, opened 2017-05-23
MySQL/Hibernate Auth, initial implementation

DwayneJengSage:
Excludes email workflow, which is coming next.

Testing done:
* added/updated unit tests
* ran integ tests against both MySQL and Stormpath. (Obviously, verify email doesn't work yet.)
* manually tested scenarios that aren't integ-testable: email signIn, account disabled

To test this locally, follow the instructions on https://github.com/Sage-Bionetworks/BridgePF/wiki/bootstrap#mysql-auth


app/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
```diff
@@ -0,0 +1,251 @@
...
+
+    /** Account health ID, which maps to health code. */
+    public String getHealthId() {
+        return healthId;
+    }
+
+    /** @see #getHealthId */
+    public void setHealthId(String healthId) {
+        this.healthId = healthId;
+    }
```
alxdarksage, line 148: In migration are we going to get rid of healthId and just work directly with healthCode? Is the accounts table encrypted?

app/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
```diff
@@ -0,0 +1,251 @@
...
+    @CollectionTable(name = "AccountRoles", joinColumns = @JoinColumn(name = "accountId", referencedColumnName = "id"))
+    @Column(name = "role")
+    @ElementCollection(fetch = FetchType.EAGER)
+    @Enumerated(EnumType.STRING)
+    public Set<Roles> getRoles() {
+        if (roles == null) {
+            roles = EnumSet.noneOf(Roles.class);
+        }
+        return roles;
+    }
```
alxdarksage, line 229: Not able to use @Embedded/@Embeddable for a set of Roles, or do you want them in a separate table?

app/org/sagebionetworks/bridge/hibernate/HibernateAccountConsent.java
```diff
@@ -0,0 +1,82 @@
+package org.sagebionetworks.bridge.hibernate;
+
+import javax.persistence.Embeddable;
+
+/**
+ * Consent, as represented by Hibernate. Note that this is different from the ConsentSignature class because of how
+ * Hibernate marshalls data to and from the database. In particular, it doesn't contain any of the keys (account ID,
+ * subpopulation GUID, signedOn). This is because Hibernate doesn't allow us to have table keys in an embeddable'
+ * object.
+ */
```
alxdarksage, line 10: I don't really understand why signedOn is a key that can't be in this record. consentCreatedOn is a key to the consent version, but signedOn is just the timestamp of an event, like withdrewOn.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountConsentKey.java
```diff
@@ -0,0 +1,61 @@
...
+ * This class does not include account ID as Hibernate embeds this object directly into HibernateAccount.
+ */
+// Note that Hibernate requires that this class be mutable and accesses the member variables directly. However,
+// Hibernate also expects this to be used as a map key. Mutability and hashability are generally at odds with each
+// other. To minimize potential problems, there are no setters on the member variables. Use with care.
+@Access(AccessType.FIELD)
+@Embeddable
+public final class HibernateAccountConsentKey {
+    private String subpopulationGuid;
+    private long signedOn;
```
alxdarksage, line 19: I would consider the keys here to be the foreign keys to the consent that was signed, and that's subpopulation Guid + consentCreatedOn, aren't they? I guess this will work too.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -0,0 +1,490 @@
...
+    @Autowired
+    public final void setHibernateHelper(HibernateHelper hibernateHelper) {
+        this.hibernateHelper = hibernateHelper;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void verifyEmail(EmailVerification verification) {
+        // Not yet implemented. See https://sagebionetworks.jira.com/browse/BRIDGE-1838
+        // Until then, the only way to verify an account in MySQL is to edit the database directly.
```
alxdarksage, line 67: throw new UnsupportedOperationException() in these methods?

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -0,0 +1,490 @@
...
+    public String createAccount(Study study, Account account, boolean sendVerifyEmail) {
+        // Initial creation of account. Fill in basic initial parameters.
+        String accountId = BridgeUtils.generateGuid();
+
+        HibernateAccount hibernateAccount = marshallAccount(account);
+        hibernateAccount.setId(accountId);
+        hibernateAccount.setStudyId(study.getIdentifier());
+        hibernateAccount.setCreatedOn(DateUtils.getCurrentMillisFromEpoch());
+        hibernateAccount.setModifiedOn(DateUtils.getCurrentMillisFromEpoch());
+        hibernateAccount.setPasswordModifiedOn(DateUtils.getCurrentMillisFromEpoch());
```
alxdarksage, line 187: Shouldn't this be done alongside setting password and nowhere else?

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -0,0 +1,490 @@
...
+
+        // Can't change study, email, createdOn, or passwordModifiedOn.
+        HibernateAccount persistedAccount = hibernateHelper.getById(HibernateAccount.class, accountId);
+        if (persistedAccount == null) {
+            throw new EntityNotFoundException(Account.class, "Account " + accountId + " not found");
+        }
+        accountToUpdate.setStudyId(persistedAccount.getStudyId());
+        accountToUpdate.setEmail(persistedAccount.getEmail());
+        accountToUpdate.setCreatedOn(persistedAccount.getCreatedOn());
+        accountToUpdate.setPasswordModifiedOn(persistedAccount.getPasswordModifiedOn());
```
alxdarksage, line 223: Would it be possible to submit JSON that inadvertently clears out roles, consents, etc, as these are not copied over.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -0,0 +1,490 @@
...
+
+        // Update modifiedOn.
+        accountToUpdate.setModifiedOn(DateUtils.getCurrentMillisFromEpoch());
+
+        // Update
+        hibernateHelper.update(accountToUpdate);
```
alxdarksage, line 229: Any optimistic locking going on here?

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -0,0 +1,490 @@
...
+                    hibernateConsentValue.setBirthdate(oneConsent.getBirthdate());
+                }
+
+                // Store in hibernate account.
+                hibernateConsentMap.put(hibernateConsentKey, hibernateConsentValue);
+            }
+        }
+
+        return hibernateAccount;
+    }
```
alxdarksage, line 414: What is the danger is simply using HibernateAccount as we do StormpathAccount (expose interface, supply as implementation of Account.create()), why do we need GenericAccount as well? Is this temporary? This seems to be leading to a lot of copying we'll need to update with changes.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -0,0 +1,490 @@
...
+
+        // Sort consents by signedOn, from oldest to newest.
+        for (List<ConsentSignature> consentSignatureListForSubpop : account.getAllConsentSignatureHistories()
+                .values()) {
+            Collections.sort(consentSignatureListForSubpop, (c1, c2) -> Long.compare(c1.getSignedOn(),
+                    c2.getSignedOn()));
+        }
+
+        return account;
+    }
```
alxdarksage, line 473: Same as above

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -0,0 +1,490 @@
...
+        } else {
+            return null;
+        }
+    }
+
+    // Helper method which marshalls a GenericAccount into a HibernateAccount.
+    // Package-scoped to facilitate unit tests.
+    static HibernateAccount marshallAccount(Account account) {
+        // Currently does not work with StormpathAccount. This is because StormpathAccount doesn't support certain
+        // behaviors we need to make this work.
```
alxdarksage, line 351: I'm guessing this comment explains the need for these methods... see below. Hopefully this'll be temporary.

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
```diff
@@ -99,6 +106,34 @@
         type = FilterType.ANNOTATION, value = Configuration.class))
 @Configuration
 public class BridgeSpringConfig {
+    @Bean
+    @Autowired
+    public AccountDao accountDao(HealthCodeService healthCodeService, HibernateHelper hibernateHelper,
+            StudyService studyService, SubpopulationService subpopService) {
+        // To avoid ambiguity and refactoring in nearly a dozen, we only want one bean with type AccountDao. This means
```
alxdarksage: trivial: "refactoring in nearly a dozen" -- dozen what? classes?

alxdarksage (commented):
Sorry I had a couple of comments outside of this review. Looks fine, I'm surprised you need a separate table for roles, unless you want it that way.

app/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
DwayneJengSage: Not immediately, no. This code change is pretty big as is. Ripping HealthID out of the system will make it even bigger. I’d leave it for a future improvement, and not necessarily one that’s tied to the Stormpath migration.

MySQL is encrypted, and I’ve verified that our RDS instance uses encryption. I believe we discussed this already and agreed that additional encryption is unnecessary.

app/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
DwayneJengSage: Embeddable still makes a separate table. Ideally, we'd use the Set MySQL type, but Hibernate doesn't seem to support this.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountConsent.java
DwayneJengSage: The user can withdraw and sign the consent again, so using consentCreatedOn as the primary key could result in conflict. The signedOn timestamp uniquely identifies each consent each time the user signs it.

As for why it can't be in the consent object, this is a limitation of Hibernate.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
DwayneJengSage: Added.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
DwayneJengSage: There is currently no code path that allows us to create an account with no password. (Except possibly by creating an account in the DB directly.) So it seemed appropriate to tag the passwordModifiedOn at initial account creation.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
DwayneJengSage: The only place that does this is ParticipantService, and we already have logic to selectively marshall over the editable attributes.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
DwayneJengSage: Hibernate does not support optimistic locking.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
DwayneJengSage: There are some very slight differences between HibernateAccount and GenericAccount, most notably around collections. For example, Stormpath doesn’t let us have the attribute map directly, while Hibernate does everything with that map directly. Also,the Account interface has a Map of Strings to a List of ConsentSignatures while Hibernate forces us to use… something else.

After we get rid of Stormpath, we could take steps to merge these together. But that largely depends on how tightly coupled you want to be to the Hibernate implementation. The consents stuff is super wonky, so I'm not sure if we want to use that in our interface.

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
DwayneJengSage: A dozen places. Comment fixed.

alxdarksage (approved):

PR #1459, opened 2017-05-23
Add BloodPressure, Height, Weight constraints

liujoshua:


app/org/sagebionetworks/bridge/models/surveys/BloodPressureConstraints.java
```diff
@@ -0,0 +1,12 @@
+package org.sagebionetworks.bridge.models.surveys;
+
+import static org.sagebionetworks.bridge.models.surveys.UIHint.BLOODPRESSURE;
+
+import java.util.EnumSet;
+
+public class BloodPressureConstraints extends Constraints {
+    public BloodPressureConstraints() {
+        setDataType(DataType.DECIMAL);
+        setSupportedHints(EnumSet.of(BLOODPRESSURE));
```
alxdarksage: Blood pressure is a number like 120/80 (possibly with an optional pulse value, it's not a decimal type. It might be its own data type, or maybe a string in a set format like X/Y Z

alxdarksage:
If surveys start to produce blood pressure as a type, it'll need to be handled by the uploading of that type. I don't know if we would prefer it to be sent as a string ("120/80 45") or as three separate values in JSON, or three separate columns. But the uploader will have to handle that.

Seems like there is also a validation component here to ensure they set a unit of measurement and that it is appropriate given isInfant and the fact that they are talking weight or height.

alxdarksage (changes requested):

app/org/sagebionetworks/bridge/models/surveys/BloodPressureConstraints.java
liujoshua: Thanks, totally meant this to be a String, like the JIRA spec'ed

liujoshua:
Curious why there would be three values? Is the 3rd one the pulse?

app/org/sagebionetworks/bridge/models/surveys/UIHint.java
```diff
@@ -1,5 +1,7 @@
 package org.sagebionetworks.bridge.models.surveys;
 
+import org.xhtmlrenderer.css.parser.property.PrimitivePropertyBuilders;
```
alxdarksage: This doesn't look right... sure we need this?

app/org/sagebionetworks/bridge/models/surveys/WeightConstraints.java
```diff
@@ -0,0 +1,21 @@
...
+    private boolean isForInfant;
+
+    public WeightConstraints() {
+        super();
+    }
+
+    public void setIsForInfant(boolean isForInfant) {
+        this.isForInfant = isForInfant;
+    }
+
```
alxdarksage, line 18: Should also have a unit (ounces, pound, kilos), shouldn't it?

app/org/sagebionetworks/bridge/services/UploadSchemaService.java
```diff
@@ -259,11 +260,12 @@ private static void addFieldDefsForSurveyQuestion(List<UploadFieldDefinition> fi
...
         // optional (not required).
         String fieldName = question.getIdentifier();
-        UploadFieldDefinition.Builder fieldDefBuilder = new UploadFieldDefinition.Builder().withName(fieldName)
-                .withRequired(false);
 
         UploadFieldType uploadFieldType;
         if (constraints instanceof MultiValueConstraints) {
+            UploadFieldDefinition.Builder fieldDefBuilder = new UploadFieldDefinition.Builder().withName(fieldName)
+                    .withRequired(false);
+
```
alxdarksage: I would move this back to where it was so you don't have to instantiate in two different places. I'm also surprised this compiles because of Java's block scoping...

app/org/sagebionetworks/bridge/models/surveys/BloodPressureConstraints.java
```diff
@@ -0,0 +1,12 @@
+package org.sagebionetworks.bridge.models.surveys;
+
+import static org.sagebionetworks.bridge.models.surveys.UIHint.BLOODPRESSURE;
+
+import java.util.EnumSet;
+
+public class BloodPressureConstraints extends Constraints {
+    public BloodPressureConstraints() {
+        setDataType(DataType.STRING);
```
alxdarksage: I would argue for DataType.BLOODPRESSURE along with documentation of the data being uploaded (systolic, diastolic, unit).

alxdarksage (changes requested):

app/org/sagebionetworks/bridge/models/surveys/UIHint.java
liujoshua: Removed. Not sure how that got there

app/org/sagebionetworks/bridge/models/surveys/WeightConstraints.java
liujoshua: It does, and currently is being inherited from DecimalConstraint. I may change this

app/org/sagebionetworks/bridge/services/UploadSchemaService.java
liujoshua: Moved

app/org/sagebionetworks/bridge/models/surveys/BloodPressureConstraints.java
liujoshua: Like we talked about offline, looks like DataType is used for mapping a Constraint to a Synapse Column Type, and I didn't see a benefit in adding one for BP

liujoshua: It turns out DataType is coupled to how Constraints are deserialized, so I added values for BLOODPRESSURE, WEIGHT, and HEIGHT

PR #1460, opened 2017-05-24
Set the time when creating a persistent task so that it can be finished multiple times in one day.

alxdarksage:
This behavior is complicated so I wrote an integration test that verifies this works, accounting for the way the scheduling service works alongside the scheduler.

### DIFF
--- a/app/org/sagebionetworks/bridge/models/schedules/PersistentActivityScheduler.java
+++ b/app/org/sagebionetworks/bridge/models/schedules/PersistentActivityScheduler.java
@@ -3,7 +3,6 @@ package org.sagebionetworks.bridge.models.schedules;
 import java.util.List;
 
 import org.joda.time.DateTime;
-import org.joda.time.LocalTime;
 
 import com.google.common.collect.Lists;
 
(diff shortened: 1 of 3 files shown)


app/org/sagebionetworks/bridge/models/schedules/PersistentActivityScheduler.java
```diff
@@ -15,24 +14,24 @@
...
-            }
             String finishedId = "activity:"+activity.getGuid()+":finished";
             DateTime scheduledTime = getFirstEventDateTime(context, finishedId+"," + schedule.getEventId());
-            
+
             if (scheduledTime != null) {
+                DateTime localDateTime = scheduledTime.withZone(context.getInitialTimeZone());
                 addScheduledActivityAtTimeForOneActivity(scheduledActivities, plan, context,
-                        scheduledTime.toLocalDate(), LocalTime.MIDNIGHT, activity);
+                        localDateTime.toLocalDate(), localDateTime.toLocalTime(), activity);
```
syoung-smallwisdom, line 34: oops

syoung-smallwisdom (approved):
looks good

PR #1462, opened 2017-05-25
fix Spring wonkiness in Auth stuff

DwayneJengSage:
[Yesterday it worked.](https://github.com/Sage-Bionetworks/BridgePF/pull/1461)

Then, today, it stopped working.

Spring is weird like that.

This change refactors the Spring stuff around Auth to restore Autowiring, uses the Primary annotation to resolve ambiguity, and the Lazy annotation to resolve circular dependencies. This should allow us to autowire up everything in a simpler more reliable way.

While testing this, I discovered a test break. One of our recent changes accidentally changed it so requestResetPassword without a study now throws an NPE instead of a 404. I fixed AuthenticationController to use the getStudyOrThrowException() (same as signUp) and added unit tests.

Testing done:
* unit tests
* integ tests for both MySQL implementation and Stormpth implementation
* manually tested verify email, reset password, and email sign-in workflows for both MySQL implementation and Stormpath implementation

### DIFF
--- a/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
+++ b/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
@@ -31,7 +31,6 @@ import com.stormpath.sdk.client.Client;
 import com.stormpath.sdk.client.ClientBuilder;
 import com.stormpath.sdk.client.Clients;
 
-import org.sagebionetworks.bridge.cache.CacheProvider;
 import org.sagebionetworks.bridge.dao.AccountDao;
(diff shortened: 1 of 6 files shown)


alxdarksage (approved):
Approved. Personally I like the dependency settlers to be non-public so I don't see them when looking at the interface of a service. They clutter things up.

DwayneJengSage:
99% of the time, public vs non-public doesn't matter. The remaining 1% of the time, I'm not yet sure which one is more likely to come back and bite us.

PR #1463, opened 2017-05-27
Throttle retrieval of External IDs

liujoshua:


conf/bridge-server.conf
```diff
@@ -52,6 +52,9 @@ dev.external.id.add.limit = 10
 uat.external.id.add.limit = 100
 prod.external.id.add.limit = 100
 
+external.id.get.limit = 1
+prod.external.id.get.limit = 10
+
 external.id.lock.duration = 30000
```
alxdarksage: This seems way too high. The request for a single page takes under a second, so why wouldn't this be something like 1000?

liujoshua: I've updated the naming and added comments to clarify that this is related to dynamo read capacity, and rate limiting queries in the DAO

(Not sure I interpreted your comment correctly)

alxdarksage (approved):

PR #1464, opened 2017-05-30
Adding worker APIs to retrieve users

alxdarksage:
And opening up emailStatus to researchers so we can do a check on sign in.

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
@@ -5,6 +5,7 @@ import static org.sagebionetworks.bridge.BridgeConstants.NO_CALLER_ROLES;
 import static org.sagebionetworks.bridge.BridgeUtils.getDateTimeOrDefault;
 import static org.sagebionetworks.bridge.BridgeUtils.getIntOrDefault;
 import static org.sagebionetworks.bridge.Roles.RESEARCHER;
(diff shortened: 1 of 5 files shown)


app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
```diff
@@ -95,8 +96,24 @@ public Result updateSelfParticipant() throws Exception {
...
-        
         Study study = studyService.getStudy(session.getStudyIdentifier());
+        
+        return getParticipantsInternal(study, offsetByString, pageSizeString, emailFilter, startDateString,
+                endDateString);
+    }
+
+    public Result getParticipantsForWorker(String studyId, String offsetByString, String pageSizeString, String emailFilter,
+            String startDateString, String endDateString) {
+        getAuthenticatedSession(WORKER);
```
liujoshua, line 107: Can we combine this with the above method by using getAuthenticatedSession(RESEARCHER, WORKER)?

liujoshua (approved):
Looks good, one comment

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
alxdarksage: The key difference of this method is that we get the study from a parameter, and not from the current user's session. And we don't want researchers to be able to retrieve content in another person's study.

PR #1465, opened 2017-05-30
fix MySQL SSL warnings

DwayneJengSage:
Fix for https://sagebionetworks.jira.com/browse/BRIDGE-1783

This adds some configuration to turn on SSL in Hibernate.

I also added the certificate truststore so the server can verify the SSL cert (which protects against man-in-the-middle attacks). This was done by downloading the RDS certs, splitting them, and creating a keystore with those certs. Then, I merged it with Java's cacerts. This *should* be safe to upload to Github, because it's all public certs.

Tested locally against an RDS instance configured for SSL only. Was not able to test on Heroku because of https://sagebionetworks.jira.com/browse/BRIDGE-1585


alxdarksage (approved):
Wow, SSL is a pain.

PR #1470, opened 2017-06-01
Fixing total in resource list, removing warnings.

alxdarksage:
This adds total to the JSON (which will fix the integration tests without modification to the tests), and cleans up some warnings introduced by your PR.

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDao.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDao.java
@@ -66,6 +66,8 @@ public class DynamoExternalIdDao implements ExternalIdDao {
     private static final String STUDY_ID = "studyId";
     private static final String ASSIGNMENT_FILTER = "assignmentFilter";
     private static final String ID_FILTER = "idFilter";
+    private static final String TOTAL_FILTER = "total";
+
 
     private int addLimit;
     private int lockDuration;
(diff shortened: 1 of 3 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDao.java
```diff
@@ -144,7 +146,8 @@ public final void setMapper(DynamoDBMapper mapper) {
 
         ForwardCursorPagedResourceList<ExternalIdentifierInfo> resourceList = new ForwardCursorPagedResourceList<>(
                 identifiers, offsetKey, pageSize)
-                .withFilter(ID_FILTER, idFilter);
+                .withFilter(ID_FILTER, idFilter)
+                .withFilter(TOTAL_FILTER,Integer.toString(identifiers.size()));
```
DwayneJengSage, line 150: Is there a unit test for this?

Also, I'm not sure this fixes the integ tests. It solves the NullPointerException, but the total will be the page size, while the integ test expects the count of all external IDs. In short, even with this fix, we still need to update the integ tests.

DwayneJengSage (changes requested):

alxdarksage:
Closing this. The ForwardCursorResourceList is indeed different and no longer has a total, and we want to move all the paging like this (next page key, no idea of the total number of items) to the ForwardCursor list JSON.

PR #1477, opened 2017-06-06
Tweaking some numbers in new external ID queries for better performance (as measured through use of the BSM).

alxdarksage:
The performance now in the UI is similar to before. It uses some of the burst capacity of the DDB table, it is probably possible to get throughput exceptions if you hit this API very hard, and of course, this is a capacity regulator per machine, so requests would be hitting separate machines that are nevertheless consuming one pool of DDB capacity.


liujoshua (approved):
Ship it! Let's see what happens

PR #1478, opened 2017-06-06
add BridgePF-prod env to aws beanstalk deployments

zaro0508:


alxdarksage (approved):
Sorry, I didn't know this was that small!

PR #1480, opened 2017-06-07
Rename String offset from offsetBy to offsetKey

liujoshua:


app/org/sagebionetworks/bridge/models/schedules/ScheduledActivityList.java
```diff
@@ -0,0 +1,20 @@
...
+public class ScheduledActivityList extends ForwardCursorPagedResourceList<ScheduledActivity>{
+    public ScheduledActivityList(
+            List<ScheduledActivity> items, String offsetBy, int pageSize) {
+        super(items, offsetBy, pageSize);
+    }
+
+    public String getOffsetBy() {
+        return getOffsetKey();
+    }
+}
```
alxdarksage, line 20: Nice solution for this

alxdarksage (approved):
Looks good, do we have an inventory of the places in the documentation where we're going to have to change this? To handle generics, these ForwardCursorResourceLists become typed lists, like ExternalIdentifierList. - UPDATE: Looks like only ExternalId and ScheduledActivity are returned in this kind of list... though I am going to add uploads next.

PR #1484, opened 2017-06-09
Add API for custom activity events

liujoshua:


app/org/sagebionetworks/bridge/play/controllers/ActivityEventController.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Created by jyliu on 6/5/2017.
+ */
+@Controller
+public class ActivityEventController extends BaseController {
+    @Autowired
+    private ActivityEventService activityEventService;
+
+    public Result createActivityEvent(String eventKey, String timestamp) {
+        UserSession session = getAuthenticatedSession();
```
alxdarksage: I think you should be consented...

app/org/sagebionetworks/bridge/play/controllers/ActivityEventController.java
```diff
@@ -0,0 +1,30 @@
...
+
+        DateTime eventTime = DateUtils.getDateTimeOrDefault(timestamp, null);
```
alxdarksage: Is it nice to allow the server to set the timestamp? If so, we'd set the default value.

conf/routes
```diff
@@ -165,6 +165,8 @@ POST   /v3/scheduleplans           @org.sagebionetworks.bridge.play.controllers.
 GET    /v3/scheduleplans/:guid     @org.sagebionetworks.bridge.play.controllers.SchedulePlanController.getSchedulePlan(guid: String)
 POST   /v3/scheduleplans/:guid     @org.sagebionetworks.bridge.play.controllers.SchedulePlanController.updateSchedulePlan(guid: String)
 DELETE /v3/scheduleplans/:guid     @org.sagebionetworks.bridge.play.controllers.SchedulePlanController.deleteSchedulePlan(guid: String)
+POST   /v3/scheduleplans/events    @org.sagebionetworks.bridge.play.controllers.ActivityEventController.createActivityEvent(eventKey: String?= null, timestamp: String?=null)
```
alxdarksage: Doesn't this need to go before POST /v3/scheduleplans/:guid ? 

I think /v3/scheduleplans/events will call that method, with a guid of "events"

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -519,6 +520,11 @@ private void checkViolationConstraints(Study study) {
                         throwConstraintViolation(plan, study);
                     }
                 }
+                for (String eventId : Schedule.EVENT_ID_SPLITTER.split(schedule.getEventId())) {
+                    if (!eventKeys.contains(eventId)){
+                        throwConstraintViolation(plan, study);
+                    }
+                }
```
alxdarksage: Also add validation that these event strings don't have colons in the StudyValidator

alxdarksage (changes requested):
Minor stuff.

app/org/sagebionetworks/bridge/play/controllers/ActivityEventController.java
liujoshua: Agreed, changed

app/org/sagebionetworks/bridge/models/activities/ActivityEventImpl.java
```diff
@@ -0,0 +1,40 @@
...
+ * Created by liujoshua on 6/19/2017.
+ */
+public class ActivityEventImpl implements ActivityEvent {
+    private final String healthCode;
+    private final String eventId;
+    private final String answerValue;
+    private final Long timestamp;
+
+    @JsonCreator
+    public ActivityEventImpl(@JsonProperty("healthCode") String healthCode, @JsonProperty("eventId") String eventId,
```
alxdarksage: health code should never be retrieved from JSON. So I would never expect to see JSON-related annotations for health code.

app/org/sagebionetworks/bridge/models/activities/CustomActivityEventRequest.java
```diff
@@ -0,0 +1,28 @@
...
+        this.timestamp = timestamp;
+    }
+
+
+    public String getEventKey() {
+        return eventKey;
+    }
+
+    public Long getTimestamp() {
+        return timestamp;
```
alxdarksage: This should be a DateTime value, that serializes to an ISO 8601 time, like all other instances in API.

app/org/sagebionetworks/bridge/play/controllers/ActivityEventController.java
```diff
@@ -0,0 +1,88 @@
...
+                    new ActivityEventImpl(
+                            null,
+                            entry.getKey(),
+                            null,
+                            timestamp == null ? null : timestamp.getMillis()
+                    ));
+        }
+        return new PagedResourceList<>(activityEventList, null, activityEventList.size(),
+                activityEventList.size());
+    }
```
alxdarksage: Will this ever be paged? Each event only has one timestamp, there would in the most complex case only be a dozen or so events (I would guess), and the records are very small. Lists that are returned from controllers using the ok() method are converted to ResourceList<T> JSON constructions without further effort on our part.

conf/routes
```diff
@@ -160,11 +160,16 @@ GET    /v3/uploadschemas/:schemaId/recent          @org.sagebionetworks.bridge.p
...
-DELETE /v3/scheduleplans/:guid     @org.sagebionetworks.bridge.play.controllers.SchedulePlanController.deleteSchedulePlan(guid: String)
+GET    /v3/scheduleplans                  @org.sagebionetworks.bridge.play.controllers.SchedulePlanController.getSchedulePlans
+POST   /v3/scheduleplans                  @org.sagebionetworks.bridge.play.controllers.SchedulePlanController.createSchedulePlan
+GET    /v3/scheduleplans/:guid            @org.sagebionetworks.bridge.play.controllers.SchedulePlanController.getSchedulePlan(guid: String)
+POST   /v3/scheduleplans/:guid            @org.sagebionetworks.bridge.play.controllers.SchedulePlanController.updateSchedulePlan(guid: String)
+DELETE /v3/scheduleplans/:guid            @org.sagebionetworks.bridge.play.controllers.SchedulePlanController.deleteSchedulePlan(guid: String)
+
+# Activity Events
+GET    /v3/activityevents               @org.sagebionetworks.bridge.play.controllers.ActivityEventController.getSelfActivityEvents
+POST   /v3/activityevents/custom        @org.sagebionetworks.bridge.play.controllers.ActivityEventController.createCustomActivityEvent
```
alxdarksage: This should be POST /v3/activityevents/:userId. You are adding an item to that collection. There isn't a need to specify "custom" because there's no alternative, and it's not a verb.

app/org/sagebionetworks/bridge/validators/StudyValidator.java
```diff
@@ -59,6 +59,10 @@ public void validate(Object obj, Errors errors) {
                 errors.rejectValue("identifier", "must be at least 2 characters");
             }
         }
+        if (study.getActivityEventKeys().stream()
+                .anyMatch(k -> !k.matches(BridgeConstants.BRIDGE_IDENTIFIER_PATTERN))) {
+            errors.rejectValue("activityEventKey", "must contain only lower-case letters and/or numbers with optional dashes");
+        }
```
alxdarksage: We changed this method to just prevent deletion of taskIdentifiers and dataGroups, you could do the same with eventKeys if you want.

alxdarksage (changes requested):

conf/routes
alxdarksage: Erin pointed out that we're adding new endpoints at v3, but there's no reason for that, because each unique URL is versioned separately (all verbs are versioned together). So really, this could be 

GET /v1/activityevents
GET /v1/activityevents/:userId
POST /v1/activityevents/:userId

app/org/sagebionetworks/bridge/models/activities/CustomActivityEventRequest.java
liujoshua: I'm guessing this means the GET operations should also have this in ISO 8601?

app/org/sagebionetworks/bridge/play/controllers/ActivityEventController.java
liujoshua: Good to know

app/org/sagebionetworks/bridge/models/activities/ActivityEventImpl.java
```diff
@@ -0,0 +1,51 @@
+package org.sagebionetworks.bridge.models.activities;
```
alxdarksage: Why is the implementation not just DynamoActivityEvent? Do we need another implementation?

app/org/sagebionetworks/bridge/models/schedules/Schedule.java
```diff
@@ -38,6 +40,8 @@
     public static final String TYPE_PROPERTY_NAME = "type";
     public static final String ACTIVITIES_PROPERTY = "activities";
     public static final String TIMES_PROPERTY = "times";
+
+    public static final Splitter EVENT_ID_SPLITTER = Splitter.on(Pattern.compile("\\s*,\\s*"));
```
alxdarksage, line 44: As a generic comma splitter, we could put in BridgeUtils.

app/org/sagebionetworks/bridge/play/controllers/ActivityEventController.java
```diff
@@ -0,0 +1,57 @@
...
+        this.participantService = participantService;
+    }
+
+    public Result createCustomActivityEvent() {
+        UserSession session = getAuthenticatedAndConsentedSession();
+        CustomActivityEventRequest activityEvent = parseJson(request(), CustomActivityEventRequest.class);
+
+        Study study = studyService.getStudy(session.getStudyIdentifier());
+
+        StudyParticipant participant = participantService.getParticipant(study, session.getId(), false);
```
alxdarksage: Don't need to load this again... session has session.getHealthCode()

app/org/sagebionetworks/bridge/play/controllers/ActivityEventController.java
```diff
@@ -0,0 +1,57 @@
...
+
+        activityEventService.publishCustomEvent(session.getStudyIdentifier(), participant.getHealthCode(),
+                activityEvent.getEventKey(), activityEvent.getTimestamp());
+
+        return okResult("Event recorded");
```
alxdarksage: return createdResult("Event recorded.") I think this is a 201, not a 200.

app/org/sagebionetworks/bridge/services/ActivityEventService.java
```diff
@@ -94,6 +121,23 @@ public void publishActivityEvent(ActivityEvent event) {
...
+            DateTime timestamp = entry.getValue();
+            activityEventList.add(
+                    new ActivityEventImpl(
+                            null,
+                            entry.getKey(),
+                            null,
+                            timestamp == null ? null : timestamp.getMillis()
+                    ));
+        }
+        return activityEventList;
```
alxdarksage: We could just return a serialization of the DynamoActivityEvent, which would be less code. See e.g. Study.STUDY_LIST_WRITER which only returns two fields of the study.

app/org/sagebionetworks/bridge/validators/StudyValidator.java
```diff
@@ -59,6 +59,10 @@ public void validate(Object obj, Errors errors) {
                 errors.rejectValue("identifier", "must be at least 2 characters");
             }
         }
+        if (study.getActivityEventKeys().stream()
+                .anyMatch(k -> !k.matches(BridgeConstants.BRIDGE_IDENTIFIER_PATTERN))) {
+            errors.rejectValue("activityEventKey", "must contain only lower-case letters and/or numbers with optional dashes");
```
alxdarksage: Could be activityEventKeys, as this is the field name of the study.

alxdarksage (changes requested):
Some minor changes. Biggest is that we only need one implementation of ActivityEvent.

alxdarksage (approved):

PR #1488, opened 2017-06-09
external developers and researchers should not be able to bypass email verification

DwayneJengSage:
See also https://github.com/Sage-Bionetworks/BridgeIntegrationTests/pull/171

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
@@ -124,15 +124,13 @@ public class ParticipantController extends BaseController {
         return okResult(page);
     }
     
-    public Result createParticipant(String verifyEmailString) throws Exception {
+    public Result createParticipant() throws Exception {
         UserSession session = getAuthenticatedSession(RESEARCHER);
         Study study = studyService.getStudy(session.getStudyIdentifier());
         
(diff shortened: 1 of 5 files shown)


app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
```diff
@@ -124,15 +124,13 @@ private Result getParticipantsInternal(Study study, String offsetByString, Strin
         return okResult(page);
     }
     
-    public Result createParticipant(String verifyEmailString) throws Exception {
+    public Result createParticipant() throws Exception {
```
alxdarksage, line 127: This will prevent admins from creating users in the UI... I don't currently switch to the admin API to create users because I think of that as a test UI, not something for production. I would set verifyEmail to true here unless the user has the admin role. This will allow me to create users through our tool.

alxdarksage (changes requested):
This will make it so I cannot onboard users as an admin without sending a couple of emails.

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
DwayneJengSage: Discussed offline. This vulnerability is an existential risk to our platform and needs to be closed. We can improve the developer onboarding experience at a later date. Will file a separate JIRA.

PR #1490, opened 2017-06-12
This is an index query so last evaluated key does not work

alxdarksage:
Instead, paging by treating the offsetKey as an index to the range key, and adjusting the date range condition appropriately. The batch loading actual content of records.


app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
```diff
@@ -118,38 +119,48 @@ public Upload getUpload(@Nonnull String uploadId) {
...
+            }
+        }
+        
+        String nextOffsetKey = null;
+        if (results.size() > pageSize) {
+            nextOffsetKey = Long.toString(Iterables.getLast(results).getRequestedOn());
+        }
 
-        return new ForwardCursorPagedResourceList<>(results, nextOffsetKey, pageSize)
+        int lastIndex = Math.min(sizeWithIndicatorRecord-1, results.size());
```
liujoshua: sizeWithIndicatorRecord-1 took a while to understand, let's use pageSize

liujoshua (approved):

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
alxdarksage: Yes, changed

PR #1492, opened 2017-06-14
admins can get and update participants

DwayneJengSage:
We recently made a change so that only admins can set the email verification status directly. However, admins can't actually get or update users. This breaks some integration tests (specifically, the one where we mark a user as *not* verified to test that unverified users can't log in). This also makes it so that we have no way of manually marking a user as verified.

This fixes that by adding admin access to get and update participants.


app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
```diff
@@ -176,7 +177,7 @@ public Result getRequestInfo(String userId) throws Exception {
     }
     
     public Result updateParticipant(String userId) {
-        UserSession session = getAuthenticatedSession(RESEARCHER);
+        UserSession session = getAuthenticatedSession(ADMIN, RESEARCHER);
         Study study = studyService.getStudy(session.getStudyIdentifier());
 
         StudyParticipant participant = parseJson(request(), StudyParticipant.class);
```
alxdarksage, line 183: At some point I may add ADMIN to getParticipants() so that in the UI, you can search for and retrieve the participant. Right now, I am usually also a researcher at the point I am setting up a study.

alxdarksage (approved):
This is fine, we should add to docs and SDK.

PR #1493, opened 2017-06-16
Survey rule changes

alxdarksage:
- rules moved from constraints to element, both questions and information screens, in a backwards compatible manner
- adjusted survey validation appropriately (only some rules applicable to info screens).
- added the "always" operator. Always do this rule... doesn't matter what the answer to the question, or if there's any answer at all (even declined);


syoung-smallwisdom (approved):
Looks good. 

It wasn't entirely clear to me whether or not this changes the json that will be sent to the client. Looks like it is the same? I am checking b/c I wouldn't want this to break existing apps. Thanks

alxdarksage:
The rules in the constraints will be mirrored one level up in the element. We should be able to switch client libraries over to the element without anything breaking in the meanwhile.

PR #1494, opened 2017-06-19
Add data group rule

alxdarksage:
This builds on the prior PR to add an "ALWAYS" operator to survey rules. That PR should be merged first. This rule specifies a condition under which a user should be tagged with a data group.


syoung-smallwisdom (approved):
looks good

PR #1497, opened 2017-06-20
Accounting for fact that elements return empty arrays, even before they are defined.

alxdarksage:
Deserialization of surveys creates empty arrays (not nulls) for SurveyElement rules. Account for this in the logic to allow migration from constraint rules to element rules.

### DIFF
--- a/test/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDaoTest.java
+++ b/test/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDaoTest.java
@@ -20,6 +20,7 @@ import javax.annotation.Resource;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


DwayneJengSage (changes requested):
I don't see test cases for:

* null rules list in question, non-null non-empty rules list in constraints
* empty rules list in question, non-null non-empty rules in constraints

alxdarksage:
Added these test cases.

test/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDaoTest.java
```diff
@@ -673,6 +678,40 @@ public void rulesMovedToSurveyElementInBackwardsCompatibleManner() throws Except
...
+        assertEquals(1, ((SurveyQuestion)element2).getConstraints().getRules().size());
+        
+        // Setting both to no rules will clear them.
+        Survey survey3 = updateRules(survey2, ImmutableList.of(), ImmutableList.of());
+        SurveyElement element3 = survey3.getElements().get(0);
+        assertEquals(0, element3.getRules().size());
+        assertEquals(0, ((SurveyQuestion)element3).getConstraints().getRules().size());
+        
+        // null rules list in question, non-null non-empty rules list in constraints
+        Survey survey4 = updateRules(survey3, null, ImmutableList.of());
```
DwayneJengSage: Comment says non-null non-empty, but the list is empty.

DwayneJengSage (changes requested):

PR #1502, opened 2017-06-23
Setup new relic for BridgePF

zaro0508:
Setup new relic agent to monitor java performance on AWS EC2 instances.

* update new relic to ver 3.39.1
* install new relic agent config to ec2 instances
* configured the following Beanstalk ENV config vars which overrides
  setting in the agent config file:

  JAVA_OPTS=-Dnewrelic.config.file=/var/app/current/newrelic/newrelic.yml -javaagent:/var/app/current/lib/com.newrelic.agent.java.newrelic-agent-3.39.1.jar
  NEW_RELIC_APP_NAME=BridgePF-dev-AWS
  NEW_RELIC_LICENSE_KEY=XXXXX
  NEW_RELIC_LOG=/var/log/newrelic_agent.log

### DIFF
--- a/dist/newrelic/newrelic.yml
+++ b/dist/newrelic/newrelic.yml
@@ -1,0 +1,302 @@
+# This file configures the New Relic Agent.  New Relic monitors
+# Java applications with deep visibility and low overhead.  For more details and additional
+# configuration options visit https://docs.newrelic.com/docs/java/java-agent-configuration.
+#
+# This configuration file is custom generated for app24538145@heroku.com_3
+#
+# This section is for settings common to all environments.
+# Do not add anything above this next line.
+common: &default_settings
+
+  # ============================== LICENSE KEY ===============================
+  # You must specify the license key associated with your New Relic
+  # account. For example, if your license key is 12345 use this:
+  # license_key: '12345'
+  # The key binds your Agent's data to your account in the New Relic service.
+  license_key: 'a999a743da64efbab3ab2a999e9f3ae9df799f66'
+
+  # Agent Enabled
+  # Use this setting to disable the agent instead of removing it from the startup command.
+  # Default is true.
+  agent_enabled: true
+
+  # Set the name of your application as you'd like it show up in New Relic.
+  # If enable_auto_app_naming is false, the agent reports all data to this application.
+  # Otherwise, the agent reports only background tasks (transactions for non-web applications)
+  # to this application. To report data to more than one application 
+  # (useful for rollup reporting), separate the application names with ";".
+  # For example, to report data to "My Application" and "My Application 2" use this:
+  # app_name: My Application;My Application 2
+  # This setting is required. Up to 3 different application names can be specified.
+  # The first application name must be unique.
+  app_name: My Application
+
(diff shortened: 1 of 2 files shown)


dist/newrelic/newrelic.yml
```diff
@@ -0,0 +1,302 @@
+# This file configures the New Relic Agent.  New Relic monitors
```
DwayneJengSage, line 1: We already have a newrelic.yml https://github.com/Sage-Bionetworks/BridgePF/blob/develop/newrelic.yml

Why do we need a second one, and why is this different from the current one?

Also, iirc, a lot of the settings shouldn't be specified in the newrelic.yml file, because a lot of the settings can be configured directly through the New Relic web UI.

DwayneJengSage (changes requested):

dist/newrelic/newrelic.yml
zaro0508: I believe the one you are referencing is for heroku. I'm not sure where it came from but I think it's just from an older version of the new relic agent.   This one is for the upgraded new relic agent (ver 3.39.1) and it's placed in the dist folder so that it will get deployed to the elastic beanstalk ec2 instances.    I guess we can attempt to consolidate them but I didn't feel the need to do that since the heroku one will just go away once we move to AWS.

In terms of setting configs with newrelic.yml vs from web ui.  There shouldn't be any problems setting configs in yml and web ui.  The web ui settings take precedence[1] over yml file configs allowing you to override the configs.   I actually prefer setting the configs in yml so we can keep it under source control.

[1] https://docs.newrelic.com/docs/agents/java-agent/configuration/java-agent-configuration-config-file

DwayneJengSage: My concerns are twofold:

1. We spent a lot of time configuring New Relic. The new configs should match the existing ones. If we're throwing away the existing configs, we should have a good reason for it.

2. If we need to deploy to Bridge every time we want to tweak the New Relic monitoring, this will severely slow down our ability to change our monitoring settings. If the actual configured monitoring settings are always going to be out-of-sync with source control, why have them in source control in the first place?

zaro0508: Replies..
1. We are not throwing any configs away.  The existing configs are in this change file.  New configs in this file are in addition to the existing configs.   I do see that enable_auto_app_naming is different so I will update to 'true' which matches existing config.

2. The setting in the file are just defaults settings which can be overriden in the new relic UI.  Keeping them in the file allows you to more easily change defaults and not have to reference the docs to determine what the defaults are for each config.  It continue to work the same as the existing config file.  You don't need to redeploy bridge if you set these configs in the new relic UI.  Anyways if you are still opposed to this then I can remove the configs and make this match the existing heroku config file.

DwayneJengSage: The "defaults" don't match what we currently have configured in New Relic, so they'll almost certainly be overridden as soon as we deploy. If we're not using the defaults anyway, why bother have them at all? Including them in git will only serve to confuse future developers.

DwayneJengSage (approved):

PR #1503, opened 2017-06-26
Data group rules

alxdarksage:
Splits element rules into a before and after set.
Add two new actions for before sets: displayIf and displayUnless
Add the ability to specify data groups to test against (rather than the answer to a question);
Add two new set operators, "any" and "all", specifically for testing against data groups
Update validation rules

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyElement.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyElement.java
@@ -27,7 +27,8 @@ public class DynamoSurveyElement implements SurveyElement {
     private String type;
     private int order;
     private JsonNode data;
-    private List<SurveyRule> rules;
+    private List<SurveyRule> beforeRules;
+    private List<SurveyRule> afterRules;
 
     public DynamoSurveyElement() {
     }
@@ -93,15 +94,23 @@ public class DynamoSurveyElement implements SurveyElement {
      */
     @DynamoDBTypeConverted(converter = SurveyRuleListMarshaller.class)
     @DynamoDBAttribute
-    public List<SurveyRule> getRules() {
-        return (this.rules == null) ? null : ImmutableList.copyOf(this.rules);
+    public List<SurveyRule> getAfterRules() {
+        return (this.afterRules == null) ? null : ImmutableList.copyOf(this.afterRules);
     }
-    public void setRules(List<SurveyRule> rules) {
-        this.rules = rules;
+    public void setAfterRules(List<SurveyRule> afterRules) {
+        this.afterRules = afterRules;
+    }
+    @DynamoDBTypeConverted(converter = SurveyRuleListMarshaller.class)
+    @DynamoDBAttribute
+    public List<SurveyRule> getBeforeRules() {
+        return (this.beforeRules == null) ? null : ImmutableList.copyOf(this.beforeRules);
+    }
+    public void setBeforeRules(List<SurveyRule> beforeRules) {
+        this.beforeRules = beforeRules;
     }
     @Override
     public int hashCode() {
-        return Objects.hash(guid, identifier, order, surveyCompoundKey, type, rules);
+        return Objects.hash(guid, identifier, order, surveyCompoundKey, type, beforeRules, afterRules);
     }
 
     @Override
@@ -116,7 +125,8 @@ public class DynamoSurveyElement implements SurveyElement {
                 Objects.equals(order, this.order) &&
                 Objects.equals(surveyCompoundKey, this.surveyCompoundKey) &&
                 Objects.equals(type, other.type) &&
-                Objects.equals(rules, other.rules);
+                Objects.equals(afterRules, other.afterRules) &&
+                Objects.equals(beforeRules, other.beforeRules);
     }
     
 }
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyElement.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyElement.java
@@ -1,7 +1,6 @@
 package org.sagebionetworks.bridge.dynamodb;
 
 import java.util.List;
-import java.util.Objects;
 
 import org.sagebionetworks.bridge.models.surveys.SurveyElement;
 import org.sagebionetworks.bridge.models.surveys.SurveyElementConstants;
@@ -108,25 +107,4 @@ public class DynamoSurveyElement implements SurveyElement {
     public void setBeforeRules(List<SurveyRule> beforeRules) {
         this.beforeRules = beforeRules;
     }
-    @Override
-    public int hashCode() {
-        return Objects.hash(guid, identifier, order, surveyCompoundKey, type, beforeRules, afterRules);
-    }
-
-    @Override
-    public boolean equals(Object obj) {
-        if (this == obj)
-            return true;
-        if (obj == null || getClass() != obj.getClass())
-            return false;
-        DynamoSurveyElement other = (DynamoSurveyElement) obj;
-        return Objects.equals(guid, other.guid) &&
-                Objects.equals(identifier, other.identifier) &&
-                Objects.equals(order, this.order) &&
-                Objects.equals(surveyCompoundKey, this.surveyCompoundKey) &&
-                Objects.equals(type, other.type) &&
-                Objects.equals(afterRules, other.afterRules) &&
-                Objects.equals(beforeRules, other.beforeRules);
-    }
-    
 }
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyInfoScreen.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyInfoScreen.java
@@ -29,6 +29,7 @@ public class DynamoSurveyInfoScreen extends DynamoSurveyElement implements Surve
         setIdentifier(entry.getIdentifier());
         setGuid(entry.getGuid());
         setData(entry.getData());
+        setBeforeRules(entry.getBeforeRules());
         setAfterRules(entry.getAfterRules());
     }
 
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyInfoScreen.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyInfoScreen.java
@@ -29,7 +29,7 @@ public class DynamoSurveyInfoScreen extends DynamoSurveyElement implements Surve
         setIdentifier(entry.getIdentifier());
         setGuid(entry.getGuid());
         setData(entry.getData());
-        setRules(entry.getRules());
+        setAfterRules(entry.getAfterRules());
     }
 
     @Override
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyInfoScreen.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyInfoScreen.java
@@ -25,6 +25,7 @@ public class DynamoSurveyInfoScreen extends DynamoSurveyElement implements Surve
     }
     
     public DynamoSurveyInfoScreen(SurveyElement entry) {
+        setSurveyCompoundKey(entry.getSurveyCompoundKey());
         setType(entry.getType());
         setIdentifier(entry.getIdentifier());
         setGuid(entry.getGuid());
@@ -97,48 +98,4 @@ public class DynamoSurveyInfoScreen extends DynamoSurveyElement implements Surve
         this.title = JsonUtils.asText(data, TITLE_PROPERTY);
         this.image = JsonUtils.asEntity(data, IMAGE_PROPERTY, Image.class);
     }
-
-    @Override
-    public int hashCode() {
-        final int prime = 31;
-        int result = super.hashCode();
-        result = prime * result + ((prompt == null) ? 0 : prompt.hashCode());
-        result = prime * result + ((promptDetail == null) ? 0 : promptDetail.hashCode());
-        result = prime * result + ((title == null) ? 0 : title.hashCode());
-        result = prime * result + ((image == null) ? 0 : image.hashCode());
-        return result;
-    }
-
-    @Override
-    public boolean equals(Object obj) {
-        if (this == obj)
-            return true;
-        if (!super.equals(obj))
-            return false;
-        if (getClass() != obj.getClass())
-            return false;
-        DynamoSurveyInfoScreen other = (DynamoSurveyInfoScreen) obj;
-        if (prompt == null) {
-            if (other.prompt != null)
-                return false;
-        } else if (!prompt.equals(other.prompt))
-            return false;
-        if (promptDetail == null) {
-            if (other.promptDetail != null)
-                return false;
-        } else if (!promptDetail.equals(other.promptDetail))
-            return false;
-        if (title == null) {
-            if (other.title != null)
-                return false;
-        } else if (!title.equals(other.title))
-            return false;
-        if (image == null) {
-            if (other.image != null)
-                return false;
-        } else if (!image.equals(other.image))
-            return false;
-        return true;
-    }
-
 }
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyQuestion.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyQuestion.java
@@ -30,11 +30,12 @@ public class DynamoSurveyQuestion extends DynamoSurveyElement implements SurveyQ
     }
     
     public DynamoSurveyQuestion(SurveyElement entry) {
-        setType( entry.getType() );
-        setIdentifier( entry.getIdentifier() );
-        setGuid( entry.getGuid() );
-        setData( entry.getData() );
-        setAfterRules( entry.getAfterRules() );
+        setType(entry.getType());
+        setIdentifier(entry.getIdentifier());
+        setGuid(entry.getGuid());
+        setData(entry.getData());
+        setBeforeRules(entry.getBeforeRules());
+        setAfterRules(entry.getAfterRules());
     }
     
     @Override
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyQuestion.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyQuestion.java
@@ -34,7 +34,7 @@ public class DynamoSurveyQuestion extends DynamoSurveyElement implements SurveyQ
         setIdentifier( entry.getIdentifier() );
         setGuid( entry.getGuid() );
         setData( entry.getData() );
-        setRules( entry.getRules() );
+        setAfterRules( entry.getAfterRules() );
     }
     
     @Override
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyQuestion.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyQuestion.java
@@ -1,12 +1,13 @@
 package org.sagebionetworks.bridge.dynamodb;
 
-import java.util.Objects;
+import java.util.List;
 
 import org.sagebionetworks.bridge.json.BridgeObjectMapper;
 import org.sagebionetworks.bridge.json.JsonUtils;
 import org.sagebionetworks.bridge.models.surveys.Constraints;
 import org.sagebionetworks.bridge.models.surveys.SurveyElement;
 import org.sagebionetworks.bridge.models.surveys.SurveyQuestion;
+import org.sagebionetworks.bridge.models.surveys.SurveyRule;
 import org.sagebionetworks.bridge.models.surveys.UIHint;
 
 import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBAttribute;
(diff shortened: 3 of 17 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyElement.java
```diff
@@ -116,7 +125,8 @@ public boolean equals(Object obj) {
                 Objects.equals(order, this.order) &&
                 Objects.equals(surveyCompoundKey, this.surveyCompoundKey) &&
                 Objects.equals(type, other.type) &&
-                Objects.equals(rules, other.rules);
+                Objects.equals(afterRules, other.afterRules) &&
+                Objects.equals(beforeRules, other.beforeRules);
```
DwayneJengSage: Should we add an equalsVerifier test for this?

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyInfoScreen.java
```diff
@@ -29,7 +29,8 @@ public DynamoSurveyInfoScreen(SurveyElement entry) {
         setIdentifier(entry.getIdentifier());
         setGuid(entry.getGuid());
         setData(entry.getData());
-        setRules(entry.getRules());
+        setBeforeRules(entry.getBeforeRules());
+        setAfterRules(entry.getAfterRules());
```
DwayneJengSage, line 34: Do we have a unit test to make sure these rules are being properly pulled in from the SurveyElement?

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyQuestion.java
```diff
@@ -30,11 +30,12 @@ public DynamoSurveyQuestion() {
...
-        setIdentifier( entry.getIdentifier() );
-        setGuid( entry.getGuid() );
-        setData( entry.getData() );
-        setRules( entry.getRules() );
+        setType(entry.getType());
+        setIdentifier(entry.getIdentifier());
+        setGuid(entry.getGuid());
+        setData(entry.getData());
+        setBeforeRules(entry.getBeforeRules());
+        setAfterRules(entry.getAfterRules());
```
DwayneJengSage, line 40: Similarly, do we have a unit test for this?

app/org/sagebionetworks/bridge/models/surveys/SurveyElement.java
```diff
@@ -17,7 +17,8 @@
     public static final String IMAGE_PROPERTY = "image";
     public static final String PROMPT_DETAIL_PROPERTY = "promptDetail";
     public static final String PROMPT_PROPERTY = "prompt";
-    public static final String RULES_PROPERTY = "rules";
+    public static final String BEFORE_RULES_PROPERTY = "beforeRules";
+    public static final String AFTER_RULES_PROPERTY = "afterRules";
```
DwayneJengSage: Minor: Fields in interfaces are implicitly public static final.

app/org/sagebionetworks/bridge/models/surveys/SurveyRule.java
```diff
@@ -98,21 +133,26 @@ public boolean equals(Object obj) {
                Objects.equals(operator, other.operator) &&
                Objects.equals(value, other.value) &&
                Objects.equals(endSurvey, other.endSurvey) &&
-               Objects.equals(assignDataGroup, other.assignDataGroup);
+               Objects.equals(assignDataGroup, other.assignDataGroup) &&
+               Objects.equals(dataGroups, other.dataGroups);
```
DwayneJengSage: hashCode() and equals() are missing displayIf and displayUnless. This is almost certainly incorrect.

app/org/sagebionetworks/bridge/models/surveys/SurveyRule.java
```diff
@@ -31,22 +33,34 @@
...
-            String assignDataGroup) {
+            String assignDataGroup, Set<String> dataGroups, Boolean displayIf, Boolean displayUnless) {
         this.operator = operator;
         this.value = value;
         this.skipToTarget = skipToTarget;
         this.endSurvey = endSurvey;
         this.assignDataGroup = assignDataGroup;
+        this.dataGroups = dataGroups;
+        this.displayIf = displayIf;
+        this.displayUnless = displayUnless;
```
DwayneJengSage, line 63: Do we have JSON serialization tests for these?

app/org/sagebionetworks/bridge/models/surveys/SurveyRule.java
```diff
@@ -98,21 +133,26 @@ public boolean equals(Object obj) {
...
+               Objects.equals(assignDataGroup, other.assignDataGroup) &&
+               Objects.equals(dataGroups, other.dataGroups);
     }
 
     @Override
     public String toString() {
         return "SurveyRule [operator=" + operator + ", value=" + value + ", skipToTarget=" + skipToTarget
-                + ", endSurvey=" + endSurvey + ", assignDataGroup=" + assignDataGroup + "]";
+                + ", endSurvey=" + endSurvey + ", assignDataGroup=" + assignDataGroup + ", dataGroups=" + 
+                dataGroups + "]";
```
DwayneJengSage: Minor: displayIf and displayUnless? Also, what happens if any of these are null?

app/org/sagebionetworks/bridge/models/surveys/SurveyRule.java
```diff
@@ -133,12 +173,29 @@ public Builder withEndSurvey(Boolean endSurvey) {
             }
             return this;
         }
+        public Builder withDisplayIf(Boolean displayIf) {
+            if (Boolean.TRUE.equals(displayIf)) {
+                this.displayIf = displayIf;
+            }
```
DwayneJengSage: Why do we need to check that this is true? Are we concerned that people might accidentally call withDisplayIf(true), and then later call withDisplayIf(false)? If so, why is leaving it to true more intuitive than leaving it to false?

Alternatively, if the goal is to default to false, it would be better for the lower case b boolean (since we'd only have two states anyway) in both the class and the builder (explicitly initialize it to false for documentation/clarity). This way, it'll be consistent with all the other builders, be easier to reason about, and easier to test.

If we _do_ keep this logic, we'll want to add unit tests to make sure that, once set, we can't clear it with withDisplayIf(false) or withDisplayIf(null).

(Here and in withDisplayUnless().)

app/org/sagebionetworks/bridge/validators/SurveySaveValidator.java
```diff
@@ -224,18 +263,30 @@ private void validateOneRuleSet(Errors errors, SurveyRule rule, Set<String> alre
...
-            errors.rejectValue(fieldPath, "must have one and only one action: skipTo, endSurvey, or assignDataGroup");
+        if (rule.getDisplayIf() != null) {
+            actionCount++;
         }
-        if (rule.getAssignDataGroup() != null && !dataGroups.contains(rule.getAssignDataGroup())) {
-            errors.rejectValue(fieldPath, "has a data group '" + rule.getAssignDataGroup()
-                    + "' that is not a valid data group: " + BridgeUtils.COMMA_SPACE_JOINER.join(dataGroups));            
+        if (rule.getDisplayUnless() != null) {
+            actionCount++;
         }
```
DwayneJengSage, line 275: It doesn't appear that we have any tests that verify that you can't have displayIf and displayUnless at the same time. We also want to avoid having 2^5 tests testing every possible combination, as that would not be maintainable.

My recommendation is to have N tests, one for each action, so that we can verify that they're each being counted by hasOneAction(), as well as a test with no actions.

test/org/sagebionetworks/bridge/validators/SurveySaveValidatorTest.java
```diff
@@ -881,11 +881,11 @@ public void operatorsInInfoScreenRulesAreValidated() {
         info.setPromptDetail("prompt detail");
         info.setIdentifier("identifier");
         info.setGuid("guid");
-        info.setRules(Lists.newArrayList(
+        info.setAfterRules(Lists.newArrayList(
                 new SurveyRule.Builder().withValue("foo").withOperator(Operator.EQ).withEndSurvey(true).build()));
```
DwayneJengSage: Should we also add beforeRules to verify that both are being tested? (Whether this logically goes here or in beforeRulesAreValidated() is up to you.)

app/org/sagebionetworks/bridge/validators/SurveySaveValidator.java
```diff
@@ -143,77 +144,115 @@ private void validateRules(Errors errors, List<SurveyElement> elements) {
...
+                // Only validate fields that are relevant for the operator, ignore the other.
+                if (SurveyRule.SET_OPERATORS.contains(rule.getOperator())) {
+                    // tests against data groups
+                    if (isEmpty(rule.getDataGroups())) {
+                        errors.rejectValue(fieldPath, "should define one or more data groups");
+                    } else if (!dataGroups.containsAll(rule.getDataGroups())) {
+                        errors.rejectValue(fieldPath,
+                            "contains dataGroups '" + COMMA_SPACE_JOINER.join(rule.getDataGroups())
+                                    + "' that are not valid data groups: " + COMMA_SPACE_JOINER.join(dataGroups));
+                    }
```
DwayneJengSage: Unit tests for these?
* empty data groups for ANY
* invalid data groups for ANY
* at least one invalid rule with ALL to verify that that's being tested

app/org/sagebionetworks/bridge/validators/SurveySaveValidator.java
```diff
@@ -143,77 +144,115 @@ private void validateRules(Errors errors, List<SurveyElement> elements) {
...
+                } else if (valueMissingForOperator(rule)) {
+                    errors.rejectValue(fieldPath, "is required");
+                }
```
DwayneJengSage: Unit test for missing value?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyElement.java
alxdarksage: I'm going to remove the hashCode/equals methods from the SurveyElement class hierarchy. It's broken, and isn't used, even for tests. Creating hashCode/equals methods for SurveyElement which has to be mutable for Dynamo, and subclassable in a way that changes equality, and includes JsonNode (which has abstract methods that need prefab values)... it's just a quagmire and not worth it. So, removed.

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyInfoScreen.java
alxdarksage: Added tests of the copy constructors for SurveyInfoScreen and SurveyQuestion

app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyQuestion.java
alxdarksage: Added

app/org/sagebionetworks/bridge/models/surveys/SurveyElement.java
alxdarksage: Removed unnecessary keywords

app/org/sagebionetworks/bridge/models/surveys/SurveyRule.java
alxdarksage: Added

app/org/sagebionetworks/bridge/models/surveys/SurveyRule.java
alxdarksage: Added (these are used for tests and SurveyRuleTest was failing... I'm rerunning all the tests before committing next round of changes).

app/org/sagebionetworks/bridge/models/surveys/SurveyRule.java
alxdarksage: Added (when they are null they just display as "someField=null" which is fine for toString).

app/org/sagebionetworks/bridge/models/surveys/SurveyRule.java
alxdarksage: If the user sends a property set to false (which they could do), we set the object field to null so that it is not serialized and sent back in future JSON. Only one boolean property can be true at a time, and this ensures only the true property will be in JSON representations. I've added a test to SurveyRuleTest to verify this.

app/org/sagebionetworks/bridge/validators/SurveySaveValidator.java
alxdarksage: Added

app/org/sagebionetworks/bridge/validators/SurveySaveValidator.java
alxdarksage: Added, and tests for null value operators that don't need a value (always, declined)

app/org/sagebionetworks/bridge/validators/SurveySaveValidator.java
alxdarksage: I added a test specifically for each action, no actions, and two actions. This forced some refactoring of the test class to reduce the duplication of creating a survey with one element and one rule in it to validate these SurveyRule validations.

test/org/sagebionetworks/bridge/validators/SurveySaveValidatorTest.java
alxdarksage: I've verified that in the three ways we validate afterRules, we validate beforeRules, I have not validated the every single validation works through both code paths because I have verified that the same code is executed on both (as both are a list of rules).

app/org/sagebionetworks/bridge/validators/SurveySaveValidator.java
DwayneJengSage: My concern is that we don't know that this code is correctly counting displayIf and displayUnless unless we test it.

test/org/sagebionetworks/bridge/validators/SurveySaveValidatorTest.java
DwayneJengSage: I see the test for validateOneRuleSet() and validateSkipToTargetExists(), but not validateOnRuleSetInInfoScreen().

app/org/sagebionetworks/bridge/models/surveys/SurveyRule.java
DwayneJengSage: As written, once this is set to true, it can never be cleared. If we want false to be null, we need an else block to set it to null.

app/org/sagebionetworks/bridge/validators/SurveySaveValidator.java
alxdarksage: I've added one test for each action to verify the action makes a valid rule... I can alter the validator to return the count from a protected method, then one test could verify that all actions are being counted.

test/org/sagebionetworks/bridge/validators/SurveySaveValidatorTest.java
alxdarksage: Okay, I've added before/after rules to the info screen and validated there are errors for both.

app/org/sagebionetworks/bridge/models/surveys/SurveyRule.java
alxdarksage: I haven't really ever called this method more than once because it's in the builder... but I can add that if you think it's a problem down the road.

test/org/sagebionetworks/bridge/dynamodb/DynamoSurveyQuestionTest.java
```diff
@@ -1,17 +1,52 @@
...
+        question.setUiHint(UIHint.COMBOBOX);
+        question.setConstraints(new IntegerConstraints());
+        question.setSurveyCompoundKey("surveyCompoundKey");
+        question.setGuid("guid");
+        question.setIdentifier("identifier");
+        question.setType("SurveyQuestion");
+        SurveyRule beforeRule = new SurveyRule.Builder().withDisplayUnless(true).withDataGroups(Sets.newHashSet("foo")).build();
+        SurveyRule afterRule = new SurveyRule.Builder().withOperator(Operator.ALWAYS).withEndSurvey(true).build();
+        question.setBeforeRules(Lists.newArrayList(beforeRule));
+        question.setAfterRules(Lists.newArrayList(afterRule));
```
DwayneJengSage, line 36: Before and after rules aren't checked in the copy.

app/org/sagebionetworks/bridge/models/surveys/SurveyRule.java
alxdarksage: Updated these to explicitly set to null if not true

test/org/sagebionetworks/bridge/models/surveys/SurveyRuleTest.java
```diff
@@ -94,4 +95,58 @@ public void canSerializeAssignDataGroupRule() throws Exception {
...
+    
+    @Test
+    public void canSerializeDisplayIf() throws Exception {
+        SurveyRule displayIf = new SurveyRule.Builder().withOperator(Operator.ANY)
+                .withDataGroups(Sets.newHashSet("foo")).withDisplayIf(Boolean.TRUE).build();
+        
+        JsonNode node = BridgeObjectMapper.get().valueToTree(displayIf);
+        assertEquals("any", node.get("operator").asText());
+        assertEquals("foo", node.get("dataGroups").get(0).asText());
+        assertTrue(node.get("displayIf").asBoolean());
```
DwayneJengSage: asBoolean() will attempt to parse the JSON value into a boolean. This means we don't know if the serializer is serializing the value as, for example, the int 1 or the string "true". Recommend using booleanValue() instead. (Likewise textValue() instead of asText(), although this matters less often.)

Similarly for canSerializeDisplayUnless().

app/org/sagebionetworks/bridge/models/surveys/SurveyRule.java
```diff
@@ -31,22 +33,34 @@
...
     
     private SurveyRule(Operator operator, Object value, String skipToTarget, Boolean endSurvey,
-            String assignDataGroup) {
+            String assignDataGroup, Set<String> dataGroups, Boolean displayIf, Boolean displayUnless) {
         this.operator = operator;
         this.value = value;
         this.skipToTarget = skipToTarget;
         this.endSurvey = endSurvey;
         this.assignDataGroup = assignDataGroup;
+        this.dataGroups = dataGroups;
```
DwayneJengSage, line 61: I don't see a serialization test for dataGroups.

test/org/sagebionetworks/bridge/validators/SurveySaveValidatorTest.java
```diff
@@ -785,89 +786,55 @@ public void willValidateEarliestDateTimeIsNotAfterLatestDateTime() {
...
     @Test
     public void endSurveyRuleValidInConstraints() {
-        survey = new TestSurvey(SurveySaveValidatorTest.class, false);
-
-        // This rule is a valid "end the survey" rule, and passes validation.
-        SurveyQuestion question = ((TestSurvey) survey).getIntegerQuestion();
         SurveyRule rule = new SurveyRule.Builder().withOperator(SurveyRule.Operator.EQ).withValue(1)
                 .withEndSurvey(Boolean.TRUE).build();
-        question.getConstraints().setRules(Lists.newArrayList(rule));
+        updateSurveyWithAfterRulesInOneQuestion(rule);
```
DwayneJengSage: This test is for constraints, but got moved to afterRules.

test/org/sagebionetworks/bridge/validators/SurveySaveValidatorTest.java
```diff
@@ -785,89 +786,55 @@ public void willValidateEarliestDateTimeIsNotAfterLatestDateTime() {
...
-        question.getConstraints().setRules(Lists.newArrayList(rule));
+        updateSurveyWithRulesInConstraints(rule);
 
-        assertValidatorMessage(validator, survey, "elements[4].constraints.rules[0]",
+        assertValidatorMessage(validator, survey, "elements[0].constraints.rules[0]",
                 "has a skipTo identifier that doesn't exist: this_does_not_exist");
     }
 
     @Test
-    public void backreferenceSkipToTargetInvalidInElement() throws Exception {
```
DwayneJengSage: Why is this test deleted?

test/org/sagebionetworks/bridge/validators/SurveySaveValidatorTest.java
```diff
@@ -881,41 +848,218 @@ public void operatorsInInfoScreenRulesAreValidated() {
...
+                "contains dataGroups 'notInStudy' that are not valid data groups: baz, foo");
+    }
+    
+    @Test
+    public void validatesMixedValidityDataGroupsForAllOperator() {
+        TreeSet<String> dataGroups = new TreeSet();
+        dataGroups.add("foo");
+        dataGroups.add("notInStudy");
+        
+        SurveyRule rule = new SurveyRule.Builder().withOperator(SurveyRule.Operator.ANY)
```
DwayneJengSage: This test is for ALL, but the code is for ANY.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #1507, opened 2017-07-11
Stop sharing deployment buckets

zaro0508:
To facilitate cloudformation deployments we want to use
separate resources for each environment.


zaro0508:
replaced with https://github.com/Sage-Bionetworks/BridgePF/pull/1510

PR #1509, opened 2017-07-11
User new email for existing account

alxdarksage:
Using the new template when we detect an account already exists.

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyQuestion.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoSurveyQuestion.java
@@ -1,13 +1,10 @@
 package org.sagebionetworks.bridge.dynamodb;
 
-import java.util.List;
-
 import org.sagebionetworks.bridge.json.BridgeObjectMapper;
 import org.sagebionetworks.bridge.json.JsonUtils;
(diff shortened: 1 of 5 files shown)


syoung-smallwisdom (commented):
I'm confused about the UX flow. How does the app know that they need to show a screen to let the user enter their new password? This is for a user who is already in the registration flow and has had a password set for them automatically. If the app doesn't get a callback to show login (rather than email verification) then the user is still hooped.

zaro0508:
Technically it looks good to me.

alxdarksage:
@syoung-smallwisdom: this happens when you sign up and must verify your email address. Right now we report success back to the client, but send the user a reset password email. Which is confusing if they are signing up a second time, having forgotten they created an account. This just changes the email to tell them they are signing up again... it still includes a reset password link.

I do not know if the client has ever gracefully handled this case, and I don't know that we've had any discussion about cases where the app tries to set the password. We should walk through that, but the situation is no worse for sending a clearer email.

The reason we do not report back anything in this API along the lines of a failure is to prevent session enumeration attacks against our public API.

liujoshua (approved):

syoung-smallwisdom (approved):

PR #1513, opened 2017-07-17
Get activities v4

alxdarksage:
The v4 API returns activities within a time range. Activities that are scheduled based on event timestamps, cannot all be calculated by the scheduler. So the v4 approach of "schedule activities, then replace persisted activities with their persisted data" will not work for activities that were based on an event timestamp that has subsequently been overwritten by a later event. The scheduler can only return the most recent activity in that sequence.

Persistent tasks, which are scheduled based on the timestamp "activity:<guid>:finished" fall squarely into this category. The scheduler can get the current persistent task in the sequence, it can't get the older ones in the time window.

This change gets all the GUIDs of all the scheduled activities, and then retrieves them from the DB after saving any newly scheduled activities. This picks up all necessary historical tasks in the time window. This is more resource intensive, so throughput may need to be increased as more apps use the v4 API.


test/org/sagebionetworks/bridge/services/ScheduledActivityServiceDuplicateTest.java
```diff
@@ -350,6 +350,7 @@ public void withNoPersistedTasksItWorksEvenOnNextDayUTC() throws Exception {
         // It's adjusted, magically it's still 6/30 I haven't figured out why yet.
         List<ScheduledActivity> list = filterByLabel(activities, "Do Persistent Activity");
         assertEquals(1, list.size());
+        System.out.println("context.getStartsOn(): " + context.getStartsOn());
```
syoung-smallwisdom: Did you mean to leave this in?

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -968,13 +1076,19 @@ private String firstTimeStampFor(int initialTZOffset, int requestTZOffset, Sched
         return activities.stream().map(ScheduledActivity::getGuid).collect(Collectors.toSet());
     }
     
+    // TODO: Make endsOn a constant, no reason for it to be here.
```
syoung-smallwisdom: Please sign and date TODO comments. That way if the TODO is still here in a year, you have a quick reference to when it was added. (Git history can change and isn't a great way to track this) Thanks!

syoung-smallwisdom (approved):
Question and a nit. Otherwise, looks good.

app/org/sagebionetworks/bridge/models/schedules/Activity.java
```diff
@@ -71,7 +71,8 @@ public SurveyReference getSurvey() {
         return survey;
     }
     public boolean isPersistentlyRescheduledBy(Schedule schedule) {
-        return schedule.schedulesImmediatelyAfterEvent() && getActivityFinishedEventId(schedule);
+        return (schedule.getScheduleType() == ScheduleType.PERSISTENT || 
+               (schedule.schedulesImmediatelyAfterEvent() && getActivityFinishedEventId(schedule)));
```
DwayneJengSage, line 75: Add test case to ActivityTest.activityKnowsWhenItIsPersistentlyScheduled().

app/org/sagebionetworks/bridge/models/schedules/PersistentActivityScheduler.java
```diff
@@ -29,7 +29,8 @@
             DateTime scheduledTime = getFirstEventDateTime(context, finishedId+"," + schedule.getEventId());
 
             if (scheduledTime != null) {
-                DateTime localDateTime = scheduledTime.withZone(context.getInitialTimeZone());
+                DateTime localDateTime = scheduledTime.withZone(context.getStartsOn().getZone());
```
DwayneJengSage: initialTimeZone is always fixed. startsOn.getZone() can change if the user changes timezones (or DST passes). Could this cause duplicate persistent events?

Additionally, in the PersistentActivityScheduleTest, startsOn is implicitly in the same time zone as initialTimeZone. Do we have a test to make sure this change does what we expect it to do?

app/org/sagebionetworks/bridge/models/schedules/ScheduledActivity.java
```diff
@@ -72,6 +72,15 @@ default ScheduledActivityStatus getStatus() {
...
         }
+        // We freeze the user's time zone to the initial time zone in order to prevent duplicates.
+        // For schedules that use the time portion of an event timestamp, this can cause odd 
+        // behavior when a user changes time zones. For the most part this resolves and is  
+        // acceptable, but persistent tasks are logically always available, even if you move in 
+        // a way that would shift the task to "scheduled" for some number of hours. Just prevent
+        // this.
+        if (getPersistent()) {
+            return ScheduledActivityStatus.AVAILABLE;
+        }
```
DwayneJengSage, line 83: Should we add test cases to DynamoScheduledActivityTest.hasValidStatusBasedOnTimestamps() to account for this new logic?

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
```diff
@@ -181,12 +181,86 @@ public ScheduledActivityList getActivityHistory(String healthCode,
...
+            
+            V4_MERGE.mergeActivityLists(saves, scheduledActivities, activity, dbActivity, i);
+        }
+        activityDao.saveActivities(saves);
+        
+        // We've removed all the persisted activities that were found by the scheduler. Those have been saved
+        // as necessary. The remaining tasks were scheduled based on user-triggered events that can be updated, 
+        // not events like "enrollment" but events like "when this activity is finished." These are now all 
+        // added to the activities that will be return.
+        scheduledActivities.addAll(dbMap.values());
```
DwayneJengSage: V3 and V4 are almost identical, differing in only a few places. How much of a breaking change would it be to have V3 also return all previous activities for a guid, so that we don't have to have two methods that are nearly identical? (Given that we're also reverting V4_MERGE to match V3_MERGE, the only functional difference between V3 and V4 is that V4 also shows finished and expired tasks.)

If we decide that showing historical tasks is a breaking change for V3, can we add a post-processing step? This would be cleaner than having two nearly identical methods.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -968,13 +1072,19 @@ private String firstTimeStampFor(int initialTZOffset, int requestTZOffset, Sched
...
-                .withStartsOn(NOW).withAccountCreatedOn(ENROLLMENT.minusHours(2)).withEndsOn(endsOn)
-                .withHealthCode(HEALTH_CODE).withUserId(USER_ID).withEvents(events);
+        return new ScheduleContext.Builder().withStudyIdentifier(TEST_STUDY)
+                .withInitialTimeZone(DateTimeZone.forOffsetHours(-7))
+                .withAccountCreatedOn(ENROLLMENT.minusHours(2))
+                .withStartsOn(STARTS_ON)
+                .withEndsOn(endsOn)
+                .withHealthCode(HEALTH_CODE)
+                .withUserId(USER_ID)
+                .withEvents(events);
```
DwayneJengSage: Why are we changing this? We specifically set up the tests this way to test the time zone fix. If we change the schedule context here, we'll need to reason through N different test cases to make sure we didn't remove a test case or accidentally break something.

I'd strongly recommend keeping this intact. If tests break because of it, it's very likely that we accidentally introduced a bug, which we need to understand.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
```diff
@@ -181,12 +181,87 @@ public ScheduledActivityList getActivityHistory(String healthCode,
...
+            
+            V4_MERGE.mergeActivityLists(saves, scheduledActivities, activity, dbActivity, i);
+        }
+        activityDao.saveActivities(saves);
+        
+        // We've removed all the persisted activities that were found by the scheduler. Those have been saved
+        // as necessary. The remaining tasks were scheduled based on user-triggered events that can be updated, 
+        // not events like "enrollment" but events like "when this activity is finished." These are now all 
+        // added to the activities that will be returned.
+        scheduledActivities.addAll(dbMap.values());
```
DwayneJengSage, line 196: V3 and V4 are almost identical, differing in only a few places. How much of a breaking change would it be to have V3 also return all previous activities for a guid, so that we don't have to have two methods that are nearly identical? (Given that we're also reverting V4_MERGE to match V3_MERGE, the only functional difference between V3 and V4 is that V4 also shows finished and expired tasks.)

If we decide that showing historical tasks is a breaking change for V3, can we add a post-processing step? This would be cleaner than having two nearly identical methods.

DwayneJengSage (changes requested):

DwayneJengSage (changes requested):
Additional issues filed as JIRAs:

https://sagebionetworks.jira.com/browse/BRIDGE-1906 - This is not a blocker, but it should be trivially to fix.

https://sagebionetworks.jira.com/browse/BRIDGE-1907 - This is a blocker for this PR. Without fixing the tests, we can't verify that your code changes are correct.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: Implemented

app/org/sagebionetworks/bridge/models/schedules/Activity.java
alxdarksage: Added. We decided the older logic can be removed after we update all the schedules in production, but I've file that for later as BRIDGE-1908.

app/org/sagebionetworks/bridge/models/schedules/PersistentActivityScheduler.java
alxdarksage: I will revert this. I don't want to change this... instead I decided just to mark persistent tasks as available, even if the user moves timezones very near to the event timestamp causing the task to shift to scheduled. I've split the endsOn and initial time zones in the persistent activities test.

app/org/sagebionetworks/bridge/models/schedules/ScheduledActivity.java
alxdarksage: Added similar test for persistent type

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: Good point. Reverted. Doesn't effect tests.

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
alxdarksage: This has been changed to use the same merge behavior.

alxdarksage: Okay looking at this, I would rather keep them separate. They're slightly different. I did combine more of the duplicated code between both the v3 and v4 methods.

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
```diff
@@ -180,13 +180,68 @@ public ScheduledActivityList getActivityHistory(String healthCode,
...
+        
+        // Compare scheduled and persisted activities, replacing scheduled with persisted where they exist
+        List<ScheduledActivity> saves = performMerge(scheduledActivities, dbMap, V4_COLLECT_SAVES);
+        activityDao.saveActivities(saves);
+        
+        // We've removed all the persisted activities that were found by the scheduler. Those have been saved
+        // as necessary. The remaining tasks were scheduled based on user-triggered events that can be updated, 
+        // not events like "enrollment" but events like "when this activity is finished." These are now all 
+        // added to the activities that will be returned.
+        scheduledActivities.addAll(dbMap.values());
```
DwayneJengSage, line 196: Reasoning through this, I'm pretty sure the bug for https://sagebionetworks.jira.com/browse/BRIDGE-1906 is still present here. We save the new tasks into DDB, but then we immediately overwrite the return values with the old DDB values, at least for one getScheduledActivities() call.

This also tells me that the unit tests need to go up a level and call getScheduledActivities() as a whole instead of calling its individual parts. These unit tests were fine back when merge logic was entirely encapsulated in updateActivitiesAndCollectSaves() https://github.com/Sage-Bionetworks/BridgePF/blob/9404a3536ac3826c6b6bc70b5f42789d2507129d/app/org/sagebionetworks/bridge/services/ScheduledActivityService.java#L132 However, now we have merge logic that exists outside of performMerge() that's no longer being tested.

Ultimately, I think what we actually want is just the logic in V3_MERGE. If there's no DDB Activity, or the DDB Activity is in SCHEDULED or AVAILABLE, we should use the newly generated activity. Otherwise, use the one from DDB. We already know this logic works and has been thoroughly tested. The only difference is that in V4, we want to also save expired tasks. It should be fine to extend this logic to V3, since V3 always hides expired tasks anyway.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -374,8 +381,9 @@ public void deleteActivitiesForSchedulePlanRejectsBadValue() {
     public void newActivitiesIncludedInSaveAndResultsV3() {
         List<ScheduledActivity> scheduled = createNewActivities("AAA", "BBB");
         List<ScheduledActivity> db = createStartedActivities("BBB");
+        Map<String, ScheduledActivity> dbMap = Maps.uniqueIndex(db, ScheduledActivity::getGuid);
         
-        List<ScheduledActivity> saves = merge(scheduled, db, V3_MERGE);
+        List<ScheduledActivity> saves = service.performMerge(scheduled, dbMap, V3_MERGE);
```
DwayneJengSage: Now that a lot of the merge logic is being handled outside of performMerge() (especially for V4), I don't believe this is sufficient. We need to move it up one level and test by calling getScheduledActivities().

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -912,17 +1055,18 @@ private String firstTimeStampFor(int initialTZOffset, int requestTZOffset, Sched
         return activities.get(0).getScheduledOn().toString();
     }
     
-    private List<ScheduledActivity> merge(List<ScheduledActivity> scheduled, List<ScheduledActivity> db, NewAndPersistedActivitiesMerger merger) {
+    /*
+    private List<ScheduledActivity> merge2(List<ScheduledActivity> scheduled, List<ScheduledActivity> db, NewAndPersistedActivitiesMerger merger) {
```
DwayneJengSage, line 1132: Commented test code should be removed.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
alxdarksage: Okay I see what you mean about updating the db tasks before retrieving them, the V3 merge logic can be used in both cases (I'm realizing now that the scheduler excludes expired tasks as well, so these won't run into the hundreds or thousands. They are calculated but not returned outside the startsOn/endsOn window, which exist for both APIs).

The persisted tasks being added after that should no longer be actionable because they weren't returned by the scheduler. So I don't think we need to update their references. In truth there are edge cases that can come up but I don't think it's worth trying to anticipate them (e.g. a custom event that is used to trigger a task, where the user creates the task and then updates the custom event without updating the task... the original task is going to stay available/expired based on its logic, maybe that's what a client developer would want).

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: Removed

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: This series of tests has been refactored to call the v3 and v4 service methods.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -372,111 +380,154 @@ public void deleteActivitiesForSchedulePlanRejectsBadValue() {
...
+        
+        List<ScheduledActivity> returnedActivities = service.getScheduledActivities(createScheduleContext(NOW).build());
+        assertNotNull(returnedActivities.get(0).getStartedOn());
+        assertActivityGuids(returnedActivities, "AAA", "BBB", "CCC");
         
-        scheduled = service.orderActivities(scheduled, V3_FILTER);
-        assertNotNull(scheduled.get(0).getStartedOn());
+        verify(activityDao).saveActivities(scheduledActivityListCaptor.capture());
+        List<ScheduledActivity> saves = scheduledActivityListCaptor.getValue();
+        assertActivityGuids(saves, "AAA", "BBB");
```
DwayneJengSage: I'm about 95% sure the purpose of this test was to test what happens if we have no new activities to save. We should remove AAA and BBB from getSchedulePlans() to make sure this test tests the same thing as the original test.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -372,111 +380,154 @@ public void deleteActivitiesForSchedulePlanRejectsBadValue() {
...
-        List<ScheduledActivity> saves = merge(scheduled, db, V3_MERGE);
-        assertActivities(saves);
+        List<ScheduledActivity> returnedActivities = service.getScheduledActivities(createScheduleContext(NOW).build());
+        assertActivityGuids(returnedActivities, "BBB");
         
-        scheduled = service.orderActivities(scheduled, V3_FILTER);
-        assertTrue(scheduled.isEmpty());
+        verify(activityDao).saveActivities(scheduledActivityListCaptor.capture());
+        List<ScheduledActivity> saves = scheduledActivityListCaptor.getValue();
+        assertActivityGuids(saves, "BBB");
```
DwayneJengSage: I'm not sure how this test is workable in the new format. The purpose of this test appears to be 3 fold:

1. AAA isn't replaced by the expired task from DDB.
2. BBB isn't saved to DDB because it's expired.
3. CCC isn't added to scheduled (though that's always the case in V3).

It looks like case 1 is preserved with AAA. I'm not sure case 3 is meaningful, but it should be trivial to also preserve this test case.

The only questionable one is case 2. Does the scheduler ever return expired tasks? Or is this a facetious test case?

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -372,111 +380,154 @@ public void deleteActivitiesForSchedulePlanRejectsBadValue() {
...
+        SchedulePlan ccc = schedulePlan("CCC");
+        
+        when(schedulePlanService.getSchedulePlans(ClientInfo.UNKNOWN_CLIENT, TEST_STUDY)).thenReturn(Lists.newArrayList(aaa,bbb,ccc));
+        
+        List<ScheduledActivity> db = createStartedActivities("AAA"+TIME_PORTION,"CCC"+TIME_PORTION);
+        when(activityDao.getActivityHistoryV2(HEALTH_CODE, "BBB", NOW, NOW, TIME_ZONE, null, API_MAXIMUM_PAGE_SIZE))
+                .thenReturn(new ScheduledActivityList(db, null, 0));
+        
+        List<ScheduledActivity> returnedActivities = service.getScheduledActivitiesV4(createScheduleContext(NOW).build());
+        assertActivityGuids(returnedActivities, "AAA", "BBB", "CCC");
```
DwayneJengSage, line 529: This test is functionally identical to the existing test, but it doesn't seem to test that AAA and CCC were replaced by their started counterparts.

DwayneJengSage (changes requested):
Overall looks good! Just a few test notes, which hopefully should be easy to fix.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: Reduced to one (expired) task.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: Added tests that startedOn has been set for AAA and CCC

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: I'm confused by this comment. At these line numbers, AAA and CCC are expired, so we expect BBB to be saved, and also the only one returned. (It's true that the scheduler should not return expired tasks, so we're verifying that expired tasks that have been persisted aren't added into the set of results that are returned.)

alxdarksage: Changed test to do what the original test did, by asking for activities in the past that are expired.

DwayneJengSage (approved):

PR #1514, opened 2017-07-17
Adding a description of password requirements to the password reset page. 

alxdarksage:
We can add this description elsewhere if needed. We could tell them that the password has to be a maximum of 999 characters, but this makes for a very hostile techie message and I'm not aware that humans create passwords in the hundreds of characters. Or any password managers either.

### DIFF
--- a/app/org/sagebionetworks/bridge/BridgeUtils.java
+++ b/app/org/sagebionetworks/bridge/BridgeUtils.java
@@ -357,7 +357,7 @@ public class BridgeUtils {
                 phrases.add("one number");
             }
             if (policy.isSymbolRequired()) {
-                phrases.add("one symoblic character");
+                phrases.add("one symbolic character");
             }
             for (int i=0; i < phrases.size(); i++) {
                 if (i == phrases.size()-1) {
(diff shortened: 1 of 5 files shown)


app/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -338,5 +340,36 @@ public static String encodeURIComponent(String component) {
...
+                phrases.add("one lower-case letter");
+            }
+            if (policy.isUpperCaseRequired()) {
+                phrases.add("one upper-case letter");
+            }
+            if (policy.isNumericRequired()) {
+                phrases.add("one number");
+            }
+            if (policy.isSymbolRequired()) {
+                phrases.add("one symbolic character");
```
liujoshua: I'd suggest specifying the set of symbolic characters

alxdarksage: It's a very ugly list and not very user friendly to read. But maybe I can give examples or something.

liujoshua (approved):

PR #1522, opened 2017-07-25
add optimistic versioning to HibernateAccount

DwayneJengSage:
This fixes https://sagebionetworks.jira.com/browse/BRIDGE-1898

If you make two simultaneous updates to the same account, or consent to two different subpops simultaneously, it's very possible that you'll hit a race condition and the "second one wins".

This adds optimistic versioning to HibernateAccount, which is handled mostly transparently by Hibernate. This way, if you make two simultaneous edits, the second one will throw a 409, alerting the end user that something bad happened.

Testing done:
* updated and ran unit tests
* no functional changes, but ran integ tests as a sanity check
* manually tested concurrent modification by injecting a 5 second wait into update, then testing various combinations like simultaneous consents, simultaneous edits, edit with a password change, etc

### DIFF
--- a/app/org/sagebionetworks/bridge/exceptions/ConcurrentModificationException.java
+++ b/app/org/sagebionetworks/bridge/exceptions/ConcurrentModificationException.java
@@ -27,7 +27,7 @@ public class ConcurrentModificationException extends BridgeServiceException {
     }
     
     public Class<? extends BridgeEntity> getEntityClass() {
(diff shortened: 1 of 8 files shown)


alxdarksage (approved):

alxdarksage:
One you merge this into dev, I will need to make sure that you can still update users in the BSM. At the least, I will need to update the version number after saves.

PR #1525, opened 2017-07-27
Rationalize paging

alxdarksage:
- Backwards compatible rationalization of paging (integration tests pass before and after changes, using new version of Java SDK and old SDK);
- `ResourceList` classes pulled into a hierarchy, with request parameters now returned under a `requestParams` property; accessors directly on the list object are now all deprecated;
- `ForwardCursorPagedResourceList` includes the `nextPageOffsetKey` necessary to retrieve the next page of results;
- `PagedResourceList` includes the `total` property indicating the total number of records across all pages (this property is deprecated and will be removed elsewhere in other list classes);
- `ScheduleActivityList` removed (only consumers of this API can switch to `offsetKey`);
- Consistency in use of startTime/startDate parameters: startDate/endDate are used for LocalDate parameters, and startTime/endTime are used for DateTime parameters;
- `/v3/participants` and `/v3/studies/:studyId/participants` now accept startTime/endTime (as well as the older startDate/endDate) so client API can be changed over;


app/org/sagebionetworks/bridge/models/ResourceList.java
```diff
@@ -1,22 +1,80 @@
...
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.collect.ImmutableMap;
 
-public final class ResourceList<T> {
+/**
+ * Basic list of items, not paged, as calculated based on the parameters that were 
+ * sent to the server and are included in the <code>ResourceList</code>.
+ */
+public class ResourceList<T> {
```
DwayneJengSage, line 21: It looks like unit tests only test serialization. We should add unit tests (for withRequestParams(), getTotal(), getDateTime(), and getLocalDate()) to make sure everything is covered.

app/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceList.java
```diff
@@ -1,88 +1,75 @@
...
+    @Deprecated
+    public String getOffsetKey() {
+        return nextPageOffsetKey;
     }
-    public int getPageSize() {
-        return pageSize;
+    @Deprecated
+    public Integer getPageSize() {
+        Object value = getRequestParams().get(PAGE_SIZE);
+        return (value == null) ? null : (int)value;
```
DwayneJengSage: Could you cast this to a capital I Integer? This way, you don't need the redundant null check.

app/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceList.java
```diff
@@ -1,88 +1,75 @@
...
+        return nextPageOffsetKey;
     }
-    @JsonAnySetter
-    private void setFilter(String key, String value) {
-        if (isNotBlank(key) && isNotBlank(value)) {
-            filters.put(key, value);    
-        }
+    @JsonProperty(HAS_NEXT)
+    public boolean hasNext() {
+        return (nextPageOffsetKey != null);
```
DwayneJengSage: Unit test for hasNext() (both cases).

app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDao.java
```diff
@@ -140,22 +141,21 @@ public final void setMapper(DynamoDBMapper mapper) {
...
-                .withFilter(ID_FILTER, idFilter);
-        if (assignmentFilter != null) {
-            resourceList = resourceList.withFilter(ASSIGNMENT_FILTER, assignmentFilter.toString());
-        }
-        return resourceList;
+        } while ((identifiers.size() < pageSize) && (nextPageOffsetKey != null));
+
+        return new ForwardCursorPagedResourceList<>(
+                identifiers, nextPageOffsetKey)
+                .withRequestParam(ResourceList.OFFSET_KEY, offsetKey)
```
DwayneJengSage, line 155: It looks like the goal here is to preserve the original offsetKey in the requestParams (which may be different from nextOffsetKey). We should test for this in the paging tests in DynamoIdDaoTest. (canRetrieveCurrentAndNextPage() and pagingWithNoRemainderWorks())

Similarly in a bunch of other Daos where we add offsetKey in addition to nextOffsetKey.

app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDao.java
```diff
@@ -140,22 +141,21 @@ public final void setMapper(DynamoDBMapper mapper) {
...
+                .withRequestParam(ResourceList.PAGE_SIZE, pageSize)
+                .withRequestParam(ResourceList.ID_FILTER, idFilter)
+                .withRequestParam(ResourceList.ASSIGNMENT_FILTER, assignmentFilter);
```
DwayneJengSage, line 158: We should add unit tests for each Dao to make sure we're correctly adding request params into our ResourceLists. This will ensure that we aren't accidentally dropping any values and making breaking changes.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
```diff
@@ -201,9 +204,11 @@ public Upload getUpload(@Nonnull String uploadId) {
...
         String nextPageOffsetKey = (page.getLastEvaluatedKey() != null) ? page.getLastEvaluatedKey().get(UPLOAD_ID).getS() : null;
         
-        return new ForwardCursorPagedResourceList<>(uploadList, nextPageOffsetKey, pageSize)
-                .withFilter("startDate", startTime.toString())
-                .withFilter("endDate", endTime.toString());
+        return new ForwardCursorPagedResourceList<>(uploadList, nextPageOffsetKey)
+                .withRequestParam(ResourceList.OFFSET_KEY, offsetKey)
+                .withRequestParam(ResourceList.PAGE_SIZE, pageSize)
+                .withRequestParam(ResourceList.START_TIME, startTime)
+                .withRequestParam(ResourceList.END_TIME, endTime);
```
DwayneJengSage, line 211: There are a few cases where we the old code had "start/endDate" even though the values were logically "start/endTimes". Changing the name would be a breaking change. Should we also add the filters as START_DATE even though that's logically incorrect? (Here and in a few other Daos as well.)

app/org/sagebionetworks/bridge/models/DateRangeResourceList.java
```diff
@@ -7,30 +7,32 @@
...
+        return getLocalDate(END_DATE);
+    }
+    @Override
+    @Deprecated
+    public Integer getTotal() {
+        // This is necessary solely to keep current integration tests passing. 
+        // The total property is going away on everything except PagedResourceList.
+        Integer total = super.getTotal();
+        return (total == null) ? 0 : total;
+    }
```
DwayneJengSage, line 32: Add unit test for getTotal() (both cases).

app/org/sagebionetworks/bridge/models/DateTimeRangeResourceList.java
```diff
@@ -10,31 +10,24 @@
...
     public DateTime getStartTime() {
-        return startTime;
+        return getDateTime(START_TIME);
     }
+    @Deprecated
     @JsonSerialize(using = DateTimeSerializer.class)
     public DateTime getEndTime() {
-        return endTime;
+        return getDateTime(END_TIME);
+    }
```
DwayneJengSage, line 28: Add unit tests for getStart/EndTime()

app/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceList.java
```diff
@@ -1,88 +1,75 @@
...
+    public DateTime getScheduledOnEnd() {
+        return getDateTime(SCHEDULED_ON_END);
     }
-    @JsonProperty("hasNext")
-    public boolean hasNext() {
-        return (offsetKey != null);
+    @Deprecated
+    public String getOffsetKey() {
+        return nextPageOffsetKey;
     }
```
DwayneJengSage: Should this return nextPageOffsetKey or the offsetKey request param? Looking at usage in the Daos, it seems like they may be different.

app/org/sagebionetworks/bridge/models/PagedResourceList.java
```diff
@@ -1,108 +1,61 @@
...
-    /**
-     * Convenience method for adding the DDB key as a filter. The key must be returned to retrieve 
-     * the next page of DDB records.
-     */
-    public PagedResourceList<T> withOffsetKey(String offsetKey) {
-        return withFilter(OFFSET_KEY_FILTER, offsetKey);
+    @Deprecated
+    public DateTime getEndTime() {
+        return getDateTime(END_TIME);
     }
```
DwayneJengSage: Unit test for start/endTime?

test/org/sagebionetworks/bridge/models/PagedResourceListTest.java
```diff
@@ -52,26 +66,34 @@ public void canSerialize() throws Exception {
                 new TypeReference<PagedResourceList<AccountSummary>>() {});
 
         assertEquals(page.getTotal(), serPage.getTotal());
-        assertEquals(page.getOffsetBy(), serPage.getOffsetBy());
-        assertEquals(page.getPageSize(), serPage.getPageSize());
-        assertEquals(page.getFilters().get("emailFilter"), serPage.getFilters().get("emailFilter"));
+        assertEquals(100, page.getPageSize());
+        assertEquals((Integer)123, page.getOffsetBy());
+        assertEquals("filterString", page.getEmailFilter());
```
DwayneJengSage: The other asserts in this block test the serialized page, but this block tests the original page. Is this intentional?

app/org/sagebionetworks/bridge/models/ReportTypeResourceList.java
```diff
@@ -7,23 +7,19 @@
...
-    public int getTotal() {
-        return items.size();
+    public ReportTypeResourceList(@JsonProperty(ITEMS) List<T> items) {
+        super(items);
     }
+    @Deprecated
     public ReportType getReportType() {
-        return reportType;
+        return (ReportType)getRequestParams().get(REPORT_TYPE);
+    }
```
DwayneJengSage, line 19: Unit test for getReportType?

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
```diff
@@ -95,34 +101,55 @@ public Result updateSelfParticipant() throws Exception {
...
+        // Similarly, we will return startTime/endTime in the top-level request parameter properties as 
+        // startDate/endDate while transitioning, to maintain backwards compatibility.
+        ObjectNode node = (ObjectNode)MAPPER.valueToTree(page);
+        Map<String,Object> rp = page.getRequestParams();
+        if (rp.get(START_TIME) != null) {
+            node.put(START_DATE, (String)rp.get(START_TIME));    
+        }
+        if (rp.get(END_TIME) != null) {
+            node.put(END_DATE, (String)rp.get(END_TIME));    
+        }
```
DwayneJengSage, line 151: Would it be cleaner to add this logic in the Controller where it's being returned? Or in the Dao where it's being created? Adding it in the Dao ensures we have both request params everywhere we need it, and we don't have to muck with raw JSON.

Either way, we need a unit test in whichever class handles it to make sure we're doing this.

app/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -289,8 +293,11 @@ public Upload getUpload(@Nonnull String uploadId) {
             return builder.build();
         }).collect(Collectors.toList());
         
-        return new ForwardCursorPagedResourceList<UploadView>(views, list.getOffsetKey(), list.getPageSize())
-                .withFilter("startTime", startTime).withFilter("endTime", endTime);
+        return new ForwardCursorPagedResourceList<UploadView>(views, list.getNextPageOffsetKey())
+                .withRequestParam(OFFSET_KEY, list.getRequestParams().get(OFFSET_KEY))
+                .withRequestParam(PAGE_SIZE, list.getRequestParams().get(PAGE_SIZE))
+                .withRequestParam(START_TIME, startTime)
+                .withRequestParam(END_TIME, endTime);
```
DwayneJengSage: Update UploadServiceMockTest.validateUploadMocks() to make sure we're not accidentally losing these parameters?

test/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceListTest.java
```diff
@@ -29,15 +31,39 @@ public void canSerialize() throws Exception {
...
-        assertEquals(2, node.get("offsetKey").asInt());
+        
+        assertEquals("nextOffsetKey", node.get("offsetKey").asText());
+        assertEquals("nextOffsetKey", node.get("nextPageOffsetKey").asText());
+        assertEquals(startTime.toString(), node.get("startTime").asText());
+        assertEquals(endTime.toString(), node.get("endTime").asText());
+        assertEquals(startTime.toString(), node.get("scheduledOnStart").asText());
+        assertEquals(endTime.toString(), node.get("scheduledOnEnd").asText());
         assertEquals(100, node.get("pageSize").asInt());
-        assertEquals("filterString", node.get("emailFilter").asText());
```
DwayneJengSage: emailFilter accidentally deleted?

test/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceListTest.java
```diff
@@ -29,15 +31,39 @@ public void canSerialize() throws Exception {
...
-        assertTrue(node.get("hasNext").asBoolean());
+        assertEquals(2, node.get("total").asInt());
```
DwayneJengSage: asInt() will automatically parse strings into ints. This means string "3" will pass, even though it arguably shouldn't. If we want to validate both value and type, we should use intValue() instead.

Likewise for asBoolean() and, to a lesser extent, asText().

test/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceListTest.java
```diff
@@ -29,15 +31,39 @@ public void canSerialize() throws Exception {
...
         assertEquals("ForwardCursorPagedResourceList", node.get("type").asText());
+        assertTrue(node.get("hasNext").asBoolean());
+        
+        JsonNode rp = node.get("requestParams");
+        assertEquals(startTime.toString(), rp.get("startTime").asText());
+        assertEquals(endTime.toString(), rp.get("endTime").asText());
+        assertEquals(startTime.toString(), rp.get("scheduledOnStart").asText());
+        assertEquals(endTime.toString(), rp.get("scheduledOnEnd").asText());
+        assertEquals(100, rp.get("pageSize").asInt());
+        assertEquals("offsetKey", rp.get("offsetKey").asText());
```
DwayneJengSage: What about emailFilter?

test/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceListTest.java
```diff
@@ -52,11 +78,23 @@ public void canSerialize() throws Exception {
...
+        assertEquals((Integer)100, serPage.getPageSize());
+        assertEquals((Integer)2, serPage.getTotal());
+        
+        Map<String,Object> params = serPage.getRequestParams();
+        assertEquals(startTime.toString(), params.get("startTime"));
+        assertEquals(endTime.toString(), params.get("endTime"));
+        assertEquals(startTime.toString(), params.get("scheduledOnStart"));
+        assertEquals(endTime.toString(), params.get("scheduledOnEnd"));
+        assertEquals(100, params.get("pageSize"));
+        assertEquals("offsetKey", params.get("offsetKey"));
```
DwayneJengSage: What about emailFilter?

test/org/sagebionetworks/bridge/models/PagedResourceListTest.java
```diff
@@ -20,24 +21,37 @@
...
+                .withRequestParam("emailFilter", "filterString");
         
         JsonNode node = BridgeObjectMapper.get().valueToTree(page);
-        assertEquals(2, node.get("offsetBy").asInt());
-        assertEquals(123, node.get("total").asInt());
+        assertEquals(123, node.get("offsetBy").asInt());
+        assertEquals(2, node.get("total").asInt());
         assertEquals(100, node.get("pageSize").asInt());
         assertEquals("filterString", node.get("emailFilter").asText());
         assertEquals("PagedResourceList", node.get("type").asText());
```
DwayneJengSage: What about start/endTime? (Here and everywhere else in this test.)

test/org/sagebionetworks/bridge/models/PagedResourceListTest.java
```diff
@@ -52,26 +66,34 @@ public void canSerialize() throws Exception {
...
     
     @Test
     public void offsetByCanBeNull() throws Exception {
         List<AccountSummary> accounts = Lists.newArrayListWithCapacity(2);
-        PagedResourceList<AccountSummary> page = new PagedResourceList<AccountSummary>(accounts, null, 100, 123)
-                .withFilter("emailFilter", "filterString");
+        PagedResourceList<AccountSummary> page = new PagedResourceList<AccountSummary>(accounts, null)
+                .withRequestParam("offsetBy", 123)
+                .withRequestParam("pageSize", 100)
+                .withRequestParam("emailFilter", "filterString");
```
DwayneJengSage: The purpose of this test was to test that offset can be null. While changing this, you made it so that total is null. Is that intentional? Is this test still worth keeping?

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
```diff
@@ -196,18 +199,43 @@ public void before() throws Exception {
...
+        verify(mockParticipantService).getPagedAccountSummaries(eq(study), eq(10), eq(20), eq("foo"),
+                startTimeCaptor.capture(), endTimeCaptor.capture());
+        assertEquals(start.toString(), startTimeCaptor.getValue().toString());
+        assertEquals(end.toString(), endTimeCaptor.getValue().toString());
+    }
+    
+    @Test
+    public void getParticipantsWithStartTimeEndTime() throws Exception {
+        DateTime start = DateTime.now();
+        DateTime end = DateTime.now();
```
DwayneJengSage: Chances are, start and end will be the same. Would it be better to define end as start.plusMinutes(1) or something like that?  (Similarly for getParticipantsForWorker().)

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/models/ResourceList.java
alxdarksage: Added tests

app/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceList.java
alxdarksage: It should return the next page offset key, in every construction of the FCPRL, it is the offset for the next page. Of all the properties though, this is one is definitely being used by clients and they will need to be migrated to nextPageOffsetKey.

app/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceList.java
alxdarksage: Fixed

app/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceList.java
alxdarksage: added

app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDao.java
alxdarksage: Went through, found all the places where we're creating ForwardCursorPagedResourceLists, and ensured there are tests for 1) offsetKey/nextPageOffsetKey being set correctly; 2) all query parameters being returned in the map.

app/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDao.java
alxdarksage: Went through, found all the places where we're creating ForwardCursorPagedResourceLists, and ensured there are tests for 1) offsetKey/nextPageOffsetKey being set correctly; 2) all query parameters being returned in the map.

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
alxdarksage: The filters are new, so no, no need to use the old names here. Migrating requires changing the parameters used (both old and new are supported by the controllers).

app/org/sagebionetworks/bridge/models/DateRangeResourceList.java
alxdarksage: added

app/org/sagebionetworks/bridge/models/DateTimeRangeResourceList.java
alxdarksage: These are tested in serialization... added tests that these are retrievable from the deprecated getters in the object.

app/org/sagebionetworks/bridge/models/PagedResourceList.java
alxdarksage: Added (and found it needed our deserializer to persist the time zone!)

test/org/sagebionetworks/bridge/models/PagedResourceListTest.java
alxdarksage: This test class was very redundant. I removed the same tests from a second test that is specifically testing that you can provide a null offsetBy, and removed a final test of this class when returned from DDB-based calls... this wasn't adding anything.

test/org/sagebionetworks/bridge/models/PagedResourceListTest.java
alxdarksage: This is wrong, fixed.

test/org/sagebionetworks/bridge/models/PagedResourceListTest.java
alxdarksage: Hmm. After redesign, total should never be null, but has to stay an Integer to be compatible with its use as a deprecated value. Added Preconditions.checkNotNull to the constructor and a test to verify it.

app/org/sagebionetworks/bridge/models/ReportTypeResourceList.java
alxdarksage: This is in ReportTypeResourceListTest

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
alxdarksage: I feel like these older values are deprecated, so when we remove startDate/endDate from the routes/controller, it'll be easier to find and remove this code. If you feel strongly about it, I can move it (this previously said can't... which was weird. I meant can).

app/org/sagebionetworks/bridge/services/UploadService.java
alxdarksage: I changed this so that the map (whatever it is) is copied entirely, so we don't have to track and test manually copying.

test/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceListTest.java
alxdarksage: Added back. At some point in this process I realized I needed to keep testing deprecated properties, I missed this one.

test/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceListTest.java
alxdarksage: I changed all these in this test.

test/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceListTest.java
alxdarksage: Added (it's there in the map)

test/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceListTest.java
alxdarksage: Added

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
alxdarksage: Made them different.

app/org/sagebionetworks/bridge/models/ResourceList.java
```diff
@@ -1,22 +1,80 @@
...
     public List<T> getItems() {
         return items;
     }
-    public int getTotal() {
-        return items.size();
+    public Map<String, Object> getRequestParams() {
+        return ImmutableMap.copyOf(requestParams);
+    }
+    public ResourceList<T> withRequestParam(String key, Object value) {
+        if (isNotBlank(key) && value != null) {
```
DwayneJengSage, line 56: Need test for null key and null value.

test/org/sagebionetworks/bridge/models/ResourceListTest.java
```diff
@@ -0,0 +1,109 @@
...
+    @Test
+    public void getDateTime() {
+        DateTime dateTime = DateTime.now();
+        
+        ResourceList<String> list = makeResourceList();
+        list.withRequestParam("dateTime1", dateTime);
+        list.withRequestParam("dateTime2", dateTime.toString());
+        
+        assertTrue(dateTime.isEqual(list.getDateTime("dateTime1")));
+        assertTrue(dateTime.isEqual(list.getDateTime("dateTime2")));
```
DwayneJengSage: isEqual() specifically ignores timestamp. This means we aren't testing the logic we use to store DateTimes. Also, we should test this using a non-default timezone to make sure we're correctly preserving timezone.

test/org/sagebionetworks/bridge/models/ResourceListTest.java
```diff
@@ -0,0 +1,109 @@
...
+    
+    @SuppressWarnings("deprecation")
+    @Test
+    public void getTotal() {
+        ResourceList<String> list = makeResourceList();
+        
+        list.withRequestParam("total", 3);
+        
+        assertEquals(3, list.getRequestParams().get("total"));
+        assertEquals((Integer)3, list.getTotal());
```
DwayneJengSage: Need test case for null or empty list.

Also, to avoid unnecessary confusion in the tests, we should skip adding "total" to the request params, since we never use that value anyway.

test/org/sagebionetworks/bridge/models/ResourceListTest.java
DwayneJengSage: Need test case for null DateTime.

test/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceListTest.java
```diff
@@ -52,11 +79,25 @@ public void canSerialize() throws Exception {
...
+        
+        assertEquals("nextOffsetKey", serPage.getNextPageOffsetKey());
+        assertEquals("nextOffsetKey", serPage.getOffsetKey());
+        assertEquals(startTime, serPage.getStartTime());
+        assertEquals(endTime, serPage.getEndTime());
+        assertEquals(startTime, serPage.getScheduledOnStart());
+        assertEquals(endTime, serPage.getScheduledOnEnd());
+        assertEquals((Integer)100, serPage.getPageSize());
+        assertEquals((Integer)2, serPage.getTotal());
+        assertTrue(serPage.hasNext());
```
DwayneJengSage: Can we separate the hasNext() test into a separate test? Combining this with the serialization test makes it harder to find and makes it conceptually more complex.

test/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDaoTest.java
```diff
@@ -304,11 +310,11 @@ public void pagingWithNoRemainderWorks() {
...
             ForwardCursorPagedResourceList<ExternalIdentifierInfo> page = dao.getExternalIds(studyId, null, 3, null, null);
             assertEquals(3, page.getItems().size());
-            assertNotNull(page.getOffsetKey());
+            assertNotNull(page.getNextPageOffsetKey());
             
-            page = dao.getExternalIds(studyId, page.getOffsetKey(), 3, null, null);
+            page = dao.getExternalIds(studyId, page.getNextPageOffsetKey(), 3, null, null);
             assertEquals(3, page.getItems().size());
-            assertNull(page.getOffsetKey());
+            assertNull(page.getNextPageOffsetKey());
```
DwayneJengSage: Should check that the original offsetKey was preserved and returned in the requestParams (both calls).

app/org/sagebionetworks/bridge/dynamodb/DynamoReportIndexDao.java
```diff
@@ -127,7 +128,8 @@ public void updateIndex(ReportIndex index) {
         DynamoDBQueryExpression<DynamoReportIndex> query =
                 new DynamoDBQueryExpression<DynamoReportIndex>().withHashKeyValues(hashKey);
 
-        return new ReportTypeResourceList<>(mapper.query(DynamoReportIndex.class, query), reportType);
+        return new ReportTypeResourceList<>(mapper.query(DynamoReportIndex.class, query))
+                .withRequestParam(ResourceList.REPORT_TYPE, reportType);
```
DwayneJengSage, line 132: Add unit tests for getReportType() and getRequestParam().get("reportType")

test/org/sagebionetworks/bridge/models/DateRangeResourceListTest.java
```diff
@@ -13,30 +15,38 @@
...
         
         list = BridgeObjectMapper.get().readValue(node.toString(), 
                 new TypeReference<DateRangeResourceList<String>>() {});
-        assertEquals(list.getStartDate(), LocalDate.parse("2016-02-03"));
-        assertEquals(list.getEndDate(), LocalDate.parse("2016-02-23"));
-        assertEquals(3, list.getTotal());
+        assertEquals(LocalDate.parse("2016-02-03"), list.getStartDate());
+        assertEquals(LocalDate.parse("2016-02-23"), list.getEndDate());
+        assertEquals(3, list.getItems().size());
+        assertEquals((Integer)3, list.getTotal());
```
DwayneJengSage: Similarly, can we move the getTotal() test to a separate test? Also, we need to test the null and empty list cases.

test/org/sagebionetworks/bridge/models/PagedResourceListTest.java
```diff
@@ -52,62 +69,26 @@ public void canSerialize() throws Exception {
...
+        assertEquals(endTime, serPage.getEndTime());
+        assertEquals("filterString", serPage.getEmailFilter());
+        
+        Map<String,Object> params = page.getRequestParams();
+        Map<String,Object> serParams = serPage.getRequestParams();
+        assertEquals(params.get("offsetBy"), serParams.get("offsetBy"));
+        assertEquals(params.get("pageSize"), serParams.get("pageSize"));
+        assertEquals(params.get("emailFilter"), serParams.get("emailFilter"));
+        assertEquals(params.get("startTime"), serParams.get("startTime"));
+        assertEquals(params.get("endTime"), serParams.get("endTime"));
```
DwayneJengSage: Can we test that the maps themselves are equal? If not, why are they different?

test/org/sagebionetworks/bridge/models/PagedResourceListTest.java
```diff
@@ -20,24 +20,41 @@
...
+                .withRequestParam("offsetBy", 123)
+                .withRequestParam("pageSize", 100)
+                .withRequestParam("startTime", startTime)
+                .withRequestParam("endTime", endTime)
+                .withRequestParam("emailFilter", "filterString");
         
         JsonNode node = BridgeObjectMapper.get().valueToTree(page);
-        assertEquals(2, node.get("offsetBy").asInt());
-        assertEquals(123, node.get("total").asInt());
+        assertEquals(123, node.get("offsetBy").asInt());
```
DwayneJengSage: asInt() automatically parses strings to ints. This means that string value "2" passes the test as does int value 2. This is probably not what we went. Recommend using intValue() instead. (And similarly booleanValue(). textValue() has the same issue, although obviously it's less of a problem.)

test/org/sagebionetworks/bridge/models/PagedResourceListTest.java
```diff
@@ -52,62 +69,26 @@ public void canSerialize() throws Exception {
...
         
         assertEquals(page.getItems(), serPage.getItems());
     }
     
-    @Test
+    @Test(expected = NullPointerException.class)
     public void offsetByCanBeNull() throws Exception {
         List<AccountSummary> accounts = Lists.newArrayListWithCapacity(2);
-        PagedResourceList<AccountSummary> page = new PagedResourceList<AccountSummary>(accounts, null, 100, 123)
-                .withFilter("emailFilter", "filterString");
```
DwayneJengSage: It looks like you removed the test for offsetByCanBeNull(). Is this intentional? Do we still need the test?

Also, the new test has the wrong name.

app/org/sagebionetworks/bridge/models/ReportTypeResourceList.java
DwayneJengSage, line 19: I'm still not seeing a unit test for getReportType().

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
```diff
@@ -735,18 +764,48 @@ public void getParticipantsForWorker() throws Exception {
...
+        assertEquals(start.toString(), startTimeCaptor.getValue().toString());
+        assertEquals(end.toString(), endTimeCaptor.getValue().toString());
+    }
+    
+    @Test
+    public void getParticipantsForWorkerUsingStartTimeEndTime() throws Exception {
+        session.setParticipant(new StudyParticipant.Builder().copyOf(session.getParticipant())
+                .withRoles(Sets.newHashSet(Roles.WORKER)).build());
+        DateTime start = DateTime.now();
+        DateTime end = DateTime.now();
```
DwayneJengSage: We also have the same start and end time here.

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
DwayneJengSage, line 151: Need unit tests to ensure that we're adding all the correct request params for both the old names and the new names.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/models/ResourceList.java
alxdarksage: added

test/org/sagebionetworks/bridge/models/ResourceListTest.java
alxdarksage: Added test for zone similarity (I don't know of a one-line method to test this, except maybe to compare the string formatted versions of the DateTime, but I added a TestUtils assertion for this.

alxdarksage: added, also for local dates

test/org/sagebionetworks/bridge/models/ResourceListTest.java
alxdarksage: Throw an NPE if the list is null (should never create this); total is null if list is empty.

test/org/sagebionetworks/bridge/models/ForwardCursorPagedResourceListTest.java
alxdarksage: Moved to a separate test, deleted the offsetKeyCanBeNull test as it was entirely redundant with the new test.

test/org/sagebionetworks/bridge/dynamodb/DynamoExternalIdDaoTest.java
alxdarksage: added

app/org/sagebionetworks/bridge/dynamodb/DynamoReportIndexDao.java
alxdarksage: added

test/org/sagebionetworks/bridge/models/DateRangeResourceListTest.java
alxdarksage: added

test/org/sagebionetworks/bridge/models/PagedResourceListTest.java
alxdarksage: changed all these to intValue. There are undoubtedly a ton of these in the test as I didn't know about these non-type-coercing methods.

test/org/sagebionetworks/bridge/models/PagedResourceListTest.java
alxdarksage: yeah, we totally can. Changed this.

test/org/sagebionetworks/bridge/models/PagedResourceListTest.java
alxdarksage: offsetBy is now only a request parameter (there's no reason to calculate nextPageOffsetBy, though we could). Renamed the test as we're now verifying that total, which is required, will throw an error immediately if not provided.

app/org/sagebionetworks/bridge/models/ReportTypeResourceList.java
alxdarksage: I think I see what you're looking for. Added in ReportTypeResourceListTest line 33 after commit, there are tests below of the serialization of this list to verify it is there in the JSON.

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
alxdarksage: made them different

app/org/sagebionetworks/bridge/play/controllers/ParticipantController.java
alxdarksage: added

app/org/sagebionetworks/bridge/models/ResourceList.java
```diff
@@ -1,22 +1,82 @@
...
+            } else {
+                requestParams.put(key, value);    
+            }
+        }
+        return this;
+    }
+    @Deprecated
+    public Integer getTotal() {
         return items.size();
     }
```
DwayneJengSage: We return items.size() for an empty list, but the test expects null for an empty list. Can you clarify?

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
```diff
@@ -98,9 +98,13 @@
     
     private static final String ACTIVITY_GUID = "activityGuid";
 
-    private static final DateTime ENDS_ON = DateTime.now();
+    //private static final DateTime ENDS_ON = DateTime.now();
     
-    private static final DateTime STARTS_ON = ENDS_ON.minusWeeks(1);
+    //private static final DateTime STARTS_ON = ENDS_ON.minusWeeks(1);
```
DwayneJengSage: Minor: Remove commented code.

DwayneJengSage (approved):
Two minor comments. Otherwise, looks good.

app/org/sagebionetworks/bridge/models/ResourceList.java
alxdarksage: this returns null for zero again

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
alxdarksage: removed

alxdarksage:
Verified the integration tests pass both before and after updating them to the new API presented here.

DwayneJengSage (approved):

PR #1526, opened 2017-07-28
Code to migrate between Redis servers

alxdarksage:
Verified against a stand-alone Redis cloud instance, but these will need to be set up before we merge for each environment.


app/org/sagebionetworks/bridge/cache/CacheProvider.java
```diff
@@ -118,11 +128,11 @@ public void setUserSession(final UserSession session) {
         final String userKey = RedisKey.USER_SESSION.getRedisKey(userId);
         
         final String sessionKey = RedisKey.SESSION.getRedisKey(sessionToken);
-        try (JedisTransaction transaction = jedisOps.getTransaction()) {
+        try (JedisTransaction transaction = newJedisOps.getTransaction()) {
             
             // If the key exists, get the remaining time to expiration. If it doesn't exist
             // then save with the full expiration period.
-            final Long ttl = jedisOps.ttl(userKey);
+            final Long ttl = newJedisOps.ttl(userKey);
```
DwayneJengSage, line 135: Is it worth adding a mock call to the unit test to make sure we're correctly getting and using the TTL from the newJedisOps?

test/org/sagebionetworks/bridge/cache/CacheProviderMigrationTest.java
```diff
@@ -0,0 +1,387 @@
...
+        assertEquals("userId", captured.getId());
+    }
+            
+    private UserSession mockSessionAndTransaction(JedisOps thisJedisOps) {
+        UserSession session = stubSession("userId");
+        
+        when(thisJedisOps.getTransaction()).thenReturn(transaction);
+        doReturn(transaction).when(transaction).setex(any(), anyInt(), any());
+        
+        when(thisJedisOps.getTransaction()).thenReturn(transaction);
```
DwayneJengSage: Duplicate call to when(thisJedisOps.getTransaction()).thenReturn(transaction)?

test/org/sagebionetworks/bridge/cache/CacheProviderMigrationTest.java
```diff
@@ -0,0 +1,387 @@
...
+
+    @Test
+    public void removeSession() {
+        mockSessionAndTransaction(newJedisOps);
+        UserSession session = mockSessionAndTransaction(oldJedisOps);
+        
+        cacheProvider.removeSession(session);
+        
+        verify(transaction, times(2)).del("sessionToken:session");
+        verify(transaction, times(2)).del("userId:session:user");
```
DwayneJengSage: It would probably be cleaner to have the new and old Jedises return different transactions. This way, we can verify that we delete from both, rather than deleting from one Jedis twice.

test/org/sagebionetworks/bridge/cache/CacheProviderMigrationTest.java
```diff
@@ -0,0 +1,387 @@
...
+        mockSessionAndTransaction(newJedisOps);
+        mockSessionAndTransaction(oldJedisOps);
+        
+        when(newJedisOps.get("userId:session:user")).thenReturn("sessionToken");
+        
+        cacheProvider.removeSessionByUserId("userId");
+        
+        verify(transaction, times(2)).del("sessionToken:session");
+        verify(transaction, times(2)).del("userId:session:user");
+    }
```
DwayneJengSage: Is it worth having all 3 cases, no session, session in old cache, session in new cache?

app/org/sagebionetworks/bridge/cache/CacheProvider.java
```diff
@@ -195,10 +208,13 @@ public void removeSession(final UserSession session) {
...
             final String userKey = RedisKey.USER_SESSION.getRedisKey(userId);
-            final String sessionToken = jedisOps.get(userKey);
+            final String sessionToken = getWithFallback(userKey, false);
             if (sessionToken != null) {
                 final String sessionKey = RedisKey.SESSION.getRedisKey(sessionToken);
-                try (JedisTransaction transaction = jedisOps.getTransaction()) {
+                try (JedisTransaction transaction = oldJedisOps.getTransaction()) {
+                    transaction.del(sessionKey).del(userKey).exec();
+                }
+                try (JedisTransaction transaction = newJedisOps.getTransaction()) {
```
DwayneJengSage: Would it be better to have two independent if blocks, if there's a session in new, delete new, if there's a session in old, delete old? This cleanly separates the newJedisOps from oldJedisOps and probably makes it easier to reason about and test.

test/org/sagebionetworks/bridge/cache/CacheProviderMigrationTest.java
```diff
@@ -0,0 +1,387 @@
...
+        study.setIdentifier("studyId");
+        String ser = MAPPER.writeValueAsString(study);
+        
+        when(oldJedisOps.get("studyId:study")).thenReturn(ser);
+        
+        Study returned = cacheProvider.getStudy("studyId");
+        assertEquals(study, returned);
+        
+        verify(newJedisOps).get("studyId:study");
+        verify(oldJedisOps).get("studyId:study");
```
DwayneJengSage: Do we also want to test that we set the expiration? (Here and getStudyNewCache()?)

test/org/sagebionetworks/bridge/cache/CacheProviderMigrationTest.java
```diff
@@ -0,0 +1,387 @@
...
+        assertEquals(study, returned);
+        
+        verify(newJedisOps).get("studyId:study");
+        verify(oldJedisOps, never()).get("studyId:study");
+    }
+
+    @Test
+    public void removeStudy() {
+        mockSessionAndTransaction(newJedisOps);
+        mockSessionAndTransaction(oldJedisOps);
```
DwayneJengSage: This code path doesn't use transactions. Is this necessary? (Here and removeString().)

app/org/sagebionetworks/bridge/config/BridgeProductionSpringConfig.java
```diff
@@ -30,15 +30,32 @@
...
     @Bean(name = "jedisPool")
     public JedisPool jedisPool() throws Exception {
-        // Configure pool
+        return createJedisPool("original server", redisProviders);
+    }
+
+    @Bean(name = "newJedisPool")
+    public JedisPool newJedisPool(List<String> providers) throws Exception {
+        return createJedisPool("new server", newRedisProviders);
+    }
```
DwayneJengSage: Minor: Unused param "providers".

app/org/sagebionetworks/bridge/config/BridgeProductionSpringConfig.java
```diff
@@ -78,7 +94,7 @@ private JedisPool constructJedisPool(final String url, final JedisPoolConfig poo
         // With changes in Redis provisioning, passwords are now parseable by Java's URI class.
         URI redisURI = new URI(url);
         String password = redisURI.getUserInfo().split(":",2)[1];
-
+        
         if (bridgeConfig.isLocal() || password.equals("AWS")) {
             return new JedisPool(poolConfig, redisURI.getHost(), redisURI.getPort(),
                     bridgeConfig.getPropertyAsInt("redis.timeout"));
```
DwayneJengSage: As long as you're in here, you should clean up this logic, which will provide a fix for https://sagebionetworks.jira.com/browse/BRIDGE-1585 and remove the "AWS" hack. Should be pretty trivial to do.

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
```diff
@@ -101,6 +101,11 @@
         return Lists.newArrayList("REDISCLOUD_URL", "REDISTOGO_URL");
     }
     
+    @Bean(name = "newRedisProviders")
+    public List<String> newRedisProviders() {
+        return Lists.newArrayList(BridgeConfigFactory.getConfig().get("elasticache.url"));
+    }
```
DwayneJengSage: redisProviders() returns the env var names that we lookup to get the Redis URL. However, newRedisProviders() seems to be getting the value of elasticache.url. Is this intentional? Is that value supposed to be another env var name? Or a Redis URL?

Also, given that we have no plans to use REDISTOGO_URL anytime soon, is it worth getting rid of these altogether and just have fixed env var names? This might simplify the configs.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/cache/CacheProvider.java
alxdarksage: Added to setUserSession test

app/org/sagebionetworks/bridge/cache/CacheProvider.java
alxdarksage: changed

test/org/sagebionetworks/bridge/cache/CacheProviderMigrationTest.java
alxdarksage: Yes, removed

test/org/sagebionetworks/bridge/cache/CacheProviderMigrationTest.java
alxdarksage: Changed this, cleaned up the setup of the tests

test/org/sagebionetworks/bridge/cache/CacheProviderMigrationTest.java
alxdarksage: It doesn't really matter since it's harmless to call delete if there's no session. We just need to be sure we call delete on both Redis instances.

test/org/sagebionetworks/bridge/cache/CacheProviderMigrationTest.java
alxdarksage: Yes, added it to all three getStudy() tests

test/org/sagebionetworks/bridge/cache/CacheProviderMigrationTest.java
alxdarksage: Setting this up in the before call now

app/org/sagebionetworks/bridge/config/BridgeProductionSpringConfig.java
alxdarksage: Removed

app/org/sagebionetworks/bridge/config/BridgeProductionSpringConfig.java
alxdarksage: Take a look and see what you think. I took the parsing out to a utility method and added some tests.

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
alxdarksage: Cleaned this up, removing all the config rigamarole around falling back to Redis To Go.

DwayneJengSage (approved):

PR #1393, opened 2017-03-13
Instant export

zaclinsage:
new one-time export feature for bridge pf;
unit tests done;
manual tests done

See Also:
* https://github.com/Sage-Bionetworks/BridgeJavaSDK/pull/393
* https://github.com/Sage-Bionetworks/BridgeExporterIntegrationTests/pull/1
* https://github.com/Sage-Bionetworks/Bridge-Exporter/pull/52
* https://github.com/Sage-Bionetworks/BridgeDocs/pull/138
* https://github.com/Sage-Bionetworks/Bridge-EX-Scheduler/pull/6

PR #1534, opened 2017-08-10
Bridge API for On-Demand Export

DwayneJengSage:
Sends a message to SQS to trigger an on-demand export. Replaces https://github.com/Sage-Bionetworks/BridgePF/pull/1393

Testing done:
* added unit tests
* manually tested

### DIFF
--- a/app/org/sagebionetworks/bridge/services/ExportService.java
+++ b/app/org/sagebionetworks/bridge/services/ExportService.java
@@ -1,0 +1,11 @@
+package org.sagebionetworks.bridge.services;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+
+import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
+
+/** Service that interfaces Bridge Server with Bridge Exporter. */
+public interface ExportService {
+    /** Kicks off an on-demand export for the given study. */
+    void startOnDemandExport(StudyIdentifier studyId) throws JsonProcessingException;
+}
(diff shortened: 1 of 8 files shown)


app/org/sagebionetworks/bridge/services/ExportService.java
```diff
@@ -0,0 +1,11 @@
...
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+
+import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
+
+/** Service that interfaces Bridge Server with Bridge Exporter. */
+public interface ExportService {
+    /** Kicks off an on-demand export for the given study. */
+    void startOnDemandExport(StudyIdentifier studyId) throws JsonProcessingException;
+}
```
alxdarksage, line 11: Given there's only one service, do you want to have an interface? We got rid of them for other services.

alxdarksage (approved):
Approved but one comment about interface for a service with only one implementation.

app/org/sagebionetworks/bridge/services/ExportService.java
DwayneJengSage: I can see it both ways.

On the one hand, this has a direct dependency on SQS, which most of the other services don't. An interface helps us keep SQS-specific details from leaking into other classes. (See also SendEmailService, UserDataDownloadService.)

On the other hand, this service encapsulates a very simple SQS call that's very easy to replace. Contrast with the DAOs, which have a lot of complexity in them around mapping data types to DDB types and CRUDs.

alxdarksage: I only mention it because we removed interfaces for all the other services on the premise that until we have more than one implementation, we don't need the extra complexity.

PR #1540, opened 2017-08-24
Activity history API using object/referent of the scheduled activities

alxdarksage:
An API for user and researchers to see the scheduled activities of an account base on the object/referent of the activities, so for example, a list of all tapping tests assigned to the user, regardless of the schedule used to schedule the activity for the user.

API is forward-only, bound by a date range, and paginated. API does not accept survey identifier, only survey GUID for this first iteration (I will check that behavior in separately).


conf/routes
```diff
@@ -46,6 +46,7 @@ POST   /v3/participants/:userId/consents/withdraw            @org.sagebionetwork
 POST   /v3/participants/:userId/consents/:guid/resendConsent @org.sagebionetworks.bridge.play.controllers.ParticipantController.resendConsentAgreement(userId: String, guid: String)
 POST   /v3/participants/:userId/consents/:guid/withdraw      @org.sagebionetworks.bridge.play.controllers.ParticipantController.withdrawConsent(userId: String, guid: String)
 GET    /v3/participants/:userId/activityEvents               @org.sagebionetworks.bridge.play.controllers.ParticipantController.getActivityEvents(userId: String)
+GET    /v4/participants/:userId/activities/:referrentGuid    @org.sagebionetworks.bridge.play.controllers.ParticipantController.getActivityHistoryV3(userId: String, activityType: String, referrentGuid: String, scheduledOnStart: String ?= null, scheduledOnEnd: String ?= null, offsetKey: String ?= null, pageSize: String ?= null)
```
liujoshua: referrent has two Rs in the routes file

app/org/sagebionetworks/bridge/dynamodb/DynamoScheduledActivityDao.java
```diff
@@ -113,6 +129,126 @@ final void setDdbMapper(DynamoDBMapper mapper) {
...
+            final ActivityType activityType, final String referentGuid, final DateTime scheduledOnStart,
+            final DateTime scheduledOnEnd, final String offsetKey, final int pageSize) {
+        checkNotNull(healthCode);
+        checkNotNull(activityType);
+        checkNotNull(referentGuid);
+        checkNotNull(scheduledOnStart);
+        checkNotNull(scheduledOnEnd);
+
+        int pageSizeWithIndicator = pageSize+1;
+        String start = createReferentGuid(activityType, referentGuid, scheduledOnStart.toLocalDateTime().toString());
```
liujoshua: referentGuid here takes on two meanings. Could we call the param activityGuid and call type:guid:timestamp referentGuid?

liujoshua (approved):

PR #1541, opened 2017-08-29
Add clientData to the user's account object and study participant model.

alxdarksage:
NOTE: Existing tables need to be updated:
    `ALTER TABLE bridgepf.Accounts MODIFY clientData MEDIUMTEXT;`

This adds a large string to the account table for whatever arbitrary JSON the client wants to store on the server. It cherry picks the columns we need for AccountSummary so this doesn't effect the performance of listing accounts.

NOTE: I'll merge this when I'm in the office. I don't have access to the RDS passwords (no vpn at home) to update the dev DB before merging this.

### DIFF
--- a/app/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
+++ b/app/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
@@ -49,12 +49,12 @@ public class HibernateAccount {
     private int version;
     private String clientData;
 
-    HibernateAccount() {}
+    public HibernateAccount() {}
     
     /**
      * Constructor to load information for the AccountSummary object.
      */
-    HibernateAccount(Long createdOn, String studyId, String firstName, String lastName, String email, String id,
+    public HibernateAccount(Long createdOn, String studyId, String firstName, String lastName, String email, String id,
             AccountStatus status) {
         this.createdOn = createdOn;
         this.studyId = studyId;
--- a/app/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
+++ b/app/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
@@ -49,10 +49,15 @@ public class HibernateAccount {
     private int version;
     private String clientData;
 
+    /**
+     * No args constructor, required and used by Hibernate for full object initialization.
+     */
     public HibernateAccount() {}
     
     /**
-     * Constructor to load information for the AccountSummary object.
+     * Constructor to load information for the AccountSummary object. Could not find a way to 
+     * construct this object with just the indicated fields using a select clause, without also 
+     * specifying a constructor.
      */
     public HibernateAccount(Long createdOn, String studyId, String firstName, String lastName, String email, String id,
             AccountStatus status) {
--- a/app/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
+++ b/app/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
@@ -47,6 +47,23 @@ public class HibernateAccount {
     private Set<Roles> roles;
     private AccountStatus status;
     private int version;
+    private String clientData;
+
+    HibernateAccount() {}
+    
+    /**
+     * Constructor to load information for the AccountSummary object.
+     */
+    HibernateAccount(Long createdOn, String studyId, String firstName, String lastName, String email, String id,
+            AccountStatus status) {
+        this.createdOn = createdOn;
+        this.studyId = studyId;
+        this.firstName = firstName;
+        this.lastName = lastName;
+        this.email = email;
+        this.id = id;
+        this.status = status;
+    }
 
     /**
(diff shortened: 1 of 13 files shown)


app/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
```diff
@@ -47,6 +47,23 @@
...
     private int version;
+    private String clientData;
+
+    public HibernateAccount() {}
+    
+    /**
+     * Constructor to load information for the AccountSummary object.
+     */
+    public HibernateAccount(Long createdOn, String studyId, String firstName, String lastName, String email, String id,
+            AccountStatus status) {
```
DwayneJengSage: This constructor appears unused except for unit tests. Should we even have this constructor? How do we decide what fields should and shouldn't be included in the constructor?

I think it would be better off to use the default no-arg constructor and just use getters and setters. Otherwise, we end up in a situation where we need to update N^2 places every time we update the constructor.

app/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
```diff
@@ -250,7 +267,18 @@ public AccountStatus getStatus() {
     public void setStatus(AccountStatus status) {
         this.status = status;
     }
-
+    
+    /** @see #getClientData */
+    @Column(columnDefinition = "mediumtext", name = "clientData", nullable = true)
+    public String getClientData() {
```
DwayneJengSage, line 278: Is this a text or a mediumtext? The wiki says text, but the pull request and the annotations say mediumtext.

app/org/sagebionetworks/bridge/models/accounts/StudyParticipant.java
```diff
@@ -273,6 +283,9 @@ public Builder copyFieldsOf(StudyParticipant participant, Set<String> fieldNames
             if (fieldNames.contains("timeZone")) {
                 this.timeZone = participant.getTimeZone();    
             }
+            if (fieldNames.contains("clientData")) {
+                this.clientData = participant.getClientData();
+            }
```
DwayneJengSage: Is it worth adding a test for copyFieldsOf()?

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -185,6 +185,7 @@ public StudyParticipant getParticipant(Study study, Account account, boolean inc
         builder.withRoles(account.getRoles());
         builder.withId(account.getId());
         builder.withHealthCode(account.getHealthCode());
+        builder.withClientData(account.getClientData());
```
DwayneJengSage, line 188: Do we need to update ParticipantServiceTest.getStudyParticipantWithAccount()?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -395,6 +397,7 @@ public void getStudyParticipant() {
         List<ConsentSignature> sigs1 = Lists.newArrayList(new ConsentSignature.Builder()
                 .withName("Name 1").withBirthdate("1980-01-01").build());
         when(account.getConsentSignatureHistory(SubpopulationGuid.create("guid1"))).thenReturn(sigs1);
+        when(account.getClientData()).thenReturn(TestUtils.getClientData());
```
DwayneJengSage: For organizational purposes, this should probably be moved up to the rest of the mock statements.

DwayneJengSage (changes requested):
Mostly good. Just a few small comments.

app/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
alxdarksage: Added comment about the fact that it seems required for Hibernate to construct the object using only the fields specified in the select clause (avoiding joins and the potentially large clientData object).

app/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
alxdarksage: Updated bootstrap document (it's mediumtext)

app/org/sagebionetworks/bridge/models/accounts/StudyParticipant.java
alxdarksage: Added tests for each field in copyFieldsOf() to verify that only the indicated field is copied.

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Added

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Yes... moved alongside the other account mock statements.

DwayneJengSage (approved):

PR #1542, opened 2017-08-30
Synchronous Health Data API

DwayneJengSage:
Synchronous health data API. Used to submit small health data payloads (such as survey responses) without incurring the overhead of creating a bunch of small files to upload to S3.

See https://sagebionetworks.jira.com/wiki/spaces/BRIDGE/pages/97517599/Upload+Format+v2#UploadFormatv2-SynchronousHealthDataSubmission for more details. (Docs to be updated shortly to reflect implementation details.

### DIFF
--- a/app/org/sagebionetworks/bridge/models/healthdata/HealthDataSubmission.java
+++ b/app/org/sagebionetworks/bridge/models/healthdata/HealthDataSubmission.java
@@ -1,0 +1,120 @@
+package org.sagebionetworks.bridge.models.healthdata;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.joda.time.DateTime;
+
+import org.sagebionetworks.bridge.json.DateTimeDeserializer;
+import org.sagebionetworks.bridge.json.DateTimeSerializer;
+import org.sagebionetworks.bridge.models.BridgeEntity;
+
+/** Health data submitted by the user. */
+@JsonDeserialize(builder = HealthDataSubmission.Builder.class)
+public class HealthDataSubmission implements BridgeEntity {
+    private final String appVersion;
+    private final DateTime createdOn;
+    private final JsonNode data;
+    private final String phoneInfo;
+    private final String schemaId;
+    private final int schemaRevision;
+
+    /** Private constructor. To construct, use builder. */
+    private HealthDataSubmission(String appVersion, DateTime createdOn, JsonNode data, String phoneInfo,
+            String schemaId, int schemaRevision) {
+        this.appVersion = appVersion;
+        this.createdOn = createdOn;
+        this.data = data;
+        this.phoneInfo = phoneInfo;
+        this.schemaId = schemaId;
+        this.schemaRevision = schemaRevision;
+    }
+
+    /**
+     * App version, as reported by the app. Generally in the form "version 1.0.0, build 2". Must be 48 chars or less.
+     */
+    public String getAppVersion() {
+        return appVersion;
+    }
+
+    /** When the health data was created, as reported by the app. */
+    @JsonSerialize(using = DateTimeSerializer.class)
+    public DateTime getCreatedOn() {
+        return createdOn;
+    }
+
+    /** Health data, as key-value pairs corresponding to the schema fields. */
+    public JsonNode getData() {
+        return data;
+    }
+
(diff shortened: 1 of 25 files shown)


app/org/sagebionetworks/bridge/upload/UploadValidationContext.java
```diff
@@ -26,6 +27,16 @@
     private Map<String, byte[]> attachmentsByFieldName;
     private String recordId;
 
+    /** Health code of the user contributing the health data. */
+    public String getHealthCode() {
+        return healthCode;
```
liujoshua, line 32: DO you think it would be helpful to do something along the longs of:

1. return upload.getUploadId() if upload is not null, or
2. set healthCode in setUpload()

app/org/sagebionetworks/bridge/models/healthdata/HealthDataSubmission.java
```diff
@@ -0,0 +1,120 @@
...
+    private final String schemaId;
+    private final int schemaRevision;
+
+    /** Private constructor. To construct, use builder. */
+    private HealthDataSubmission(String appVersion, DateTime createdOn, JsonNode data, String phoneInfo,
+            String schemaId, int schemaRevision) {
+        this.appVersion = appVersion;
+        this.createdOn = createdOn;
+        this.data = data;
+        this.phoneInfo = phoneInfo;
```
alxdarksage, line 28: nomenclature: is this related to clientInfo? It's possible we'll have non-"phone" (native) clients, like mobile browsers, at some pointing the future.

app/org/sagebionetworks/bridge/services/HealthDataService.java
```diff
@@ -27,22 +51,200 @@
...
+        // ones that these handlers will be using.
+        UploadValidationContext uploadValidationContext = new UploadValidationContext();
+        uploadValidationContext.setAttachmentsByFieldName(attachmentMap);
+        uploadValidationContext.setHealthCode(participant.getHealthCode());
+        uploadValidationContext.setHealthDataRecord(record);
+        uploadValidationContext.setStudy(studyId);
+
+        // Strict Validation Handler. If this throws, this is an invalid upload (400).
+        try {
+            strictValidationHandler.handle(uploadValidationContext);
```
liujoshua, line 140: The sequence here of strictValidationHandler -> transcribeConsentHandler -> uploadArtifactsHandler is the same as the end of uploadValidationHandlerList. Would it make sense to create something like a healthDataValidationHandlerList to capture and reuse this handler sequence?

One challenger here is in this case, we are wanting to short circuit if strictValidationHandler fails

liujoshua (approved):
Looks good to me. Minor thoughts

app/org/sagebionetworks/bridge/models/healthdata/HealthDataSubmission.java
DwayneJengSage: It's actually not. The name comes from the original ResearchKit/AppCore uploads, where they had a field called phoneInfo in the info.json. This is being written to Synapse as phoneInfo.

We could try to merge phoneInfo and ClientInfo (and appVersion) together into a single unified construct, but at this point, it would be more confusing than not, and I think that level of change is outside the scope of this code change.

app/org/sagebionetworks/bridge/upload/UploadValidationContext.java
DwayneJengSage: I don't think so, for several reasons:

1. Extra logic we'd have to maintain and test. It would be better for a simple getter to remain a simple getter. getUploadId() exists primarily because it's a common enough construct that it's worth having rather than adding a bunch of null checks everywhere.
2. Health Code exists independently of Uploads. We were originally getting it from the Upload probably because we were too lazy to make it its own top-level field. In contrast, Upload ID is meaningful only when there's an Upload.

app/org/sagebionetworks/bridge/services/HealthDataService.java
DwayneJengSage: Personally, I think the configured handler list, while clever, was actually over-engineered. It would be a lot more useful if we had multiple upload paths and wanted to be able to hot swap handlers quickly. As it is, the 8 handlers have only changed once, and it was to create a new version of the IosSchemaValidationHandler. The Synchronous Upload API only has 3 steps (4 if we want to encode all the prep work as a handler), so I don't think it needs something as sophisticated as UploadValidationTask.

If we do need something more sophisticated, it wouldn't be difficult to refactor this code there and then.

PR #1544, opened 2017-09-06
Participant reports update

alxdarksage:
Adding new APIs:

1) clients can now save and retrieve participant reports, using DateTime instead of LocalDate values (so "2017-09-07T18:01:20.398Z" in addition to "2017-09-07"), retrieving the reports using a paginated API.

2) get participants now has a v4 version of the API that can accept either DateTime/LocalDate values, and is also paginated

3) The other verbs on these v4 endpoints are also versioned at v4, according to our guidelines for versioning URLs

### DIFF
--- a/app/org/sagebionetworks/bridge/models/RangeTuple.java
+++ b/app/org/sagebionetworks/bridge/models/RangeTuple.java
@@ -1,0 +1,17 @@
+package org.sagebionetworks.bridge.models;
+
+public final class RangeTuple<S> {
+    private final S start;
+    private final S end;
+    
+    public RangeTuple(S start, S end) {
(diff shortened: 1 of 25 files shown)


app/org/sagebionetworks/bridge/models/RangeTuple.java
```diff
@@ -0,0 +1,17 @@
+package org.sagebionetworks.bridge.models;
+
+public final class RangeTuple<S> {
+    private final S start;
+    private final S end;
+    
+    public RangeTuple(S start, S end) {
```
liujoshua, line 7: This is nice

liujoshua (approved):

PR #1551, opened 2017-09-08
BRIDGE-1946: set JVM hostname on new relic UI.

zaro0508:
The APM -> JVM dashboard lists aws EC2 machines by IP addresses[1].
Change it to list the machines with the EC2 instance IDs.

[1] https://imgur.com/a/POZWt


DwayneJengSage (approved):
Looks good to me. In general, what do EC2 Instance IDs look like?

Also, since this is in .ebextensions, this means it has no effect on our Heroku env, right?

zaro0508:
Correct, this change has no effect on Heroku env.   EC2 instances look something like this "i-08c6f1a9a8ca2fb02"

PR #1545, opened 2017-09-07
BRIDGE-1937: Configure new relic application settings

zaro0508:
Configure the AWS NR application settings to match settings
that are on the Heroku new relic BridgePF-prod account.
This includes the setting to ignore 4XX status codes.

PR #1552, opened 2017-09-08
BRIDGE-1943: workaround to ignore 4xx status codes

zaro0508:
AWS currently does not provide a way to ignore 4xx errors from enhanced
health monitoring metrics[1].  This is a workaround from stackoverflow[2]
that will make AWS ignore all 4xx errors.

This means that EB will never degrade the environment health due to
4xx errors because it just won't know that they're occurring. This also
means that the "Health" page in the EB environment will always display
0 for the 4xx response code count.

[1] https://forums.aws.amazon.com/thread.jspa?messageID=716284
[2] https://stackoverflow.com/questions/36398456/elastic-beanstalk-disable-health-state-change-based-on-4xx-responses

### DIFF
--- a/dist/.ebextensions/ignore_4xx.config
+++ b/dist/.ebextensions/ignore_4xx.config
@@ -1,0 +1,6 @@
+container_commands:
+  01-patch-healthd:
+    command: "sudo /bin/sed -i 's/\\# normalize units to seconds with millisecond resolution/if status \\&\\& status.index(\"4\") == 0 then next end/g' /opt/elasticbeanstalk/lib/ruby/lib/ruby/gems/2.2.0/gems/healthd-appstat-1.0.1/lib/healthd-appstat/plugin.rb"
+  02-restart-healthd:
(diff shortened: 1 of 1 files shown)


DwayneJengSage (approved):
Looks good to me.

One clarifying question. Does this affect our 4XX alarms? We don't want Elastic Beanstalk to degrade our service health in response to the inevitable flood of 401s, but we do (for example) want to selectively alarm on a sudden increase in 412s.

zaro0508:
hmm, i'm not sure but probably unlikely since cloudwatch metrics come from cloudwatch logs.  The workaround also says selectively ignoring specific 4XX health errors is possible.   We currently have new relic setup to ignore "400, 401, 403, 404, 409, 410, 412, 423, 429"[3] so i've configured it to ignore all 4XX errors.   Would you prefer to selectively ignore in this case? 

[3] https://github.com/Sage-Bionetworks/BridgePF/pull/1545

DwayneJengSage:
If we can't configure the alarm rates, then we should just ignore all 4XX errors in the Elastic Beanstalk health checks.

We should definitely test our 4XX alarms after making this change, and fix them if needed.

PR #1554, opened 2017-09-12
excludeStudyIdInExport flag

DwayneJengSage:
See https://sagebionetworks.jira.com/browse/BRIDGE-1942

This is needed because we want new studies to not include studyId in the exported data, but we don't want to break old studies. This flag is only changeable by admins. It defaults to false, and new studies are created with this flag set to true.

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
@@ -29,7 +29,7 @@ public final class DynamoStudy implements Study {
     private String name;
     private String sponsorName;
     private String identifier;
-    private boolean excludeStudyIdInExport;
+    private boolean studyIdExcludedInExport;
     private String supportEmail;
     private Long synapseDataAccessTeamId;
     private String synapseProjectId;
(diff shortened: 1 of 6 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
```diff
@@ -122,6 +123,18 @@ public void setVersion(Long version) {
...
     @Override
+    public boolean getExcludeStudyIdInExport() {
+        return excludeStudyIdInExport;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void setExcludeStudyIdInExport(boolean excludeStudyIdInExport) {
+        this.excludeStudyIdInExport = excludeStudyIdInExport;
+    }
```
alxdarksage: isExcludeStudyId vs isStudyIdExcludedInExport, latter is a little clearer to me

DwayneJengSage: Renamed

liujoshua (approved):

PR #1555, opened 2017-09-14
Sequence period in recurring schedules

alxdarksage:
JIRA: https://sagebionetworks.jira.com/browse/BRIDGE-1953

Sets a limit (relative to the start of a scheduling series) to the number of tasks that are issued to a user. Limit is declared using a period. This is similar to the time window, but that is in absolute, UTC time, whereas this can be used to create a sequence in a crossover study, where the user is moved through different protocols over the course of the study.

### DIFF
--- a/app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
+++ b/app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
@@ -164,9 +164,11 @@ public abstract class ActivityScheduler {
     private boolean scheduledOnAfterSequencePeriod(ScheduleContext context, DateTime scheduledTime, int currentCount) {
         // Don't know why you'd specify a sequence, then force N number of activities to be returned, but the 
         // minimum activities setting is part of the public API and it would be unexpected if it didn't work.
-        
+        /*
         boolean minMet = context.getMinimumPerSchedule() == 0 || currentCount >= context.getMinimumPerSchedule();
         return sequenceEndsOn != null && !scheduledTime.isBefore(sequenceEndsOn) && minMet;
+        */
+        return sequenceEndsOn != null && !scheduledTime.isBefore(sequenceEndsOn);
     }
     
     private boolean scheduledOnAfterEndsOnNoMinimumCount(ScheduleContext context, DateTime scheduledTime) {
--- a/app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
+++ b/app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
@@ -14,6 +14,7 @@ import org.sagebionetworks.bridge.BridgeUtils;
 public abstract class ActivityScheduler {
     
     protected final Schedule schedule;
+    protected DateTime sequenceEndsOn;
     
     ActivityScheduler(Schedule schedule) {
         this.schedule = schedule;
@@ -39,6 +40,11 @@ public abstract class ActivityScheduler {
         if (schedule.getDelay() != null) {
             eventTime = eventTime.plus(schedule.getDelay());
         }
+        // Sequence period is measured from the first timestamp plus the delay, to the end of the period.
+        if (schedule.getSequencePeriod() != null) {
+            sequenceEndsOn = eventTime.plus(schedule.getSequencePeriod());
+        }
+        
         return eventTime;
     }
     
@@ -140,6 +146,9 @@ public abstract class ActivityScheduler {
     protected boolean shouldContinueScheduling(ScheduleContext context, DateTime scheduledTime,
             List<ScheduledActivity> scheduledActivities) {
 
+        if (scheduledOnAfterSequencePeriod(context, scheduledTime, scheduledActivities.size())) {
+            return false;
+        }
         if (scheduledOnAfterEndsOnNoMinimumCount(context, scheduledTime)) {
             return false;
         }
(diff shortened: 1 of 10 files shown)


app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -152,6 +161,14 @@ protected boolean shouldContinueScheduling(ScheduleContext context, DateTime sch
         return boundaryNotMet;
     }
     
+    private boolean scheduledOnAfterSequencePeriod(ScheduleContext context, DateTime scheduledTime, int currentCount) {
+        // Don't know why you'd specify a sequence, then force N number of activities to be returned, but the 
+        // minimum activities setting is part of the public API and it would be unexpected if it didn't work.
+        
+        boolean minMet = context.getMinimumPerSchedule() == 0 || currentCount >= context.getMinimumPerSchedule();
+        return sequenceEndsOn != null && !scheduledTime.isBefore(sequenceEndsOn) && minMet;
```
DwayneJengSage: As per IM conversation, this behavior is counter-intuitive and is not consistent with how schedule.endsOn works. We should remove the interaction between minimumPerSchedule and sequencePeriod.

test/org/sagebionetworks/bridge/models/schedules/ScheduleTest.java
```diff
@@ -36,26 +36,28 @@ public void canRountripSerialize() throws Exception {
         schedule.setExpires(Period.parse("P2D"));
         schedule.setStartsOn(DateTime.parse("2015-02-02T10:10:10.000Z"));
         schedule.setEndsOn(DateTime.parse("2015-01-01T10:10:10.000Z"));
-        schedule.setEventId("eventId");
+        schedule.setEventId(Schedule.EVENT_ID_PROPERTY);
         schedule.setInterval(Period.parse("P3D"));
-        schedule.setLabel("label");
+        schedule.setSequencePeriod(Period.parse("P3W"));
```
DwayneJengSage, line 41: Should also add a test that we can parse sequencePeriod back from JSON.

test/org/sagebionetworks/bridge/models/schedules/ScheduleTest.java
```diff
@@ -36,26 +36,28 @@ public void canRountripSerialize() throws Exception {
...
+        assertEquals("recurring", node.get(Schedule.SCHEDULE_TYPE_PROPERTY).asText());
+        assertEquals("eventId", node.get(Schedule.EVENT_ID_PROPERTY).asText());
+        assertEquals("0 0 8 ? * TUE *", node.get(Schedule.CRON_TRIGGER_PROPERTY).asText());
+        assertEquals("P1D", node.get(Schedule.DELAY_PROPERTY).asText());
+        assertEquals("P3D", node.get(Schedule.INTERVAL_PROPERTY).asText());
+        assertEquals("P2D", node.get(Schedule.EXPIRES_PROPERTY).asText());
+        assertEquals("P3W", node.get(Schedule.SEQUENCE_PERIOD_PROPERTY).asText());
+        assertEquals("2015-02-02T10:10:10.000Z", node.get(Schedule.STARTS_ON_PROPERTY).asText());
+        assertEquals("2015-01-01T10:10:10.000Z", node.get(Schedule.ENDS_ON_PROPERTY).asText());
+        assertFalse(node.get(Schedule.PERSISTENT_PROPERTY).asBoolean());
```
DwayneJengSage: Minor: booleanValue() instead of asBoolean(). This checks for type in addition to value. Otherwise, string "false" and int 0 will pass.

test/org/sagebionetworks/bridge/validators/ScheduleValidatorTest.java
```diff
@@ -32,6 +32,41 @@ public void before() {
...
+    public void sequencePeriodForOnetimeSchedule() {
+        Schedule schedule = new Schedule();
+        schedule.setSequencePeriod("P3D");
+        schedule.setScheduleType(ScheduleType.ONCE);
+        
+        assertValidatorMessage(validator, schedule, Schedule.SEQUENCE_PERIOD_PROPERTY, "cannot be set unless schedule is recurring");
+    }
+    
+    @Test
+    public void dPeriodSameAsIntervalOK() {
```
DwayneJengSage: test name?

DwayneJengSage (changes requested):
One major comment regarding the interaction between minimumPerSchedule and sequencePeriod, and three minor comments. Otherwise, looks good.

test/org/sagebionetworks/bridge/models/schedules/ScheduleTest.java
alxdarksage: Added to ScheduleTest.

test/org/sagebionetworks/bridge/models/schedules/ScheduleTest.java
alxdarksage: Changed all of these to the non-converting *Value() calls. Still trying to break this habit.

test/org/sagebionetworks/bridge/validators/ScheduleValidatorTest.java
alxdarksage: Fixed.

alxdarksage:
Updated based on comments. Thanks!

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -152,6 +161,16 @@ protected boolean shouldContinueScheduling(ScheduleContext context, DateTime sch
         return boundaryNotMet;
     }
     
+    private boolean scheduledOnAfterSequencePeriod(ScheduleContext context, DateTime scheduledTime, int currentCount) {
+        // Don't know why you'd specify a sequence, then force N number of activities to be returned, but the 
+        // minimum activities setting is part of the public API and it would be unexpected if it didn't work.
+        /*
+        boolean minMet = context.getMinimumPerSchedule() == 0 || currentCount >= context.getMinimumPerSchedule();
+        return sequenceEndsOn != null && !scheduledTime.isBefore(sequenceEndsOn) && minMet;
+        */
```
DwayneJengSage: Minor: Delete commented code and update comment.

DwayneJengSage (approved):
One minor comment that probably doesn't need further review.

PR #1556, opened 2017-09-15
getActivities to not throw when a reference is broken

DwayneJengSage:
See https://sagebionetworks.jira.com/browse/BRIDGE-1939

For a variety of reasons, we can't defend against broken references in schedule plans. The best we can do is ensure that broken references don't cause all getActivities calls to fail with a 404. Rather, this converts the 404 into a log error, so we can detect it, and the tasks are still vended to the user.

This might result in a 404 Survey Not Found later down the line, but it will be isolated to the offending survey rather than all activities.

Testing done:
* unit tests
* integ tests
* manually tested by creating a schedule with (a) broken schema ref (b) broken survey ref (c) broken compound activity ref (d) valid compound activity with a broken schema ref and broken survey ref. (Broken survey refs were created by creating a valid survey ref, then deleting the survey from DDB directly.) Called getActivities and verified the call succeeded. Verified the expected log errors in the logs.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
+++ b/app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
@@ -28,12 +28,15 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


DwayneJengSage:
Note that I removed an optimization in resolveListsInCompoundActivity(). This optimization doesn't get us much, and removing the optimization means one less code path to test.

alxdarksage (approved):

PR #1558, opened 2017-09-15
better S3 error handling for upload complete API

DwayneJengSage:
Currently, every S3 error in the upload complete API is treated as a 404 Not Found. This means that S3 internal server errors and permissions issues are reported and logged incorrectly.

This fix only reports 404s as 404s, and everything else is a 500 Internal Server Error (as it's either a back-end error or a Bridge configuration error).

Testing done:
* unit tests
* integ tests
* manually tested (a) file not in S3 (b) S3 permission denied (c) normal success case

### DIFF
--- a/app/org/sagebionetworks/bridge/services/UploadService.java
+++ b/app/org/sagebionetworks/bridge/services/UploadService.java
@@ -17,10 +17,10 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.Resource;
 
-import com.amazonaws.AmazonClientException;
 import com.amazonaws.HttpMethod;
 import com.amazonaws.auth.AWSStaticCredentialsProvider;
 import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.model.AmazonS3Exception;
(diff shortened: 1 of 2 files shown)


app/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -344,8 +345,13 @@ public void uploadComplete(StudyIdentifier studyId, UploadCompletionClient compl
...
             obj = s3Client.getObjectMetadata(uploadBucket, objectId);
-        } catch (AmazonClientException ex) {
-            throw new NotFoundException(ex);
+        } catch (AmazonS3Exception ex) {
+            if (ex.getStatusCode() == 404) {
+                throw new NotFoundException(ex);
+            } else {
+                // Only S3 404s are mapped to 404s. Everything else is an internal server error.
+                throw new BridgeServiceException(ex);
+            }
```
alxdarksage, line 354: Approved. I wonder if we could use a generic mapping class to convert all AWS constraint violation exceptions, all AWS not found exceptions, etc. in one place, so if fixed, it would be fixed everywhere in our codebase where it could occur.

alxdarksage (approved):

PR #1560, opened 2017-09-16
Reference new relic env vars in EB

zaro0508:
This PR depends on https://github.com/Sage-Bionetworks/BridgePF-infra/pull/35

Reference new relic parameters from AWS elastic beanstalk instead of defining
them in this repo.


zaro0508:
Better approach is to move config to /etc/newrelic/newrelic.yml

zaro0508:
This PR is being replaced by 1561

PR #1561, opened 2017-09-18
BRIDGE-1947: Refactor for easier NR agent upgrades

zaro0508:
The current setup requires coordinated config changes to both BridgePF and
BridgePF-infra on a new relic agent upgrade. The purpose of this change is
to make it so that NR agent upgrades only require changes to BridgePF.

* This PR depends on https://github.com/Sage-Bionetworks/BridgePF-infra/pull/37
* Setup new relic agent without the version info.
* Reference new relic parameters from AWS env vars instead of duplicating
them in both repos.

### DIFF
--- a/dist/.ebextensions/newrelic-server.config
+++ b/dist/.ebextensions/newrelic-server.config
@@ -7,12 +7,10 @@ container_commands:
   "01SetLicenseKey":
     command: "nrsysmond-config --set license_key=$NEW_RELIC_LICENSE_KEY"
   "02SetNRServerInstanceId":
-    command: "export INSTANCE_ID=$(curl http://instance-data/latest/meta-data/instance-id) && echo hostname=$INSTANCE_ID && echo hostname=$INSTANCE_ID >> $NEW_RELIC_SERVER_CONFIG"
+    command: "export INSTANCE_ID=$(curl http://instance-data/latest/meta-data/instance-id) && echo hostname=$INSTANCE_ID && echo hostname=$INSTANCE_ID >> /etc/newrelic/nrsysmond.cfg"
   "03CopyNRApmAgent":
     command: "/bin/mkdir -p /usr/local/lib/newrelic && cp -rf /var/app/staging/lib/com.newrelic.agent.java.newrelic-agent-3.32.0.jar $NEW_RELIC_APM_AGENT"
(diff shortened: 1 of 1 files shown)


DwayneJengSage:
I'm having trouble understanding what this accomplishes. It looks like you're moving a bunch of configs to environment variables, defining those variables in BridgePF-infra, and then consuming them here.

What's the difference between updating those values here vs in BridgePF-infra? When do we expect the values to be different?

zaro0508:
Yes, you are understanding this change series correctly.  To update the NR agent I would need to make changes to sbt lib ref in BridgePF and JAVA_OPTS env var in BridgePF-infra.  I'm trying to setup the repos so that future updates to NR agent only needs to be done in BridgePF.  An additional PR is needed after this series merges.  That PR would update JAVA_OPTS to the NR vars as well.  It will be something like this on BridgePF-infra:

`
ParameterKey=JavaOpts,ParameterValue=-Dnewrelic.config.file=$NewRelicApmConfig\ -javaagent:$NewRelicApmAgent
`

However setting JAVA_OPTS in this series would cause a failed deployment.  To avoid downtime we will need to update JAVA_OPTS after this setup is merged. 

The sequence is..
1. merge PR https://github.com/Sage-Bionetworks/BridgePF-infra/pull/37
2. merge this PR
3. merge PR to change JAVA_OPTS

DwayneJengSage:
I'm still a bit confused. All the env vars are things that don't change, while the thing that does change is the New Relic JAR. Wouldn't it make more sense for just the New Relic Jar to be parameterized by env vars?

zaro0508:
Yes.  I can keep this refactor to only NEW_RELIC_AGENT.  I thought since i was parameterizing that one I would also do the others as well, NEW_RELIC_APM_CONFIG & NEW_RELIC_SERVER_CONFIG.   NEW_RELIC_APM_CONFIG is also set in both repos, in newrelic-server.sh in BridgePF and JAVA_OPTS in BridgePF-infra.

DwayneJengSage:
I prefer to keep things simple. If we don't expect those values to change, I don't think we need to parameterize them. If we later find out we do need to change them, it'll be easy to parameterize later.

DwayneJengSage (approved):

PR #1565, opened 2017-09-20
Upload Metadata

DwayneJengSage:
See JIRA https://sagebionetworks.jira.com/browse/BRIDGE-1431 and design doc https://docs.google.com/a/sagebase.org/document/d/1dnZ8PaJ7G9H93fBOS7OqSzj2V_MbouGx1nwAaIYa2ZE/edit?usp=sharing

This change adds the upload metadata fields to the Study as well as the Upload API and the Synchronous Health Data API.

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
@@ -1,5 +1,6 @@
 package org.sagebionetworks.bridge.dynamodb;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -63,6 +64,7 @@ public final class DynamoStudy implements Study {
     private boolean disableExport;
 
     public DynamoStudy() {
+        uploadMetadataFieldDefinitions = new ArrayList<>();
         profileAttributes = new HashSet<>();
         taskIdentifiers = new HashSet<>();
         activityEventKeys = new HashSet<>();
@@ -215,7 +217,8 @@ public final class DynamoStudy implements Study {
     /** {@inheritDoc} */
     @Override
     public void setUploadMetadataFieldDefinitions(List<UploadFieldDefinition> uploadMetadataFieldDefinitions) {
-        this.uploadMetadataFieldDefinitions = uploadMetadataFieldDefinitions;
+        this.uploadMetadataFieldDefinitions = (uploadMetadataFieldDefinitions == null) ? new ArrayList<>() :
+                uploadMetadataFieldDefinitions;
     }
 
     /** {@inheritDoc} */
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
@@ -2,6 +2,7 @@ package org.sagebionetworks.bridge.dynamodb;
 
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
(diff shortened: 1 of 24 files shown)


DwayneJengSage:
Also added Worker to the get study by ID calls. BridgeEX will need to call Bridge Server to get the study to get upload metadata field defs.

app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
```diff
@@ -203,6 +206,19 @@ public void setUsesCustomExportSchedule(boolean usesCustomExportSchedule) {
     }
 
     /** {@inheritDoc} */
+    @DynamoDBTypeConverted(converter = DynamoUploadSchema.FieldDefinitionListMarshaller.class)
+    @Override
+    public List<UploadFieldDefinition> getUploadMetadataFieldDefinitions() {
+        return uploadMetadataFieldDefinitions;
```
alxdarksage, line 214: Prevent return of null?

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -458,6 +466,12 @@ public Study updateStudy(Study study, boolean isAdminUpdate) {
 
         Validate.entityThrowingException(validator, study);
 
+        // Only admins can delete or modify upload metadata fields. Check this after validation, so we don't have to
+        // deal with duplicates.
+        if (!isAdminUpdate) {
+            checkUploadMetadataConstraints(originalStudy, study);
+        }
```
alxdarksage: But client devs can add?

app/org/sagebionetworks/bridge/validators/HealthDataRecordValidator.java
```diff
@@ -106,6 +106,12 @@ public void validate(Object object, Errors errors) {
                 errors.rejectValue("uploadDate", Validate.CANNOT_BE_NULL);
             }
 
+            // userMetadata is optional, but if specified, it must be an object node.
+            JsonNode userMetadata = record.getUserMetadata();
+            if (userMetadata != null && !userMetadata.isObject()) {
+                errors.rejectValue("userMetadata", "must be an object node");
+            }
```
alxdarksage, line 113: So, do we ever validate that the JSON expresses the FieldDefinitions defined for a study? Shouldn't we? At least the keys of the object should match.

app/org/sagebionetworks/bridge/validators/UploadFieldDefinitionListValidator.java
```diff
@@ -0,0 +1,133 @@
...
+
+import org.sagebionetworks.bridge.BridgeUtils;
+import org.sagebionetworks.bridge.models.upload.UploadFieldDefinition;
+import org.sagebionetworks.bridge.models.upload.UploadFieldType;
+import org.sagebionetworks.bridge.upload.UploadUtil;
+
+/**
+ * Validates a List of UploadFieldDefinitions. This doesn't use the Validator class, because (1) the typing for a List
+ * of UploadFieldDefinitions is a little wonky and (2) we needed to pass in a little more context to get this to work.
+ */
```
alxdarksage, line 20: This is fine. I usually pass it in via the constructor, which means we create the validator in a call and then dispose of it.

DwayneJengSage: I prefer the opposite: singletons with no state whose behavior can be determined entirely by their inputs.

app/org/sagebionetworks/bridge/validators/HealthDataRecordValidator.java
DwayneJengSage: No. As it turns out, too much validation makes things too rigid and slows down app development. It's the same reason why most schema fields end up being marked required=false and why StrictValidation, in hindsight, was a mistake.

app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
DwayneJengSage: Changed to look like the other collections in this class.

app/org/sagebionetworks/bridge/services/StudyService.java
DwayneJengSage: Yes. Added to the comment "Anyone (admin or developer) can add or re-order fields."

alxdarksage (approved):

PR #1573, opened 2017-09-27
Using app config (if it exists) to resolve scheduled activities.

alxdarksage:
This finishes app configs by getting the app config for the request access scheduled activities, and referring to it before looking up the most recent schema/survey. Logic to resolve schedules was moved into separate class to make it easier to test (ReferenceResolver). I did not remove tests of reference resolution in the other tests, let me know if we should do this.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/AppConfigService.java
+++ b/app/org/sagebionetworks/bridge/services/AppConfigService.java
@@ -67,7 +67,7 @@ public class AppConfigService {
         return appConfigDao.getAppConfig(studyId, guid);
     }
     
-    public AppConfig getAppConfigForUser(CriteriaContext context) {
+    public AppConfig getAppConfigForUser(CriteriaContext context, boolean throwException) {
         checkNotNull(context);
 
         List<AppConfig> appConfigs = getAppConfigs(context.getStudyIdentifier());
@@ -78,10 +78,12 @@ public class AppConfigService {
           .collect(BridgeCollectors.toImmutableList());
 
         // Should have matched one and only one app config.
-        // The goal of the following code is not to introduce production exceptions when changing app configs.
         if (matches.isEmpty()) {
-            // If there are no matches, return the "default" app config (TBD: for now, throw exception)
-            throw new EntityNotFoundException(AppConfig.class);
+            if (throwException) {
+                throw new EntityNotFoundException(AppConfig.class);    
+            } else {
+                return null;
+            }
         } else if (matches.size() != 1) {
             // If there is more than one match, return the one created first, but log an error
             LOG.error("CriteriaContext matches more than one app config: criteriaContext=" + context + ", appConfigs="+matches);
--- a/app/org/sagebionetworks/bridge/services/ReferenceResolver.java
+++ b/app/org/sagebionetworks/bridge/services/ReferenceResolver.java
@@ -1,0 +1,216 @@
+package org.sagebionetworks.bridge.services;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.joda.time.DateTime;
+import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
+import org.sagebionetworks.bridge.models.ClientInfo;
+import org.sagebionetworks.bridge.models.schedules.Activity;
+import org.sagebionetworks.bridge.models.schedules.ActivityType;
+import org.sagebionetworks.bridge.models.schedules.CompoundActivity;
+import org.sagebionetworks.bridge.models.schedules.CompoundActivityDefinition;
+import org.sagebionetworks.bridge.models.schedules.ScheduledActivity;
+import org.sagebionetworks.bridge.models.schedules.SchemaReference;
+import org.sagebionetworks.bridge.models.schedules.SurveyReference;
+import org.sagebionetworks.bridge.models.schedules.TaskReference;
+import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
+import org.sagebionetworks.bridge.models.surveys.Survey;
+import org.sagebionetworks.bridge.models.upload.UploadSchema;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Activity references can be "abstract" in the sense that they need not specify the version of a survey 
+ * or schema that the client should use. When scheduled activities are generated, we "resolve" these 
+ * references to the most recent schema (that matches the app version of the requestor), or the most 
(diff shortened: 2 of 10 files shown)


app/org/sagebionetworks/bridge/services/AppConfigService.java
```diff
@@ -78,10 +78,12 @@ public AppConfig getAppConfigForUser(CriteriaContext context) {
...
         // Should have matched one and only one app config.
-        // The goal of the following code is not to introduce production exceptions when changing app configs.
         if (matches.isEmpty()) {
-            // If there are no matches, return the "default" app config (TBD: for now, throw exception)
-            throw new EntityNotFoundException(AppConfig.class);
+            if (throwException) {
+                throw new EntityNotFoundException(AppConfig.class);    
+            } else {
+                return null;
+            }
```
DwayneJengSage, line 86: Is it worth adding a unit test for throwException=false just for branch coverage?

test/org/sagebionetworks/bridge/services/ReferenceResolverTest.java
```diff
@@ -0,0 +1,329 @@
...
+        resolver.resolve(scheduledActivity);
+        
+        verify(compoundActivityDefinitionService, times(1)).getCompoundActivityDefinition(STUDY_ID, TASK_ID);
+    }
+    
+    @Test
+    public void compoundActivityFullyResolvedFromAppConfig() {
+        schemaReferences.put(SCHEMA_ID, RESOLVED_SCHEMA_REF);
+        surveyReferences.put(SURVEY_GUID, RESOLVED_SURVEY_REF);
+        scheduledActivity.setActivity(activityBuilder.withCompoundActivity(UNRESOLVED_COMPOUND_ACTIVITY).build());
```
DwayneJengSage, line 252: I think it's worth using the skinny ref here (and probably in the next 3 tests). This will test that we recursively resolve schemas and surveys inside a compound activity ref.

Alternatively, one extra test cases that specifically tests this case.

test/org/sagebionetworks/bridge/services/ReferenceResolverTest.java
```diff
@@ -0,0 +1,329 @@
...
+        
+        resolver.resolve(scheduledActivity);
+        
+        assertNull(scheduledActivity.getActivity().getTask().getSchema().getRevision());
+    }
+    
+    @Test
+    public void unresolvableCompoundActivityReturnedAsIs() {
+        scheduledActivity.setActivity(activityBuilder.withCompoundActivity(COMPOUND_ACTIVITY_SKINNY_REF).build());
+        when(compoundActivityDefinitionService.getCompoundActivityDefinition(STUDY_ID, TASK_ID)).thenThrow(new EntityNotFoundException(CompoundActivityDefinition.class));
```
DwayneJengSage, line 358: Should we also test a compound activity with unresolvable schemas and surveys?

app/org/sagebionetworks/bridge/services/ReferenceResolver.java
```diff
@@ -0,0 +1,216 @@
...
+            SurveyReference resolvedSurveyRef = resolveSurvey(surveyRef);
+
+            if (resolvedSurveyRef != null && !resolvedSurveyRef.equals(surveyRef)) {
+                resolvedActivity = new Activity.Builder().withActivity(activity).withSurvey(resolvedSurveyRef).build();
+            }
+        } else if (activityType == ActivityType.TASK) {
+            TaskReference taskRef = activity.getTask();
+            SchemaReference schemaRef = taskRef.getSchema();
+
+            if (schemaRef != null) {
```
DwayneJengSage, line 89: Is it worth adding a test case for a taskRef without a schemaRef, for branch coverage?

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
```diff
@@ -366,189 +358,32 @@ private int byteLength(JsonNode node) {
...
 
         List<SchedulePlan> plans = schedulePlanService.getSchedulePlans(context.getCriteriaContext().getClientInfo(),
                 context.getCriteriaContext().getStudyIdentifier());
-
+        
+        AppConfig appConfig = appConfigService.getAppConfigForUser(context.getCriteriaContext(), false);
+        Map<String, SurveyReference> surveyReferences = (appConfig == null) ? ImmutableMap.of()
+                : Maps.uniqueIndex(appConfig.getSurveyReferences(), SurveyReference::getGuid);
+        Map<String, SchemaReference> schemaReferences = (appConfig == null) ? ImmutableMap.of()
+                : Maps.uniqueIndex(appConfig.getSchemaReferences(), SchemaReference::getId);
```
DwayneJengSage, line 370: Is it worth adding a minimal test to ScheduledActivityServiceResolveLinksTest with non-null AppConfig and non-empty reference maps, just to make sure we're constructing the maps correctly?

DwayneJengSage (changes requested):
Looks good. All my notes are test notes.

I'm okay with leaving the tests in ScheduledActivityServiceResolveLinksTest as they are, since they run pretty fast as is.

I'd also be okay with removing most of the tests in ScheduledActivityServiceResolveLinksTest, since they are redundant. As long as we leave in minimal tests to test both the AppConfig null and non-null cases.

app/org/sagebionetworks/bridge/services/AppConfigService.java
alxdarksage: Added test for null case.

test/org/sagebionetworks/bridge/services/ReferenceResolverTest.java
alxdarksage: I've added a test of a reference compound activity that retrieves a definition where the references aren't resolved, and verify that those in turn are verified.

test/org/sagebionetworks/bridge/services/ReferenceResolverTest.java
alxdarksage: Added test where compound activity def is returned but the references are broken (not found by the services).

app/org/sagebionetworks/bridge/services/ReferenceResolver.java
alxdarksage: added

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
alxdarksage: Added a test where there's an AppConfig, verify the app configs references are used to resolve the scheduled activities, so the map should be correct.

DwayneJengSage (approved):

PR #1574, opened 2017-09-28
decouple survey upload format from original RK/AC implementation

DwayneJengSage:
See https://docs.google.com/a/sagebase.org/document/d/1rW2wbx7UnPD9EMhbXcer5PzyreA4fQ5IG4Rswg3E158/edit?usp=sharing

### DIFF
--- a/app/org/sagebionetworks/bridge/upload/GenericUploadFormatHandler.java
+++ b/app/org/sagebionetworks/bridge/upload/GenericUploadFormatHandler.java
@@ -1,0 +1,217 @@
+package org.sagebionetworks.bridge.upload;
+
+import java.util.Iterator;
+import java.util.Map;
+import javax.annotation.Nonnull;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.apache.commons.lang3.StringUtils;
+import org.joda.time.DateTime;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import org.sagebionetworks.bridge.json.BridgeObjectMapper;
+import org.sagebionetworks.bridge.json.DateUtils;
+import org.sagebionetworks.bridge.json.JsonUtils;
+import org.sagebionetworks.bridge.models.GuidCreatedOnVersionHolderImpl;
+import org.sagebionetworks.bridge.models.healthdata.HealthDataRecord;
+import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
+import org.sagebionetworks.bridge.models.surveys.Survey;
+import org.sagebionetworks.bridge.models.upload.UploadFieldDefinition;
+import org.sagebionetworks.bridge.models.upload.UploadFieldType;
+import org.sagebionetworks.bridge.models.upload.UploadSchema;
+import org.sagebionetworks.bridge.services.SurveyService;
+import org.sagebionetworks.bridge.services.UploadSchemaService;
+
+/**
+ * <p>
+ * Processes uploads for the v2_generic format. This handler reads from
+ * {@link org.sagebionetworks.bridge.upload.UploadValidationContext#getUnzippedDataMap} and
+ * {@link org.sagebionetworks.bridge.upload.UploadValidationContext#getJsonDataMap} and updates the existing record in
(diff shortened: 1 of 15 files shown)


app/org/sagebionetworks/bridge/upload/GenericUploadFormatHandler.java
```diff
@@ -0,0 +1,217 @@
...
+        this.uploadSchemaService = uploadSchemaService;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void handle(@Nonnull UploadValidationContext context) throws UploadValidationException {
+        Map<String, byte[]> attachmentMap = context.getAttachmentsByFieldName();
+        Map<String, JsonNode> jsonDataMap = context.getJsonDataMap();
+        HealthDataRecord record = context.getHealthDataRecord();
+        ObjectNode dataMap = (ObjectNode) record.getData();
```
alxdarksage, line 70: Always an object node and not an array, string, etc?

app/org/sagebionetworks/bridge/upload/GenericUploadFormatHandler.java
```diff
@@ -0,0 +1,217 @@
...
+        // Parse into a Joda DateTime.
+        DateTime createdOn = null;
+        if (StringUtils.isNotBlank(createdOnString)) {
+            try {
+                createdOn = DateTime.parse(createdOnString);
+            } catch (IllegalArgumentException ex) {
+                // Write a message to the validation context, but there's no need to log or throw.
+                context.addMessage("Invalid date-time: " + createdOnString);
+            }
+        }
```
alxdarksage, line 140: Could use BridgeUtils.getDateTimeOrDefault(createdOnString, null), and when null, write context message. Handles blank value and try/catch

test/org/sagebionetworks/bridge/upload/UploadUtilTest.java
```diff
@@ -588,4 +645,19 @@ public void iosTimestamp() {
...
+        // Use an immutable map. If we attempt to modify the original map, this will throw.
+        // Use a map of strings, for simplicity of testing.
+        Map<String, String> inputMap = ImmutableMap.<String, String>builder().put("foo", "bar")
+                .put("sanitize!@#$this", "sanitize this's value").build();
+
+        // Execute and validate.
+        Map<String, String> outputMap = UploadUtil.sanitizeFieldNames(inputMap);
+        assertEquals(2, outputMap.size());
+        assertEquals("bar", outputMap.get("foo"));
+        assertEquals("sanitize this's value", outputMap.get("sanitize____this"));
```
alxdarksage, line 661: Is there any danger of two strings like "a!@#b" and "a^&*b" both being sanitized to a___b and a___b? If so, should you opt for some kind of escaping of special characters so the fields remain unique?

alxdarksage (approved):

app/org/sagebionetworks/bridge/upload/GenericUploadFormatHandler.java
DwayneJengSage: This is guaranteed by InitRecordHandler to always be an ObjectNode. Not sure why it's typed as a JsonNode in HealthDataRecord.

It might be worth refactoring so getData() is type ObjectNode instead of JsonNode, but that's a lot of files it would touch. Filed JIRA https://sagebionetworks.jira.com/browse/BRIDGE-1960 to track this refactor.

app/org/sagebionetworks/bridge/upload/GenericUploadFormatHandler.java
DwayneJengSage: BridgeUtils.getDateTimeOrDefault() throws on parse error, so it'd be the same amount of code either way.

test/org/sagebionetworks/bridge/upload/UploadUtilTest.java
DwayneJengSage: The limitation is that you can't create Synapse columns with these characters, and we enforce that same limitation in schema fields. There would be no point trying to preserve these special characters in field names, since we can't use them anyway.

alxdarksage: Not preserve but escape, so for example using URL escaping, "a!%40%23b" and "a%5E%26*b", which remain different but also safe for use. However, changes of this being an issue seem very slim.

PR #1575, opened 2017-09-29
Remove new relic server

zaro0508:
New relic server agent (newrelic-sysmond) is being replaced by
the new relic infrastructure[1] agent.  We already have infrastructure
agent setup and running so we no longer need the server agent.
The infrastructure agent reports both apm and infrastructure
metrics.

[1] https://docs.newrelic.com/docs/infrastructure/new-relic-infrastructure/installation/transition-servers-new-relic-infrastructure


zaro0508:
Tests indicate that both NR server and NR infra agents are needed by infrastructure.

PR #1582, opened 2017-10-09
Ensure that finished tasks remain finished in v4 API.

alxdarksage:
This was changed to ensure that users see all the activities that are scheduled in the period they are looking for. This change removed functionality to attempt to load all the scheduled activities from the database.

The problem is that once these tasks fall outside the time range, they are not retrieved and their state is reset. This is a problem for one-time tasks that sit around for awhile... or any other task with a long periodicity (longer than the query window for activities when scheduling).

### DIFF
--- a/app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
+++ b/app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
@@ -261,7 +261,7 @@ public class ScheduledActivityService {
         
         Map<String,ScheduledActivity> dbMap = Maps.newHashMap();
         // IA-545: If a schedule has an identical activity but a new GUID (say if we change the schedule on the user), the user can 
-        // lose existing activities. So during the time window the user is looking at, we will return any activities that exist.x`
+        // lose existing activities. So during the time window the user is looking at, we will return any activities that exist.
         for (String activityGuid : activityGuids) {
             ForwardCursorPagedResourceList<ScheduledActivity> list = activityDao.getActivityHistoryV2(
                     context.getCriteriaContext().getHealthCode(), activityGuid,
--- a/app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
+++ b/app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
@@ -260,6 +260,8 @@ public class ScheduledActivityService {
         }).collect(Collectors.toSet());
         
         Map<String,ScheduledActivity> dbMap = Maps.newHashMap();
+        // IA-545: If a schedule has an identical activity but a new GUID (say if we change the schedule on the user), the user can 
+        // lose existing activities. So during the time window the user is looking at, we will return any activities that exist.
         for (String activityGuid : activityGuids) {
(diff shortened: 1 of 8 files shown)


test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -665,6 +665,23 @@ public void getActivityHistoryV3ValidatesDateRange() {
...
     }
     
+    @Test
+    public void oneTimeActivityDisappearsOutOfTimeRange() throws Exception {
+        DateTime endsOn = DateTime.now().plusDays(3);
+        ScheduleContext context = createScheduleContext(endsOn).build();
+        
+        mockAllCallsForDbActivities(Lists.newArrayList());
+        
+        service.getScheduledActivitiesV4(context);
```
DwayneJengSage: Should our unit test also verify that we return the expected activity, with the finishedOn timestamps and everything?

DwayneJengSage (changes requested):
One test note. Otherwise, looks good.

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
```diff
@@ -260,6 +260,8 @@ protected DateTime getDateTime() {
         }).collect(Collectors.toSet());
         
         Map<String,ScheduledActivity> dbMap = Maps.newHashMap();
+        // IA-545: If a schedule has an identical activity but a new GUID (say if we change the schedule on the user), the user can 
+        // lose existing activities. So during the time window the user is looking at, we will return any activities that exist.x`
```
DwayneJengSage: Minor: Stray "x`" at the end of the comment.

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -1214,8 +1261,10 @@ private String firstTimeStampFor(int initialTZOffset, int requestTZOffset, Sched
         Map<String,DateTime> events = Maps.newHashMap();
         events.put("enrollment", ENROLLMENT);
         
-        return new ScheduleContext.Builder().withStudyIdentifier(TEST_STUDY).withInitialTimeZone(DateTimeZone.UTC)
-                .withStartsOn(NOW).withAccountCreatedOn(ENROLLMENT.minusHours(2)).withEndsOn(endsOn)
+        DateTimeZone zone = DateTimeZone.forOffsetHours(-7); // zone of the original times
+        // Initial time zone and the time zone of enrollment should be the same. They are in reality.
+        return new ScheduleContext.Builder().withStudyIdentifier(TEST_STUDY).withInitialTimeZone(zone)
+                .withStartsOn(NOW).withAccountCreatedOn(ENROLLMENT.minusHours(2).withZone(zone)).withEndsOn(endsOn)
```
DwayneJengSage: I'm not sure if we should be changing the timezone here. iirc, we wrote the tests like this specifically to ensure that changing timezones doesn't break anything. This is being used by a lot of tests, and I'm not sure what the impact is of changing this.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #1585, opened 2017-10-11
Improve Upload Validation

DwayneJengSage:
See https://sagebionetworks.jira.com/browse/BRIDGE-1865

Strict Validation as originally intended to filter out unusable data, such as Walking Activity with no accelerometer data, and to aid in app development by having missing (or incorrectly named) fields be rejected.

In practice, we just ended up setting most fields to required=false, which defeats the whole purpose of Strict Validation. In addition, researchers have asked for incomplete data, so they can do data quality analysis.

Our solution is to change the boolean Strict Validation into a three-state enum, and the default option will be to permit everything, but write the validation errors to Synapse so incomplete data is properly marked as such.

Next steps: Update BridgeEX to read record.validationErrors and write it to Synapse.

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
@@ -24,6 +24,7 @@ import org.sagebionetworks.bridge.models.studies.Study;
 import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
 import org.sagebionetworks.bridge.models.studies.StudyIdentifierImpl;
 import org.sagebionetworks.bridge.models.upload.UploadFieldDefinition;
+import org.sagebionetworks.bridge.models.upload.UploadValidationStrictness;
 
 @DynamoDBTable(tableName = "Study")
 @BridgeTypeName("Study")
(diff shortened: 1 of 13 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
```diff
@@ -39,6 +40,7 @@
     private String technicalEmail;
     private boolean usesCustomExportSchedule;
     private List<UploadFieldDefinition> uploadMetadataFieldDefinitions;
+    private UploadValidationStrictness uploadValidationStrictness;
```
alxdarksage, line 43: What happens to strictUploadValidationEnabled? Will it eventually go away? Should it be removed from the BSM UI when this is put in?

alxdarksage (approved):

app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
DwayneJengSage: strictUploadValidationEnabled is deprecated in favor of the new Enum. I have no plans to remove it at the moment, and I'm not sure what migration will look like.

That said, in the Bridge Study Manager, we should be replacing strictUploadValidationEnabled with uploadValidationStrictnes.

PR #1587, opened 2017-10-13
Reauth API

alxdarksage:
See: https://sagebionetworks.jira.com/browse/BRIDGE-1956

This also removes the feature where you can change a password when signing in via email. Instead, after signing in via this channel, you can use the reauthorization token to re-acquire a session.

This update requires modifying SQL tables, run:

```
ALTER TABLE `Accounts`
  ADD `reauthTokenAlgorithm` ENUM('STORMPATH_HMAC_SHA_256', 'BCRYPT', 'PBKDF2_HMAC_SHA_256') NULL,
  ADD `reauthTokenHash` VARCHAR(255) NULL,
  ADD `reauthTokenModifiedOn` BIGINT NULL;
```

I will add these to the on boarding document when we're ready to merge. I will also update existing RDS instances with these columns.


app/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -113,6 +114,10 @@ public static String resolveTemplate(String template, Map<String,String> values)
     public static String generateGuid() {
         return UUID.randomUUID().toString();
     }
+    
+    public static String generateRandomString() {
+        return STRING_GENERATOR.nextString();
+    }
```
DwayneJengSage: Minor: Instead of having a singleton StringIdGenerator in BridgeUtils and a wrapper function, would it be cleaner to just make StringIdGenerator public, have the singleton live in that class, and call StringIdGenerator.INSTANCE.nextString() directly?

app/org/sagebionetworks/bridge/StringIdGenerator.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.security.SecureRandom;
+import java.util.Objects;
+import java.util.Random;
+
+/**
+ * This code came from Stack Exchange, reformatted to our standards. Unfortunately I then lost 
+ * the reference to the page I took it from. Cleaned up to our formatting standards.
+ */
+class StringIdGenerator {
```
DwayneJengSage: Minor: Would "SecureTokenGenerator" be a more descriptive name for this?

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -283,13 +311,53 @@ public Account getAccount(Study study, String id) {
...
+            throw new BridgeServiceException("Error validating "+type+": " + ex.getMessage(), ex);
+        }        
+    }
+    
+    private String hashCredential(PasswordAlgorithm algorithm, String value) {
+        String hash = null;
+        try {
+            hash = algorithm.generateHash(value);
+        } catch (InvalidKeyException | InvalidKeySpecException | NoSuchAlgorithmException ex) {
+            throw new BridgeServiceException("Error creating reauthentication token: " + ex.getMessage(), ex);
```
DwayneJengSage: Minor: Error message references re-auth tokens, but this method is called in other code paths as well.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -135,36 +131,71 @@ public void changePassword(Account account, String newPassword) {
...
+            String hash, String credentialValue, String credentialName) {
+
+        verifyCredential(hibernateAccount.getId(), credentialName, algorithm, hash, credentialValue);
 
         // Unmarshall account
-        validateHealthCode(study.getStudyIdentifier(), hibernateAccount);
-        return unmarshallAccount(hibernateAccount);
+        Account account = unmarshallAccount(hibernateAccount);
+        validateHealthCode(new StudyIdentifierImpl(hibernateAccount.getStudyId()), hibernateAccount, account, false);
+        updateReauthToken(hibernateAccount, account);
```
DwayneJengSage: This updates the reauth token even during password logins. Is this something that we want?

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -135,36 +131,71 @@ public void changePassword(Account account, String newPassword) {
...
+    private Account authenticateInternal(HibernateAccount hibernateAccount, PasswordAlgorithm algorithm,
+            String hash, String credentialValue, String credentialName) {
+
+        verifyCredential(hibernateAccount.getId(), credentialName, algorithm, hash, credentialValue);
 
         // Unmarshall account
-        validateHealthCode(study.getStudyIdentifier(), hibernateAccount);
-        return unmarshallAccount(hibernateAccount);
+        Account account = unmarshallAccount(hibernateAccount);
+        validateHealthCode(new StudyIdentifierImpl(hibernateAccount.getStudyId()), hibernateAccount, account, false);
```
DwayneJengSage: Is there any reason why we reversed the order of unmarshallAccount() and validateHealthCode()? If we validate before we unmarshall, we only need to update the HibernateAccount, and unmarshal() will create the GenericAccount with the healthCode/Id. If we validate _after_, then we need to update both the HibernateAccount and GenericAccount. (Here and 3 other places.)

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
```diff
@@ -228,64 +241,234 @@ public void authenticateAccountNotFound() throws Exception {
...
+        assertTrue(PasswordAlgorithm.DEFAULT_PASSWORD_ALGORITHM.checkHash(captured.getReauthTokenHash(),
+                account.getReauthToken()));
+    }
+    
+    @Test
+    public void signOut() throws Exception {
+        HibernateAccount hibernateAccount = makeValidHibernateAccount(false, true);
+        hibernateAccount.setReauthTokenAlgorithm(PasswordAlgorithm.DEFAULT_PASSWORD_ALGORITHM);
+        hibernateAccount.setReauthTokenHash("AAA");
+        hibernateAccount.setReauthTokenModifiedOn(DateTime.now().getMillis());
```
DwayneJengSage: We create a reauth token in makeValidHibernateAccount() and then immediately throw it away. This seems like an unnecessary use of CPU cycles, especially since hashing passwords is CPU intensive.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -217,9 +213,30 @@ public UserSession signIn(Study study, CriteriaContext context, SignIn signIn) t
...
+            throw new ConsentRequiredException(session);
+        }
+        return session;
+    }
 
     public void signOut(final UserSession session) {
         if (session != null) {
+            accountDao.signOut(session.getStudyIdentifier(), session.getParticipant().getEmail());
             cacheProvider.removeSession(session);
         }
```
DwayneJengSage, line 241: Add a unit test?

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -279,23 +281,43 @@ public void emailSignInThrowsConsentRequired() {
...
-        doReturn(account).when(accountDao).getAccountWithEmail(study, RECIPIENT_EMAIL);
-        doReturn(PARTICIPANT).when(participantService).getParticipant(study, account, false);
+    public void reauthentication() {
+        ((GenericAccount)account).setId(USER_ID);
+
+        StudyParticipant participant = new StudyParticipant.Builder().withEmail(RECIPIENT_EMAIL).build();
         doReturn(CONSENTED_STATUS_MAP).when(consentService).getConsentStatuses(any());
+        doReturn(account).when(accountDao).reauthenticate(study, REAUTH_REQUEST);
+        doReturn(participant).when(participantService).getParticipant(study, account, false);
+        service.reauthenticate(study, CONTEXT, REAUTH_REQUEST);
```
DwayneJengSage: Verify that returned session contains reauthToken?

Do we also need this test for signIn and emailSignIn?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/BridgeUtils.java
alxdarksage: Okay, changed this. Now as an instance it only generate 21 character random strings... but that's all we'll ever probably need.

app/org/sagebionetworks/bridge/StringIdGenerator.java
alxdarksage: Renamed

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: Yes. Changed this in all four places.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: Yes. The token is rotated on any successful authentication. At least, that's my design at the moment, we could revisit this.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: Added the name of the credential being hashed to the method signature.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage: Added both normal and null session cases.

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
alxdarksage: Changed makeValidateHibernateAccount flag to false to prevent creating and hashing value that is immediately replaced.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: Added this for all three paths.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -279,23 +319,45 @@ public void emailSignInThrowsConsentRequired() {
...
-        doReturn(PARTICIPANT).when(participantService).getParticipant(study, account, false);
+    public void reauthentication() {
+        ((GenericAccount)account).setId(USER_ID);
+        account.setReauthToken(REAUTH_TOKEN);
+
+        StudyParticipant participant = new StudyParticipant.Builder().withEmail(RECIPIENT_EMAIL).build();
         doReturn(CONSENTED_STATUS_MAP).when(consentService).getConsentStatuses(any());
+        doReturn(account).when(accountDao).reauthenticate(study, REAUTH_REQUEST);
+        doReturn(participant).when(participantService).getParticipant(study, account, false);
+        service.reauthenticate(study, CONTEXT, REAUTH_REQUEST);
```
DwayneJengSage: We should test that service.reathenticate() returns the UserSession we expect with the reauthToken we expect. This tests that the method does what we expect it to do.

DwayneJengSage (changes requested):
Looks good. Just one comment about a test.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: Updated to assign and examine returned session

DwayneJengSage (approved):

PR #1591, opened 2017-10-20
Adjustments to email sign in

alxdarksage:
- removing all the parameters that have to be returned to sign in except for the token. This is backwards compatible with the existing implementation (we'll just ignore extra params);
- enable an unverified account if the correct token is sent (rather than throwing exception);
- change behavior just to send another email with the same token if API is called before last call expires (allows token to be resent, which is more useful than throwing a rate-limiting exception... I will add proper rate limiting in a later checkin);
- if no account is found, add a pause before returning to foil account enumeration attacks;
- Added a builder to sign in, which I'd like to rename Credentials. Rationalized the validation to accommodate phone-based authentication, which is coming. A lot of the changes are switching over  to the use of the builder rather than a constructor.


app/org/sagebionetworks/bridge/json/SignInDeserializer.java
```diff
@@ -12,21 +12,28 @@
...
         String reauthToken = JsonUtils.asText(node, "reauthToken");
         
         String accountIdentifier = (StringUtils.isNotBlank(username)) ? username : email;
         
-        return new SignIn(studyId, accountIdentifier, password, token, reauthToken);
+        return new SignIn.Builder().withStudy(studyId).withEmail(accountIdentifier).withPhone(phone)
+                .withPassword(password).withToken(token).withReauthToken(reauthToken).build();
     }
 
 }
```
alxdarksage: I could actually add the logic to capture "username" in the builder, and eliminate this class.

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
```diff
@@ -32,26 +35,27 @@
...
-            throw new BadRequestException("Study identifier is required.");
-        }
-        Study study = studyService.getStudy(signInRequest.getStudyId());
-        verifySupportedVersionOrThrowException(study);
+        // We'll provide the study once we reconstitute the transaction
+        ClientInfo clientInfo = getClientInfoFromUserAgentHeader();
+        LinkedHashSet<String> langs = getLanguagesFromAcceptLanguageHeader();
         
-        CriteriaContext context = getCriteriaContext(study.getStudyIdentifier());
+        UserSession session = authenticationService.emailSignIn(clientInfo, langs, tokenHolder);
```
alxdarksage: Need to capture consent exception, and throw the kill switch exception instead, if necessary, or else re throw the consent exception. Just to be consistent with other methods

DwayneJengSage (changes requested):
As per discussion, it looks like we're adding a ton of complexity to remove the studyId and email address from the emailSignIn URL (but leaving the authentication token, which is arguably the most sensitive part).

It looks like we're adding a lot of complexity and specialized codepaths to support that, and I'm having trouble seeing the benefits of doing this work. I think it would be easier to simply leave the studyId and email address in the URL, which reduces the net complexity and code churn.

PR #1592, opened 2017-10-23
Fewer features in email cleanup.

alxdarksage:
- enable an unverified account if the correct token is sent (rather than throwing exception);
- if no account is found, add a pause before returning to foil account enumeration attacks;

### DIFF
--- a/app/org/sagebionetworks/bridge/services/AuthenticationService.java
+++ b/app/org/sagebionetworks/bridge/services/AuthenticationService.java
@@ -132,7 +132,6 @@ public class AuthenticationService {
             try {
                 // The not found case returns *much* faster than the normal case. To prevent account enumeration 
                 // attacks, measure time of a successful case and delay for that period before returning.
-                LOG.info("Email submitted for sign in that is not registered: " + signIn.getEmail());
                 TimeUnit.MILLISECONDS.sleep(emailSignInRequestInMillis.get());            
             } catch(InterruptedException e) {
                 // Just return, the thread was killed by the connection, the server died, etc.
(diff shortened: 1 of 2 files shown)


app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -116,13 +127,21 @@ public void requestEmailSignIn(SignIn signIn) {
             throw new LimitExceededException("Email currently pending confirmation.");
         }
         
-        // check that email is in the study, if not, return quietly to prevent session enumeration attacks
+        // check that email is in the study, if not, return quietly to prevent account enumeration attacks
         if (accountDao.getAccountWithEmail(study, signIn.getEmail()) == null) {
+            try {
+                // The not found case returns *much* faster than the normal case. To prevent account enumeration 
+                // attacks, measure time of a successful case and delay for that period before returning.
+                LOG.info("Email submitted for sign in that is not registered: " + signIn.getEmail());
```
DwayneJengSage: I'm not sure if we should be logging this. On the one hand, this only logs email addresses that we know are _not_ part of the study. On the other hand, we probably shouldn't be logging email addresses into the logs in general.

DwayneJengSage (commented):
One minor comment about logging an email address that's not found. Otherwise, looks good.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -116,13 +127,21 @@ public void requestEmailSignIn(SignIn signIn) {
             throw new LimitExceededException("Email currently pending confirmation.");
```
alxdarksage: I need to remove this. If the email fails to send and the user tries to trigger another email, it just seems unreasonable to throw an error rather than resending the email. This isn't a good way to rate-limit requests.

PR #1593, opened 2017-10-24
SignIn refactor

alxdarksage:
- Moved to the builder pattern for constructing SignIn instances, and deleted the deserializer (in keeping with the more common style of the codebase);
- Cleaned up SignInValidator so it is easier to understand what should be validated, and added validation for the phone-based authentication that will come once phone is added.


app/org/sagebionetworks/bridge/validators/SignInValidator.java
```diff
@@ -1,32 +1,53 @@
...
     }
     
-    private Type type;
+    private Set<RequiredFields> requiredFields;
 
-    public SignInValidator(Type type) {
-        checkNotNull(type);
-        this.type = type;
+    private SignInValidator(RequiredFields... fields) {
+        requiredFields = Sets.newHashSet(fields);
```
DwayneJengSage: Nitpick: This works fine as is, and probably has virtually no impact on performance, but EnumSets are usually preferred for enums.

The only downside is that EnumSets require you to use the EnumSet.of(E first, E...rest) pattern, which could be a problem if for whatever reason, we need an empty SignInValidator. I'll leave it to your judgement.

DwayneJengSage (approved):
Looks good. One minor comment.

app/org/sagebionetworks/bridge/validators/SignInValidator.java
alxdarksage: Changed to directly pass enumset to constructor, to avoid an intermediate list.

PR #1594, opened 2017-10-24
Add phone

alxdarksage:
JIRA: https://sagebionetworks.jira.com/browse/BRIDGE-1975

However, while you can add a phone number now when creating an account, we are not using it. You can't use it to sign in with a password, and there's no way to sign in via SMS. These are coming next.

Adding a phone number requires updating the Accounts table:

ALTER TABLE Accounts
    ADD phone VARCHAR(20),
    ADD phoneRegion VARCHAR(2),
    ADD phoneVerified BOOLEAN,
    ADD emailVerified BOOLEAN;
CREATE UNIQUE INDEX `Accounts-StudyId-Phone-Index` ON Accounts (`studyId` ASC, `phone` ASC);

Note that with the addition of an index, you can't create two accounts with the same phone number without getting a conflict error. Right now that error is "odd" as it is returned from the API, but my next checkin will be to implement finding people by their phone numbers and I will clarify the exception then. We're not using phone numbers yet so this should not harm anything.

THINGS YOU CAN'T DO YET: Sign in using a phone/SMS; verify your phone number; sign in with a phone number + password.


alxdarksage:
I'm not seeing code to set the emailVerified flag to true if a user successfully signs in via email. This will have the same issue that updateAccount() doesn't allow you to change these flags. I want to add this but also clean it up relative to the verifyEmail() path itself where I moved code to avoid the issue in updateAccount().

app/org/sagebionetworks/bridge/models/accounts/Phone.java
```diff
@@ -0,0 +1,72 @@
...
+import javax.persistence.Transient;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.i18n.phonenumbers.PhoneNumberUtil;
+import com.google.i18n.phonenumbers.PhoneNumberUtil.PhoneNumberFormat;
+import com.google.i18n.phonenumbers.Phonenumber.PhoneNumber;
+
+@Embeddable
+public class Phone {
```
DwayneJengSage, line 15: This class exposes Hibernate implementation and mixes data with behavior. Would it be better to split the Phone class into the Phone interface, HibernatePhone implementation, and PhoneUtils?

Or are we moving away from the model of using interfaces to hide implementation-specific details?

app/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
```diff
@@ -106,6 +111,46 @@ public String getEmail() {
...
+    public Boolean getEmailVerified() {
+        // For accounts prior to the introduction of the email/phone verification flags, where 
+        // the flag was not set on creation or verification of the email address, return the right value.
+        if (emailVerified == null) {
+            if (status == AccountStatus.ENABLED) {
+                return Boolean.TRUE;
+            } else if (status == AccountStatus.UNVERIFIED) {
+                return Boolean.FALSE;
+            }
+        }
```
DwayneJengSage: What does this look like for legacy accounts? Will this effectively cause the emailVerified flag to be automatically updated on the next write?

Would it be cleaner to put this in HibernateAccountDao.unmarshallAccount()? This way, we can separate data from behaviors.

test/org/sagebionetworks/bridge/hibernate/HibernateAccountTest.java
```diff
@@ -130,4 +131,19 @@ public void accountSummaryConstructor() {
...
+        HibernateAccount account = new HibernateAccount();
+        account.setStatus(AccountStatus.ENABLED);
+        
+        assertEquals(Boolean.TRUE, account.getEmailVerified());
+        
+        account.setStatus(AccountStatus.UNVERIFIED);
+        assertEquals(Boolean.FALSE, account.getEmailVerified());
+        
+        account.setStatus(null);
+        assertNull(account.getEmailVerified());
```
DwayneJengSage: Test case for if emailVerified is non-null? At the very least, this tests the other branch of the if-statement.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -85,7 +88,38 @@ public final void setHibernateHelper(HibernateHelper hibernateHelper) {
...
+        checkNotNull(account);
+        
+        boolean shouldUpdateEmailVerified = (account.getEmailVerified() != Boolean.TRUE);
+        boolean shouldUpdateStatus = (account.getStatus() == AccountStatus.UNVERIFIED);
+        
+        if (shouldUpdateEmailVerified || shouldUpdateStatus) {
+            HibernateAccount hibernateAccount = hibernateHelper.getById(HibernateAccount.class, account.getId());
+            if (hibernateAccount == null) {
+                return;
+            }
```
DwayneJengSage: If the account isn't found in the database, would this be an error?

Should we add a unit test to verify this branch?

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
```diff
@@ -475,7 +560,7 @@ public void getAccountAfterAuthenticateReturnsNull() throws Exception {
     @Test
     public void constructAccount() throws Exception {
         // execute and validate
-        GenericAccount account = (GenericAccount) dao.constructAccount(STUDY, EMAIL, DUMMY_PASSWORD);
+        GenericAccount account = (GenericAccount) dao.constructAccount(STUDY, EMAIL, PHONE, DUMMY_PASSWORD);
```
DwayneJengSage, line 571: Unit test should verify that the constructed account contains the expected phone and the verified flags are set correctly.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -276,13 +313,16 @@ public void updateAccount(Account account) {
...
+        // Can't change study, email, phone, emailVerified, phoneVerified, createdOn, or passwordModifiedOn.
         HibernateAccount persistedAccount = hibernateHelper.getById(HibernateAccount.class, accountId);
         if (persistedAccount == null) {
             throw new EntityNotFoundException(Account.class, "Account " + accountId + " not found");
         }
         accountToUpdate.setStudyId(persistedAccount.getStudyId());
         accountToUpdate.setEmail(persistedAccount.getEmail());
+        accountToUpdate.setPhone(persistedAccount.getPhone());
+        accountToUpdate.setEmailVerified(persistedAccount.getEmailVerified());
+        accountToUpdate.setPhoneVerified(persistedAccount.getPhoneVerified());
```
DwayneJengSage, line 330: Update HibernateAccountDaoTest.updateSuccess() to make sure we're preventing updates to phone and the verified flags?

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -457,6 +497,9 @@ static HibernateAccount marshallAccount(Account account) {
         HibernateAccount hibernateAccount = new HibernateAccount();
         hibernateAccount.setId(genericAccount.getId());
         hibernateAccount.setEmail(genericAccount.getEmail());
+        hibernateAccount.setPhone(genericAccount.getPhone());
+        hibernateAccount.setEmailVerified(genericAccount.getEmailVerified());
+        hibernateAccount.setPhoneVerified(genericAccount.getPhoneVerified());
```
DwayneJengSage, line 507: Update marshall() and unmarshall() unit tests to make sure we're un/marshalling phone and verified flags correctly?

test/org/sagebionetworks/bridge/models/accounts/GenericAccountTest.java
```diff
@@ -145,4 +146,19 @@ public void roles() {
...
+        GenericAccount account = new GenericAccount();
+        account.setStatus(AccountStatus.ENABLED);
+        
+        assertEquals(Boolean.TRUE, account.getEmailVerified());
+        
+        account.setStatus(AccountStatus.UNVERIFIED);
+        assertEquals(Boolean.FALSE, account.getEmailVerified());
+        
+        account.setStatus(null);
+        assertNull(account.getEmailVerified());
```
DwayneJengSage: Similarly, test case for when emailVerified is not null?

app/org/sagebionetworks/bridge/models/accounts/GenericAccount.java
```diff
@@ -143,6 +146,51 @@ public void setEmail(String email) {
...
+    public Boolean getEmailVerified() {
+        // For accounts prior to the introduction of the email/phone verification flags, where 
+        // the flag was not set on creation or verification of the email address, return the right value.
+        if (emailVerified == null) {
+            if (status == AccountStatus.ENABLED) {
+                return Boolean.TRUE;
+            } else if (status == AccountStatus.UNVERIFIED) {
+                return Boolean.FALSE;
+            }
+        }
```
DwayneJengSage: Similarly, if we moved this logic to the unmarshall() code, we wouldn't have to write or test this twice.

app/org/sagebionetworks/bridge/models/accounts/Phone.java
```diff
@@ -0,0 +1,72 @@
...
+    
+    public static final boolean isValid(Phone phone) {
+        try {
+            PhoneNumber phoneNumber = PHONE_UTIL.parse(phone.getNumber(), phone.getRegionCode());
+            return PHONE_UTIL.isValidNumber(phoneNumber);
+        } catch (Exception e) {
+        }
+        return false;
```
DwayneJengSage: Unit tests for isValid?

Also, what cases throw an Exception? Do we want to catch all exceptions or just specific exceptions like NumberFormatException?

app/org/sagebionetworks/bridge/models/accounts/Phone.java
```diff
@@ -0,0 +1,72 @@
...
+    
+    // For Hibernate. 
+    Phone() {
+    }
+    
+    @JsonCreator
+    public Phone(@JsonProperty("number") String number, @JsonProperty("regionCode") String regionCode) {
+        this.number = number;
+        this.regionCode = regionCode;
+    }
```
DwayneJengSage: Why couldn't Jackson use the default constructor and use getters and setters for attributes?

test/org/sagebionetworks/bridge/models/accounts/PhoneTest.java
```diff
@@ -0,0 +1,38 @@
...
+    @Test
+    public void canSerialize() throws Exception {
+        Phone phone = new Phone("408.258.8569", "US");
+        assertEquals("(408) 258-8569", phone.getNationalFormat());
+        
+        JsonNode node = BridgeObjectMapper.get().valueToTree(phone);
+        assertEquals("+14082588569", node.get("number").textValue());
+        assertEquals("US", node.get("regionCode").textValue());
+        assertEquals("Phone", node.get("type").textValue());
+        assertEquals(4, node.size());
```
DwayneJengSage: The JSON node has 4 entries, but you're only testing 3 of them?

test/org/sagebionetworks/bridge/models/accounts/PhoneTest.java
```diff
@@ -0,0 +1,38 @@
...
+import org.sagebionetworks.bridge.json.BridgeObjectMapper;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+public class PhoneTest {
+
+    @Test
+    public void canSerialize() throws Exception {
+        Phone phone = new Phone("408.258.8569", "US");
+        assertEquals("(408) 258-8569", phone.getNationalFormat());
```
DwayneJengSage: We should have separate explicit tests for getNationalFormat() and getNumber() instead of merging them with JSON serialization tests. As is, we only indirectly verify getNumber() returns the data we expect it to.

app/org/sagebionetworks/bridge/models/accounts/Phone.java
```diff
@@ -0,0 +1,72 @@
...
+        if (this.number != null && this.regionCode != null) {
+            try {
+                PhoneNumber phoneNumber = PHONE_UTIL.parse(this.number, this.regionCode);
+                if (PHONE_UTIL.isValidNumber(phoneNumber)) {            
+                    return PHONE_UTIL.format(phoneNumber, format);
+                }
+            } catch (Exception e) {
+            }
+        }
+        return number;
```
DwayneJengSage: Should this throw if number or regionCode are null?

What happens if the number or regionCode are invalid?

What are the cases that this will throw? Should we be catching all Exceptions or just specific exceptions like NumberFormatException?

We should have unit tests to test the various edge cases.

app/org/sagebionetworks/bridge/models/accounts/StudyParticipant.java
```diff
@@ -214,6 +234,9 @@ public Builder copyOf(StudyParticipant participant) {
             this.firstName = participant.getFirstName();
             this.lastName = participant.getLastName();
             this.email = participant.getEmail();
+            this.phone = participant.getPhone();
+            this.emailVerified = participant.getEmailVerified();
+            this.phoneVerified = participant.getPhoneVerified();
```
DwayneJengSage, line 239: Add to StudyParticipantTest.canCopy()?

app/org/sagebionetworks/bridge/models/accounts/StudyParticipant.java
```diff
@@ -241,6 +264,15 @@ public Builder copyFieldsOf(StudyParticipant participant, Set<String> fieldNames
...
             }
+            if (fieldNames.contains("phone")) {
+                withPhone(participant.getPhone());
+            }
+            if (fieldNames.contains("emailVerified")) {
+                withEmailVerified(participant.getEmailVerified());
+            }
+            if (fieldNames.contains("phoneVerified")) {
+                withPhoneVerified(participant.getPhoneVerified());
+            }
```
DwayneJengSage, line 275: Add a assertCopyField() test to StudyParticipantTest?

app/org/sagebionetworks/bridge/models/accounts/StudyParticipant.java
```diff
@@ -300,6 +332,18 @@ public Builder withEmail(String email) {
...
             return this;
         }
+        public Builder withPhone(Phone phone) {
+            this.phone = phone;
+            return this;
+        }
+        public Builder withEmailVerified(Boolean emailVerified) {
+            this.emailVerified = emailVerified;
+            return this;
+        }
```
DwayneJengSage, line 342: This is the inverse of the HibernateAccount/GenericAccount case. In this case, we're storing StudyParticipants in Redis, which have verified emails but don't have the emailVerified flag? What will happen in this case? Is there any way to set the flag appropriately when read from Redis?

The simplest thing to do would be to clear Redis when we push to Prod, to get rid of legacy sessions, but that has the unfortunate side effect of logging everyone out.

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -178,6 +178,9 @@ public StudyParticipant getParticipant(Study study, Account account, boolean inc
         builder.withFirstName(account.getFirstName());
         builder.withLastName(account.getLastName());
         builder.withEmail(account.getEmail());
+        builder.withPhone(account.getPhone());
+        builder.withEmailVerified(account.getEmailVerified());
+        builder.withPhoneVerified(account.getPhoneVerified());
```
DwayneJengSage, line 183: Add test case to ParticipantServiceTest.getStudyParticipant()?

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #1595, opened 2017-10-27
Remove Redis migration code.

alxdarksage:
https://sagebionetworks.jira.com/browse/BRIDGE-1978


zaro0508 (changes requested):
Looks good.  should "rediscloud.url" from bridge-server.conf be removed as well?  Also should we remove REDIS_Cloud_URL var from BridgePF-infra as well?

zaro0508 (approved):

PR #1597, opened 2017-11-02
Phone lookup

alxdarksage:
Introduces an AccountId object which is used by AccountDao to retrieve accounts. It contains a study ID and one of: an ID, email or phone number. You still can't sign in using a phone #, that'll come next.

Once you start tugging at this thread you realize we have many, many methods that take a studyIdentifier, and then a userId, email, or eventually a phone number. So I just stopped at a place where everything compiles, calling AccountDao and the related AccountWorflowService. This is already a lot of files.

The only functional change was one place where we were returning a 404 if someone requested a password reset email. We should not do this, it allows you to determine which emails are in the study. It also fit better with a refactor to use AccountId.

The AccountId will throw an exception if you use a getter on the object for a field that has not been initialized. This should make it easier to verify that it's not being constructed with one identifier, then being called to retrieve a different identifier (e.g. made with a userId, then called later to retrieve an email address). The only place this isn't relevant is in the private getHibernateAccount() method where we'll query the DB based on what we find in AccountId. There's an accessor on AccountId to get a version that doesn't have these precondition checks.

### DIFF
--- a/app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
+++ b/app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
@@ -137,8 +137,8 @@ public class HibernateAccountDao implements AccountDao {
 
     /** {@inheritDoc} */
     @Override
-    public void requestResetPassword(AccountId accountId) {
-        accountWorkflowService.requestResetPassword(accountId);
+    public void requestResetPassword(Study study, AccountId accountId) {
+        accountWorkflowService.requestResetPassword(study, accountId);
     }
 
     /** {@inheritDoc} */
@@ -422,10 +422,12 @@ public class HibernateAccountDao implements AccountDao {
     /** {@inheritDoc} */
     @Override
     public void deleteAccount(AccountId accountId) {
-        HibernateAccount hibernateAccount = getHibernateAccount(accountId);
-        if (hibernateAccount != null) {
-            hibernateHelper.deleteById(HibernateAccount.class, hibernateAccount.getId());    
+        String userId = accountId.getUnguardedAccountId().getId();
+        if (userId == null) {
+            HibernateAccount hibernateAccount = getHibernateAccount(accountId);
+            userId = hibernateAccount.getId();
         }
+        hibernateHelper.deleteById(HibernateAccount.class, userId);    
     }
 
     /** {@inheritDoc} */
--- a/app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
+++ b/app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
@@ -289,10 +289,10 @@ public class HibernateAccountDao implements AccountDao {
      * happen outside of this method. This is used during migration so that Stormpath does ID generation and so we
      * don't verify email twice.
      */
-    public void createAccountForMigration(Study study, Account account, String accountId, boolean sendVerifyEmail) {
+    public void createAccountForMigration(Study study, Account account, String id, boolean sendVerifyEmail) {
         // Initial creation of account. Fill in basic initial parameters.
         HibernateAccount hibernateAccount = marshallAccount(account);
-        hibernateAccount.setId(accountId);
+        hibernateAccount.setId(id);
         hibernateAccount.setStudyId(study.getIdentifier());
         hibernateAccount.setCreatedOn(DateUtils.getCurrentMillisFromEpoch());
         hibernateAccount.setModifiedOn(DateUtils.getCurrentMillisFromEpoch());
@@ -304,7 +304,8 @@ public class HibernateAccountDao implements AccountDao {
             hibernateHelper.create(hibernateAccount);
         } catch (ConcurrentModificationException ex) {
             // account exists, but we don't have the userId, load the account
-            HibernateAccount otherAccount = getHibernateAccount(AccountId.forEmail(study.getIdentifier(), account.getEmail()));
+            AccountId accountId = AccountId.forEmail(study.getIdentifier(), account.getEmail());
+            HibernateAccount otherAccount = getHibernateAccount(accountId);
             if (otherAccount != null) {
                 throw new EntityAlreadyExistsException(Account.class, "userId", otherAccount.getId());
             } else {
@@ -395,6 +396,8 @@ public class HibernateAccountDao implements AccountDao {
 
     // Helper method to get a single account for a given study and id, email address, or phone number.
     private HibernateAccount getHibernateAccount(AccountId accountId) {
+        // This is the only method where accessing null values is not an error, since we're searching for 
+        // the value that was provided. There will be one.
         AccountId unguarded = accountId.getUnguardedAccountId();
         if (unguarded.getId() != null) {
             return hibernateHelper.getById(HibernateAccount.class, unguarded.getId());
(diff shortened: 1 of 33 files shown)


app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -403,27 +394,38 @@ private String hashCredential(PasswordAlgorithm algorithm, String type, String v
...
 
     /** {@inheritDoc} */
     @Override
-    public void deleteAccount(Study study, String id) {
-        hibernateHelper.deleteById(HibernateAccount.class, id);
+    public void deleteAccount(AccountId accountId) {
+        HibernateAccount hibernateAccount = getHibernateAccount(accountId);
+        if (hibernateAccount != null) {
+            hibernateHelper.deleteById(HibernateAccount.class, hibernateAccount.getId());    
+        }
```
DwayneJengSage: This is functionally identical, but in the new code, in the string ID case, we'll query the DB for an account, and then only use the string ID (which we already have). Is it worth adding a hasId() to AccountId and optimizing on the case where we only need the string ID?

test/org/sagebionetworks/bridge/models/accounts/AccountIdTest.java
```diff
@@ -0,0 +1,70 @@
...
+                .equals(AccountId.forEmail(TEST_STUDY_IDENTIFIER, "email")));
+    }
+    
+    @Test
+    public void factoryMethodsWork() {
+        String number = PHONE.getNumber();
+        assertEquals("one", AccountId.forId(TEST_STUDY_IDENTIFIER, "one").getId());
+        assertEquals("one", AccountId.forEmail(TEST_STUDY_IDENTIFIER, "one").getEmail());
+        assertEquals(number, AccountId.forPhone(TEST_STUDY_IDENTIFIER, PHONE).getPhone().getNumber());
+    }
```
DwayneJengSage, line 38: Is it worth adding tests that get() throws if the value is null?

test/org/sagebionetworks/bridge/models/accounts/SignInTest.java
```diff
@@ -89,5 +89,23 @@ public void acceptsUsernameAsEmail() throws Exception {
...
+    }
+    
+    @Test
+    public void signInAccountIdWithPhone() {
+        SignIn signIn = new SignIn.Builder().withStudy(TestConstants.TEST_STUDY_IDENTIFIER)
+                .withPhone(TestConstants.PHONE).withPassword("password").build();
+        AccountId accountId = signIn.getAccountId();
+        assertEquals(TestConstants.TEST_STUDY_IDENTIFIER, accountId.getStudyId());
+        assertEquals(TestConstants.PHONE.getNumber(), accountId.getPhone().getNumber());
+    }
```
DwayneJengSage, line 111: Is it worth adding a case for getAccountId() with neither email nor phone?

app/org/sagebionetworks/bridge/services/AccountWorkflowService.java
```diff
@@ -168,29 +167,29 @@ public void notifyAccountExists(Study study, Email email) {
...
-        checkNotNull(email);
+    public void requestResetPassword(AccountId accountId) {
+        checkNotNull(accountId);
         
-        Account account = accountDao.getAccountWithEmail(study, email.getEmail());
+        Account account = accountDao.getAccount(accountId);
         if (account != null) {
-            sendPasswordResetRelatedEmail(study, email, study.getResetPasswordTemplate());    
+            Study study = studyService.getStudy(account.getStudyIdentifier());
+            sendPasswordResetRelatedEmail(study, account.getEmail(), study.getResetPasswordTemplate());
```
DwayneJengSage: It looks like in the old code, we already take in study. Would it be better to keep that, instead of calling studyService to get the study again?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -724,16 +730,17 @@ public void requestResetPassword() {
         
         participantService.requestResetPassword(STUDY, ID);
         
-        verify(accountDao).requestResetPassword(eq(STUDY), emailCaptor.capture());
+        verify(accountDao).requestResetPassword(accountIdCaptor.capture());
```
DwayneJengSage: Nitpick: Would eq(ACCOUNT_ID) be better than use a capture?

DwayneJengSage (changes requested):
Mostly good, a few small comments.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: Changed to check for ID and only query if it has something else

test/org/sagebionetworks/bridge/models/accounts/AccountIdTest.java
alxdarksage: Added. Also changed the guard because it was throwing IllegalArgumentException on a getter with no parameters, which seems weird. Now it throws an NPE.

app/org/sagebionetworks/bridge/services/AccountWorkflowService.java
alxdarksage: Okay, carried study over from services. Added a precondition check that study ID and account ID are the same.

test/org/sagebionetworks/bridge/models/accounts/SignInTest.java
alxdarksage: Added. Verified we validate before calling getAccountId() in all pathways since this will generate a server error.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Yes, it's better, changed.

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
```diff
@@ -850,13 +851,72 @@ public void getByEmailNotFound() {
...
+        
+        verify(mockHibernateHelper).deleteById(HibernateAccount.class, ACCOUNT_ID);
+    }
+    
+    @Test
+    public void deleteWithId() throws Exception {
+        // Directly deletes with the ID it has
+        dao.deleteAccount(ACCOUNT_ID_WITH_ID);
+        
         verify(mockHibernateHelper).deleteById(HibernateAccount.class, ACCOUNT_ID);
```
DwayneJengSage, line 920: Is it worth adding a check that we *don't* re-query Hibernate for the account?

test/org/sagebionetworks/bridge/models/accounts/SignInTest.java
```diff
@@ -89,5 +89,36 @@ public void acceptsUsernameAsEmail() throws Exception {
...
+    
+    @Test
+    public void signInAccountIncomplete() {
+        SignIn signIn = new SignIn.Builder().withStudy(TestConstants.TEST_STUDY_IDENTIFIER)
+                .withPassword("password").build();
+        // SignIn should be validated to hold either email or phone before we 
+        // retrieve accountId 
+        try {
+            signIn.getAccountId();    
+        } catch(IllegalArgumentException e) {
```
DwayneJengSage: Need a fail() here. Otherwise, this test passes if getAccountId() doesn't throw.

DwayneJengSage (approved):
Just two more small comments about test. Otherwise, looks good.

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
alxdarksage: Added

test/org/sagebionetworks/bridge/models/accounts/SignInTest.java
alxdarksage: good catch, added

PR #1599, opened 2017-11-08
Phone-only accounts; phone-only sign-ins

alxdarksage:
BRIDGE-1955; BRIDGE-1970; 
NOTE: You will need to run the following command as email is no longer required with this check-in:

ALTER TABLE `BridgeDB`.`Accounts` CHANGE COLUMN `email` `email` VARCHAR(255) NULL;

- Either email or phone (or both) must be provided on sign up. Providing a password is now optional;
- added 10 character app "short name" for SMS messages;
- added phone sign in "channel" similar to email-based sign in (unlike email-based sign in, you cannot currently disable it for a study... we can add this if we feel we need it);
- added phone to the AccountSummary object. Absolutely need it for the BSM to display these Accounts;
- request reset password will send either an email or SMS message depending on what is available, preferring email; it silently fails if neither has been verified;
- User data download service returns 400 if the account in question does not have a verified email address;
- email verification silently fails if the user has no email address (should not be triggered during sign up, but there's an API call to resend the token... this does nothing and returns 200);
- Notify account exists now either sends an email message or an SMS text with the link to reset the password;
- Consenting, or requesting that a consent be emailed again to the user, now silently do nothing for accounts with no email addresses. We need an alternative pathway for this, which is filed as BRIDGE-1988.

### DIFF
--- a/app/org/sagebionetworks/bridge/BridgeConstants.java
+++ b/app/org/sagebionetworks/bridge/BridgeConstants.java
@@ -94,6 +94,12 @@ public class BridgeConstants {
     
     /**
      * This whitelist adds a few additional tags and attributes that are used by the CKEDITOR options we have 
+     * Limited to 140 characters (Java is UTF-16, so two bytes per character, assuming for now that 
+     * SNS converts these to ASCII.
+     */
+    public static final int SMS_CHARACTER_LIMIT = 140;
+
+    /**
      * displayed in the UI.
      */
     public static final Whitelist CKEDITOR_WHITELIST = Whitelist.relaxed()
--- a/app/org/sagebionetworks/bridge/BridgeConstants.java
+++ b/app/org/sagebionetworks/bridge/BridgeConstants.java
@@ -93,14 +93,14 @@ public class BridgeConstants {
     public static final Set<Roles> NO_CALLER_ROLES = ImmutableSet.of();
     
     /**
-     * This whitelist adds a few additional tags and attributes that are used by the CKEDITOR options we have 
      * Limited to 140 characters (Java is UTF-16, so two bytes per character, assuming for now that 
      * SNS converts these to ASCII.
      */
     public static final int SMS_CHARACTER_LIMIT = 140;
 
     /**
-     * displayed in the UI.
+     * This whitelist adds a few additional tags and attributes that are used by the CKEDITOR options 
+     * we have displayed in the UI.
      */
     public static final Whitelist CKEDITOR_WHITELIST = Whitelist.relaxed()
             .addTags("hr", "s", "caption")
--- a/app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
+++ b/app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
@@ -1,6 +1,8 @@
 package org.sagebionetworks.bridge.hibernate;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static org.sagebionetworks.bridge.services.AuthenticationService.ChannelType.EMAIL;
+import static org.sagebionetworks.bridge.services.AuthenticationService.ChannelType.PHONE;
 
 import java.io.IOException;
 import java.security.InvalidKeyException;
@@ -50,6 +52,7 @@ import org.sagebionetworks.bridge.models.studies.StudyIdentifierImpl;
 import org.sagebionetworks.bridge.models.subpopulations.ConsentSignature;
 import org.sagebionetworks.bridge.models.subpopulations.SubpopulationGuid;
 import org.sagebionetworks.bridge.services.AccountWorkflowService;
+import org.sagebionetworks.bridge.services.AuthenticationService;
 import org.sagebionetworks.bridge.services.HealthCodeService;
 
 import com.fasterxml.jackson.databind.JsonNode;
@@ -61,9 +64,9 @@ public class HibernateAccountDao implements AccountDao {
     private static final Logger LOG = LoggerFactory.getLogger(HibernateAccountDao.class);
 
     static final String ACCOUNT_SUMMARY_QUERY_PREFIX = "select new " + HibernateAccount.class.getCanonicalName() +
-            "(createdOn, studyId, firstName, lastName, email, id, status) ";
+            "(createdOn, studyId, firstName, lastName, email, phone, id, status) ";
     static final String EMAIL_QUERY = "from HibernateAccount where studyId='%s' and email='%s'";
-    static final String PHONE_QUERY = "from HibernateAccount where studyId='%s' and phone='%s'";
+    static final String PHONE_QUERY = "from HibernateAccount where studyId='%s' and phone.number='%s' and phone.regionCode='%s'";
     
     private AccountWorkflowService accountWorkflowService;
     private HealthCodeService healthCodeService;
@@ -91,7 +94,7 @@ public class HibernateAccountDao implements AccountDao {
     @Override
     public void verifyEmail(EmailVerification verification) {
         Account account = accountWorkflowService.verifyEmail(verification);
-        verifyEmail(account);
+        verifyChannel(EMAIL, account);
     }
     
     /**
@@ -100,7 +103,8 @@ public class HibernateAccountDao implements AccountDao {
      * address.
      */
     @Override
-    public void verifyEmail(Account account) {
+    public void verifyChannel(AuthenticationService.ChannelType channelType, Account account) {
+        checkNotNull(channelType);
         checkNotNull(account);
         
         // Do not modify the account if it is disabled (all email verification workflows are 
@@ -108,10 +112,13 @@ public class HibernateAccountDao implements AccountDao {
         if (account.getStatus() == AccountStatus.DISABLED) {
             return;
         }
-        boolean shouldUpdateEmailVerified = (account.getEmailVerified() != Boolean.TRUE);
+        
+        // Avoid updating on every sign in by examining object state first.
+        boolean shouldUpdateEmailVerified = (channelType == EMAIL && account.getEmailVerified() != Boolean.TRUE);
+        boolean shouldUpdatePhoneVerified = (channelType == PHONE && account.getPhoneVerified() != Boolean.TRUE);
         boolean shouldUpdateStatus = (account.getStatus() == AccountStatus.UNVERIFIED);
         
-        if (shouldUpdateEmailVerified || shouldUpdateStatus) {
+        if (shouldUpdatePhoneVerified || shouldUpdateEmailVerified || shouldUpdateStatus) {
             HibernateAccount hibernateAccount = hibernateHelper.getById(HibernateAccount.class, account.getId());
             if (hibernateAccount == null) {
                 throw new EntityNotFoundException(Account.class);
@@ -119,6 +126,10 @@ public class HibernateAccountDao implements AccountDao {
             if (shouldUpdateEmailVerified) {
                 account.setEmailVerified(Boolean.TRUE);
                 hibernateAccount.setEmailVerified(Boolean.TRUE);
+            }
+            if (shouldUpdatePhoneVerified) {
+                account.setPhoneVerified(Boolean.TRUE);
+                hibernateAccount.setPhoneVerified(Boolean.TRUE);
             }
             if (shouldUpdateStatus) {
                 account.setStatus(AccountStatus.ENABLED);
@@ -173,7 +184,7 @@ public class HibernateAccountDao implements AccountDao {
     /** {@inheritDoc} */
     @Override
     public Account authenticate(Study study, SignIn signIn) {
-        HibernateAccount hibernateAccount = fetchHibernateAccount(study, signIn);
+        HibernateAccount hibernateAccount = fetchHibernateAccount(signIn);
         return authenticateInternal(hibernateAccount, hibernateAccount.getPasswordAlgorithm(),
                 hibernateAccount.getPasswordHash(), signIn.getPassword(), "password");
     }
@@ -181,7 +192,7 @@ public class HibernateAccountDao implements AccountDao {
     /** {@inheritDoc} */
     @Override
     public Account reauthenticate(Study study, SignIn signIn) {
-        HibernateAccount hibernateAccount = fetchHibernateAccount(study, signIn);
+        HibernateAccount hibernateAccount = fetchHibernateAccount(signIn);
         return authenticateInternal(hibernateAccount, hibernateAccount.getReauthTokenAlgorithm(),
                 hibernateAccount.getReauthTokenHash(), signIn.getReauthToken(), "reauth token");
     }
@@ -201,7 +212,7 @@ public class HibernateAccountDao implements AccountDao {
     @Override
     public Account getAccountAfterAuthentication(AccountId accountId) {
         HibernateAccount hibernateAccount = getHibernateAccount(accountId);
-        
+
         if (hibernateAccount != null) {
             validateHealthCode(hibernateAccount, false);
             Account account = unmarshallAccount(hibernateAccount);
@@ -260,13 +271,14 @@ public class HibernateAccountDao implements AccountDao {
         account.setPhoneVerified(Boolean.FALSE);
         account.setHealthId(healthId);
 
-        // Hash password.
-        PasswordAlgorithm passwordAlgorithm = PasswordAlgorithm.DEFAULT_PASSWORD_ALGORITHM;
-        String passwordHash = hashCredential(passwordAlgorithm, "password", password);
+        // Hash password if it has been supplied.
+        if (password != null) {
+            PasswordAlgorithm passwordAlgorithm = PasswordAlgorithm.DEFAULT_PASSWORD_ALGORITHM;
+            String passwordHash = hashCredential(passwordAlgorithm, "password", password);
 
-        account.setPasswordAlgorithm(passwordAlgorithm);
-        account.setPasswordHash(passwordHash);
-
+            account.setPasswordAlgorithm(passwordAlgorithm);
+            account.setPasswordHash(passwordHash);
+        }
         return account;
     }
 
@@ -303,14 +315,20 @@ public class HibernateAccountDao implements AccountDao {
         try {
             hibernateHelper.create(hibernateAccount);
         } catch (ConcurrentModificationException ex) {
-            // account exists, but we don't have the userId, load the account
-            AccountId accountId = AccountId.forEmail(study.getIdentifier(), account.getEmail());
+            // Account can conflict because studyId + email or studyId + phone has been used for an 
+            // existing account. 
+            AccountId accountId = null;
+            if (account.getEmail() != null) {
+                accountId = AccountId.forEmail(study.getIdentifier(), account.getEmail());
+            } else if (account.getPhone() != null) {
+                accountId = AccountId.forPhone(study.getIdentifier(), account.getPhone());
+            }
             HibernateAccount otherAccount = getHibernateAccount(accountId);
             if (otherAccount != null) {
                 throw new EntityAlreadyExistsException(Account.class, "userId", otherAccount.getId());
             } else {
                 throw new BridgeServiceException("Conflict creating an account, but can't find an existing " +
-                        "account with the same study and email");
+                        "account with the same study and email or phone");
             }
         }
     }
@@ -355,7 +373,7 @@ public class HibernateAccountDao implements AccountDao {
         }
     }
 
-    private HibernateAccount fetchHibernateAccount(Study study, SignIn signIn) {
+    private HibernateAccount fetchHibernateAccount(SignIn signIn) {
         // Fetch account
         HibernateAccount hibernateAccount = getHibernateAccount(signIn.getAccountId());
         if (hibernateAccount == null || hibernateAccount.getStatus() == AccountStatus.UNVERIFIED) {
@@ -406,7 +424,7 @@ public class HibernateAccountDao implements AccountDao {
         if (unguarded.getEmail() != null) {
             query = String.format(EMAIL_QUERY, unguarded.getStudyId(), unguarded.getEmail());
         } else {
-            query = String.format(PHONE_QUERY, unguarded.getStudyId(), unguarded.getPhone().getNumber());
+            query = String.format(PHONE_QUERY, unguarded.getStudyId(), unguarded.getPhone().getNumber(), unguarded.getPhone().getRegionCode());
         }
         List<HibernateAccount> accountList = hibernateHelper.queryGet(query, null, null, HibernateAccount.class);
         if (accountList.isEmpty()) {
@@ -699,7 +717,7 @@ public class HibernateAccountDao implements AccountDao {
 
         // Unmarshall single account
         return new AccountSummary(hibernateAccount.getFirstName(), hibernateAccount.getLastName(),
-                hibernateAccount.getEmail(), hibernateAccount.getId(), createdOn, hibernateAccount.getStatus(),
-                studyId);
+                hibernateAccount.getEmail(), hibernateAccount.getPhone(), hibernateAccount.getId(), createdOn,
+                hibernateAccount.getStatus(), studyId);
     }
 }
--- a/app/org/sagebionetworks/bridge/models/accounts/SignIn.java
+++ b/app/org/sagebionetworks/bridge/models/accounts/SignIn.java
@@ -3,6 +3,7 @@ package org.sagebionetworks.bridge.models.accounts;
 import org.sagebionetworks.bridge.models.BridgeEntity;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 
 @JsonDeserialize(builder = SignIn.Builder.class)
@@ -24,6 +25,9 @@ public final class SignIn implements BridgeEntity {
         this.reauthToken = reauthToken;
     }
     
+    // Serializing this property as study allows us to use SignIn to construct the JSON payload that we 
+    // are accepting from clients, which is useful for tests to avoid manually constructing the JSON string.
+    @JsonProperty("study")
     public String getStudyId() {
         return studyId;
     }
--- a/app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
@@ -182,8 +182,8 @@ public class AuthenticationController extends BaseController {
                 }
                 throw e;
             }
-            logAuthenticationSuccess(session);
         }
+        logAuthenticationSuccess(session);
 
         // You can proceed if 1) you're some kind of system administrator (developer, researcher), or 2)
         // you've consented to research.
@@ -194,7 +194,7 @@ public class AuthenticationController extends BaseController {
     }
 
     private void logAuthenticationSuccess(UserSession session) {
-        writeSessionInfoToMetrics(session);
+        writeSessionInfoToMetrics(session);  
         // We have removed the cookie in the past, only to find out that clients were unknowingly
         // depending on the cookie to preserve the session token. So it remains.
         response().setCookie(BridgeConstants.SESSION_TOKEN_HEADER, session.getSessionToken(),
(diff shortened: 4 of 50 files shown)


app/org/sagebionetworks/bridge/BridgeConstants.java
```diff
@@ -94,6 +94,12 @@
     
     /**
      * This whitelist adds a few additional tags and attributes that are used by the CKEDITOR options we have 
+     * Limited to 140 characters (Java is UTF-16, so two bytes per character, assuming for now that 
+     * SNS converts these to ASCII.
+     */
```
DwayneJengSage: Nitpick: It looks like the first line of the next comment got absorbed into this one.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -303,14 +315,20 @@ public void createAccountForMigration(Study study, Account account, String id, b
...
-            // account exists, but we don't have the userId, load the account
-            AccountId accountId = AccountId.forEmail(study.getIdentifier(), account.getEmail());
+            // Account can conflict because studyId + email or studyId + phone has been used for an 
+            // existing account. 
+            AccountId accountId = null;
+            if (account.getEmail() != null) {
+                accountId = AccountId.forEmail(study.getIdentifier(), account.getEmail());
+            } else if (account.getPhone() != null) {
+                accountId = AccountId.forPhone(study.getIdentifier(), account.getPhone());
+            }
```
DwayneJengSage, line 325: Do we need an else case for an impossible codepath? Or is a NullPointerException acceptable in this situation?

app/org/sagebionetworks/bridge/models/accounts/SignIn.java
```diff
@@ -24,6 +25,9 @@ private SignIn(String studyId, String email, Phone phone, String password, Strin
         this.reauthToken = reauthToken;
     }
     
+    // Serializing this property as study allows us to use SignIn to construct the JSON payload that we 
+    // are accepting from clients, which is useful for tests to avoid manually constructing the JSON string.
+    @JsonProperty("study")
```
DwayneJengSage, line 30: Our tests assume that SignIn will only ever be deserialized, never serialized. This annotation implies this is no longer true. Should we add serialization tests in SignInTest?

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
```diff
@@ -52,10 +52,36 @@ public Result emailSignIn() {
         CriteriaContext context = getCriteriaContext(study.getStudyIdentifier());
         
         UserSession session = authenticationService.emailSignIn(context, signInRequest);
-        
+        logAuthenticationSuccess(session);
+
```
DwayneJengSage, line 56: Is it worth adding some clauses to the emailSignIn() unit test to make sure we're writing metrics, setting cookie, caching request info, etc? To make it easier, maybe a helper method in AuthenticationControllerMockTest verifyCommonSignInFunctionality()?

Similarly for phoneSignIn, reauth, and regular signIn.

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerMockTest.java
```diff
@@ -699,38 +710,115 @@ private void mockResetPasswordRequest() throws Exception {
...
+    public void phoneSignIn() throws Exception {
+        TestUtils.mockPlayContextWithJson(PHONE_SIGN_IN);
+        
+        Metrics metrics = new Metrics(TEST_REQUEST_ID);
+        doReturn(metrics).when(controller).getMetrics();
+        
+        when(authenticationService.phoneSignIn(any(), any())).thenReturn(userSession);
+        
+        Result result = controller.phoneSignIn();
+        assertEquals(200, result.status());
```
DwayneJengSage, line 766: Is it worth verifying that the result contains a user session?

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
```diff
@@ -52,10 +52,36 @@ public Result emailSignIn() {
...
+    }
+
+    public Result phoneSignIn() {
+        SignIn signInRequest = parseJson(request(), SignIn.class);
+
+        if (isBlank(signInRequest.getStudyId())) {
+            throw new BadRequestException("Study identifier is required.");
+        }
+        Study study = studyService.getStudy(signInRequest.getStudyId());
+        verifySupportedVersionOrThrowException(study);
```
DwayneJengSage, line 75: Is it worth adding a unit test that phoneSignIn() checks the study shutoff switch?

app/org/sagebionetworks/bridge/services/AccountWorkflowService.java
```diff
@@ -203,16 +238,28 @@ private void sendPasswordResetRelatedEmail(Study study, String email, EmailTempl
     public void resetPassword(PasswordReset passwordReset) {
         checkNotNull(passwordReset);
         
-        String cacheKey = passwordReset.getSptoken() + ":" + passwordReset.getStudyIdentifier();
-        String email = cacheProvider.getString(cacheKey);
+        // This pathway is unusual as the account may have an email address or a phone number, so test for both.
+        String emailCacheKey = passwordReset.getSptoken() + ":email:" + passwordReset.getStudyIdentifier();
+        String phoneCacheKey = passwordReset.getSptoken() + ":phone:" + passwordReset.getStudyIdentifier();
```
DwayneJengSage: This has the side effect that if someone requests a password reset before we deploy, and then we deploy, their password reset token is effectively invalidated. Is this an acceptable scenario?

Once potential workaround is to just "grandfather in" sptoken:study as implicitly meaning email, rather than trying to migrate the token.

test/org/sagebionetworks/bridge/services/AccountWorkflowServiceTest.java
```diff
@@ -194,16 +215,34 @@ public void notifyAccountExists() throws Exception {
...
+    public void notifyAccountExistsForPhone() throws Exception {
+        AccountId accountId = AccountId.forPhone(TEST_STUDY_IDENTIFIER, TestConstants.PHONE);
+        when(service.createTimeLimitedToken()).thenReturn("ABC");
+        when(mockAccount.getPhone()).thenReturn(TestConstants.PHONE);
+        when(mockAccount.getPhoneVerified()).thenReturn(Boolean.TRUE);
+        when(mockAccountDao.getAccount(accountId)).thenReturn(mockAccount);
+        
+        service.notifyAccountExists(study, accountId);
+        
+        verify(mockNotificationsService).sendSMSMessage(eq(study.getStudyIdentifier()),
```
DwayneJengSage: Verify mock cache provider?

test/org/sagebionetworks/bridge/services/AccountWorkflowServiceTest.java
```diff
@@ -228,23 +318,47 @@ public void requestResetPasswordInvalidEmailFailsQuietly() throws Exception {
...
+        service.requestResetPassword(study, ACCOUNT_ID_WITH_PHONE);
+        
+        verifyNoMoreInteractions(mockSendMailService);
+        verifyNoMoreInteractions(mockNotificationsService);
+    }
+    
+    @Test
     public void resetPassword() {
-        when(mockCacheProvider.getString("sptoken:api")).thenReturn(EMAIL);
+        when(mockCacheProvider.getString("sptoken:email:api")).thenReturn(EMAIL);
```
DwayneJengSage: Unit test for resetting password via phone? (Including invalidating the phone cache key.)

test/org/sagebionetworks/bridge/services/AccountWorkflowServiceTest.java
```diff
@@ -217,6 +256,57 @@ public void requestResetPassword() throws Exception {
...
+        when(mockStudyService.getStudy(TEST_STUDY)).thenReturn(study);
+        when(mockAccount.getPhone()).thenReturn(TestConstants.PHONE);
+        when(mockAccount.getPhoneVerified()).thenReturn(Boolean.TRUE);        
+        when(mockAccount.getStudyIdentifier()).thenReturn(TEST_STUDY);
+        
+        service.requestResetPassword(study, ACCOUNT_ID_WITH_PHONE);
+        
+        verify(mockCacheProvider).setString(eq("ABC:phone:api"), stringCaptor.capture(), eq(60*60*2));
+        
+        Phone captured = BridgeObjectMapper.get().readValue(stringCaptor.getValue(), Phone.class);
```
DwayneJengSage: stringCaptor is used twice here, so we never validate what's actually being written to mockCacheProvider.

DwayneJengSage (changes requested):
Only reviewed through AccountWorkflowService. Will continue reviewing on Monday.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -176,14 +208,20 @@ public void requestEmailSignIn() throws Exception {
...
+        // In this case, where there is a value and an account, we do't generate a new one,
+        // we just send the message again.
         doReturn("something").when(cacheProvider).getString(CACHE_KEY);
+        doReturn(account).when(accountDao).getAccount(any());
         
-        service.requestEmailSignIn(SIGN_IN_REQUEST);
+        service.requestEmailSignIn(SIGN_IN_REQUEST_WITH_EMAIL);
+        
+        verify(cacheProvider, never()).setString(any(), any(), anyInt());
+        verify(sendMailService).sendEmail(any());
```
DwayneJengSage: Is it worth adding an ArgumentCaptor to verify the email passed to sendMailService.sendEmail() contains the token we expect?

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -373,4 +411,113 @@ public void reauthThrowsUnconsentedException() {
...
+        when(service.getPhoneToken()).thenReturn("123456");
+        
+        service.requestPhoneSignIn(SIGN_IN_REQUEST_WITH_PHONE);
+        
+        verify(cacheProvider).getString(cacheKey);
+        verify(cacheProvider).setString(eq(cacheKey), stringCaptor.capture(), eq(300));
+        verify(notificationsService).sendSMSMessage(eq(study.getStudyIdentifier()), eq(TestConstants.PHONE),
+                stringCaptor.capture());
+        assertEquals("123456", stringCaptor.getAllValues().get(0));
+        assertEquals("Enter 123-456 to sign in to AppName", stringCaptor.getAllValues().get(1));
```
DwayneJengSage: I would recommend using two different ArgumentCaptors for the different verify() checks. This way, you can name them something meaningful, and you won't accidentally break something by adding another use of stringCaptor.capture() in between.

Alternatively, since you know what the string is, you can just use eq() and match the string exactly.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -373,4 +411,113 @@ public void reauthThrowsUnconsentedException() {
...
+    
+    @Test
+    public void phoneSignIn() {
+        String cacheKey = TestConstants.PHONE.getNumber() + ":api:phoneSignInRequest";
+        when(cacheProvider.getString(cacheKey)).thenReturn(TOKEN);
+        when(accountDao.getAccountAfterAuthentication(ACCOUNT_ID_WITH_PHONE)).thenReturn(account);
+        when(participantService.getParticipant(study, account, false)).thenReturn(PARTICIPANT);
+        when(consentService.getConsentStatuses(any())).thenReturn(CONSENTED_STATUS_MAP);
+        
+        service.phoneSignIn(CONTEXT, SIGN_IN_WITH_PHONE);
```
DwayneJengSage: Is it worth verifying that service.phoneSignIn() returns a UserSessionInfo with the basic attributes we expect?

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
```diff
@@ -319,4 +320,40 @@ public void accountFailureConsistent() {
...
+
+        StudyParticipant noEmail = new StudyParticipant.Builder().copyOf(participant).withEmail(null).build();
+        Withdrawal withdrawal = new Withdrawal("reason");
+        CriteriaContext context = new CriteriaContext.Builder().withUserId(ID)
+                .withStudyIdentifier(study.getStudyIdentifier()).build();
+        
+        consentService.withdrawConsent(study, SUBPOP_GUID, noEmail, context, withdrawal, SIGNED_ON);
+        
+        verify(sendMailService, never()).sendEmail(any());
+    }
```
DwayneJengSage, line 415: Should we also add a unit for withdrawAllConsent w/o email?

app/org/sagebionetworks/bridge/services/NotificationsService.java
```diff
@@ -168,6 +184,35 @@ public void sendNotificationToUser(StudyIdentifier studyId, String healthCode, N
...
+    public void sendSMSMessage(StudyIdentifier studyId, Phone phone, String message) {
+        checkNotNull(studyId);
+        checkNotNull(phone);
+        checkNotNull(message);
+        
+        // Limited to 140 characters (Java is UTF-16, so two bytes per character, assuming for now that 
+        // SNS converts these to ASCII, requires integration testing
+        if (message.length() > BridgeConstants.SMS_CHARACTER_LIMIT) {
+            throw new BridgeServiceException("SMS message cannot be longer than 140 UTF-8/ASCII characters.");
+        }
```
DwayneJengSage: Do we know whether the limit is bytes or characters? If we want to count bytes, we can always do message.getBytes(Charsets.UTF-whatever).length. Do we know what encoding SMS sends the message with?

app/org/sagebionetworks/bridge/services/UserAdminService.java
```diff
@@ -117,50 +118,51 @@ public UserSession createUser(Study study, StudyParticipant participant, Subpopu
...
-                newUserSession = authenticationService.getSession(study, context);
             }
-            if (!signUserIn) {
-                authenticationService.signOut(newUserSession);
-                newUserSession.setAuthenticated(false);
+            if (signUserIn) {
+                SignIn signIn = new SignIn.Builder().withStudy(study.getIdentifier()).withEmail(participant.getEmail())
+                        .withPassword(participant.getPassword()).build();
+                return authenticationService.signIn(study, context, signIn);
             }
```
DwayneJengSage: This only works if the user is created with an email. Without it, it'll throw 400 InvalidEntityException.

Do we know for sure this is only called for users with emails? Should we have a check for if participant.getEmail() exists and throw a different exception? Fail silently?

app/org/sagebionetworks/bridge/services/UserAdminService.java
```diff
@@ -117,50 +118,51 @@ public UserSession createUser(Study study, StudyParticipant participant, Subpopu
             boolean signUserIn, boolean consentUser) {
         checkNotNull(study, "Study cannot be null");
         checkNotNull(participant, "Participant cannot be null");
-        checkNotNull(participant.getEmail(), "Sign up email cannot be null");
```
DwayneJengSage, line 120: Should we add a unit test for creating a user without an email address?

What happens if we create a user with neither email nor phone?

app/org/sagebionetworks/bridge/validators/SignInValidator.java
```diff
@@ -29,14 +31,18 @@
     /** Sign in using token sent through SMS. */
     public static final SignInValidator PHONE_SIGNIN = new SignInValidator(EnumSet.of(STUDY, PHONE, TOKEN));
 
+    /** Request a reset password link via email or SMS. */
+    public static final SignInValidator REQUEST_RESET_PASSWORD = new SignInValidator(EnumSet.of(STUDY, EMAIL_OR_PHONE));
```
DwayneJengSage, line 35: Should we add unit tests for REQUEST_RESET_PASSWORD?

test/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
```diff
@@ -93,9 +86,11 @@ public void validatesUpdate() {
             assertNull(e.getErrors().get("email"));
             assertNull(e.getErrors().get("externalId"));
             assertNull(e.getErrors().get("password"));
-            assertError(e, "dataGroups", 0, " 'badGroup' is not defined for study (use group1, group2, bluebell)");
-            assertError(e, "attributes", 0, " 'badValue' is not defined for study (use attr1, attr2, phone)");
+            //assertError(e, "dataGroups", 0, " 'badGroup' is not defined for study (use group1, group2, bluebell)");
+            //assertError(e, "attributes", 0, " 'badValue' is not defined for study (use attr1, attr2, phone)");
```
DwayneJengSage: Minor: Remove commented out tests.

test/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
```diff
@@ -120,108 +115,95 @@ public void validPasses() {
...
-        assertCorrectMessage(withEmail(null), "email", "email is required");
+        assertValidatorMessage(validator, withEmail(null), "StudyParticipant", "email or phone is required");
     }
     
     @Test
     public void passwordRequired() {
         validator = new StudyParticipantValidator(study, true);
-        assertCorrectMessage(withPassword(""), "password", "password is required");
+        assertValidatorMessage(validator, withPassword(""), "password", "is required");
     }
```
DwayneJengSage: Unit test for null password okay?

test/org/sagebionetworks/bridge/TestUserAdminHelper.java
```diff
@@ -112,9 +112,10 @@ public void deleteUser(TestUser testUser) {
         checkNotNull(testUser);
         
         if (testUser.getSession() != null) {
+            String userId = testUser.getId();
             authService.signOut(testUser.getSession());
+            deleteUser(testUser.getStudy(), userId);
         }
-        deleteUser(testUser.getStudy(), testUser.getId());
     }
```
DwayneJengSage: Shouldn't the user be deleted even if they have no session?

test/org/sagebionetworks/bridge/TestUtils.java
```diff
@@ -117,7 +117,13 @@ public static void assertValidatorMessage(Validator validator, Object object, St
             Validate.entityThrowingException(validator, object);
             fail("Should have thrown exception");
         } catch(InvalidEntityException e) {
-            assertEquals(fieldNameAsLabel+error, e.getErrors().get(fieldName).get(0));
+            for (int i=0; i < e.getErrors().get(fieldName).size(); i++) {
+                String msg = e.getErrors().get(fieldName).get(i);
+                if ((fieldNameAsLabel+error).equals(msg)) {
+                    return;
+                }
+            }
```
DwayneJengSage: Would e.getErrors().get(fieldName).contains(fieldNameAsLabel+error) work here?

test/org/sagebionetworks/bridge/services/UserAdminServiceMockTest.java
```diff
@@ -149,9 +152,7 @@ public void creatingUserConsentsToAllRequiredConsents() {
         assertEquals(participant.getEmail(), signIn.getEmail());
         assertEquals(participant.getPassword(), signIn.getPassword());
         
-        for (SubpopulationGuid guid : session.getConsentStatuses().keySet()) {
-            verify(consentService).consentToResearch(eq(study), eq(guid), any(StudyParticipant.class), any(), eq(SharingScope.NO_SHARING), eq(false));
-        }
```
DwayneJengSage, line 154: Why did we remove this from the test? Does creating a user with the consentUser no longer function?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/BridgeConstants.java
alxdarksage: Yes, fixed this.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: If we got a concurrent modification exception, this account should really have an email or phone number... it seems fine to allow an NPE further along in the code path.

app/org/sagebionetworks/bridge/models/accounts/SignIn.java
alxdarksage: Added serialization test to SignInTest.

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
alxdarksage: Added verifyCommonLoggingForSignIns() for these methods.

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
alxdarksage: Added

app/org/sagebionetworks/bridge/services/AccountWorkflowService.java
alxdarksage: Yes good point. I've changed email cacheKey back to the original format, and added phone as a new cache key (which is being deployed with the phone changes). Should be backwards-compatible.

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerMockTest.java
alxdarksage: Added

test/org/sagebionetworks/bridge/services/AccountWorkflowServiceTest.java
alxdarksage: Added

test/org/sagebionetworks/bridge/services/AccountWorkflowServiceTest.java
alxdarksage: Re-arranged these

test/org/sagebionetworks/bridge/services/AccountWorkflowServiceTest.java
alxdarksage: Added

app/org/sagebionetworks/bridge/services/NotificationsService.java
alxdarksage: It's 140 bytes in GSM, which is not a standard  encoding in the JDK, but it is a 7-bit-per-character encoding with oddities, so we can approximate a measure with ASCII. It might throw an error or send a split message if too long... which is probably acceptable as it should ultimately come out during testing.

app/org/sagebionetworks/bridge/services/UserAdminService.java
alxdarksage: Good questions: failures occur. Moved the code around to validate that the minimum (study + email or phone) is included to create the user. Added test for this.

app/org/sagebionetworks/bridge/services/UserAdminService.java
alxdarksage: No, we don't, can also be called with an account using only phone. Changed and verified with test.

app/org/sagebionetworks/bridge/validators/SignInValidator.java
alxdarksage: Added

test/org/sagebionetworks/bridge/TestUserAdminHelper.java
alxdarksage: I've restored this. This is definitely an error.

test/org/sagebionetworks/bridge/TestUtils.java
alxdarksage: Yep. Changed.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: Added.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: Changed to verify these directly.

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
alxdarksage: Added, along with a test for the case where an email exists (which is covered currently in the test that starts a Spring context).

test/org/sagebionetworks/bridge/services/UserAdminServiceMockTest.java
alxdarksage: I see no reason, I've restored it and updated the tests to pass (the method was itself reorganized).

test/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
alxdarksage: Removed

test/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
alxdarksage: Yes it's now different from this case. Added an additional test.

test/org/sagebionetworks/bridge/services/AccountWorkflowServiceTest.java
```diff
@@ -217,6 +259,56 @@ public void requestResetPassword() throws Exception {
...
+        when(mockAccountDao.getAccount(ACCOUNT_ID_WITH_PHONE)).thenReturn(mockAccount);
+        when(mockStudyService.getStudy(TEST_STUDY)).thenReturn(study);
+        when(mockAccount.getPhone()).thenReturn(TestConstants.PHONE);
+        when(mockAccount.getPhoneVerified()).thenReturn(Boolean.TRUE);        
+        when(mockAccount.getStudyIdentifier()).thenReturn(TEST_STUDY);
+        
+        service.requestResetPassword(study, ACCOUNT_ID_WITH_PHONE);
+        
+        verify(mockCacheProvider).setString(eq("ABC:phone:api"), stringCaptor.capture(), eq(60*60*2));
+        verify(mockNotificationsService).sendSMSMessage(eq(TEST_STUDY), eq(TestConstants.PHONE), stringCaptor.capture());
```
DwayneJengSage: We should use 2 separate ArgumentCaptors here. That would make it more readable, easier to update, and less likely to have faulty tests.

app/org/sagebionetworks/bridge/validators/SignInValidator.java
```diff
@@ -29,14 +31,21 @@
...
+    
+    /** The basics of a sign in that must be present for the admin create user API. */
+    public static final SignInValidator MINIMAL = new SignInValidator(EnumSet.of(STUDY, EMAIL_OR_PHONE));
```
DwayneJengSage, line 38: MINIMAL and REQUEST_RESET_PASSWORD are the same. Should they be the same validator instance? Or do we want to keep the separate for organizational purposes and/or potential future modifications?

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -373,4 +418,111 @@ public void reauthThrowsUnconsentedException() {
...
```
DwayneJengSage: Is it worth verifying that service.phoneSignIn() returns a UserSessionInfo with the basic attributes we expect?

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerMockTest.java
```diff
@@ -536,22 +542,18 @@ public void signInNewSessionUnconsentedWorker() throws Exception {
...
-        doReturn(metrics).when(controller).getMetrics();
-
         // mock AuthenticationService
         UserSession session = createSession(TestConstants.REQUIRED_SIGNED_CURRENT, null);
         when(authenticationService.getSession(TEST_SESSION_TOKEN)).thenReturn(session);
 
         // execute and validate
         Result result = controller.signOut();
         assertEquals(HttpStatus.SC_OK, result.status());
-        assertSessionInfoInMetrics(metrics);
```
DwayneJengSage, line 554: verifyMetrics()?

DwayneJengSage (approved):
Looks good. Just some minor comments.

PR #1603, opened 2017-11-17
Switch to using new Bridge Exporter queue

zaro0508:
The infra for Bridge-Exporter app is now automated with
Bridge-Exporter-infra[1] repo.  This project uses cloudformation
to setup the EB worker env and SQS queues. We should switch to using
this queue and deprecate the manually created SQS queues.

[1] https://github.com/Sage-Bionetworks/Bridge-Exporter-infra


DwayneJengSage (approved):

DwayneJengSage:
The Bridge Worker migration was a huge mess. It's taking a significant amount of manual dev effort to deal with all the issues that came about as a result of it.

Given that the Bridge Exporter is a critical component, and given that we already have anecdotal evidence that this style of migration will go badly, I think we need a better plan to do a phased migration of the Bridge Exporter. Closing this, so we don't accidentally push it out.

PR #1607, opened 2017-11-30
Update synapse client

alxdarksage:
Fix for BRIDGE-1985, and some other observed issues with the project/team as they are created:

- Update synapse client version to eliminate problem where users show up in Synapse UI with "custom" permissions;
- Reduce permissions of the access team so new team members get read/download permissions (only);
- Do not send email verification emails while creating Bridge participants, because the email you add to the study will almost always be unverified by SES, and this throws an error while creating study. Instead, we're going to create these initial users verified, and instruct them in a follow-up email to go reset the password.
- Fixed the names, which were ugly, and added a random string at the end, to ensure they never conflict with any other GLOBALLY (!?) unique team or project name in Synapse. If there's any failure, the systems are left in an inconsistent state which is highly undesirable.

### DIFF
--- a/app/org/sagebionetworks/bridge/BridgeUtils.java
+++ b/app/org/sagebionetworks/bridge/BridgeUtils.java
@@ -4,7 +4,6 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static java.lang.Integer.parseInt;
 import static org.apache.commons.lang3.StringUtils.isBlank;
-import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static org.springframework.util.StringUtils.commaDelimitedListToSet;
 
 import java.io.UnsupportedEncodingException;
@@ -41,7 +40,6 @@ import org.springframework.core.annotation.AnnotationUtils;
 import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper.FailedBatch;
 import com.google.common.base.Function;
 import com.google.common.base.Joiner;
-import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
--- a/app/org/sagebionetworks/bridge/BridgeUtils.java
+++ b/app/org/sagebionetworks/bridge/BridgeUtils.java
@@ -1,8 +1,10 @@
 package org.sagebionetworks.bridge;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static java.lang.Integer.parseInt;
 import static org.apache.commons.lang3.StringUtils.isBlank;
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static org.springframework.util.StringUtils.commaDelimitedListToSet;
 
 import java.io.UnsupportedEncodingException;
@@ -39,6 +41,7 @@ import org.springframework.core.annotation.AnnotationUtils;
 import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper.FailedBatch;
 import com.google.common.base.Function;
 import com.google.common.base.Joiner;
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
@@ -399,9 +402,11 @@ public class BridgeUtils {
     }
     
     public static String toSynapseFriendlyName(String input) {
-        if (input == null) {
-            return "";
-        }
-        return input.replaceAll("[^a-zA-Z0-9\\.\\-_\\s]", " ").replaceAll("\\s+", " ").trim();
+        checkNotNull(input);
+        
+        String value = input.replaceAll("[^a-zA-Z0-9\\.\\-_\\s]", " ").replaceAll("\\s+", " ").trim();
+        checkArgument(StringUtils.isNotBlank(value));
+        
+        return value; 
     }
 }
--- a/build.sbt
+++ b/build.sbt
@@ -19,7 +19,7 @@ libraryDependencies ++= Seq(
   filters,
   // Sage packages
   "org.sagebionetworks" % "bridge-base" % "2.7.6",
-  "org.sagebionetworks" % "synapseJavaClient" % "161.0-4-g843de2c",
+  "org.sagebionetworks" % "synapseJavaClient" % "206.0-9-gb27c77a5f7",
   // AWS
   "com.amazonaws" % "aws-java-sdk-s3" % "1.11.198",
   "com.amazonaws" % "aws-java-sdk-sqs" % "1.11.198",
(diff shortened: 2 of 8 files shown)


app/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -397,5 +397,11 @@ public static String createReferentGuidIndex(Activity activity, LocalDateTime lo
         }
         throw new BridgeServiceException("Invalid activityType specified");    
     }
-
+    
+    public static String toSynapseFriendlyName(String input) {
+        if (input == null) {
+            return "";
+        }
```
DwayneJengSage: I'm pretty sure "" is probably not a Synapse-friendly name either. (Or if it is, it's not a particularly useful one.) Would it be better to just throw on a null input?

test/org/sagebionetworks/bridge/BridgeUtilsTest.java
```diff
@@ -387,6 +387,22 @@ public void getLocalDateWithError() {
...
+    
+    @Test
+    public void nullToSynapseFriendlyName() {
+        assertEquals("", BridgeUtils.toSynapseFriendlyName(null));
+    }
+    
+    @Test
+    public void emptyStringToSynapseFriendlyName() {
+        assertEquals("", BridgeUtils.toSynapseFriendlyName("  #"));
+    }
```
DwayneJengSage: If the result is an empty string, should that be an error?

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
```diff
@@ -484,12 +492,15 @@ public void createStudyAndUser() throws SynapseException {
...
-        verify(participantService).createParticipant(eq(study), eq(mockUser2.getRoles()), eq(mockUser2), eq(true));
+        verify(participantService).createParticipant(eq(study), eq(mockUser1.getRoles()), eq(mockUser1), eq(false));
+        verify(participantService).createParticipant(eq(study), eq(mockUser2.getRoles()), eq(mockUser2), eq(false));
         verify(participantService, times(2)).requestResetPassword(eq(study), eq(mockIdentifierHolder.getIdentifier()));
         verify(mockSynapseClient, times(2)).newAccountEmailValidation(any(), eq(SYNAPSE_REGISTER_END_POINT));
         verify(service).createStudy(study);
         verify(service).createSynapseProjectTeam(TEST_ADMIN_IDS, study);
+        
+        assertTrue(projectCaptor.getValue().getName().contains(TEST_PROJECT_NAME));
+        assertTrue(teamCaptor.getValue().getName().contains(TEST_TEAM_NAME));
```
DwayneJengSage: Minor: Is it worth using startsWith() as a stronger test than contains()?

Minor: Is it worth checking that the name isn't exactly TEST_NNNN_NAME to make sure we're appending that extra random bit?

build.sbt
```diff
@@ -19,7 +19,7 @@ libraryDependencies ++= Seq(
   filters,
   // Sage packages
   "org.sagebionetworks" % "bridge-base" % "2.7.7",
-  "org.sagebionetworks" % "synapseJavaClient" % "161.0-4-g843de2c",
+  "org.sagebionetworks" % "synapseJavaClient" % "206.0-9-gb27c77a5f7",
```
DwayneJengSage: I believe the most up-to-date version is 206.0. The "-9-whatever" seems to be a one-off incremental build.

DwayneJengSage (changes requested):

build.sbt
alxdarksage: Changed this, works fine

app/org/sagebionetworks/bridge/BridgeUtils.java
alxdarksage: This now throws exceptions now if the value is blank or null. It could happen if the study name was particularly strange, but almost certainly will not. Also added tests to verify these are converted to BadRequestException by the service.

test/org/sagebionetworks/bridge/BridgeUtilsTest.java
alxdarksage: Now throws an IllegalArgumentException

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
alxdarksage: Switched to startsWith(). The TEST_PROJECT_NAME/TEST_TEAM_NAME constants include the strings we're appending to the study name as converted.

test/org/sagebionetworks/bridge/BridgeUtilsTest.java
```diff
@@ -387,6 +387,22 @@ public void getLocalDateWithError() {
...
+    @Test
+    public void toSynapseFriendlyName() {
+        assertEquals("This is a .-_ synapse Friendly Name3",
+                BridgeUtils.toSynapseFriendlyName("This (is a).-_ synapse Friendly Name3 "));
+    }
+    
+    @Test(expected = NullPointerException.class)
+    public void nullToSynapseFriendlyNameThrowsException() {
+        assertEquals("", BridgeUtils.toSynapseFriendlyName(null));
+    }
```
DwayneJengSage: Minor: No need for the assert here or in emptyStringToSynapseFriendlyName(), because BridgeUtils will throw before the assert is called.

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
```diff
@@ -484,12 +492,15 @@ public void createStudyAndUser() throws SynapseException {
...
-        verify(participantService).createParticipant(eq(study), eq(mockUser2.getRoles()), eq(mockUser2), eq(true));
+        verify(participantService).createParticipant(eq(study), eq(mockUser1.getRoles()), eq(mockUser1), eq(false));
+        verify(participantService).createParticipant(eq(study), eq(mockUser2.getRoles()), eq(mockUser2), eq(false));
         verify(participantService, times(2)).requestResetPassword(eq(study), eq(mockIdentifierHolder.getIdentifier()));
         verify(mockSynapseClient, times(2)).newAccountEmailValidation(any(), eq(SYNAPSE_REGISTER_END_POINT));
         verify(service).createStudy(study);
         verify(service).createSynapseProjectTeam(TEST_ADMIN_IDS, study);
+        
+        assertTrue(projectCaptor.getValue().getName().startsWith(TEST_PROJECT_NAME));
+        assertTrue(teamCaptor.getValue().getName().startsWith(TEST_TEAM_NAME));
```
DwayneJengSage: Don't we also append a random string to the project name and team name? Should we be testing that as well? Otherwise, if we somehow fail to append the random string, this test will still pass, because TEST_PROJECT_NAME starts with itself.

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/BridgeUtilsTest.java
alxdarksage: Yep, removed these

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
alxdarksage: Factored out random string generation to a method that can be mocked with the spy, updated the test constants to include the random bit and converted these to assertEquals tests.

DwayneJengSage (approved):

PR #1608, opened 2017-12-01
Oauth APIs.

alxdarksage:
Specifically to support contacting Fitbit.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/OAuthProviderService.java
+++ b/app/org/sagebionetworks/bridge/services/OAuthProviderService.java
@@ -1,0 +1,265 @@
+package org.sagebionetworks.bridge.services;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.UnsupportedEncodingException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.http.NameValuePair;
+import org.apache.http.client.entity.UrlEncodedFormEntity;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.message.BasicNameValuePair;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.sagebionetworks.bridge.BridgeUtils;
+import org.sagebionetworks.bridge.exceptions.BadRequestException;
+import org.sagebionetworks.bridge.exceptions.BridgeServiceException;
+import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
+import org.sagebionetworks.bridge.exceptions.UnauthorizedException;
+import org.sagebionetworks.bridge.json.BridgeObjectMapper;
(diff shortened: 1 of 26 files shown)


app/org/sagebionetworks/bridge/services/OAuthService.java
```diff
@@ -0,0 +1,125 @@
...
+            // If no grant was saved or successfully returned from a grant, it's not found.
+            if (grant == null) {
+                throw new EntityNotFoundException(OAuthAccessGrant.class);
+            } else if (getDateTime().isAfter(grant.getExpiresOn())) {
+                // If there's a grant record, but it has expired, attempt to refresh it
+                grant = providerService.refreshAccessGrant(provider, vendorId, grant.getRefreshToken());
+            }
+        } catch(Exception e) {
+            // any error, delete the grant because it is in an unknown and probably bad state.
+            grantDao.deleteAccessGrant(study.getStudyIdentifier(), vendorId, healthCode);
```
liujoshua, line 112: It seems we should not delete for errors which may succeed if retried later (like 503s). Right now there's no mechanism to ask the user to redo the grant flow (which seems like a gap) and it's more conservative to keep around a grant that may fail again, than delete a grant that may succeed later.

liujoshua (approved):
Looks good to me

app/org/sagebionetworks/bridge/services/OAuthService.java
alxdarksage: This can be changed, but it's absolutely the case that we can lose the right to access the API, and the application design needs to account for that or it'll just be broken.

app/org/sagebionetworks/bridge/services/OAuthService.java
```diff
@@ -0,0 +1,125 @@
...
+            if (authToken != null && authToken.getAuthToken() != null) {
+                grant = providerService.requestAccessGrant(provider, authToken);
+            } else {
+                // If not, start first by seeing if a grant has been saved
+                grant = grantDao.getAccessGrant(study.getStudyIdentifier(), vendorId, healthCode);
+            }
+            // If no grant was saved or successfully returned from a grant, it's not found.
+            if (grant == null) {
+                throw new EntityNotFoundException(OAuthAccessGrant.class);
+            } else if (getDateTime().isAfter(grant.getExpiresOn())) {
```
liujoshua, line 106: It's possible this grant has not expired but has been revoked. If the worker detects this, should it be able to notify bridge?

app/org/sagebionetworks/bridge/services/OAuthService.java
```diff
@@ -0,0 +1,125 @@
...
+                // If there's a grant record, but it has expired, attempt to refresh it
+                grant = providerService.refreshAccessGrant(provider, vendorId, grant.getRefreshToken());
```
liujoshua, line 108: If this part fails, a mechanism to get a new grant from the participant would be useful

app/org/sagebionetworks/bridge/services/OAuthProviderService.java
```diff
@@ -0,0 +1,271 @@
...
+        // Other 403 exceptions indicate a permissions issue, possibly based on scope or something else
+        // that Bridge doesn't control.
+        else if (statusCode == 403) {
+            throw new UnauthorizedException(jsonToErrorMessage(response.getBody()));
+        } 
+        // Other bad request, are bad requests... possibly due to input from the client
+        else if (statusCode > 399 && statusCode < 500) {
+            throw new BadRequestException(jsonToErrorMessage(response.getBody()));
+        } 
+        // And everything, for now, can be treated as Bridge server error.
```
alxdarksage, line 205: "And everything else"

PR #1609, opened 2017-12-05
Changing order and kind of exceptions that are thrown for disabled/unverified accounts.

alxdarksage:
Creates a v4 of signIn and reauthenticate with a couple of changes:

- failures to sign in always thrown "not found exceptions" now, and not exceptions indicating the account is unverified or disabled
- v4 versions of the APIs throw UnauthorizedException if the account is not verified (neither email or phone has been verified). v3 still returns a 404, but the 403 allows the client to know that the account exists, but the user has yet to verify it.


liujoshua (approved):
Look great. Thanks

PR #1613, opened 2017-12-09
Generalizing CacheProvider in preparation to cache subpopulations.

alxdarksage:
Creating a generic set of methods to cache objects, and refactoring to use them. Because otherwise I would have to add subpopulation-specific methods, which is unnecessary.

And also the intent to participate API, where I'd like to use the same object-based methods to cache the Intent to Participate.

### DIFF
--- a/app/org/sagebionetworks/bridge/cache/CacheProvider.java
+++ b/app/org/sagebionetworks/bridge/cache/CacheProvider.java
@@ -6,6 +6,7 @@ import java.util.List;
 
 import javax.annotation.Resource;
 
+import org.sagebionetworks.bridge.BridgeConstants;
 import org.sagebionetworks.bridge.config.BridgeConfigFactory;
 import org.sagebionetworks.bridge.exceptions.BridgeServiceException;
 import org.sagebionetworks.bridge.json.BridgeObjectMapper;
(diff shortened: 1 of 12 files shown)


app/org/sagebionetworks/bridge/cache/CacheProvider.java
```diff
@@ -215,72 +187,104 @@ public void removeSessionByUserId(final String userId) {
...
-            if (!"OK".equals(result)) {
-                throw new BridgeServiceException("Study storage error");
-            }
-        } catch (Throwable e) {
-            promptToStartRedisIfLocal(e);
-            throw new BridgeServiceException(e);
-        }
+        checkNotNull(study);
+        String redisKey = RedisKey.STUDY.getRedisKey(study.getIdentifier());
+        setObject(redisKey, study);
```
zaro0508: shouldn't this set an expiration?

app/org/sagebionetworks/bridge/cache/CacheProvider.java
```diff
@@ -215,72 +187,104 @@ public void removeSessionByUserId(final String userId) {
...
     }
 
     public Study getStudy(String identifier) {
+        checkNotNull(identifier);
+        String redisKey = RedisKey.STUDY.getRedisKey(identifier);
+        return getObject(redisKey, Study.class);
```
zaro0508: shouldn't this set an expiration?

zaro0508 (approved):
LGTM

PR #1615, opened 2017-12-12
Cache subpopulations

alxdarksage:
Some logic that is more business logic than persistence was moved into the service, which makes it far easier to cache at the level of persistence only.


app/org/sagebionetworks/bridge/cache/CacheProvider.java
```diff
@@ -220,6 +221,21 @@ public void removeStudy(String identifier) {
         return null;
     }
     
+    public <T> T getObject(String cacheKey, TypeReference<T> typeRef) {
```
zaro0508, line 224: This looks ok, but it doesn't seem to be used anywhere?

test/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDaoMockTest.java
```diff
@@ -125,43 +124,44 @@ public void getSubpopulationConstructsCriteriaIfNotSaved() {
...
+        doReturn(defaultSubpop).when(dao).getSubpopulation(TEST_STUDY, SUBPOP_GUID);
+        doReturn(Criteria.create()).when(criteriaDao).getCriteria(any());
+        
+        dao.deleteSubpopulation(TEST_STUDY, SUBPOP_GUID, true, false);
     }
     
     @Test
-    public void logicalDeleteSubpopulationDoesNotDeleteCriteria() {
-        dao.deleteSubpopulation(TEST_STUDY, SUBPOP_GUID, false);
+    public void allowDeleteOfDefault() {
```
zaro0508: Looks like this checks physical delete of default subpopulation.  Is there a case where you would do a logical delete of default?  it doesn't look like that case is covered.

app/org/sagebionetworks/bridge/cache/CacheProvider.java
alxdarksage: It is used in SubpopulationService line 172 to get the cached list of subpopulations

test/org/sagebionetworks/bridge/dynamodb/DynamoSubpopulationDaoMockTest.java
alxdarksage: Added test that logical delete also cannot delete the default (unless you're calling from the delete everything service method which sets the flag to override this, used for tests).

app/org/sagebionetworks/bridge/cache/CacheProvider.java
zaro0508: thanks.  i really should figure out how to setup intelliji to make it correctly compile the project so i can use the 'usage' feature.

zaro0508 (approved):
LGTM

PR #1616, opened 2017-12-12
Intent to participate API.

alxdarksage:
- Prior to creating an account, you can submit a consent signature for a study/subpopulation. On sign up, if the client requests it in the signUp payload, we'll look for this consent and consent the user, preventing the return of a 412 response.

- The API can also be configured to send an SMS message to the caller of the intent API. This is expected to be a link to an App Store, but it does not need to be.


app/org/sagebionetworks/bridge/models/itp/IntentToParticipate.java
```diff
@@ -0,0 +1,98 @@
...
+        this.phone = phone;
+        this.subpopGuid = subpopGuid;
+        this.scope = scope;
+        this.osName = osName;
+        this.consentSignature = consentSignature;
+    }
+
+    public String getStudy() {
+        return study;
+    }
```
DwayneJengSage: Nitpick: studyId instead of study?

app/org/sagebionetworks/bridge/models/itp/IntentToParticipate.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+        public Builder withConsentSignature(ConsentSignature consentSignature) {
+            this.consentSignature = consentSignature;
+            return this;
+        }
+        public IntentToParticipate build() {
+            // Same adjustments that we do for ClientInfo, to normalize the name if it changes.
+            if (OperatingSystem.SYNONYMS.containsKey(osName)) {
+                osName = OperatingSystem.SYNONYMS.get(osName);
+            }
```
DwayneJengSage, line 94: Is it worth making a unit test (separate from the JSON test, as it gets buried in there) that tests that we have OS synonyms hooked up?

app/org/sagebionetworks/bridge/play/controllers/IntentController.java
```diff
@@ -0,0 +1,40 @@
...
+        
+        // An early hack in the system was that sharing scope was added to the consent signature 
+        // JSON even though it is not part of the signature. We need to move that value because 
+        // the client API continues to treat sharing as part of the consent signature.
+        JsonNode requestNode = requestToJSON(request());
+        if (requestNode != null && requestNode.has("consentSignature")) {
+            SharingOption sharing = SharingOption.fromJson(requestNode.get("consentSignature"), 2);
+            intent = new IntentToParticipate.Builder().copyOf(intent)
+                    .withScope(sharing.getSharingScope()).build();
+        }
```
DwayneJengSage, line 34: Would it be cleaner to put this logic in IntentToParticipate.Builder.build()? This centralizes the invariants of ITP and ensures that scope is always present in getScope() or not at all.

test/org/sagebionetworks/bridge/services/AuthenticationServiceTest.java
```diff
@@ -132,6 +141,43 @@ public void signInInvalidCredentials() throws Exception {
...
                 new SignIn.Builder().withStudy(study.getIdentifier()).withEmail("foobar").withPassword("bar").build());
     }
+    
+    @Test
+    public void signUpWithIntentToParticipate() throws Exception {
+        String email = TestUtils.makeRandomTestEmail(AuthenticationServiceTest.class);
+        StudyParticipant participant = new StudyParticipant.Builder()
+                .withEmail(email)
+                .withPassword("P@ssword`1").build();
+        IdentifierHolder holder = null;
```
DwayneJengSage: We want to move away from real dependencies in unit tests. Would it be better to have a test in AuthenticationServiceMockTest that tests dataflow and wiring, and have this test be in Integ Tests instead?

app/org/sagebionetworks/bridge/services/IntentService.java
```diff
@@ -0,0 +1,139 @@
...
+        this.notificationsService = notificationsService;
+    }
+    
+    @Autowired
+    final void setAccountDao(AccountDao accountDao) {
+        this.accountDao = accountDao;
+    }
+    
+    public void submitIntentToParticipate(IntentToParticipate intent) {
+        Validate.entityThrowingException(IntentToParticipateValidator.INSTANCE, intent);
```
DwayneJengSage, line 74: Add unit test for invalid intent?

app/org/sagebionetworks/bridge/validators/StudyValidator.java
```diff
@@ -157,6 +157,17 @@ public void validate(Object obj, Errors errors) {
...
+        // Links in installedLinks are length-constrained by SMS.
+        if (!study.getInstallLinks().isEmpty()) {
+            for (Map.Entry<String,String> entry : study.getInstallLinks().entrySet()) {
+                if (StringUtils.isBlank(entry.getValue())) {
+                    errors.rejectValue("installLinks", "cannot be blank");
+                } else if (entry.getValue().length() > BridgeConstants.SMS_CHARACTER_LIMIT) {
+                    errors.rejectValue("installLinks", "cannot be longer than "+BridgeConstants.SMS_CHARACTER_LIMIT+" characters");
+                }
+            }
+        }
```
DwayneJengSage: Add to StudyValidatorTest?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/models/itp/IntentToParticipate.java
alxdarksage: Changed

app/org/sagebionetworks/bridge/models/itp/IntentToParticipate.java
alxdarksage: Added a separate test to verify this

app/org/sagebionetworks/bridge/play/controllers/IntentController.java
alxdarksage: Because this is about converting the external format (JSON) to the internal format (IntentToParticipate), and scope is not part of ConsentSignature, we'd have to introduce a fromJson factory method, which is more work to maintain. I would prefer to translate this in the controller.

app/org/sagebionetworks/bridge/services/IntentService.java
alxdarksage: Added

DwayneJengSage (approved):

app/org/sagebionetworks/bridge/validators/StudyValidator.java
```diff
@@ -144,30 +144,6 @@ public void validate(Object obj, Errors errors) {
...
-        // Links in installedLinks are length-constrained by SMS.
-        if (!study.getInstallLinks().isEmpty()) {
-            for (Map.Entry<String,String> entry : study.getInstallLinks().entrySet()) {
-                if (StringUtils.isBlank(entry.getValue())) {
-                    errors.rejectValue("installLinks", "cannot be blank");
-                } else if (entry.getValue().length() > BridgeConstants.SMS_CHARACTER_LIMIT) {
-                    errors.rejectValue("installLinks", "cannot be longer than "+BridgeConstants.SMS_CHARACTER_LIMIT+" characters");
-                }
-            }
-        }
```
DwayneJengSage: It looks like we deleted the install links from the validator. Given that we still have unit tests for these, this seems unintentional.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #1618, opened 2017-12-14
Adding ability to filter accounts by phone number substring.

alxdarksage:
At some point the number of parameters to the participant search might warrant a parameter object of some kind. Otherwise this change is trivial.


test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -328,41 +328,41 @@ public void createParticipantWithInvalidParticipant() {
     
     @Test
     public void getPagedAccountSummaries() {
-        participantService.getPagedAccountSummaries(STUDY, 1100, 50, "foo", START_DATE, END_DATE);
+        participantService.getPagedAccountSummaries(STUDY, 1100, 50, "foo", "bar", START_DATE, END_DATE);
```
zaro0508, line 331: [optional] maybe use string that's more representative of a phone number?

zaro0508 (approved):

PR #1623, opened 2018-01-04
Refactoring accountWorkflowService, pulling it up to the service layer to make adding email/phone easier.

alxdarksage:
On an update it will be possible to add an email address (or phone number). When that happens we'll need to send the email verification email for studies that are configured to send it, and that in turn is easier if we refactor to call the AccountWorkflowService from other services, and not from the AccountDao. So this refactor does that in a first check in.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/AuthenticationService.java
+++ b/app/org/sagebionetworks/bridge/services/AuthenticationService.java
@@ -293,7 +293,8 @@ public class AuthenticationService {
         checkNotNull(verification);
 
         Validate.entityThrowingException(EmailVerificationValidator.INSTANCE, verification);
-        accountDao.verifyEmail(verification);
+        Account account = accountWorkflowService.verifyEmail(verification);
+        accountDao.verifyEmail(account);
     }
     
     public void resendEmailVerification(StudyIdentifier studyIdentifier, Email email) {
@@ -303,7 +304,7 @@ public class AuthenticationService {
         Validate.entityThrowingException(EmailValidator.INSTANCE, email);
         try {
             AccountId accountId = AccountId.forEmail(studyIdentifier.getIdentifier(), email.getEmail());
-            accountDao.resendEmailVerificationToken(accountId);    
+            accountWorkflowService.resendEmailVerificationToken(accountId);    
         } catch(EntityNotFoundException e) {
             // Suppress this. Otherwise it reveals if the account does not exist
             LOG.info("Resend email verification for unregistered email in study '"+studyIdentifier.getIdentifier()+"'");
@@ -317,7 +318,7 @@ public class AuthenticationService {
         // validate the data in signIn, then convert it to an account ID which we know will be valid.
         Validate.entityThrowingException(SignInValidator.REQUEST_RESET_PASSWORD, signIn);
         try {
-            accountDao.requestResetPassword(study, signIn.getAccountId());    
+            accountWorkflowService.requestResetPassword(study, signIn.getAccountId());    
         } catch(EntityNotFoundException e) {
             // Suppress this. Otherwise it reveals if the account does not exist
             LOG.info("Request reset password request for unregistered email in study '"+signIn.getStudyId()+"'");
@@ -329,7 +330,7 @@ public class AuthenticationService {
 
         Validate.entityThrowingException(passwordResetValidator, passwordReset);
         
-        accountDao.resetPassword(passwordReset);
+        accountWorkflowService.resetPassword(passwordReset);
     }
     
     protected String getEmailToken() {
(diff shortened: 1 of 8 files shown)


test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
```diff
@@ -723,17 +690,12 @@ public void createAccountSuccess() {
...
-        verify(mockAccountWorkflowService, never()).sendEmailVerificationToken(any(), any(), any());
     }
 
     @Test
     public void createAccountVerifyEmail() {
-        // Most of this is tested in createAccountSuccess(). Just test email workflow.
-        String accountId = dao.createAccount(STUDY, makeValidGenericAccount(), true);
-        verify(mockAccountWorkflowService).sendEmailVerificationToken(STUDY, accountId, EMAIL);
-
+        dao.createAccount(STUDY, makeValidGenericAccount());
```
DwayneJengSage: Now that we don't have verify email in the AccountDao, is this redundant with createAccountSuccess()?

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -305,7 +305,8 @@ public void verifyEmail(EmailVerification verification) {
         checkNotNull(verification);
 
         Validate.entityThrowingException(EmailVerificationValidator.INSTANCE, verification);
-        accountDao.verifyEmail(verification);
+        Account account = accountWorkflowService.verifyEmail(verification);
+        accountDao.verifyEmail(account);
```
DwayneJengSage, line 309: It looks like we don't have unit tests for a lot of the methods in AuthenticationService. Should we add them?

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -260,10 +268,18 @@ public IdentifierHolder createParticipant(Study study, Set<Roles> callerRoles, S
...
+        account.setStatus(sendVerifyEmail ? AccountStatus.UNVERIFIED : AccountStatus.ENABLED);
+        
+        String accountId = accountDao.createAccount(study, account);
+
         externalIdService.assignExternalId(study, participant.getExternalId(), account.getHealthCode());
         optionsService.setAllOptions(study.getStudyIdentifier(), account.getHealthCode(), options);
+        // send verify email
+        if (sendVerifyEmail) {
+            accountWorkflowService.sendEmailVerificationToken(study, accountId, account.getEmail());
+        }
```
DwayneJengSage, line 282: Is it worth adding unit tests for (1) requestSendVerifyEmail=false (2)requestSendVerifyEmail=true and (3) requestSendVerifyEmail=true but isEmailVerificationEnabled=false?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -938,21 +933,6 @@ public void throwLimitExceededException() {
         participantService.createParticipant(STUDY, CALLER_ROLES, PARTICIPANT, false);
     }
     
-    private void verifyStatusCreate(Set<Roles> callerRoles) {
-        mockHealthCodeAndAccountRetrieval();
-        
-        StudyParticipant participant = new StudyParticipant.Builder().copyOf(PARTICIPANT)
-                .withStatus(AccountStatus.ENABLED).build();
```
DwayneJengSage, line 945: Why were these tests deleted?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage: Between AuthenticationServiceTest and AuthenticationServiceMockTest, all the methods are covered except verifyEmail(...), so I added a couple of tests to the mock test for that.

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Added 4 tests specifically for this. Only the case where email verification is enabled, and the caller wants email verification, calls the workflow service (and account status is tied 1:1 to this, account is only unverified if we're going to send the email).

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
alxdarksage: Yes it is, removed it.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: These tests were supposed to verify that status was not changed, but it is now changed in the service (and that is now verified in other tests, like those that call verifyStatusUpdate... which looks at impact of roles on whether or not you can change the status, and the tests you just asked me to add to test calls to accountWorkflowService).

DwayneJengSage (approved):

PR #1625, opened 2018-01-10
Fix persistence of email and phone sign in sessions.

alxdarksage:
On a phone sign in everything works but the session wasn't persisted. Same would be true of email-based sign in. Fixed with tests.


DwayneJengSage (approved):
Looks good. Is it worth also writing an integ test to test this scenario?

PR #1626, opened 2018-01-11
Add identifier pt2

alxdarksage:
NOTE: I think we may close this an introduce an API specifically to add or maybe even change identifier values like phone/email, where you have to authenticate as part of the call. This would be to prevent someone from intercepting a session token and using it to gain control of the account.

It is now possible to add an email or phone number on an update to an account. If the study is configured for it, we will send out an email to verify the email address when added to an existing account.

### DIFF
--- a/app/org/sagebionetworks/bridge/hibernate/HibernateHelper.java
+++ b/app/org/sagebionetworks/bridge/hibernate/HibernateHelper.java
@@ -10,12 +10,12 @@ import javax.persistence.PersistenceException;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
 import org.hibernate.Transaction;
-import org.hibernate.exception.ConstraintViolationException;
 import org.hibernate.query.Query;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 import org.sagebionetworks.bridge.exceptions.ConcurrentModificationException;
+import org.sagebionetworks.bridge.exceptions.ConstraintViolationException;
 
 /** Encapsulates common scenarios for using Hibernate to make it easier to use. */
 @Component
@@ -38,7 +38,7 @@ public class HibernateHelper {
         } catch (PersistenceException ex) {
             // If you try to create a row that already exists, Hibernate will throw a PersistenceException wrapped in a
             // ConstraintViolationException.
-            if (ex.getCause() instanceof ConstraintViolationException) {
+            if (ex.getCause() instanceof org.hibernate.exception.ConstraintViolationException) {
                 throw new ConcurrentModificationException(
                         "Attempting to write a new row that violates key constraints");
             } else {
@@ -116,8 +116,9 @@ public class HibernateHelper {
             throw new ConcurrentModificationException("Row has the wrong version number; it may have " + 
                     "been saved in the background.");
         } catch (PersistenceException ex) {
-            if (ex.getCause() instanceof ConstraintViolationException) {
-                throw new ConcurrentModificationException(ex.getMessage());
+            if (ex.getCause() instanceof org.hibernate.exception.ConstraintViolationException) {
+                throw new ConstraintViolationException.Builder()
+                        .withMessage("Email or phone number conflicts with an existing account.").build();
             } else {
                 throw ex;
             }
--- a/app/org/sagebionetworks/bridge/hibernate/HibernateHelper.java
+++ b/app/org/sagebionetworks/bridge/hibernate/HibernateHelper.java
@@ -113,8 +113,14 @@ public class HibernateHelper {
                 return null;
             });
         } catch (OptimisticLockException ex) {
-            throw new ConcurrentModificationException("Row has the wrong version number; it may have been saved in " +
-                    "the background.");
+            throw new ConcurrentModificationException("Row has the wrong version number; it may have " + 
+                    "been saved in the background.");
+        } catch (PersistenceException ex) {
+            if (ex.getCause() instanceof ConstraintViolationException) {
+                throw new ConcurrentModificationException(ex.getMessage());
+            } else {
+                throw ex;
+            }
         }
     }
 
--- a/app/org/sagebionetworks/bridge/services/ParticipantService.java
+++ b/app/org/sagebionetworks/bridge/services/ParticipantService.java
@@ -271,6 +271,7 @@ public class ParticipantService {
         boolean sendVerifyEmail = requestSendVerifyEmail && study.isEmailVerificationEnabled();
         if (sendVerifyEmail) {
             account.setStatus(AccountStatus.UNVERIFIED);
+            account.setEmailVerified(Boolean.FALSE);
         } else {
             account.setStatus(AccountStatus.ENABLED);
             if (account.getEmail() != null) {
--- a/app/org/sagebionetworks/bridge/services/ParticipantService.java
+++ b/app/org/sagebionetworks/bridge/services/ParticipantService.java
@@ -269,8 +269,16 @@ public class ParticipantService {
         updateAccountOptionsAndRoles(study, callerRoles, options, account, participant);
         
         boolean sendVerifyEmail = requestSendVerifyEmail && study.isEmailVerificationEnabled();
-        
-        account.setStatus(sendVerifyEmail ? AccountStatus.UNVERIFIED : AccountStatus.ENABLED);
+        if (sendVerifyEmail) {
+            account.setStatus(AccountStatus.UNVERIFIED);
+        } else {
+            account.setStatus(AccountStatus.ENABLED);
+            if (account.getEmail() != null) {
+                account.setEmailVerified(Boolean.TRUE);
+            }
+            // If phone + password account, you will have to verify the number by signing in with it 
+            // before you can do things like requesting a password reset. See BRIDGE-2061.
+        }
         
         String accountId = accountDao.createAccount(study, account);
 
--- a/app/org/sagebionetworks/bridge/services/ParticipantService.java
+++ b/app/org/sagebionetworks/bridge/services/ParticipantService.java
@@ -284,8 +284,10 @@ public class ParticipantService {
 
         externalIdService.assignExternalId(study, participant.getExternalId(), account.getHealthCode());
         optionsService.setAllOptions(study.getStudyIdentifier(), account.getHealthCode(), options);
-        // send verify email
-        if (sendVerifyEmail) {
+        // send verify email. However even if that's desirable, don't try it if the account used a 
+        // phone number instead. The account is UNVERIFIED until the account signs in successfully with 
+        // a phone.
+        if (sendVerifyEmail && participant.getEmail() != null) {
             accountWorkflowService.sendEmailVerificationToken(study, accountId, account.getEmail());
         }
         return new IdentifierHolder(accountId);
--- a/app/org/sagebionetworks/bridge/services/ParticipantService.java
+++ b/app/org/sagebionetworks/bridge/services/ParticipantService.java
@@ -291,6 +291,9 @@ public class ParticipantService {
         Validate.entityThrowingException(new StudyParticipantValidator(study, false), participant);
         
         Account account = getAccountThrowingException(study, participant.getId());
+        
+        boolean sendVerificationEmail = (account.getEmail() == null && participant.getEmail() != null);
+        
         Map<ParticipantOption, String> options = Maps.newHashMap();
 
         // Do this first because if the ID has been taken or is invalid, we do not want to update anything else.
@@ -308,6 +311,10 @@ public class ParticipantService {
         }
         accountDao.updateAccount(account);
         optionsService.setAllOptions(study.getStudyIdentifier(), account.getHealthCode(), options);
+        
+        if (study.isEmailVerificationEnabled() && sendVerificationEmail) {
+            accountWorkflowService.sendEmailVerificationToken(study, account.getId(), participant.getEmail());
+        }
     }
 
     private void throwExceptionIfLimitMetOrExceeded(Study study) {
@@ -331,6 +338,17 @@ public class ParticipantService {
         account.setFirstName(participant.getFirstName());
         account.setLastName(participant.getLastName());
         account.setClientData(participant.getClientData());
+        // If the existing value has not been set and a value is being provided, then set it, but otherwise,
+        // we do not update the fields.
+        if (account.getEmail() == null && participant.getEmail() != null) {
+            account.setEmail(participant.getEmail());
+            account.setEmailVerified(Boolean.FALSE);
+        }
+        if (account.getPhone() == null && participant.getPhone() != null) {
+            account.setPhone(participant.getPhone());
+            account.setPhoneVerified(Boolean.FALSE);
+        }
+
         for (String attribute : study.getUserProfileAttributes()) {
             String value = participant.getAttributes().get(attribute);
             account.setAttribute(attribute, value);
--- a/test/org/sagebionetworks/bridge/hibernate/HibernateHelperTest.java
+++ b/test/org/sagebionetworks/bridge/hibernate/HibernateHelperTest.java
@@ -22,7 +22,6 @@ import com.google.common.collect.ImmutableList;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
 import org.hibernate.Transaction;
-import org.hibernate.exception.ConstraintViolationException;
 import org.hibernate.query.Query;
 import org.junit.Before;
 import org.junit.Test;
@@ -61,7 +60,7 @@ public class HibernateHelperTest {
     public void createConcurrentModificationException() {
         // mock session to throw - Need to mock the ConstraintViolationException, because the exception itself is
         // pretty heavy-weight.
-        PersistenceException ex = new PersistenceException(mock(ConstraintViolationException.class));
+        PersistenceException ex = new PersistenceException(mock(org.hibernate.exception.ConstraintViolationException.class));
         when(mockSession.save(any())).thenThrow(ex);
 
         // setup and execute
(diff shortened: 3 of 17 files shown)


test/org/sagebionetworks/bridge/hibernate/HibernateHelperTest.java
```diff
@@ -22,7 +22,6 @@
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
 import org.hibernate.Transaction;
-import org.hibernate.exception.ConstraintViolationException;
```
DwayneJengSage, line 25: Minor: Why did we remove this import? Was there a type conflict?

app/org/sagebionetworks/bridge/hibernate/HibernateHelper.java
```diff
@@ -113,8 +113,14 @@ public void update(Object obj) {
                 return null;
             });
         } catch (OptimisticLockException ex) {
-            throw new ConcurrentModificationException("Row has the wrong version number; it may have been saved in " +
-                    "the background.");
+            throw new ConcurrentModificationException("Row has the wrong version number; it may have " + 
+                    "been saved in the background.");
+        } catch (PersistenceException ex) {
+            if (ex.getCause() instanceof ConstraintViolationException) {
+                throw new ConcurrentModificationException(ex.getMessage());
```
DwayneJengSage: What error message will this be? Should we add our own error message to make it easier to read in the logs?

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
```diff
@@ -516,6 +520,36 @@ public void partialUpdateSelfParticipant() throws Exception {
...
+        String json = MAPPER.writeValueAsString(updatedParticipant);
+        mockPlayContextWithJson(json);
+
+        Result result = controller.updateSelfParticipant();
+        
+        assertEquals(200, result.status());
+        JsonNode node = TestUtils.getJson(result);
+        assertEquals("UserSessionInfo", node.get("type").textValue());
+        assertEquals(TestConstants.PHONE.getNumber(), node.get("phone").get("number").textValue());
+        assertEquals(EMAIL, node.get("email").textValue());
```
DwayneJengSage, line 550: Shouldn't this test return the updated participant, not the original one?

Also, is it worth adding a check that sessionUpdateService was called with the updated participant?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -362,6 +371,7 @@ public void createParticipantEmailEnabledNoVerificationWanted() {
         
         verify(accountWorkflowService, never()).sendEmailVerificationToken(any(), any(), any());
         verify(account).setStatus(AccountStatus.ENABLED);
+        verify(account).setEmailVerified(Boolean.TRUE);
```
DwayneJengSage, line 375: Should createParticipantEmailEnabledVerificationWanted() also be updated?

Also, what about the case where we don't request send verify email, and we also don't provide an email?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -626,6 +636,92 @@ public void notSettingStatusDoesntClearStatus() {
...
+        StudyParticipant participant = new StudyParticipant.Builder().copyOf(PARTICIPANT)
+                .withEmail("updated@email.com").withEmailVerified(Boolean.FALSE).build();
+        
+        participantService.updateParticipant(STUDY, ImmutableSet.of(), participant);
+        
+        verify(accountDao).updateAccount(accountCaptor.capture());
+        Account account = accountCaptor.getValue();
+        verify(account, never()).setEmail("updated@email.com");
+        verify(account, never()).setEmailVerified(Boolean.FALSE);
+        verify(accountWorkflowService, never()).sendEmailVerificationToken(STUDY, ID, "updated@email.com");
```
DwayneJengSage: It might be worthwhile to check that we never call setEmail(anyString()) and sendEmailVerificationToken(anyString()). This will test that we don't also send an email verification email to the original email address either.

DwayneJengSage: Is it worthwhile to add tests for

* deleting email address?
* email address unchanged?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -626,6 +636,92 @@ public void notSettingStatusDoesntClearStatus() {
...
+        StudyParticipant participant = new StudyParticipant.Builder().copyOf(PARTICIPANT)
+                .withPhone(OTHER_PHONE).withEmailVerified(Boolean.FALSE).build();
+        
+        participantService.updateParticipant(STUDY, ImmutableSet.of(), participant);
+        
+        verify(accountDao).updateAccount(accountCaptor.capture());
+        Account account = accountCaptor.getValue();
+        verify(account, never()).setPhone(PHONE);
+        verify(account, never()).setPhoneVerified(Boolean.TRUE);
+        verify(accountWorkflowService, never()).sendEmailVerificationToken(STUDY, ID, EMAIL);
```
DwayneJengSage: Similarly, is it worth changing that we never call sendEmailVerificationToken(any()), just to make sure we're not calling it with null? (Here and in the next test.)

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/hibernate/HibernateHelper.java
alxdarksage: Ugh okay, this should be translated to *our* ConstraintViolationException, and that's why I removed the hibernate ConstraintViolationException from imports (below). I'll change the exception and add a message specific to this case. (The message will be based on duplicate phone/email, which is the only constraint on the table that seems likely to be triggered by user input.)

test/org/sagebionetworks/bridge/hibernate/HibernateHelperTest.java
alxdarksage: Removed so we can import our ConstraintViolationException (which I'm now using).

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
alxdarksage: We're testing here the case where the updates aren't accepted (like email for an account that already has an email). We want the participant as it is returned from the service (asUpdated object) to be set in cache and returned to the user.

Added verification that this same object is being used calling sessionUpdateService

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: When we set it to unverified, we don't set the emailVerified flag at all. I thought about it... we could. But we don't really need to.

Added createParticipantWithPhoneEmailEnabledVerificationWanted() test for case of a phone being supplied where we would attempt to verify an email address.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Makes sense, changed.

alxdarksage: Added tests of deleting email or phone (you can't). We have updateParticipantEmailCannotBeChanged() to test changing of email (and same with phone).

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Yes, changed here as well (and in the other tests that are verifying the method was never called).

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
```diff
@@ -516,6 +520,41 @@ public void partialUpdateSelfParticipant() throws Exception {
...
```
DwayneJengSage, line 550: Should we add a test case for when the participant is updated? Since that's the case that we're changing the code for.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -362,6 +373,7 @@ public void createParticipantEmailEnabledNoVerificationWanted() {
         
         verify(accountWorkflowService, never()).sendEmailVerificationToken(any(), any(), any());
         verify(account).setStatus(AccountStatus.ENABLED);
+        verify(account).setEmailVerified(Boolean.TRUE);
```
DwayneJengSage, line 375: We test that emailVerified is set to true here, but we don't test that emailVerified is _not_ set to true in the createParticipantEmailEnabledVerificationWanted(). We probably should, to make sure that if branch is properly tested.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -375,6 +387,23 @@ public void createParticipantEmailEnabledVerificationWanted() {
...
     
+    @Test
+    public void createParticipantWithPhoneEmailEnabledVerificationWanted() {
+        STUDY.setEmailVerificationEnabled(true);
+        mockHealthCodeAndAccountRetrieval(null, null);
+
+        StudyParticipant phoneOnly = new StudyParticipant.Builder()
+                .copyOf(PARTICIPANT).withEmail(null).build();
+        
+        participantService.createParticipant(STUDY, CALLER_ROLES, phoneOnly, true);
```
DwayneJengSage, line 398: What about the inverse case, where we don't provide an email and we _don't_ request email verification?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -626,6 +655,126 @@ public void notSettingStatusDoesntClearStatus() {
...
+        when(account.getPhoneVerified()).thenReturn(Boolean.TRUE);
+        
+        StudyParticipant participant = new StudyParticipant.Builder().copyOf(PARTICIPANT)
+                .withPhone(OTHER_PHONE).withEmailVerified(Boolean.FALSE).build();
+        
+        participantService.updateParticipant(STUDY, ImmutableSet.of(), participant);
+        
+        verify(accountDao).updateAccount(accountCaptor.capture());
+        Account account = accountCaptor.getValue();
+        verify(account, never()).setPhone(PHONE);
```
DwayneJengSage: Should this be setPhone(any()) to make sure we're not setting this to any other phone number?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -626,6 +655,126 @@ public void notSettingStatusDoesntClearStatus() {
...
+        verify(account, never()).setPhoneVerified(Boolean.TRUE);
+        verify(accountWorkflowService, never()).sendEmailVerificationToken(STUDY, ID, EMAIL);
```
DwayneJengSage: Should this be sendEmailVerificationToken(any())?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -626,6 +655,126 @@ public void notSettingStatusDoesntClearStatus() {
         verify(account, never()).setStatus(any());
     }
 
+    @Test
+    public void updateParticipantEmailCannotBeChanged() {
```
DwayneJengSage, line 690: Do we have a test case for when they update the participant but don't change the email and/or phone?

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
alxdarksage: It would be the same, as the logic about whether the participant is changed or not is in the service. We want to verify that we reload the participant object rather than use the one submitted, so that's why the test is verifying if the participant object is different from what was submitted, that's what we see cached and returned.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Added.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Added (on update, should not touch the setters or send an email).

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Yes it seems any case of using never() should be paired with any(). Did a search and replace of a few more in this test.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Changed

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -362,6 +372,7 @@ public void createParticipantEmailEnabledNoVerificationWanted() {
         
         verify(accountWorkflowService, never()).sendEmailVerificationToken(any(), any(), any());
         verify(account).setStatus(AccountStatus.ENABLED);
+        verify(account).setEmailVerified(Boolean.TRUE);
```
DwayneJengSage, line 375: We test that emailVerified is set to true here, but we don't test that emailVerified is not set to true in the createParticipantEmailEnabledVerificationWanted(). We probably should, to make sure that if branch is properly tested.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -375,6 +386,40 @@ public void createParticipantEmailEnabledVerificationWanted() {
...
+        StudyParticipant phoneOnly = new StudyParticipant.Builder()
+                .copyOf(PARTICIPANT).withEmail(null).build();
+        
+        participantService.createParticipant(STUDY, CALLER_ROLES, phoneOnly, false);
+        
+        verify(accountWorkflowService, never()).sendEmailVerificationToken(any(), any(), any());
+        // The account will not be verified.
+        verify(account).setStatus(AccountStatus.ENABLED);
+        verify(account).setPhone(PHONE);
+        verify(account).setPhoneVerified(Boolean.FALSE);
```
DwayneJengSage, line 421: We should also check that setEmailVerified() is not set to true, to make sure that the if-statement is correct.

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Okay I added it to the case where the account should be unverified, and specifically set emailVerified=false (and added it to the test).

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: I think I covered this case now in createParticipantEmailEnabledVerificationWanted(). I also verified in a couple of update tests that the user cannot set emailVerified = true through the object they submit (it always starts out as false).

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -373,6 +384,41 @@ public void createParticipantEmailEnabledVerificationWanted() {
...
```
DwayneJengSage, line 421: This is actually not covered by createParticipantEmailEnabledVerificationWanted(). That's an entirely separate if-branch.

As far as I can tell, this is the only test that covers the branch where we don't ask for email verification and we don't submit an email address. We have an if-check that makes sure we don't call setEmailVerified(true), and as far as I can tell, there's nothing in this unit test that verifies that that's (not) happening.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -626,6 +686,127 @@ public void notSettingStatusDoesntClearStatus() {
...
+    }
+
+    @Test
+    public void updateParticipantEmailCanBeAddedWithVerification() {
+        STUDY.setEmailVerificationEnabled(true);
+        mockHealthCodeAndAccountRetrieval(null, PHONE);
+        
+        // Set emailVerified = true to test that this is always set to false, regardless of submitted object
+        StudyParticipant participant = new StudyParticipant.Builder().copyOf(PARTICIPANT).withEmail("updated@email.com")
+                .withEmailVerified(Boolean.TRUE).withEmailVerified(Boolean.FALSE).build();
```
DwayneJengSage, line 714: This doesn't test what you think it tests. You call .withEmailVerified(true) and then immediately call .withEmailVerified(false), so the participant you're submitting has emailVerified = false.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -626,6 +686,127 @@ public void notSettingStatusDoesntClearStatus() {
...
+        verify(accountWorkflowService).sendEmailVerificationToken(STUDY, ID, "updated@email.com");
+    }
+
+    @Test
+    public void updateParticipantEmailCanBeAddedNoVerification() {
+        STUDY.setEmailVerificationEnabled(false);
+        mockHealthCodeAndAccountRetrieval(null, PHONE);
+        
+        StudyParticipant participant = new StudyParticipant.Builder().copyOf(PARTICIPANT).withEmail("updated@email.com")
+                .withEmailVerified(Boolean.TRUE).withEmailVerified(Boolean.TRUE).build();
```
DwayneJengSage, line 731: Minor: You call .withEmailVerified(true) twice here.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -269,15 +269,26 @@ public IdentifierHolder createParticipant(Study study, Set<Roles> callerRoles, S
...
-        
-        account.setStatus(sendVerifyEmail ? AccountStatus.UNVERIFIED : AccountStatus.ENABLED);
+        if (sendVerifyEmail) {
+            account.setStatus(AccountStatus.UNVERIFIED);
+            account.setEmailVerified(Boolean.FALSE);
+        } else {
+            account.setStatus(AccountStatus.ENABLED);
+            if (account.getEmail() != null) {
+                account.setEmailVerified(Boolean.TRUE);
+            }
```
alxdarksage, line 279: Do we really want this? We haven't verified the email address. They should trigger and answer an email verification request. This should always be disabled and unverified *unless* it's an admin creation where we're specifically trying to bypass validation.

DwayneJengSage: I'm about 99% sure that only admins can call this API. This call pattern is currently being used for integration tests, where without it, it's impossible to test things like UserDataDownload.

PR #1629, opened 2018-01-17
Adding type property to requestParams map.

alxdarksage:


app/org/sagebionetworks/bridge/models/ResourceList.java
```diff
@@ -46,6 +48,7 @@
     public ResourceList(@JsonProperty(ITEMS) List<T> items) {
         checkNotNull(items);
         this.items = items;
+        this.requestParams.put(TYPE, REQUEST_PARAMS);
```
DwayneJengSage, line 52: Minor: It might be worth making requestParams final, to ensure that it can never be overwritten, so we can guarantee that requestParams contains the TYPE parameter.

DwayneJengSage (approved):
One minor comment. Otherwise, looks good as is.

DwayneJengSage (approved):

PR #1633, opened 2018-01-24
require email validation; flag to suppress email verification email on sign-up

DwayneJengSage:
Studies using email sign-in shouldn't disable email verification, as this removes the server-side check to prevent email bounces. Instead,we introduce a flag to separately suppress the email verification email.

This also reverts https://github.com/Sage-Bionetworks/BridgePF/pull/1624

Testing done:
* unit tests
* manually tested the email sign-in scenario both with and without the flag and verified that (a) email sign-in works and (b) email sign-in automatically verifies the email on first sign-in

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
@@ -58,6 +58,7 @@ public final class DynamoStudy implements Study {
     private String shortName;
     private String sponsorName;
     private String identifier;
+    private boolean autoVerificationEmailSuppressed;
     private boolean studyIdExcludedInExport;
     private String supportEmail;
     private Long synapseDataAccessTeamId;
@@ -170,6 +171,18 @@ public final class DynamoStudy implements Study {
 
     public void setVersion(Long version) {
         this.version = version;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public boolean isAutoVerificationEmailSuppressed() {
+        return autoVerificationEmailSuppressed;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void setAutoVerificationEmailSuppressed(boolean autoVerificationEmailSuppressed) {
+        this.autoVerificationEmailSuppressed = autoVerificationEmailSuppressed;
     }
 
     /** {@inheritDoc} */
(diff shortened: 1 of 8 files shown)


app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
```diff
@@ -172,6 +173,18 @@ public void setVersion(Long version) {
...
+    @Override
+    public boolean isAutoVerificationEmailSuppressed() {
+        return autoVerificationEmailSuppressed;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void setAutoVerificationEmailSuppressed(boolean autoVerificationEmailSuppressed) {
+        this.autoVerificationEmailSuppressed = autoVerificationEmailSuppressed;
+    }
```
alxdarksage, line 186: isSendVerificationEmail[OnSignUp]Enabled, to be consistent with other flags?

app/org/sagebionetworks/bridge/validators/StudyValidator.java
```diff
@@ -143,7 +143,20 @@ public void validate(Object obj, Errors errors) {
         validateEmails(errors, study.getTechnicalEmail(), "technicalEmail");
         validateEmails(errors, study.getConsentNotificationEmail(), "consentNotificationEmail");
         validateDataGroupNamesAndFitForSynapseExport(errors, study.getDataGroups());
-        
+
+        // emailVerificationEnabled=true (public study):
+        //     externalIdValidationEnabled and externalIdRequiredOnSignup can vary independently
+        // emailVerificationEnabled=false:
+        //     externalIdValidationEnabled and externalIdRequiredOnSignup must both be true
```
alxdarksage, line 150: This is probably not 100% true now that you can have phone number accounts. In that case we're assuming that the phone is not verified until you successfully use it to sign in, and that's the behavior we're seeking with these flags for email-based studies.

alxdarksage (changes requested):
Seems fine but I would name the flag in the same way as the other ones in terms of behavior is enabled/disabled, not suppressed.

app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
DwayneJengSage: The reason why I named the flag this way is because I wanted the default value to be false so we can avoid writing code to backfill all the previous studies. I could change the name from “suppressed” to “disabled”, but it would still be backwards relative to all the other flags.

app/org/sagebionetworks/bridge/validators/StudyValidator.java
DwayneJengSage: That's technically true but (a) we have code that keys off of account status and email verification and nothing currently looks at phone verification and (b) this change is meant to be a quick change so we can patch the email verification vulnerability, not an exhaustive feature change.

alxdarksage (approved):

PR #1636, opened 2018-01-29
Test content type of all tested controller methods, fix cases where it is wrong.

alxdarksage:
BRIDGE-2090.

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/BaseController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/BaseController.java
@@ -5,6 +5,7 @@ import static org.sagebionetworks.bridge.BridgeConstants.SESSION_TOKEN_HEADER;
 import static org.sagebionetworks.bridge.dao.ParticipantOption.LANGUAGES;
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import java.io.IOException;
 import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -42,6 +43,8 @@ import org.sagebionetworks.bridge.services.SessionUpdateService;
 import org.sagebionetworks.bridge.services.StudyService;
 
 import org.apache.commons.lang3.StringUtils;
+import org.codehaus.jackson.JsonGenerationException;
+import org.codehaus.jackson.map.JsonMappingException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -57,6 +60,7 @@ import play.mvc.Result;
 import com.amazonaws.util.Throwables;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectWriter;
 import com.google.common.base.Strings;
 import com.google.common.collect.Sets;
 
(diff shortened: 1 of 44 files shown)


app/org/sagebionetworks/bridge/play/controllers/BaseController.java
```diff
@@ -42,6 +43,8 @@
 import org.sagebionetworks.bridge.services.StudyService;
 
 import org.apache.commons.lang3.StringUtils;
+import org.codehaus.jackson.JsonGenerationException;
+import org.codehaus.jackson.map.JsonMappingException;
```
DwayneJengSage: These are Jackson v1 exceptions. Are they compatible with Jackson v2? We should probably use Jackson v2 for consistency.

app/org/sagebionetworks/bridge/play/controllers/BaseController.java
```diff
@@ -312,6 +316,11 @@ Result okResult(Object obj) {
         return ok((JsonNode)MAPPER.valueToTree(new ResourceList<T>(list)));
     }
     
+    Result okResult(ObjectWriter writer, Object object) throws JsonGenerationException, JsonMappingException, IOException {
+        String jsonString = writer.writeValueAsString(object);
+        return okResult(BridgeObjectMapper.get().readTree(jsonString));
+    }
```
DwayneJengSage: Would it be cleaner to just do this: https://stackoverflow.com/questions/22593471/setting-content-type-header-with-play-framework-java

test/org/sagebionetworks/bridge/play/controllers/ApplicationControllerMockTest.java
```diff
@@ -87,18 +87,22 @@ public void before() {
     }
     
     @Test
-    public void verifyEmailWorks() {
+    public void verifyEmailWorks() throws Exception {
         Result result = controller.verifyEmail("test-study");
         
+        assertEquals("text/html; charset=utf-8", result.header("Content-Type"));
+        assertEquals(200, result.status());
```
DwayneJengSage, line 94: Minor: This should use TestUtils.assertStatus(). (Here and in resetPasswordWorks().)

test/org/sagebionetworks/bridge/TestUtils.java
```diff
@@ -128,13 +128,19 @@ public static void assertValidatorMessage(Validator validator, Object object, St
...
+        
         JsonNode node = BridgeObjectMapper.get().readTree(Helpers.contentAsString(result));
         String resultMessage = node.get("message").asText();
-        assertEquals(statusCode, result.status());
         assertEquals("application/json", result.contentType());
         assertEquals(message, resultMessage);
     }
 
+    public static void assertResult(Result result, int statusCode) throws Exception {
+        assertEquals(BridgeConstants.JSON_MIME_TYPE, result.header("Content-Type"));
```
DwayneJengSage, line 139: Minor: The two assertResult() methods both test content type in a different way. Is it worth making them both look the same for consistency?

DwayneJengSage (changes requested):
Mostly good. A few comments.

app/org/sagebionetworks/bridge/play/controllers/BaseController.java
alxdarksage: I found a way to create a Result object with the right content type that doesn't involve serializing/parsing again.

app/org/sagebionetworks/bridge/play/controllers/BaseController.java
alxdarksage: changed

test/org/sagebionetworks/bridge/TestUtils.java
alxdarksage: Yes that first one is a duplicate because that method calls the one below it, removing

test/org/sagebionetworks/bridge/play/controllers/ApplicationControllerMockTest.java
alxdarksage: I don't see this method in TestUtils... it doesn't use assertResult just because the content type is "application/html"

DwayneJengSage: My bad. This one looks like it's actually different. Disregard.

DwayneJengSage (approved):

PR #1640, opened 2018-02-02
Do not store healthCode in the SNS registration, it can generate an unlikely error where a phone is used between two users.

alxdarksage:
Addresses BRIDGE-2084

Invalid parameter: Token Reason: Endpoint arn:aws:sns:us-east-1:649232250620:endpoint/APNS_SANDBOX/ios-sdk-int-tests/fec2fffa-93cd-34de-81b5-43ac908f445c already exists with the same Token, but different attributes."

The different attribute is the healthCode of the prior user when someone creates a new account on the same phone. We don't actually need the healthCode in the registration, it was just to aid troubleshooting, so we can remove it to avoid this error.


Erin-Mounts (approved):
looks fine

PR #1647, opened 2018-02-08
Ensuring Android can specify a study-specific path so apps don't clash on the device.

alxdarksage:
PLEASE UPDATE CODE TO USE THIS PATH. Otherwise down the road, apps will conflict trying to intercept an email sign in link on the phone. Best case scenario, the user gets a dialog to choose the application, but that would be confusing.


liujoshua (approved):
Thanks!

PR #1649, opened 2018-02-09
Improvements to email-based sign in workflows

alxdarksage:
1) Adding the ${emailSignInUrl} variable to the "account exists" template, alongside the now renamed ${resetPasswordUrl} variable (${url} is maintained for backwards compatibility). For email-only sign ins without passwords, they'd like to craft the "account exists" message to sign the user in instead.

2) Adding ${url} support the email sign in template now that we know we don't have to create a separate host for each application. We can entirely construct this URL on the server so it works end-to-end. Still support existing variables for existing templates out there in study configs.

Took the opportunity to refactor AccountWorkflowService to clean up and consolidate a lot of duplicated code.

BRIDGE-2101


app/org/sagebionetworks/bridge/validators/StudyValidator.java
```diff
@@ -279,12 +280,19 @@ private void validateTemplate(Errors errors, EmailTemplate template, String fiel
             if (StringUtils.isBlank(template.getBody())) {
                 errors.rejectValue("body", "is required");
             } else {
-                if (!template.getBody().contains(requiredVariable)) {
-                    errors.rejectValue("body", "must contain the "+requiredVariable+" template variable");
+                boolean missingTemplateVariable = true;
+                for (int i=0; i < templateVariables.length; i++) {
+                    if (template.getBody().contains(templateVariables[i])) {
+                        missingTemplateVariable = false;
```
DwayneJengSage, line 286: Minor: break here, because we don't need to check any more template variables.

test/org/sagebionetworks/bridge/validators/StudyValidatorTest.java
```diff
@@ -294,12 +303,6 @@ public void requiresEmailSignInTemplateWithBody() {
         assertValidatorMessage(INSTANCE, study, "emailSignInTemplate.body", "is required");
     }
     
-    @Test
-    public void requiresEmailSignInTemplateRequiresToken() {
-        study.setEmailSignInTemplate(new EmailTemplate("subject", "body with no token", MimeType.HTML));
-        assertValidatorMessage(INSTANCE, study, "emailSignInTemplate.body", "must contain the ${token} template variable");
-    }
```
DwayneJengSage, line 301: Was deleting this test intentional? Should we instead rewrite this test with an updated error message?

conf/study-defaults/email-sign-in.txt
```diff
@@ -1 +1 @@
-Click here to sign in: <a href="https://${host}/mobile/startSession.html?email=${email}&study=${studyId}&token=${token}">https://${host}/mobile/startSession.html?email=${email}&study=${studyId}&token=${token}</a>
+Click here to sign in: ${url}
```
DwayneJengSage, line 1: Do we need the a href tag?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/validators/StudyValidator.java
alxdarksage: Added

conf/study-defaults/email-sign-in.txt
alxdarksage: No, none of the others have it and most email clients automatically link URLs when they are presented. However, it looks like users can add them in the editor and they work, as the BSM does save the template as HTML. We don't do it with the defaults though.

test/org/sagebionetworks/bridge/validators/StudyValidatorTest.java
alxdarksage: I don't see another test looking at emailSignInTemplate.body, this is an error. Restored it with the updated error message.

DwayneJengSage (approved):

PR #1653, opened 2018-02-14
Consent Notification Email Verification

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2097

Testing done:
* Updated unit tests
* Added integ tests for basic API calls
* Manually tested email verification

Manual tests:
* can't create with verified=true
* create study automatically sends verification email
* can't flip verified flag from false to true with updateStudy()
* updating email automatically sends verification email
* updating email invalidates old verification token
* unverified email doesn't send consent or withdraw emails
* verified email sends consent and withdarw emails
* backwards-compatibility: old studies without the flag are treated as verified; automatically set to tru on update


app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
```diff
@@ -68,6 +68,7 @@
     private List<UploadFieldDefinition> uploadMetadataFieldDefinitions;
     private UploadValidationStrictness uploadValidationStrictness;
     private String consentNotificationEmail;
+    private Boolean consentNotificationEmailVerified;
```
alxdarksage, line 71: We do not have a supportEmailVerified flag. Should we? I don't think we record this, we just check it and store it in Redis. If we want the same, let's file a JIRA.

app/org/sagebionetworks/bridge/services/ConsentService.java
```diff
@@ -268,19 +261,32 @@ public void consentToResearch(Study study, SubpopulationGuid subpopGuid, StudyPa
...
+    // Helper method, which abstracts away logic for sending withdraw notification email.
+    private void sendWithdrawEmail(Study study, String externalId, Account account, Withdrawal withdrawal,
+            long withdrewOn) {
+        if (account.getEmail() == null) {
+            // Withdraw email provider currently doesn't support non-email accounts. Skip.
+            return;
+        }
+
+        Boolean consentNotificationEmailVerified = study.isConsentNotificationEmailVerified();
+        if (consentNotificationEmailVerified != null && !consentNotificationEmailVerified) {
```
alxdarksage: if (consentNotificationEmailVerified != Boolean.TRUE)

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -68,25 +74,33 @@
...
 
 @Component("studyService")
 public class StudyService {
-
     private static Logger LOG = LoggerFactory.getLogger(StudyService.class);
 
+    private static final String BASE_URL = BridgeConfigFactory.getConfig().get("webservices.url");
+    static final String CONFIG_KEY_SUPPORT_EMAIL_PLAIN = "support.email.plain";
+    private static final String CONSENT_NOTIFICATION_VERIFY_EMAIL_URL =
+            "%s/mobile/verifyConsentNotificationEmail.html?study=%s&sptoken=%s";
```
alxdarksage: I just put some effort into moving all these workflow emails into AccountWorkflowService, so we could remove sending mail directly from this service. It allows us to reuse things like BASE_URL, formatting and cache services in that one service. Can this logic be moved there?

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -670,4 +732,120 @@ protected EmailTemplate sanitizeEmailTemplate(EmailTemplate template) {
...
+        // Set verified flag and update study.
+        study.setConsentNotificationEmailVerified(true);
+        updateAndCacheStudy(study);
+    }
+
+    // Creates a random token for consent notification email verification. Package-scoped so it can be mocked by unit
+    // tests.
+    String createTimeLimitedToken() {
+        return SecureTokenGenerator.INSTANCE.nextToken();
+    }
```
alxdarksage, line 832: for example, this exists in accountWorkflowService already.

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -670,4 +732,120 @@ protected EmailTemplate sanitizeEmailTemplate(EmailTemplate template) {
...
+        // Study ID that we want to verify email for.
+        public String getStudyId() {
+            return studyId;
+        }
+
+        // Email address that we want to verify.
+        public String getEmail() {
+            return email;
+        }
+    }
```
alxdarksage, line 885: ditto, already in AccountWorkflowService. Let's use what's there

app/views/verifyConsentNotificationEmail.scala.html
```diff
@@ -0,0 +1,60 @@
...
+        window.assetsHost = "@assetsHost";
+    </script>
+    <link href="//@assetsHost/mobile/styles/normalize.css" type="text/css" rel="stylesheet"/>
+    <link href="//@assetsHost/mobile/styles/mobile.css" type="text/css" rel="stylesheet"/>
+</head>
+<body>
+
+<div class="logo_box"><img id="logo" onerror="this.style.visibility='hidden'"/></div>
+<div class="message" id="m1">Verifying...</div>
+<div class="message" id="m2"></div>
```
alxdarksage: I cleaned up loading the logo so it doesn't happen in JS; we can check this in and I'll make the change, it's trivial.

conf/routes
```diff
@@ -244,6 +245,8 @@ POST   /v3/studies/init                @org.sagebionetworks.bridge.play.controll
 GET    /v3/studies/self                @org.sagebionetworks.bridge.play.controllers.StudyController.getCurrentStudy
 GET    /v3/studies/self/publicKey      @org.sagebionetworks.bridge.play.controllers.StudyController.getStudyPublicKeyAsPem
 GET    /v3/studies/self/emailStatus    @org.sagebionetworks.bridge.play.controllers.StudyController.getEmailStatus
+POST   /v3/studies/self/resendVerifyConsentNotificationEmail @org.sagebionetworks.bridge.play.controllers.StudyController.resendVerifyConsentNotificationEmail
+POST   /v3/studies/self/verifyConsentNotificationEmail @org.sagebionetworks.bridge.play.controllers.StudyController.verifyConsentNotificationEmail
```
alxdarksage: Possibly we could use a format that'll allow us to add the support email later:

/v3/studies/self/emails/verify?type=[support|consentNotification]
/v3/studies/self/emails/resendVerify?type=[support|consentNotification]

alxdarksage (changes requested):

app/org/sagebionetworks/bridge/services/StudyService.java
alxdarksage: Maybe we should rename that the WorkflowService or something, If we'd no longer just using it for account-specific emails.

app/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
DwayneJengSage: As per IM conversation, we don't need this, since we don't send any emails _to_ the support email. If in the future, we find that we _do_ need this, we can always add it at that time.

app/org/sagebionetworks/bridge/services/ConsentService.java
DwayneJengSage: The equivalent logic would actually be if (consentNotificationEmailVerified == Boolean.FALSE). I've updated this.

app/org/sagebionetworks/bridge/services/StudyService.java
DwayneJengSage: As per IM conversation, from a code consolidation perspective, the only thing we'd be able to re-use is BASE_URL and createTimeLimitedToken(). Maybe saveVerification() and restoreVerification() can be refactored to take in class and expiration time to make it generalized.

From a code organization purpose, I can see a benefit here if we renamed AccountWorkflowService to something like ChannelVerificationService. Regardless, as per IM conversation, we decided this was best left for a future refactor.

app/org/sagebionetworks/bridge/services/StudyService.java
DwayneJengSage: The one in AccountWorkflowService uses userId rather than email. The I could use email in place of userId, but that seems more like a hack.

conf/routes
DwayneJengSage: Changed, to allow future additions to this. Note that in the code, we currently only support consentNotification.

alxdarksage (approved):

app/org/sagebionetworks/bridge/play/controllers/StudyController.java
```diff
@@ -190,8 +193,42 @@ public Result getEmailStatus() throws Exception {
...
+
+    // Helper method to parse and validate the email type for study email verification workflow. We do verification
+    // here so that the service can just deal with a clean enum.
+    private static StudyEmailType parseEmailType(String typeStr) {
+        if (StringUtils.isBlank(typeStr)) {
+            throw new BadRequestException("Email type must be specified");
+        }
+
+        try {
+            return StudyEmailType.valueOf(typeStr);
```
alxdarksage: Because all the enums exposed in the API are lowercase, I would use typeStr.toUpperCase() here

DwayneJengSage: Changed.

PR #1654, opened 2018-02-15
Public api to retrieve an app config.

alxdarksage:
Change the retrieval of an app config using request criteria, from being an authenticated to a public (no auth) call. This limits what you can filter on, but the client team only anticipates filtering on version by platform. In addition, caching the app configs so every request doesn't hit DDB.

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/AppConfigController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/AppConfigController.java
@@ -37,7 +37,7 @@ public class AppConfigController extends BaseController {
     }
 
     @Resource(name = "genericViewCache")
-    public void setViewCache(ViewCache viewCache) {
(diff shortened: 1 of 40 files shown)


app/org/sagebionetworks/bridge/play/controllers/AppConfigController.java
```diff
@@ -20,19 +29,52 @@
...
     
+    private ViewCache viewCache;
+
     @Autowired
     final void setAppConfigService(AppConfigService appConfigService) {
         this.appConfigService = appConfigService;
     }
+
+    @Resource(name = "genericViewCache")
+    public void setViewCache(ViewCache viewCache) {
```
DwayneJengSage: final

DwayneJengSage (approved):
One super minor comment. Otherwise, looks good.

DwayneJengSage (approved):

PR #1656, opened 2018-02-16
Migrate participant options to accounts table

alxdarksage:
Loading and saving an account will move the values in the participant options table to the accounts table. Separately from this, we will need to run a migration to move over the values. New accounts write to the account table.

This still requires more testing, but all existing/added unit and integration tests pass.


test/org/sagebionetworks/bridge/dao/ParticipantOptionTest.java
```diff
@@ -93,14 +95,64 @@ public void canRetrieveUTCFromParticipant() {
     
     @Test
     public void canRetrieveEmptyValuesFromStudyParticipant() {
```
DwayneJengSage: Nitpick: The test names for fromAccount and fromParticipant seem to be switched.

test/org/sagebionetworks/bridge/dao/ParticipantOptionTest.java
```diff
@@ -93,14 +95,64 @@ public void canRetrieveUTCFromParticipant() {
...
+        account.setLanguages(TestUtils.newLinkedHashSet("en","de"));
+        account.setSharingScope(SharingScope.ALL_QUALIFIED_RESEARCHERS);
+        account.setTimeZone(DateTimeZone.forOffsetHours(-7));
+        
+        String result = ParticipantOption.DATA_GROUPS.fromAccount(account);
+        assertTrue(result.contains("group1"));
+        assertTrue(result.contains("group2"));
+        
+        result = ParticipantOption.EMAIL_NOTIFICATIONS.fromAccount(account);
+        assertEquals("true", result);
```
DwayneJengSage: We should test a non-default value for this test.

app/org/sagebionetworks/bridge/services/ConsentService.java
```diff
@@ -257,6 +257,9 @@ public void consentToResearch(Study study, SubpopulationGuid subpopGuid, StudyPa
         // Do this first, as it directly impacts the export of data, and if nothing else, we'd like this to succeed.
         optionsService.setEnum(study.getStudyIdentifier(), account.getHealthCode(), SHARING_SCOPE, SharingScope.NO_SHARING);
         
+        // Prevent optimistic locking exception until operations are combined into one operation. 
+        account = accountDao.getAccount(AccountId.forId(study.getIdentifier(), account.getId()));
```
DwayneJengSage, line 261: Previously, we set the sharing scope and the consents separately because they came from different sources. Now that they're both in RDS, does this still need to be true?

If not, can we have the same accountDao.updateAccount() call be used to do both? accountDao.getAccount() already populates the account object with participant options, so all we need to do is account.setSharingScope(SharingScope.NO_SHARING).

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceTest.java
```diff
@@ -58,23 +69,23 @@ public void before() {
...
+        verify(mockAccountDao).updateAccount(accountCaptor.capture(), eq(false));
+        
+        Account captured = accountCaptor.getValue();
+        assertEquals(true, captured.getNotifyByEmail());
     }
     
     @Test
     public void getBoolean() {
-        when(mockDao.getOptions(HEALTH_CODE)).thenReturn(new ParticipantOptionsLookup(map(EMAIL_NOTIFICATIONS, "true")));
+        account.setNotifyByEmail(Boolean.TRUE);
```
DwayneJengSage: Minor: Would it be better to consolidate all of these "get" tests into a single test so all of the test is in a single place?

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceTest.java
```diff
@@ -58,23 +69,23 @@ public void before() {
...
         
-        assertTrue(service.getOptions(HEALTH_CODE).getBoolean(EMAIL_NOTIFICATIONS));
+        assertTrue(service.getOptions(TestConstants.TEST_STUDY, HEALTH_CODE).getBoolean(EMAIL_NOTIFICATIONS));
```
DwayneJengSage: Would it be better to use a non-default value for the test?

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceTest.java
```diff
@@ -149,101 +158,45 @@ public void setStringSet() {
...
-        when(mockDao.getOptions(HEALTH_CODE)).thenReturn(new ParticipantOptionsLookup(map(LANGUAGES, "en,fr")));
-        
-        LinkedHashSet<String> langs = service.getOptions(HEALTH_CODE).getOrderedStringSet(LANGUAGES);
-        Iterator<String> i = langs.iterator();
-        assertEquals("en", i.next());
-        assertEquals("fr", i.next());
-        
-        langs = TestUtils.newLinkedHashSet("fr","en","kl");
+    public void setLinkedHashSet() {
+        LinkedHashSet<String> langs = TestUtils.newLinkedHashSet("fr","en","kl");
```
DwayneJengSage: For completeness, do we also want a test for getLinkedHashSet?

app/org/sagebionetworks/bridge/services/ParticipantOptionsService.java
```diff
@@ -12,53 +18,66 @@
...
+            Map<String,String> map = Maps.newHashMap();
+            map.put(TIME_ZONE.name(), TIME_ZONE.fromAccount(account));
+            map.put(SHARING_SCOPE.name(), SHARING_SCOPE.fromAccount(account));
+            map.put(EMAIL_NOTIFICATIONS.name(), EMAIL_NOTIFICATIONS.fromAccount(account));
+            map.put(EXTERNAL_IDENTIFIER.name(), EXTERNAL_IDENTIFIER.fromAccount(account));
+            map.put(DATA_GROUPS.name(), DATA_GROUPS.fromAccount(account));
+            map.put(LANGUAGES.name(), LANGUAGES.fromAccount(account));
+            return new ParticipantOptionsLookup(map);
+        }
+        return new ParticipantOptionsLookup(Maps.newHashMap());
```
DwayneJengSage, line 65: Unit test for account not found?

app/org/sagebionetworks/bridge/services/ParticipantOptionsService.java
```diff
@@ -12,53 +18,66 @@
...
-    }
-
     /**
      * Persist a boolean participant option.
      */
     public void setBoolean(StudyIdentifier studyIdentifier, String healthCode, ParticipantOption option, boolean value) {
         checkNotNull(studyIdentifier);
         checkArgument(isNotBlank(healthCode));
-        checkNotNull(option);
+        checkNotNull(option == ParticipantOption.EMAIL_NOTIFICATIONS);
```
DwayneJengSage: I'm not sure this is the check you're looking for. option==ParticipantOption.EMAIL_NOTIFICATIONS will always be non-null. (Here and a few other places in this class.)

app/org/sagebionetworks/bridge/services/ParticipantOptionsService.java
```diff
@@ -12,53 +18,66 @@
...
         
-        optionsDao.setOption(studyIdentifier, healthCode, option, Boolean.toString(value));
+        AccountId accountId = AccountId.forHealthCode(studyIdentifier.getIdentifier(), healthCode);
+        Account account = accountDao.getAccount(accountId);
+        if (account != null) {
+            account.setNotifyByEmail(value);
+            accountDao.updateAccount(account, false);
+        }
```
DwayneJengSage: If the account is not found, is that an error?

Also, unit tests for the account not found case? (Here and in several other places in this class.)

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/dao/ParticipantOptionTest.java
alxdarksage: Swapped them

test/org/sagebionetworks/bridge/dao/ParticipantOptionTest.java
alxdarksage: changed to false

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceTest.java
alxdarksage: This service is going to go away in a subsequent check in, so I wouldn't invest effort in refactoring.

app/org/sagebionetworks/bridge/services/ConsentService.java
alxdarksage: Yes, we definitely want to do this. I am holding off to keep the size of this PR down, but a subsequent PR will remove the options service, and in most cases, that'll lead to consolidating calls to update the account.

test/org/sagebionetworks/bridge/services/ParticipantOptionsServiceTest.java
alxdarksage: added

app/org/sagebionetworks/bridge/services/ParticipantOptionsService.java
alxdarksage: added

app/org/sagebionetworks/bridge/services/ParticipantOptionsService.java
alxdarksage: Yeow! changed to checkArgument.

app/org/sagebionetworks/bridge/services/ParticipantOptionsService.java
alxdarksage: Now throws an exception, added unit tests for all the setter methods to verify.

DwayneJengSage (approved):

PR #1661, opened 2018-03-01
BRIDGE-2123 BRIDGE-2125 and related

alxdarksage:
- added security flags to the cookie as well as a domain (both secure and http only, so cookie is only sent back to originating server across HTTPS);

- removed the backfill code. No one uses it and it exposes web pages and APIs I don't want to audit for security;

- created our own interceptor (rather than the play filter) to add the Content-Security-Policy header. This wasn't supported by the Play filter;

- changed our HTML pages (the only thing that is really effected by many of these security headers) so that we don't load any third party scripts on our pages. These headers _might_ also make the Bridge Study Manager less likely to be an attack vehicle for corrupting data on the server.

Tested this manually using CURL and a browser, where appropriate (it's the browser that changes its behavior based on these headers). The only potential problem these changes might cause is with clients who are relying on the cookie, of which there should be none, but somehow there are. Still, those client HTTP libraries would have to respect these headers, and I don't think they do.


app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
```diff
@@ -211,7 +211,8 @@ private void setCookieAndRecordMetrics(UserSession session) {
         // We have removed the cookie in the past, only to find out that clients were unknowingly
         // depending on the cookie to preserve the session token. So it remains.
         response().setCookie(BridgeConstants.SESSION_TOKEN_HEADER, session.getSessionToken(),
-                BridgeConstants.BRIDGE_SESSION_EXPIRE_IN_SECONDS, "/");
+                BridgeConstants.BRIDGE_SESSION_EXPIRE_IN_SECONDS, "/", 
+                bridgeConfig.getHostnameWithPostfix("webservices"), true, true);
```
DwayneJengSage: This doesn't work on local.

1. You want to use getHostnameWithPostfix() resolves to webservices-local.sagebridge.org, which doesn't resolve to anything. I think you want getWebservicesUrl().
2. Local doesn't use HTTPS. Can we set these flags based on bridgeConfig.getEnv() != LOCAL?

(Same as in BaseController.)

app/org/sagebionetworks/bridge/play/controllers/BackfillController.java
```diff
@@ -1,50 +0,0 @@
...
-import org.slf4j.LoggerFactory;
-import org.springframework.beans.BeansException;
-import org.springframework.context.ApplicationContext;
-import org.springframework.context.ApplicationContextAware;
-import org.springframework.stereotype.Controller;
-
-import play.mvc.Result;
-
-@Controller
-public class BackfillController extends BaseController implements ApplicationContextAware  {
```
DwayneJengSage: Backfill is used to redrive failed uploads. We can file a separate JIRA to build a new redrive system that doesn't rely on HTTP, but I think for now, we need to keep this.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/play/controllers/BackfillController.java
alxdarksage: Sorry I didn't know this. I'll restore those files.

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
alxdarksage: Our system doesn't use cookies so it doesn't effect us, but I can change this for localhost. I would like to remove the cookie, but it apparently breaks some Android apps that have been built against the Bridge server.

alxdarksage: Changed this to not set the SSL-related flags to true in local environment

app/org/sagebionetworks/bridge/play/controllers/BackfillController.java
alxdarksage: Restored

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerMockTest.java
```diff
@@ -446,7 +448,8 @@ private void signInNewSession(boolean isConsented, Roles role) throws Exception
 
         verify(cacheProvider).updateRequestInfo(requestInfoCaptor.capture());
         verify(mockResponse).setCookie(BridgeConstants.SESSION_TOKEN_HEADER, session.getSessionToken(),
-                BridgeConstants.BRIDGE_SESSION_EXPIRE_IN_SECONDS, "/");
+                BridgeConstants.BRIDGE_SESSION_EXPIRE_IN_SECONDS, "/",
+                BridgeConfigFactory.getConfig().get("webservices.url"), false, false);
```
DwayneJengSage: I think our build runs in dev, so this will fail in Travis. We should probably mock the env to dev, kinda like how we do it in BaseControllerTest. (2 places in AuthenticationControllerMockTest and 1 in BaseControllerTest.)

conf/routes
```diff
@@ -274,10 +274,6 @@ GET    /v3/studies/:studyId/oauth/:vendorId/:healthCode                 @org.sag
 POST   /v3/studies/:studyId/users                                       @org.sagebionetworks.bridge.play.controllers.UserManagementController.createUserWithStudyId(studyId: String)
 DELETE /v3/studies/:studyId/uploadschemas/:schemaId                     @org.sagebionetworks.bridge.play.controllers.UploadSchemaController.deleteAllRevisionsOfUploadSchema(studyId: String, schemaId: String)
 
-# Backfills
-GET    /v3/backfill/:name          @org.sagebionetworks.bridge.play.controllers.BackfillController.backfill(name: String)
-POST   /v3/backfill/:name/start    @org.sagebionetworks.bridge.play.controllers.BackfillController.start(name: String)
```
DwayneJengSage: These should be restored.

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerMockTest.java
```diff
@@ -538,6 +545,25 @@ public void signInBlockedByVersionKillSwitch() throws Exception {
...
+        when(controller.bridgeConfig.get("webservices.url")).thenReturn("http://localhost:9000");
+        
+        UserSession session = createSession(null, null);
+        when(authenticationService.signIn(any(), any(), any())).thenReturn(session);
+        
+        controller.signIn();
+        
+        verify(response).setCookie(BridgeConstants.SESSION_TOKEN_HEADER, TEST_SESSION_TOKEN,
+                BridgeConstants.BRIDGE_SESSION_EXPIRE_IN_SECONDS, "/",
+                BridgeConfigFactory.getConfig().get("webservices.url"), false, false);
```
DwayneJengSage: This will fail in Travis, since we mock bridgeConfig to return localhost, but in travis, this picks up the dev config.

test/org/sagebionetworks/bridge/play/controllers/BaseControllerTest.java
```diff
@@ -670,12 +673,45 @@ public void ifClientSendsCookieRetrieveAndResetIt() {
...
         BaseController controller = new SchedulePlanController();
+        controller.setBridgeConfig(BridgeConfigFactory.getConfig());
         
         String token = controller.getSessionToken();
         assertEquals("ABC", token);
         
         verify(mockResponse).setCookie(BridgeConstants.SESSION_TOKEN_HEADER, "ABC",
-                BridgeConstants.BRIDGE_SESSION_EXPIRE_IN_SECONDS, "/");
+                BridgeConstants.BRIDGE_SESSION_EXPIRE_IN_SECONDS, "/",
+                BridgeConfigFactory.getConfig().get("webservices.url"), false, false);
```
DwayneJengSage: This will fail in Travis because travis runs in dev.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #1664, opened 2018-03-05
Process Uploads on Disk

DwayneJengSage:
See https://sagebionetworks.jira.com/browse/BRIDGE-2083

This process uploads on disk, so that we don't have to parse potentially large uploads in memory. This improves the reliability and robustness of the upload validation pipeline, and allows us to handle larger and larger files (provided we don't need to parse them).

This also removes Health Data Attachment (unused table) and streamlines record IDs and attachment IDs to all match the upload ID (see https://sagebionetworks.jira.com/browse/BRIDGE-1725).

Testing done:
* Updated unit tests
* Ran Bridge Integ Tests
* Ran Exporter Integ Tests
* Manually tests to ensure uploads still worked as expected: health data submission, v1 data, v1 survey, v2 generic


app/org/sagebionetworks/bridge/services/UploadArchiveService.java
```diff
@@ -86,21 +120,43 @@ public void setCmsEncryptorCache(LoadingCache<String, CmsEncryptor> cmsEncryptor
...
+     * stream of decrypted data. The caller is responsible for closing both streams.
+     */
+    public InputStream decrypt(String studyId, InputStream source) {
+        // validate
+        if (StringUtils.isBlank(studyId)) {
+            throw new BadRequestException(String.format(Validate.CANNOT_BE_BLANK, "studyId"));
+        }
+        if (source == null) {
+            throw new BadRequestException(String.format(Validate.CANNOT_BE_NULL, "source"));
+        }
```
alxdarksage: Does this happen on a user request? If not, can checkNotNull be used here?

app/org/sagebionetworks/bridge/services/UploadArchiveService.java
```diff
@@ -137,12 +193,23 @@ public void setCmsEncryptorCache(LoadingCache<String, CmsEncryptor> cmsEncryptor
      * @throws BridgeServiceException
      *         if zipping fails
      */
-    public byte[] zip(@Nonnull Map<String, byte[]> dataMap) throws BridgeServiceException {
+    public byte[] zip(Map<String, byte[]> dataMap) throws BridgeServiceException {
+        // Validate input
         if (dataMap == null) {
             throw new BadRequestException(String.format(Validate.CANNOT_BE_NULL, "dataMap"));
         }
```
alxdarksage: Same here. I would reserve BRE for exceptions that will ultimately return an HTTP error to a user. I'm not clear if this is true here or not. Here and elsewhere.

app/org/sagebionetworks/bridge/upload/UploadFileHelper.java
```diff
@@ -0,0 +1,200 @@
...
+                    fieldNode = BridgeObjectMapper.get().readTree(fileInputStream);
+                } catch (IOException ex) {
+                    throw new UploadValidationException("Error parsing field file, uploadId=" + uploadId +
+                            ", fieldName=" + fieldName, ex);
+                }
+            }
+        } else {
+            // Case 2: The field refers to a top-level JSON key in one of the files. The fieldName will have one of
+            // the file names as a prefix. Iterate the file map to find that prefix. (Note: If the schema is
+            // ambiguously defined, there may be more than one. Hopefully, this never happens in the real world.)
```
alxdarksage, line 115: Can we preclude it through schema validation?

app/org/sagebionetworks/bridge/upload/UploadValidationContext.java
```diff
@@ -100,57 +100,66 @@ public void addMessage(String msg) {
...
+        return unzippedDataFileMap;
     }
 
-    /** @see #getUnzippedDataMap */
-    public void setUnzippedDataMap(Map<String, byte[]> unzippedDataMap) {
-        this.unzippedDataMap = unzippedDataMap;
+    /** @see #getUnzippedDataFileMap */
+    public void setUnzippedDataFileMap(Map<String, File> unzippedDataFileMap) {
+        this.unzippedDataFileMap = unzippedDataFileMap;
     }
```
alxdarksage, line 144: What are the error conditions for writing to file? Can the drive fill up? Do we want to write to an EFS "drive"? Do we know what the hard limit is on filesystem storage is on our instances?

alxdarksage (approved):

alxdarksage (approved):

app/org/sagebionetworks/bridge/upload/UploadValidationContext.java
DwayneJengSage: We can configure an EBS drive in Elastic Beanstalk and/or EC2. It's currently set to "container default", and for Exporter, that looks to be about 8gb, 75% of which is free.

We also delete temp dirs when we're done using them, and the temp dirs are created in _the_ temp dir, so I don't anticipate running out of disk.

app/org/sagebionetworks/bridge/upload/UploadFileHelper.java
DwayneJengSage: Not very easily. Unfortunately, the schema system doesn't have a delimiter between file name and field name, so it's not obvious what the filename is.

app/org/sagebionetworks/bridge/services/UploadArchiveService.java
DwayneJengSage: Changed.

app/org/sagebionetworks/bridge/services/UploadArchiveService.java
DwayneJengSage: Changed.

PR #1666, opened 2018-03-07
Removing participant options.

alxdarksage:


app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -254,7 +243,10 @@ private void updateReauthToken(HibernateAccount hibernateAccount, Account accoun
         hibernateAccount.setReauthTokenAlgorithm(passwordAlgorithm);
         hibernateAccount.setReauthTokenModifiedOn(DateUtils.getCurrentMillisFromEpoch());
         
-        hibernateHelper.update(hibernateAccount);
+        // We must get the current version to return from the DAO, or subsequent updates to the 
+        // account, even in the same call, will fail (e.g. to update languages captured from a request).
```
DwayneJengSage, line 245: Minor: Note in the comment that this is safe specifically because this method only updates reauth attributes, and no other codepath can update reauth attributes, so there's no chance of a version conflict stomping on our changes.

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
```diff
@@ -502,7 +493,9 @@ public void reauthenticateSuccess() throws Exception {
         assertEquals(TestConstants.TEST_STUDY, account.getStudyIdentifier());
         assertEquals(EMAIL, account.getEmail());
         assertNotEquals(originalReauthTokenHash, account.getReauthToken());
-
+        // This has been incremented by the reauth token update
+        assertEquals(2, account.getVersion());
```
DwayneJengSage, line 497: Should we have the same check in the getAccountAfterAuthentication() tests?

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -369,6 +361,17 @@ public void updateAccount(Account account, boolean allowIdentifierUpdates) {
...
+    
+    /** {@inheritDoc} */
+    @Override
+    public void editAccount(StudyIdentifier studyId, String healthCode, Consumer<Account> accountEdits) {
+        AccountId accountId = AccountId.forHealthCode(studyId.getIdentifier(), healthCode);
+        Account account = getAccount(accountId);
+        if (account != null) {
+            accountEdits.accept(account);
+            updateAccount(account, false);
+        }
```
DwayneJengSage, line 376: editAccount() unit tests?

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -694,6 +685,15 @@ static Account unmarshallAccount(HibernateAccount hibernateAccount) {
...
         account.setMigrationVersion(hibernateAccount.getMigrationVersion());
         
+        // sharing scope defaults to no sharing
+        if (account.getSharingScope() == null) {
+            account.setSharingScope(SharingScope.NO_SHARING);
+        }
+        // email notifications are opt-out
+        if (account.getNotifyByEmail() == null) {
+            account.setNotifyByEmail(true);
+        }
```
DwayneJengSage, line 698: Add test to unmarshall tests for default values?

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -304,7 +298,8 @@ private UserSession getSessionFromAccount(Study study, CriteriaContext context,
         if (participant.getLanguages().isEmpty() && !context.getLanguages().isEmpty()) {
             participant = new StudyParticipant.Builder().copyOf(participant)
                     .withLanguages(context.getLanguages()).build();
-            optionsService.setOrderedStringSet(study, account.getHealthCode(), LANGUAGES, context.getLanguages());
+            account.setLanguages(context.getLanguages());
+            accountDao.updateAccount(account, false);
```
DwayneJengSage: This will fail, for example, if reauth is enabled and we update the reauth token, or if it's a channel signIn that verifies the channel. It would be better to use accountDao.editAccount().

DwayneJengSage: Unit tests for setting languages?

app/org/sagebionetworks/bridge/services/ConsentService.java
```diff
@@ -182,13 +173,21 @@ public void consentToResearch(Study study, SubpopulationGuid subpopGuid, StudyPa
...
+        return getConsentStatuses(context, account);
+    }
+    
+    /**
+     * Get all the consent status objects for this user. From these, we determine if the user 
+     * has consented to the right consents to have access to the study, and whether or not those 
+     * consents are up-to-date. 
+     */
+    public Map<SubpopulationGuid,ConsentStatus> getConsentStatuses(CriteriaContext context, Account account) {
+        checkNotNull(context);
```
DwayneJengSage, line 190: Do we need unit tests for this?

app/org/sagebionetworks/bridge/services/ConsentService.java
```diff
@@ -226,16 +225,14 @@ public void consentToResearch(Study study, SubpopulationGuid subpopGuid, StudyPa
         if(!withdrawSignatures(account, subpopGuid, withdrewOn)) {
             throw new EntityNotFoundException(ConsentSignature.class);
         }
+        Map<SubpopulationGuid,ConsentStatus> statuses = getConsentStatuses(context, account);
+        if (!ConsentStatus.isUserConsented(statuses)) {
+            account.setSharingScope(SharingScope.NO_SHARING);
+        }
```
DwayneJengSage, line 231: Unit test for this?

test/org/sagebionetworks/bridge/services/FPHSServiceTest.java
```diff
@@ -73,59 +75,51 @@ public void registerIdThrowsException() throws Exception {
...
     @Test
     public void registerExternalIdentifier() throws Exception {
-        doReturn(new ParticipantOptionsLookup(Maps.newHashMap())).when(optionsService).getOptions(TEST_STUDY, "BBB");
+        TestUtils.mockEditAccount(mockAccountDao, mockAccount);
         
         service.registerExternalIdentifier(TEST_STUDY, "BBB", externalId);
-        verify(dao).registerExternalId(externalId);
-        verify(optionsService).setString(TEST_STUDY, "BBB", EXTERNAL_IDENTIFIER, externalId.getIdentifier());
+        verify(mockDao).registerExternalId(externalId);
+        verify(mockAccount).setExternalId(externalId.getIdentifier());
```
DwayneJengSage, line 86: Should this also test that we set the data group?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: Added

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
alxdarksage: added

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: Added unit tests.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: Added test for default sharing scope/email notifications

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage: I though here to avoid loading the account one more time. During account load we "go back to the well" in two places to update the account before returning it. We do that if it's missing a health code (presumably rare, I only found one account without a health code during migration), and then again to rotate the auth token. You are right that using the editAccount function here would ignore the version of the account at this point in the code, though I believe now it is correct after fixing an integration test by transferring the version to the Account object after the reauth update. Do you still want to change to editAccount()?

alxdarksage: Added test specifically for this (it's tested obliquely for a sign up in the non-mock test, but not for any subsequent request).

app/org/sagebionetworks/bridge/services/ConsentService.java
alxdarksage: This was just a refactor so if we had an account, we could pass it in rather than loading it again, and it's tested in ConsentServiceTest quite a bit

app/org/sagebionetworks/bridge/services/ConsentService.java
alxdarksage: This is tested in ConsentServiceTest (the test that loads Spring, not the mock tests)

test/org/sagebionetworks/bridge/services/FPHSServiceTest.java
alxdarksage: Added

app/org/sagebionetworks/bridge/services/AuthenticationService.java
DwayneJengSage: I'm not convinced this won't fail if this is the same request that, for example, verifies a channel. I think we should use editAccount() here.

app/org/sagebionetworks/bridge/services/ConsentService.java
DwayneJengSage, line 231: I'm not seeing the test in ConsentServiceTest (either pre-existing or new).

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage: Okay, changed to editAccount

app/org/sagebionetworks/bridge/services/ConsentService.java
alxdarksage: Ok, added 4 explicit tests of this in the mock class.

DwayneJengSage (approved):

PR #1667, opened 2018-03-07
Add expirationPeriod to all the emails that are sent out.

alxdarksage:
We're inconsistent in including this and we don't include the time period, so if we change it in the code we'd have to update all the templates everywhere.

I am going to go through, update all the templates at some point, and then delete the older expiration variable, ${expirationWindow}. But there's no rush on that.

### DIFF
--- a/app/org/sagebionetworks/bridge/BridgeUtils.java
+++ b/app/org/sagebionetworks/bridge/BridgeUtils.java
@@ -57,6 +57,32 @@ public class BridgeUtils {
     
     private static final SecureRandom SECURE_RANDOM = new SecureRandom();
 
+    private static final int ONE_HOUR = 60*60;
+    private static final int ONE_DAY = 60*60*24;
+    private static final int ONE_MINUTE = 60;
+    
+    /**
+     * Convert expiration measures in seconds to an English language explanation of
+     * the expiration time. This is not intended to cover odd cases, our expirations 
+     * are always in minutes or more commonly, hours 
+     */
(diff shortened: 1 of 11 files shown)


app/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -54,9 +54,34 @@
...
+    /**
+     * Convert expiration measures in seconds to an English language explanation of
+     * the expiration time. This is not intended to cover odd cases--our expirations 
+     * are in minutes, hours, or possibly days. 
+     */
+    public static String secondsToPeriodString(int seconds) {
+        if (seconds >= (ONE_DAY*2) && seconds % ONE_DAY == 0) {
+            return Integer.toString(seconds/ONE_DAY) + " days";
+        } else if (seconds >= ONE_DAY && seconds % ONE_DAY == 0) {
+            return Integer.toString(seconds/ONE_DAY) + " day";
```
DwayneJengSage, line 72: Minor: This is only possible if seconds == ONE_DAY, and the output is always "1 day". Do we want to change this to that? (Same with 1 hour and 1 minute.)

DwayneJengSage (approved):
Looks good. One minor comment.

app/org/sagebionetworks/bridge/BridgeUtils.java
alxdarksage: Yes, this just ensures that 1 second, 1 minute, 1 hour and 1 day are all singular and not plural.

PR #1670, opened 2018-03-13
Sms templates

alxdarksage:
All the SMS messages now have templates, configured in the study object. These are short, but have the same template variables as email messages.

I'm also renaming the "short URLs" so that the variables can be combined in templates (as was recently needed in the "account exists" templates where you can direct the user to reset their password, and/or sign in via email or phone).


app/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -118,6 +118,7 @@ public static AccountId parseAccountId(String studyId, String identifier) {
                 Iterables.getFirst(commaListToOrderedSet(study.getTechnicalEmail()), ""));
         map.put("consentEmail", 
                 Iterables.getFirst(commaListToOrderedSet(study.getConsentNotificationEmail()), ""));
+        map.put("host", BridgeConfigFactory.getConfig().getHostnameWithPostfix("ws"));
```
DwayneJengSage, line 121: Do we want to add this to BridgeUtilsTest.studyTemplateVariblesWorks?

app/org/sagebionetworks/bridge/models/studies/Study.java
```diff
@@ -267,6 +267,40 @@ static Study create() {
...
+    /**
+     * The template for an SMS message sent to a user that triggers the reset password workflow, 
+     * on an account that only has a phone number.
+     */
+    SmsTemplate getResetPasswordSmsTemplate();
+    void setResetPasswordSmsTemplate(SmsTemplate template);
+    
+    /**
+     * The template for the SMS message sent to a user as part of a sign in via phone workflow.
+     * @return
```
DwayneJengSage: Nitpick: Dangling return annotation.

test/org/sagebionetworks/bridge/services/AccountWorkflowServiceTest.java
```diff
@@ -378,11 +391,12 @@ public void notifyAccountExistsForPhone() throws Exception {
...
                 AccountWorkflowService.EXPIRE_IN_SECONDS);
-        verify(mockNotificationsService).sendSMSMessage(eq(study.getStudyIdentifier()), 
-                eq(TestConstants.PHONE), stringCaptor.capture());
-        String message = stringCaptor.getValue();
+        verify(mockNotificationsService).sendSMSMessage(smsMessageProviderCaptor.capture());
+        
+        String message = smsMessageProviderCaptor.getValue().getSmsRequest().getMessage();
         assertTrue(message.contains("Account for ShortName already exists. Reset password: "));
         assertTrue(message.contains("/rp?study=api&sptoken="+SPTOKEN));
+        assertTrue(message.contains(" or ABC-DEF")); // token is formatted
```
DwayneJengSage: Notify account exists for phone also now sends a phone sign-in link. Should we test for that here?

test/org/sagebionetworks/bridge/sms/SmsMessageProviderTest.java
```diff
@@ -0,0 +1,61 @@
...
+        study.setConsentNotificationEmail("consent@email.com,consent2@email.com");
+
+        SmsTemplate template = new SmsTemplate("${studyShortName} ${url} ${supportEmail}");
+        
+        // Create
+        SmsMessageProvider provider = new SmsMessageProvider.Builder()
+            .withStudy(study)
+            .withPhone(TestConstants.PHONE)
+            .withSmsTemplate(template)
+            .withToken("url", "some-url").build();
```
DwayneJengSage, line 37: Should this include withExpirationPeriod()?

test/org/sagebionetworks/bridge/sms/SmsMessageProviderTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+        // Check email
+        PublishRequest request = provider.getSmsRequest();
+        assertEquals("ShortName some-url support@email.com", request.getMessage());
+        
+        assertEquals(study.getShortName(),
+                request.getMessageAttributes().get(BridgeConstants.SENDER_ID).getStringValue());
+        assertEquals(BridgeConstants.SMS_TYPE_TRANSACTIONAL,
+                request.getMessageAttributes().get(BridgeConstants.SMS_TYPE).getStringValue());
```
DwayneJengSage: Should we also check the provider's getters as a sanity test?

app/org/sagebionetworks/bridge/sms/SmsMessageProvider.java
```diff
@@ -0,0 +1,99 @@
...
+
+    public Study getStudy() {
+        return study;
+    }
+    public SmsTemplate getTemplate() {
+        return template;
+    }
+    public Phone getPhone() {
+        return phone;
+    }
```
DwayneJengSage, line 41: Should we have a getter for the token map to facilitate testing?

DwayneJengSage (changes requested):
Overall looks good. A few notes.

app/org/sagebionetworks/bridge/BridgeUtils.java
alxdarksage: Added

app/org/sagebionetworks/bridge/models/studies/Study.java
alxdarksage: removed

test/org/sagebionetworks/bridge/services/AccountWorkflowServiceTest.java
alxdarksage: For phone, a token is sent back to be entered into the app. That's what the " or ABC-DEF" test is, reformatted it to " or " + TOKEN

test/org/sagebionetworks/bridge/sms/SmsMessageProviderTest.java
alxdarksage: yes, added

app/org/sagebionetworks/bridge/sms/SmsMessageProvider.java
alxdarksage: I didn't use it so I removed it, but I can add it back. It's consistent with MimeTypeEmailProvider. Also added direct tests of the map.

I realized as well that the map should be initialized after you call build(), not after you call the constructor method to get an sms message or email message. So I moved that and fixed the case where null values are added to the map, which will break ImmutableMap.of().

DwayneJengSage (approved):

PR #1672, opened 2018-03-14
upload file size limits

DwayneJengSage:
Impose file size limits for parsing uploads in memory, so that we can raise general file size limits.


alxdarksage (approved):
Looks fine. But we can't parse files over 20mb even if we use disk space?

PR #1674, opened 2018-03-16
External id refactor

alxdarksage:
We no longer reserve external IDs because assignment of the ID and an email with the ID in it are to the same table, and there's a unique index on external IDs. So two people can never have the same ID with conflicting email addresses. We update the ExternalId table after the save to the accounts table.

Also recasting external ID to be an identifier like email or phone: it can be added on create, but cannot be changed or added on update. It can be added through the identifiers API.

If the ID is in the ExternalId table, it is always assigned. If validation is turned on and the ID is NOT in the table, then the new account will not validate. As before, if external ID is required on sign on, we check that during validation as well.


app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -313,22 +293,25 @@ public void createAccountForMigration(Study study, Account account, String id) {
...
-            // Account can conflict because studyId + email or studyId + phone has been used for an 
+            // Account can conflict because studyId + email|phone|externalId|healthCode have been used for an 
             // existing account. 
             AccountId accountId = null;
             if (hibernateAccount.getEmail() != null) {
                 accountId = AccountId.forEmail(study.getIdentifier(), account.getEmail());
             } else if (hibernateAccount.getPhone() != null) {
                 accountId = AccountId.forPhone(study.getIdentifier(), account.getPhone());
+            } else if (hibernateAccount.getExternalId() != null) {
+                accountId = AccountId.forExternalId(study.getIdentifier(), account.getExternalId());
```
DwayneJengSage, line 304: Should we add a unit test to HibernateAccountDaoTest in the style of createAccountAlreadyExists() and createAccountAlreadyExistsForPhoneAccount()?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -1139,11 +1151,12 @@ public void updateIdentifiersDoNotOverwriteExistingIdentifiers() {
         account.setEmailVerified(Boolean.TRUE);
         account.setPhone(TestConstants.PHONE);
         account.setPhoneVerified(Boolean.TRUE);
+        account.setExternalId(EXTERNAL_ID);
         when(accountDao.authenticate(STUDY, PHONE_PASSWORD_SIGN_IN)).thenReturn(account);
         when(accountDao.getAccount(any())).thenReturn(account);
         
         IdentifierUpdate update = new IdentifierUpdate(PHONE_PASSWORD_SIGN_IN, "updated@email.com",
-                new Phone("4082588569", "US"));
+                new Phone("4082588569", "US"), EXTERNAL_ID);
```
DwayneJengSage: Should this be a different external ID? Otherwise, we can't distinguish between "we didn't overwrite the external ID" and "we overwrote the external ID with the same value".

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -530,6 +529,9 @@ public StudyParticipant updateIdentifiers(Study study, CriteriaContext context,
             !study.isAutoVerificationEmailSuppressed()) {
             accountWorkflowService.sendEmailVerificationToken(study, account.getId(), account.getEmail());
         }
+        if (assignExternalId) {
+            externalIdService.assignExternalId(study, account.getExternalId(), account.getHealthCode());
+        }
```
DwayneJengSage, line 537: Should there be a test that verifies that not assigning an external ID doesn't try to assign the external ID again?

app/org/sagebionetworks/bridge/validators/IdentifierUpdateValidator.java
```diff
@@ -39,13 +39,15 @@ public void validate(Object object, Errors errors) {
         if (update.getEmailUpdate() != null) {
             updateFields++;
         }
+        if (update.getExternalIdUpdate() != null) {
+            updateFields++;
+        }
```
DwayneJengSage: Should we add a unit test for valid ExternalId update? (Test is also missing valid phone update.)

test/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
```diff
@@ -94,122 +104,164 @@ public void validatesUpdate() {
...
+        ExternalIdentifier externalIdentifier = new DynamoExternalIdentifier();
+        when(externalIdService.getExternalId(study.getStudyIdentifier(), "foo")).thenReturn(externalIdentifier);
+
+        StudyParticipant participant = new StudyParticipant.Builder().withEmail("email@email.com")
+                .withExternalId("foo").withId("id").withPassword("pAssword1@").build();
+        
+        validator = new StudyParticipantValidator(externalIdService, study, false);
+        study.setExternalIdValidationEnabled(true);
+        Validate.entityThrowingException(validator, participant);
+    }
```
DwayneJengSage, line 310: What about the following cases?

* External ID validation required, but not specified?
* External ID specified, but validation not required?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -250,14 +255,16 @@ public void createParticipant() {
         STUDY.setEmailVerificationEnabled(true);
         mockHealthCodeAndAccountRetrieval();
         
+        when(externalIdService.getExternalId(STUDY.getStudyIdentifier(), EXTERNAL_ID))
+                .thenReturn(new DynamoExternalIdentifier());
```
DwayneJengSage: ExternalIdentifier.create() to avoid leaking DDB implementation to non-DDB classes. (Here and several other places in this PR.)

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -250,14 +255,16 @@ public void createParticipant() {
...
+        
         IdentifierHolder idHolder = participantService.createParticipant(STUDY, CALLER_ROLES, PARTICIPANT, true);
         assertEquals(ID, idHolder.getIdentifier());
         
-        verify(externalIdService).reserveExternalId(STUDY, EXTERNAL_ID, HEALTH_CODE);
         verify(externalIdService).assignExternalId(STUDY, EXTERNAL_ID, HEALTH_CODE);
         
-        verify(accountDao).constructAccount(STUDY, EMAIL, PHONE, PASSWORD);
```
DwayneJengSage: Do we want to verify the call to constructAccount()?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: Added createAccountAlreadyExistsForExternalIdAccount() where you're creating an account with only an external ID. It's used to retrieve the existing account.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Yes definitely, made that a different string

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Added, and added more verifications that the externalIdService is not called in scenarios where no assignment of the ext ID should be taking place.

app/org/sagebionetworks/bridge/validators/IdentifierUpdateValidator.java
alxdarksage: added tests for both these cases

test/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
alxdarksage: Added test for the first, the second is the externalIdForNewParticipantValidatedDoesNotExist() test. Also added a test that adding an invalid external ID on updates is ignored (because you can't change email, phone or external ID once an account is created unless you go through updateIdentifiers, which uses a different validator).

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Changed these (all new DynamoStudy() --> Study.create()).

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Added this back

alxdarksage: Added this back

test/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
DwayneJengSage: I think that's a different test case. That's for if the external ID is specified but doesn't exist. We're missing a test case for if external ID validation is enabled, but no external ID is specified. Which if I'm reading the code correctly, would succeed.

test/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
```diff
@@ -94,122 +102,186 @@ public void validatesUpdate() {
...
+    public void validateExistingManagedExternalId() {
+        when(externalIdService.getExternalId(study.getStudyIdentifier(), "foo")).thenReturn(EXT_ID);
+
+        StudyParticipant participant = new StudyParticipant.Builder().withEmail("email@email.com")
+                .withExternalId("foo").withId("id").withPassword("pAssword1@").build();
+        
+        validator = new StudyParticipantValidator(externalIdService, study, false);
+        study.setExternalIdValidationEnabled(true);
+        Validate.entityThrowingException(validator, participant);
+    }
```
DwayneJengSage, line 259: Still need a test case for External ID validation required, but not specified.

DwayneJengSage (changes requested):
Needs one more test. Otherwise, looks good.

test/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
alxdarksage: Added this test (it correctly validates the missing externalID is required).

alxdarksage: Added a test that a missing externalId on sign up, where it is required, is an error. The reason we don't validate an external ID that is added on an update is that we no longer allow you to update the external ID this way. HOWEVER you're right, we don't validate that external ID when it is added in the same way that we do with this validator, so I've added that to the IdentifierUpdateValidator, and added tests for it. While we're not really supporting this yet, it's there and could be called.

test/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
alxdarksage: Added (external ID is required but no ID is provided).

test/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
```diff
@@ -94,122 +102,200 @@ public void validatesUpdate() {
...
+        
+        StudyParticipant participant = new StudyParticipant.Builder().withEmail("email@email.com")
+                .withId("id").withPassword("pAssword1@").build();
+        
+        // This should fail. External ID is required on sign up but there's no externalId
+        validator = new StudyParticipantValidator(externalIdService, study, true);
+        study.setExternalIdValidationEnabled(true);
+        study.setExternalIdRequiredOnSignup(true);
+        assertValidatorMessage(validator, participant, "externalId", "is required");
+    }
```
DwayneJengSage, line 297: I think we're still missing a test case. If external ID validation is enabled, but no external ID is specified, the validation would succeed. However, there is no test that tests this case.

DwayneJengSage (changes requested):
Still missing the test case.

DwayneJengSage (approved):

PR #1677, opened 2018-03-21
participant signOut API can delete the session without deleting the reauth token

DwayneJengSage:
Currently, it's very difficult to test reauth, because we have to wait 12 hours for the session to expire before we can trigger the reauth code. This change changes that by adding a deleteReauthToken flag to the participant signOut API.

This is added to the researcher participant signOut API, and not to the standard signOut API, because we don't want to open this up to end users.

deleteReauthToken defaults to true for backwards compatibility.

Testing done:
* unit tests
* integ tests
* manual tests

### DIFF
--- a/app/org/sagebionetworks/bridge/dao/AccountDao.java
+++ b/app/org/sagebionetworks/bridge/dao/AccountDao.java
@@ -23,7 +23,7 @@ import org.sagebionetworks.bridge.services.AuthenticationService;
  */
 public interface AccountDao {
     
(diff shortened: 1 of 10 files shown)


liujoshua (approved):

alxdarksage:
I'm not sure what the security considerations are for this, but we should consider before we document this as public.

alxdarksage (approved):

alxdarksage:
I'll add this to the Bridge Study Manager when I get into the office...

PR #1678, opened 2018-03-22
Phone verification

alxdarksage:
Phone verification workflow added with a flag to suppress sending a verification SMS message on sign up.


app/org/sagebionetworks/bridge/models/accounts/Identifier.java
```diff
@@ -6,25 +6,39 @@
...
+    private final String email;
+    private final Phone phone;
     
-    public Email(@JsonProperty("study") String study, @JsonProperty("email") String email) {
+    public Identifier(@JsonProperty("study") String study, @JsonProperty("email") String email,
+            @JsonProperty("phone") Phone phone) {
         this.studyIdentifier = (study != null) ? new StudyIdentifierImpl(study) : null;
         this.email = email;
+        this.phone = phone;
     }
```
DwayneJengSage: Are we going to add new identifiers to this in the future, such as external ID? If we do, are we just going to add more and more constructors to this?

Would we be better off trying to consolidate this class with AccountId instead?

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerMockTest.java
```diff
@@ -656,10 +659,37 @@ public void resendEmailVerificationAppVersionDisabled() throws Exception {
...
+        study.getMinSupportedAppVersions().put(OperatingSystem.IOS, 20);
+        
+        controller.resendPhoneVerification();
+    }
+
+    @Test(expected = EntityNotFoundException.class)
+    public void resendPhoneVerificationNoStudy() throws Exception {
+        mockPlayContextWithJson(new Identifier((StudyIdentifier) null, TestConstants.PHONE));
+        controller.resendPhoneVerification();
+    }
```
DwayneJengSage: Add unit test for verifyPhone?

app/org/sagebionetworks/bridge/services/AccountWorkflowService.java
```diff
@@ -225,17 +227,45 @@ public void sendEmailVerificationToken(Study study, String userId, String recipi
...
+        
+        if (phone == null) {
+            return;
+        }
+        if (isRequestThrottled(ThrottleRequestType.VERIFY_PHONE, userId)) {
+            // Too many requests. Throttle.
+            return;
+        }
+        String sptoken = getNextToken();
+        String formattedSpToken = sptoken.substring(0,3) + "-" + sptoken.substring(3,6);
```
DwayneJengSage: We generate the token, but we don't actually store it anywhere. So when the user goes to verify their phone, they will always get the "token invalid" error.

Also, VerificationData currently doesn't distinguish between phone and email. This allows users to receive the token on their email, and then use it to verify their phone,

test/org/sagebionetworks/bridge/services/AccountWorkflowServiceTest.java
```diff
@@ -205,14 +207,47 @@ public void sendEmailVerificationTokenThrottled() {
         verify(mockSendMailService, times(2)).sendEmail(any());
     }
 
+    @Test
+    public void sendPhoneVerificationToken() throws Exception {
+        when(service.getNextToken()).thenReturn("012345");
+        
+        service.sendPhoneVerificationToken(study, USER_ID, TestConstants.PHONE);
+        
+        verify(mockNotificationsService).sendSMSMessage(smsMessageProviderCaptor.capture());
```
DwayneJengSage, line 225: This test should also verify that we save the token to Redis. (Incidentally, this makes me realize that I neglected to do the same for email verification.)

test/org/sagebionetworks/bridge/services/AccountWorkflowServiceTest.java
```diff
@@ -205,14 +207,47 @@ public void sendEmailVerificationTokenThrottled() {
...
+        Map<String,String> tokens = provider.getTokenMap();
+        assertEquals("012-345", tokens.get("sptoken"));
+        assertEquals("2 hours", tokens.get("phoneVerificationExpirationPeriod"));
+        
+        String message = provider.getSmsRequest().getMessage();
+        assertTrue(message.contains("012-345"));
+    }
+    
+    @Test
+    public void sendPhoneVerificationTokenNoEmail() throws Exception {
```
DwayneJengSage: Nitpick: NoEmail -> NoPhone

app/org/sagebionetworks/bridge/services/AccountWorkflowService.java
```diff
@@ -225,17 +227,45 @@ public void sendEmailVerificationToken(Study study, String userId, String recipi
...
         
         Study study = studyService.getStudy(accountId.getStudyId());
         Account account = accountDao.getAccount(accountId);
         if (account != null) {
-            sendEmailVerificationToken(study, account.getId(), account.getEmail());
+            if (type == ChannelType.EMAIL) {
+                sendEmailVerificationToken(study, account.getId(), account.getEmail());
+            } else if (type == ChannelType.PHONE) {
+                sendPhoneVerificationToken(study, account.getId(), account.getPhone());
+            }
```
DwayneJengSage: Should we have an else that throws a "impossible code path" exception? This way, if we ever add another channel and don't update this code, we'll get an exception instead of a silent failure.

app/org/sagebionetworks/bridge/services/AccountWorkflowService.java
```diff
@@ -247,12 +277,12 @@ public void resendEmailVerificationToken(AccountId accountId) {
...
-    public Account verifyEmail(EmailVerification verification) {
+    public Account verifyChannel(Verification verification) {
         checkNotNull(verification);
 
         VerificationData data = restoreVerification(verification.getSptoken());
         if (data == null) {
-            throw new BadRequestException(VERIFY_EMAIL_TOKEN_EXPIRED);
+            throw new BadRequestException(VERIFY_TOKEN_EXPIRED);
         }
         Study study = studyService.getStudy(data.getStudyId());
```
DwayneJengSage: verifyChannel() needs to know what the channel is, as does the token. Otherwise, someone could use the token from their email to verify their phone (or vice versa).

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -219,28 +218,31 @@ public IdentifierHolder signUp(Study study, StudyParticipant participant, boolea
...
     }
     
-    public void resendEmailVerification(StudyIdentifier studyIdentifier, Email email) {
-        checkNotNull(studyIdentifier);
-        checkNotNull(email);
+    public void resendVerification(ChannelType type, Identifier identifier) {
+        checkNotNull(identifier);
         
-        Validate.entityThrowingException(EmailValidator.INSTANCE, email);
+        Validate.entityThrowingException(new IdentifierValidator(type), identifier);
```
DwayneJengSage: Instead of creating a new validator with each request, would it be better to have singletons for each channel, and then a map that maps the channel to the validator?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -887,9 +969,9 @@ public void deleteActivitiesNoUserThrowsCorrectException() {
     public void resendEmailVerification() {
         mockHealthCodeAndAccountRetrieval();
         
-        participantService.resendEmailVerification(STUDY, ID);
+        participantService.resendVerification(STUDY, ChannelType.EMAIL, ID);
         
-        verify(accountWorkflowService).resendEmailVerificationToken(accountIdCaptor.capture());
+        verify(accountWorkflowService).resendVerificationToken(eq(ChannelType.EMAIL), accountIdCaptor.capture());
```
DwayneJengSage, line 932: Do we also need to add a version of this test for resendPhoneVerification?

app/org/sagebionetworks/bridge/validators/IdentifierValidator.java
```diff
@@ -0,0 +1,42 @@
+package org.sagebionetworks.bridge.validators;
+
+import org.apache.commons.lang3.StringUtils;
+import org.sagebionetworks.bridge.models.accounts.Identifier;
+import org.sagebionetworks.bridge.models.accounts.Phone;
+import org.sagebionetworks.bridge.services.AuthenticationService.ChannelType;
+import org.springframework.validation.Errors;
+import org.springframework.validation.Validator;
+
+public class IdentifierValidator implements Validator {
```
DwayneJengSage: Unit tests?

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -512,4 +533,38 @@ public void languagesArePersistedFromContext() {
...
+        verify(accountWorkflowService).resendVerificationToken(eq(ChannelType.PHONE), accountIdCaptor.capture());
+        
+        assertEquals(TestConstants.TEST_STUDY_IDENTIFIER, accountIdCaptor.getValue().getStudyId());
+        assertEquals(TestConstants.PHONE, accountIdCaptor.getValue().getPhone());
+    }
+    
+    @Test(expected = InvalidEntityException.class)
+    public void resendPhoneVerificationInvalid() {
+        Identifier identifier = new Identifier(null, (Phone)null);
+        service.resendVerification(ChannelType.EMAIL, identifier);
```
DwayneJengSage: Minor: Should be ChannelType.PHONE

app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -537,6 +537,7 @@ public Study updateStudy(Study study, boolean isAdminUpdate) {
             study.setReauthenticationEnabled(originalStudy.isReauthenticationEnabled());
             study.setAccountLimit(originalStudy.getAccountLimit());
             study.setStudyIdExcludedInExport(originalStudy.isStudyIdExcludedInExport());
+            study.setAutoVerificationPhoneSuppressed(originalStudy.isAutoVerificationPhoneSuppressed());
```
DwayneJengSage: Non-admins can set AutoVerificationEmailSuppressed. Should this be made consistent?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -403,6 +412,53 @@ public void createParticipantPhoneNoEmailVerificationWanted() {
...
+        mockHealthCodeAndAccountRetrieval(null, PHONE);
+
+        participantService.createParticipant(study, CALLER_ROLES, PARTICIPANT, true);
+
+        verify(accountWorkflowService, never()).sendPhoneVerificationToken(any(), any(), any());
+        assertEquals(AccountStatus.UNVERIFIED, account.getStatus());
+        assertNull(account.getPhoneVerified());
+    }
+
+    public void createParticipantEmailNoPhoneVerificationWanted() {
```
DwayneJengSage: Missing Test annotation?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/models/accounts/Identifier.java
alxdarksage: Switched to using AccountId wherever Identifier is located, works as is.

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerMockTest.java
alxdarksage: Not sure what we're verifying... phone is validated in the service?

DwayneJengSage: There are currently no tests on verifyPhone(). There should be tests on verifyPhone().

app/org/sagebionetworks/bridge/services/AccountWorkflowService.java
alxdarksage: Added that, added verification of the cache call in sendEmailVerificationToken and sendPhoneVerificationToken tests (because that's obviously missing). Added the channel type to the verifyChannel method, and now check it on verification. Added tests to verify this is enforced.

test/org/sagebionetworks/bridge/services/AccountWorkflowServiceTest.java
alxdarksage: Added, and I went through AccountWorkflowServiceTest and added verifications of the cache provider in any other tests where they were missing.

test/org/sagebionetworks/bridge/services/AccountWorkflowServiceTest.java
alxdarksage: changed

app/org/sagebionetworks/bridge/services/AccountWorkflowService.java
alxdarksage: Added everywhere we branch like this. Discovered there are no tests for AccountIdValidator, added that.

app/org/sagebionetworks/bridge/services/AccountWorkflowService.java
alxdarksage: This has been changed. Thanks for catching this.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage: Added a factory method to retrieve singletons

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Added

app/org/sagebionetworks/bridge/validators/IdentifierValidator.java
alxdarksage: Now AccountIdValidator, added AccountIdValidatorTest

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: This should be fixed

app/org/sagebionetworks/bridge/services/StudyService.java
alxdarksage: Seems fine for this (and any other setting that adjusts workflow) to be a developer setting. Removed and removed from tests.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Yes, added!

test/org/sagebionetworks/bridge/play/controllers/AuthenticationControllerMockTest.java
alxdarksage: Added test

test/org/sagebionetworks/bridge/services/AccountWorkflowServiceTest.java
```diff
@@ -172,6 +175,12 @@ public void sendEmailVerificationToken() throws Exception {
         service.sendEmailVerificationToken(study, USER_ID, EMAIL);
         
         verify(mockSendMailService).sendEmail(emailProviderCaptor.capture());
+        verify(mockCacheProvider).setObject(eq(SPTOKEN), stringCaptor.capture(), eq(AccountWorkflowService.VERIFY_OR_RESET_EXPIRE_IN_SECONDS));
+        
+        String string = stringCaptor.getValue();
+        JsonNode node = BridgeObjectMapper.get().readTree(string);
+        assertEquals("api", node.get("studyId").textValue());
+        assertEquals("userId", node.get("userId").textValue());
```
DwayneJengSage, line 183: Should this include channel? (Here and in sendPhoneVerificationToken())

test/org/sagebionetworks/bridge/services/StudyServiceTest.java
```diff
@@ -361,6 +363,7 @@ private void assertStudyDefaults(Study study) {
     private void assertStudyChanged(Study study) {
         assertFalse(study.isStudyIdExcludedInExport());
         assertFalse(study.isEmailVerificationEnabled());
+        assertTrue(study.isAutoVerificationPhoneSuppressed());
```
DwayneJengSage, line 366: Minor: It's probably not necessary to have this in assertStudyChanged(), setStudyDefaults(), or changeStudyDefaults().

test/org/sagebionetworks/bridge/validators/AccountIdValidatorTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.sagebionetworks.bridge.json.BridgeObjectMapper;
+import org.sagebionetworks.bridge.models.accounts.AccountId;
+import org.sagebionetworks.bridge.services.AuthenticationService.ChannelType;
+
+public class AccountIdValidatorTest {
+    @Test
+    public void validAccountId() {
+        AccountId accountId = AccountId.forEmail(TestConstants.TEST_STUDY_IDENTIFIER, "email@email.com");
+        Validate.entityThrowingException(AccountIdValidator.getInstance(ChannelType.EMAIL), accountId);
+    }
```
DwayneJengSage, line 17: We should probably also have the equivalent test for valid phone.

DwayneJengSage (approved):
Looks good. A few minor test notes.

PR #1681, opened 2018-03-29
Update after latest code review.

alxdarksage:


app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -397,8 +397,6 @@ public void resendVerification(Study study, ChannelType type, String userId) {
         } else if (type == ChannelType.PHONE && participant.getPhone() != null) {
             AccountId accountId = AccountId.forPhone(study.getIdentifier(), participant.getPhone());
             accountWorkflowService.resendVerificationToken(type, accountId);
-        } else {
-            throw new UnsupportedOperationException("Channel type not implemented");
```
DwayneJengSage: What's the problem with throwing this exception? If we're not passing in a channel, this is almost certainly a bug in the server. If this happens, it would be better to fast-fail the request with a 500 Internal Server Error than to do nothing and mask the error.

test/org/sagebionetworks/bridge/services/StudyServiceTest.java
```diff
@@ -332,7 +332,15 @@ public void problematicHtmlIsRemovedFromTemplates() {
...
-        setStudyDefaults(study);
+        study.setStudyIdExcludedInExport(true);
+        study.setEmailVerificationEnabled(true);
+        study.setExternalIdValidationEnabled(false);
+        study.setExternalIdRequiredOnSignup(false);
+        study.setEmailSignInEnabled(false);
+        study.setPhoneSignInEnabled(false);
+        study.setReauthenticationEnabled(false);
+        study.setAccountLimit(0);
+        study.setAutoVerificationPhoneSuppressed(false);
```
DwayneJengSage: It's probably not necessary to include autoVerificationPhoneSuppressed in this test (set or assert), or in changeStudyDefaults().

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: You get a 500 error back if the participant does not have an email or phone address, which just leaks information about the account. This isn't a pure test of the enumeration handling alone.

alxdarksage: Why don't I throw a BadRequestException which says "this is not a valid request to resent a verification." This seems... vague enough not to indicate anything about the account.

DwayneJengSage: Okay, I see it now. It's the compound clauses on the if statements. I think you need nested if statements.

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -398,7 +398,7 @@ public void resendVerification(Study study, ChannelType type, String userId) {
             AccountId accountId = AccountId.forPhone(study.getIdentifier(), participant.getPhone());
             accountWorkflowService.resendVerificationToken(type, accountId);
         } else {
-            throw new UnsupportedOperationException("Channel type not implemented");
+            throw new BadRequestException("Invalid request to resend verification");
```
alxdarksage: Note: This will fail an integration test as before, this would silently "succeed" while doing nothing.

DwayneJengSage (approved):

PR #1682, opened 2018-03-30
Allow external ID to be updated

alxdarksage:
Erin and I verified yesterday that mPower adds an external ID _after_ account creation, using the existing APIs (not the newer identifier update API). Consequently we have to relax this restriction and allow external IDs to be added on update to accounts that don't have external IDs already set.

I rewrote all the StudyParticipantValidator tests for external ID validation to make sure I'm covering all the cases.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/ParticipantService.java
+++ b/app/org/sagebionetworks/bridge/services/ParticipantService.java
@@ -302,6 +302,10 @@ public class ParticipantService {
 
         // Prevent optimistic locking exception until operations are combined into one operation. 
         account = accountDao.getAccount(AccountId.forId(study.getIdentifier(), account.getId()));
+        // Allow external ID to be added on an update if it doesn't exist.
+        if (account.getExternalId() == null) {
+            account.setExternalId(participant.getExternalId());    
+        }
         updateAccountAndRoles(study, callerRoles, account, participant);
         
         // Only Admin and Worker accounts controlled by us should be able to bypass email verification. This is
@@ -313,6 +317,8 @@ public class ParticipantService {
             }
         }
         accountDao.updateAccount(account, false);
+        
+        externalIdService.assignExternalId(study, account.getExternalId(), account.getHealthCode());    
     }
 
     private void throwExceptionIfLimitMetOrExceeded(Study study) {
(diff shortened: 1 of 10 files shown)


app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -313,6 +317,8 @@ public void updateParticipant(Study study, Set<Roles> callerRoles, StudyParticip
             }
         }
         accountDao.updateAccount(account, false);
+        
+        externalIdService.assignExternalId(study, account.getExternalId(), account.getHealthCode());
```
DwayneJengSage: Should we only assign the external ID if it changes?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -1382,18 +1368,19 @@ public void createUnmanagedExternalIdWillAssign() {
...
         StudyParticipant participant = new StudyParticipant.Builder().copyOf(PARTICIPANT)
                 .withExternalId("newExternalId").build();
         
         participantService.updateParticipant(STUDY, CALLER_ROLES, participant);
         
-        assertEquals(EXTERNAL_ID, account.getExternalId());
-        verify(externalIdService, never()).assignExternalId(any(), any(), any());
+        assertEquals("newExternalId", account.getExternalId());
+        // This is always called, but it does nothing since there's no record in the table.
+        verify(externalIdService).assignExternalId(any(), eq("newExternalId"), any());
```
DwayneJengSage: We should also have test cases for

* deleting external ID
* external ID changes
* same external ID

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Changed this, this does prevent loading the ID just to verify it is assigned to this person over and over

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Added tests or renamed for consistency:

For the managed external ID update case:
addingManagedExternalIdOnUpdateOK()
changingManagedExternalIdIgnored()
sameManagedExternalIdOnUpdateIgnored()
removingManagedExternalIdIgnored()

Also missingExternalIdThrows() and updateMissingManagedExternalIdFails()

For the unmanaged external ID update case:
addingUnmanagedExternalIdOnUpdateOK()
changingUnmanagedExternalIdIgnored()
sameUnmanagedExternalIdOnUpdateIgnored()
removingUnmanagedExternalIdIgnored()

DwayneJengSage (approved):

PR #1685, opened 2018-04-04
Adding API to retrieve uploads by upload or record ID.

alxdarksage:
There's also an API to retrieve all study uploads, so I will add an "Uploads" section to the BSM where you can look up a specific upload record.

### DIFF
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
@@ -218,12 +218,14 @@ public class DynamoUploadDao implements UploadDao {
         if (offsetKey != null) {
             // load table again to get the one last evaluated upload
             DynamoUpload2 retLastEvaluatedUpload = mapper.load(DynamoUpload2.class, offsetKey);
-
-            Map<String,AttributeValue> map = new HashMap<>();
-            map.put(UPLOAD_ID, new AttributeValue().withS(offsetKey));
-            map.put(REQUESTED_ON, new AttributeValue().withN(String.valueOf(retLastEvaluatedUpload.getRequestedOn())));
-            map.put(STUDY_ID, new AttributeValue().withS(studyId.getIdentifier()));
-            query.withExclusiveStartKey(map);
+            
+            if (retLastEvaluatedUpload != null) {
+                Map<String,AttributeValue> map = new HashMap<>();
+                map.put(UPLOAD_ID, new AttributeValue().withS(offsetKey));
+                map.put(REQUESTED_ON, new AttributeValue().withN(String.valueOf(retLastEvaluatedUpload.getRequestedOn())));
+                map.put(STUDY_ID, new AttributeValue().withS(studyId.getIdentifier()));
+                query.withExclusiveStartKey(map);
+            }
         }
         query.withLimit(pageSize);
         return query;
--- a/app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
+++ b/app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
@@ -218,14 +218,14 @@ public class DynamoUploadDao implements UploadDao {
         if (offsetKey != null) {
             // load table again to get the one last evaluated upload
             DynamoUpload2 retLastEvaluatedUpload = mapper.load(DynamoUpload2.class, offsetKey);
-            
-            if (retLastEvaluatedUpload != null) {
-                Map<String,AttributeValue> map = new HashMap<>();
-                map.put(UPLOAD_ID, new AttributeValue().withS(offsetKey));
-                map.put(REQUESTED_ON, new AttributeValue().withN(String.valueOf(retLastEvaluatedUpload.getRequestedOn())));
-                map.put(STUDY_ID, new AttributeValue().withS(studyId.getIdentifier()));
-                query.withExclusiveStartKey(map);
+            if (retLastEvaluatedUpload == null) {
+                throw new BadRequestException("Invalid offsetKey: " + offsetKey);
             }
+            Map<String,AttributeValue> map = new HashMap<>();
+            map.put(UPLOAD_ID, new AttributeValue().withS(offsetKey));
+            map.put(REQUESTED_ON, new AttributeValue().withN(String.valueOf(retLastEvaluatedUpload.getRequestedOn())));
+            map.put(STUDY_ID, new AttributeValue().withS(studyId.getIdentifier()));
+            query.withExclusiveStartKey(map);
         }
         query.withLimit(pageSize);
         return query;
--- a/app/org/sagebionetworks/bridge/play/controllers/UploadController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/UploadController.java
@@ -4,6 +4,7 @@ import java.io.IOException;
 
 import org.sagebionetworks.bridge.Roles;
 import org.sagebionetworks.bridge.dao.HealthCodeDao;
+import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
 import org.sagebionetworks.bridge.exceptions.UnauthorizedException;
 import org.sagebionetworks.bridge.json.DateUtils;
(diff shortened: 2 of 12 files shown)


DwayneJengSage (changes requested):
I think there's some confusion who this API is for. It's not for external developers or researchers. It's for Bridge developers, to figure out what's going on with uploads and records, specifically so we don't need direct database access anymore.

With that in mind, we need APIs for get upload by upload ID, and get record by record ID. (And possibly convenience methods to get upload by record ID and, if present, record by upload ID.) And these APIs need to show _everything_. Otherwise, we just end up going directly into the database again.

Also, there are a lot of fields that we definitely _don't_ want to show external developers and researchers, so we want to lock these APIs down so that only Bridge developers can look at them.

alxdarksage:
Changed the permissions to be researcher or admin, and included the health data record when you retrieve an individual UploadView object (what we call Upload in the Java SDK, which has also been updated).

test/org/sagebionetworks/bridge/models/upload/UploadViewTest.java
```diff
@@ -23,6 +30,29 @@
     
     @Test
     public void canSerialize() throws Exception {
+        DynamoHealthDataRecord record = new DynamoHealthDataRecord();
```
DwayneJengSage: Minor: HealthDataRecord.create()

app/org/sagebionetworks/bridge/play/controllers/UploadController.java
```diff
@@ -135,4 +145,28 @@ public Result uploadComplete(String uploadId, String synchronous) throws Excepti
         // Upload validation status may contain the health data record. Use the filter to filter out health code.
         return okResult(HealthDataRecord.PUBLIC_RECORD_WRITER, validationStatus);
     }
+    
+    public Result getUpload(String uploadId) throws Exception {
+        UserSession session = getAuthenticatedSession(Roles.RESEARCHER, Roles.ADMIN);
```
DwayneJengSage: Do we want to expose this to researchers? I don't think they have any need or reason to be looking at uploads

app/org/sagebionetworks/bridge/play/controllers/UploadController.java
```diff
@@ -135,4 +145,28 @@ public Result uploadComplete(String uploadId, String synchronous) throws Excepti
...
+            }
+            uploadId = record.getUploadId();
+        }
+        UploadView uploadView = uploadService.getUploadView(uploadId);
+
+        // Even if the ID is valid, ultimately you cannot view an upload outside of your study
+        String studyId = session.getStudyIdentifier().getIdentifier();
+        if (!uploadView.getUpload().getStudyId().equals(studyId)) {
+            throw new UnauthorizedException();
+        }
```
DwayneJengSage: To make this consistent with other admin APIs, I think we want to allow looking up uploads across studies. Otherwise, we'll need to create and manage dozens of admin accounts, and that's not sustainable.

app/org/sagebionetworks/bridge/play/controllers/UploadController.java
```diff
@@ -135,4 +145,28 @@ public Result uploadComplete(String uploadId, String synchronous) throws Excepti
...
+        // This filters out healthCode
+        return okResult(HealthDataRecord.PUBLIC_RECORD_WRITER, uploadView);
```
DwayneJengSage: Do we want to filter out healthCode? It's important for us to see _everything_. Otherwise, there's no point in having this API anyway.

test/org/sagebionetworks/bridge/play/controllers/UploadControllerTest.java
```diff
@@ -267,7 +278,78 @@ public void getValidationStatusEnforcesHealthCodeMatch() throws Exception {
         doReturn(otherUserSession).when(controller).getSessionEitherConsentedOrInRole(Roles.RESEARCHER);
         controller.getValidationStatus(UPLOAD_ID);
     }
+    
+    @Test
+    public void getUploadById() throws Exception {
+        TestUtils.mockPlayContext();
+        doReturn(researcherSession).when(controller).getAuthenticatedSession(RESEARCHER, ADMIN);
+        
+        DynamoHealthDataRecord record = new DynamoHealthDataRecord();
```
DwayneJengSage: Minor: HealthDataRecord.create()

app/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -299,19 +307,10 @@ public Upload getUpload(@Nonnull String uploadId) {
...
-                    builder.withSchemaRevision(record.getSchemaRevision());
-                    builder.withHealthRecordExporterStatus(record.getSynapseExporterStatus());
-                }
-            }
-            return builder.build();
-        }).collect(Collectors.toList());
+        // This summary view is accessible to developers, so we do not include details of the health data record.
+        List<UploadView> views = list.getItems().stream()
+                .map(upload -> uploadToUploadView(upload, false))
+                .collect(Collectors.toList());
```
DwayneJengSage, line 313: Should our unit test verify that the health data record is not included?

DwayneJengSage (changes requested):
Also, a lot of fields in DynamoUpload are still tagged with JsonIgnore.

app/org/sagebionetworks/bridge/services/UploadService.java
alxdarksage: Added

app/org/sagebionetworks/bridge/play/controllers/UploadController.java
alxdarksage: Changed to only expose to admins

app/org/sagebionetworks/bridge/play/controllers/UploadController.java
alxdarksage: I removed this constraint. There may be certain things down the road that don't work in the Bridge Study Manager, like linking the health code back to a participant view, because the BSM doesn't really can't assume study switching like this. However the only pagination of uploads available for the UI is per study, so if you are browsing and not looking up a specific record regardless of the study you are in, it will all work as expected.

app/org/sagebionetworks/bridge/play/controllers/UploadController.java
alxdarksage: Added it back in.

test/org/sagebionetworks/bridge/play/controllers/UploadControllerTest.java
alxdarksage: Switched, here and elsewhere in test

alxdarksage:
Also removed @JsonIgnore annotations from DynamoUpload record.

test/org/sagebionetworks/bridge/models/upload/UploadViewTest.java
alxdarksage: Changed here and elsewhere in the test

app/org/sagebionetworks/bridge/dynamodb/DynamoUploadDao.java
```diff
@@ -218,12 +218,14 @@ public Upload getUpload(@Nonnull String uploadId) {
...
-            map.put(STUDY_ID, new AttributeValue().withS(studyId.getIdentifier()));
-            query.withExclusiveStartKey(map);
+            
+            if (retLastEvaluatedUpload != null) {
+                Map<String,AttributeValue> map = new HashMap<>();
+                map.put(UPLOAD_ID, new AttributeValue().withS(offsetKey));
+                map.put(REQUESTED_ON, new AttributeValue().withN(String.valueOf(retLastEvaluatedUpload.getRequestedOn())));
+                map.put(STUDY_ID, new AttributeValue().withS(studyId.getIdentifier()));
+                query.withExclusiveStartKey(map);
+            }
```
DwayneJengSage: What is the effect of adding this if block? Do we need a unit test to test this case?

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #1686, opened 2018-04-05
Suppress consent email

alxdarksage:
- users don't have to receive an email copy of their consent when they consent or withdraw (can be disabled per subpopulation);
- study administrators don't have to provide an email to receive consents every time they are signed. That's why we have them all in a database.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/ConsentService.java
+++ b/app/org/sagebionetworks/bridge/services/ConsentService.java
@@ -27,7 +27,6 @@ import org.sagebionetworks.bridge.models.subpopulations.StudyConsentView;
 import org.sagebionetworks.bridge.models.subpopulations.Subpopulation;
 import org.sagebionetworks.bridge.models.subpopulations.SubpopulationGuid;
 import org.sagebionetworks.bridge.services.email.ConsentEmailProvider;
-import org.sagebionetworks.bridge.services.email.MimeTypeEmailProvider;
 import org.sagebionetworks.bridge.services.email.WithdrawConsentEmailProvider;
 import org.sagebionetworks.bridge.validators.ConsentAgeValidator;
 import org.sagebionetworks.bridge.validators.Validate;
@@ -159,13 +158,16 @@ public class ConsentService {
         // Publish an enrollment event, set sharing scope 
         activityEventService.publishEnrollmentEvent(participant.getHealthCode(), withConsentCreatedOnSignature);
         
-        // Send email, if required.
-        if (sendEmail && !subpop.isAutoSendConsentSuppressed() && participant.getEmail() != null) {
-            MimeTypeEmailProvider consentEmail = new ConsentEmailProvider(study, participant.getTimeZone(),
-                    participant.getEmail(), withConsentCreatedOnSignature, sharingScope,
+        if (sendEmail) {
+            // If the user's email exists and email notification is not suppressed, add the user to the provider
+            String participantEmail = subpop.isAutoSendConsentSuppressed() ? null : participant.getEmail();
+            ConsentEmailProvider consentEmail = new ConsentEmailProvider(study, participant.getTimeZone(),
+                    participantEmail, withConsentCreatedOnSignature, sharingScope,
                     studyConsent.getDocumentContent(), consentTemplate);
-
-            sendMailService.sendEmail(consentEmail);
+            
+            if (!consentEmail.getRecipients().isEmpty()) {
+                sendMailService.sendEmail(consentEmail);
+            }
         }
     }
 
@@ -231,10 +233,7 @@ public class ConsentService {
         }
         accountDao.updateAccount(account, false);
         
-        Subpopulation subpop = subpopService.getSubpopulation(study.getStudyIdentifier(), subpopGuid);
-        if (!subpop.isAutoSendConsentSuppressed()) {
-            sendWithdrawEmail(study, participant.getExternalId(), account, withdrawal, withdrewOn);
-        }
+        sendWithdrawEmail(study, participant.getExternalId(), account, withdrawal, withdrewOn);
 
         return statuses;
     }
@@ -259,12 +258,8 @@ public class ConsentService {
         account.setSharingScope(SharingScope.NO_SHARING);
         accountDao.updateAccount(account, false);
 
-        for (SubpopulationGuid subpopGuid : account.getAllConsentSignatureHistories().keySet()) {
-            Subpopulation subpop = subpopService.getSubpopulation(study.getStudyIdentifier(), subpopGuid);
-            if (!subpop.isAutoSendConsentSuppressed()) {
-                sendWithdrawEmail(study, participant.getExternalId(), account, withdrawal, withdrewOn);
-            }
-        }
+        sendWithdrawEmail(study, participant.getExternalId(), account, withdrawal, withdrewOn);
+
         // But we don't need to query, we know these are all withdraw.
         return getConsentStatuses(context, account);
     }
@@ -280,9 +275,11 @@ public class ConsentService {
             // For backwards-compatibility, a null value means the email is verified.
             return;
         }
-        MimeTypeEmailProvider consentEmail = new WithdrawConsentEmailProvider(study, externalId, account, withdrawal,
-                withdrewOn);
-        sendMailService.sendEmail(consentEmail);
+        WithdrawConsentEmailProvider consentEmail = new WithdrawConsentEmailProvider(study, externalId, account,
+                withdrawal, withdrewOn);
+        if (!consentEmail.getRecipients().isEmpty()) {
+            sendMailService.sendEmail(consentEmail);    
+        }
     }
 
     /**
@@ -302,10 +299,11 @@ public class ConsentService {
         
         String htmlTemplate = studyConsentService.getActiveConsent(subpop).getDocumentContent();
         
-        if (participant.getEmail() != null) {
-            MimeTypeEmailProvider consentEmail = new ConsentEmailProvider(study, participant.getTimeZone(),
-                    participant.getEmail(), consentSignature, sharingScope, htmlTemplate, consentTemplate);
-            sendMailService.sendEmail(consentEmail);
+        String participantEmail = subpop.isAutoSendConsentSuppressed() ? null : participant.getEmail();
+        ConsentEmailProvider consentEmail = new ConsentEmailProvider(study, participant.getTimeZone(),
+                participantEmail, consentSignature, sharingScope, htmlTemplate, consentTemplate);
+        if (!consentEmail.getRecipients().isEmpty()) {
+            sendMailService.sendEmail(consentEmail);    
         }
     }
 
--- a/app/org/sagebionetworks/bridge/services/ConsentService.java
+++ b/app/org/sagebionetworks/bridge/services/ConsentService.java
@@ -299,9 +299,8 @@ public class ConsentService {
         
         String htmlTemplate = studyConsentService.getActiveConsent(subpop).getDocumentContent();
         
-        String participantEmail = subpop.isAutoSendConsentSuppressed() ? null : participant.getEmail();
         ConsentEmailProvider consentEmail = new ConsentEmailProvider(study, participant.getTimeZone(),
-                participantEmail, consentSignature, sharingScope, htmlTemplate, consentTemplate);
+                participant.getEmail(), consentSignature, sharingScope, htmlTemplate, consentTemplate);
         if (!consentEmail.getRecipients().isEmpty()) {
             sendMailService.sendEmail(consentEmail);    
         }
--- a/app/org/sagebionetworks/bridge/services/ConsentService.java
+++ b/app/org/sagebionetworks/bridge/services/ConsentService.java
@@ -160,7 +160,7 @@ public class ConsentService {
         activityEventService.publishEnrollmentEvent(participant.getHealthCode(), withConsentCreatedOnSignature);
         
         // Send email, if required.
-        if (sendEmail && participant.getEmail() != null) {
+        if (sendEmail && !subpop.isAutoSendConsentSuppressed() && participant.getEmail() != null) {
             MimeTypeEmailProvider consentEmail = new ConsentEmailProvider(study, participant.getTimeZone(),
                     participant.getEmail(), withConsentCreatedOnSignature, sharingScope,
                     studyConsent.getDocumentContent(), consentTemplate);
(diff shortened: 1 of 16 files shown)


app/org/sagebionetworks/bridge/services/ConsentService.java
```diff
@@ -256,8 +259,12 @@ public void consentToResearch(Study study, SubpopulationGuid subpopGuid, StudyPa
...
         accountDao.updateAccount(account, false);
 
-        sendWithdrawEmail(study, participant.getExternalId(), account, withdrawal, withdrewOn);
-
+        for (SubpopulationGuid subpopGuid : account.getAllConsentSignatureHistories().keySet()) {
+            Subpopulation subpop = subpopService.getSubpopulation(study.getStudyIdentifier(), subpopGuid);
+            if (!subpop.isAutoSendConsentSuppressed()) {
+                sendWithdrawEmail(study, participant.getExternalId(), account, withdrawal, withdrewOn);
+            }
+        }
```
DwayneJengSage: This used to only send one email. Now it sends an email for every subpop. This is probably not our intent.

Incidentally, we probably want a unit test case with multiple subpops. (Both send and don't send cases.)

app/org/sagebionetworks/bridge/services/ConsentService.java
```diff
@@ -160,7 +160,7 @@ public void consentToResearch(Study study, SubpopulationGuid subpopGuid, StudyPa
         activityEventService.publishEnrollmentEvent(participant.getHealthCode(), withConsentCreatedOnSignature);
         
         // Send email, if required.
-        if (sendEmail && participant.getEmail() != null) {
+        if (sendEmail && !subpop.isAutoSendConsentSuppressed() && participant.getEmail() != null) {
```
DwayneJengSage: Setting this flag means we send for neither to the participant nor to the study consent notification email. Is this the desired behavior? Should that flag be controlling both?

app/org/sagebionetworks/bridge/services/email/WithdrawConsentEmailProvider.java
```diff
@@ -45,9 +45,11 @@ public MimeTypeEmail getMimeTypeEmail() throws MessagingException {
...
         final String sendFromEmail = getFormattedSenderEmail();
         builder.withSender(sendFromEmail);
-
-        Set<String> emailAddresses = BridgeUtils.commaListToOrderedSet(getStudy().getConsentNotificationEmail());
-        builder.withRecipients(emailAddresses);
+        
+        if (getStudy().getConsentNotificationEmail() != null) {
+            Set<String> emailAddresses = BridgeUtils.commaListToOrderedSet(getStudy().getConsentNotificationEmail());
+            builder.withRecipients(emailAddresses);
+        }
```
DwayneJengSage: If there is no consent notification email, then there are no recipients for this email. What happens in this case?

There's also no corresponding unit test for this change.

test/org/sagebionetworks/bridge/validators/StudyValidatorTest.java
```diff
@@ -230,12 +230,6 @@ public void userProfileAttributesCannotBeEmpty() {
         assertValidatorMessage(INSTANCE, study, "userProfileAttributes", "'' must contain only digits, letters, underscores and dashes, and cannot start with a dash");
     }
     
-    @Test
-    public void requiresMissingConsentNotificationEmail() {
-        study.setConsentNotificationEmail(null);
-        assertValidatorMessage(INSTANCE, study, "consentNotificationEmail", "is required");
-    }
```
DwayneJengSage: Should this test instead verify that a null consent notification email is valid?

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
```diff
@@ -1459,6 +1461,47 @@ public void testAllFourTemplatesAreSanitized() {
...
+        
+        Study update = TestUtils.getValidStudy(StudyServiceMockTest.class);
+        update.setConsentNotificationEmail(newEmail);
+        // just assume this is true for the test so defaults aren't set
+        update.setConsentNotificationEmailVerified(true);
+        
+        service.updateStudy(update, true);
+        
+        boolean hasChanged = !Objects.equals(original.getConsentNotificationEmail(),
+                update.getConsentNotificationEmail());
```
DwayneJengSage: The test and the code both have the same code. If there's a bug in the code, the same bug is present in the test. Instead of trying to calculate the hasChanged, we should pass in the expected result and check against that instead.

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
```diff
@@ -1459,6 +1461,47 @@ public void testAllFourTemplatesAreSanitized() {
...
+        
+        if (update.getConsentNotificationEmail() == null) {
+            verify(sendMailService, never()).sendEmail(any());
```
DwayneJengSage: In a similar vein, it would be better to pass in an "expectedSendEmail" rather than trying to calculate it in the test.

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
```diff
@@ -1459,6 +1461,47 @@ public void testAllFourTemplatesAreSanitized() {
...
+            assertEquals(hasChanged, !update.isConsentNotificationEmailVerified());
```
DwayneJengSage: update.isConsentNotificationEmailVerified() is always the opposite of hasChanged. Would it be better to pull this out of the if-else block?

DwayneJengSage (changes requested):

DwayneJengSage:
Also, should new subpops have this flag set to true by default or false by default?

If it's false by default, would people just end up not bothering setting the flag, and we end up sending consent notification emails by default anyway?

app/org/sagebionetworks/bridge/services/ConsentService.java
alxdarksage: I felt like this should be the intended behavior, but now I'm pausing on the case where one consent is superseded by another one. It might be strange to get a withdrawal email for both if you reconnected to a new consent. Unfortunately there's no way to determine which consents still hold and which don't right now.

I can revert to sending one withdrawal, and then suppressing if *all* the consents suppress notification. In a study that has never relied on email for example.

alxdarksage: Added test of two consents, one of which suppresses email, and two where both suppress email. Behavior is as expected.

alxdarksage: I mistook the purpose of this email, it is actually only sent to the study consent administrator, not to the study participant. I will walk back the changes to withdrawal notifications... the only thing they need to do is not send if the study has no consent email set.

app/org/sagebionetworks/bridge/services/ConsentService.java
alxdarksage: No, it would be better for these to be separate. What I will implement instead is that the provider will assemble the recipients (user, study admin, or both, or neither) and then the SendMailService only sends email to the recipients that are listed by the provider.

test/org/sagebionetworks/bridge/validators/StudyValidatorTest.java
alxdarksage: Added a test to verify it can be left blank

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
alxdarksage: Changed to specify the expected result

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
alxdarksage: Refactor this code to take the two flags you suggest, and testing the sending of email and the email verification flag separately

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
alxdarksage: Created two separate blocks

alxdarksage:
We also discussed the default setting for subpopulation.autoSendConsentSuppressed being false, to match existing behavior, and decided it was okay to go with this. Probably we'll set it to true in any GUI for creating new studies.

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
```diff
@@ -491,6 +483,129 @@ public void withdrawFromOneOfTwoOptionalConsentsDoesNotChangeSharing() throws Ex
...
+        // easiest to test this if we null out the study consent email.
+        study.setConsentNotificationEmail(null);
+        when(subpopulation.isAutoSendConsentSuppressed()).thenReturn(true);
+        
+        // sendEmail = true because the system would otherwise send it based on the call, but hasn't looked to suppress yet.
+        consentService.consentToResearch(study, SUBPOP_GUID, participant, consentSignature,
+                SharingScope.ALL_QUALIFIED_RESEARCHERS, true);
+        
+        verify(sendMailService, never()).sendEmail(any());
+    }
```
DwayneJengSage, line 529: Should there also be a test case with sendMail=false?

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
```diff
@@ -491,6 +483,129 @@ public void withdrawFromOneOfTwoOptionalConsentsDoesNotChangeSharing() throws Ex
...
+    public void withdrawAllConsentsNoConsentAdministratorEmail() {
+        setupWithdrawTest(true, true);
+        study.setConsentNotificationEmail(null);
+        
+        consentService.withdrawAllConsents(study, participant, context, WITHDRAWAL, WITHDREW_ON);
+        
+        verify(sendMailService, never()).sendEmail(any());
+    }
+    
+    public void emailConsentAgreementNoConsentAdministratorEmail() {
```
DwayneJengSage: Missing Test annotation.

app/org/sagebionetworks/bridge/services/ConsentService.java
```diff
@@ -295,10 +299,11 @@ public void emailConsentAgreement(Study study, SubpopulationGuid subpopGuid, Stu
         
         String htmlTemplate = studyConsentService.getActiveConsent(subpop).getDocumentContent();
         
-        if (participant.getEmail() != null) {
-            MimeTypeEmailProvider consentEmail = new ConsentEmailProvider(study, participant.getTimeZone(),
-                    participant.getEmail(), consentSignature, sharingScope, htmlTemplate, consentTemplate);
-            sendMailService.sendEmail(consentEmail);
+        String participantEmail = subpop.isAutoSendConsentSuppressed() ? null : participant.getEmail();
```
DwayneJengSage: We shouldn't apply the suppress flag here. If the user requests the consent email, we need to send it to them.

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
```diff
@@ -1459,6 +1461,45 @@ public void testAllFourTemplatesAreSanitized() {
...
+        
+        Study update = TestUtils.getValidStudy(StudyServiceMockTest.class);
+        update.setConsentNotificationEmail(newEmail);
+        // just assume this is true for the test so defaults aren't set
+        update.setConsentNotificationEmailVerified(true);
+        
+        service.updateStudy(update, true);
+        
+        if (shouldBeChanged && expectedSendEmail) {
+            verify(sendMailService).sendEmail(any());
```
DwayneJengSage: This test should just key off expectedSendEmail. If we ever get into a case where shouldBeChanged=false and expectedSendEmail=true, that probably means we have an incorrect assumption somewhere.

app/org/sagebionetworks/bridge/services/email/ConsentEmailProvider.java
```diff
@@ -61,6 +63,22 @@ public ConsentEmailProvider(Study study, DateTimeZone userTimeZone, String userE
         this.consentTemplate = consentTemplate;
     }
 
+    public List<String> getRecipients() {
+        List<String> recipients = Lists.newArrayList();
```
DwayneJengSage: Calculating this on the fly means we end up executing this twice. Would it be better to just calculate this in the constructor and save it to a variable?

app/org/sagebionetworks/bridge/services/email/WithdrawConsentEmailProvider.java
```diff
@@ -36,6 +39,19 @@ public WithdrawConsentEmailProvider(Study study, String externalId, Account acco
...
+    public List<String> getRecipients() {
+        List<String> recipients = Lists.newArrayList();
+        // Check if consent notification email is verified. For backwards-compatibility, a null 
+        // value means the email is verified.
+        Boolean consentNotificationEmailVerified = getStudy().isConsentNotificationEmailVerified();
+        if (consentNotificationEmailVerified == null || consentNotificationEmailVerified) {
+            // Must wrap in new list because set from BridgeUtils.commaListToSet() is immutable
+            Set<String> studyRecipients = BridgeUtils.commaListToOrderedSet(getStudy().getConsentNotificationEmail());
+            recipients.addAll( studyRecipients );
+        }
```
DwayneJengSage: Unit tests for verified flag?

Unit tests for null withdraw email?

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
alxdarksage: Added

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
alxdarksage: Added. Thanks for catching that. Also have to set up the consent signature for that test.

app/org/sagebionetworks/bridge/services/ConsentService.java
alxdarksage: That makes sense. Removing and adjusting tests.

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
alxdarksage: Removed shouldBeChanged

app/org/sagebionetworks/bridge/services/email/ConsentEmailProvider.java
alxdarksage: Moved to constructor

app/org/sagebionetworks/bridge/services/email/WithdrawConsentEmailProvider.java
alxdarksage: Added

DwayneJengSage (approved):

PR #1687, opened 2018-04-05
Cache the reauthentication token to create a grace period where it can be re-used to retrieve the existing session.

alxdarksage:
Also tested this manually with CURL.

### DIFF
--- a/app/org/sagebionetworks/bridge/models/accounts/UserSession.java
+++ b/app/org/sagebionetworks/bridge/models/accounts/UserSession.java
@@ -54,7 +54,6 @@ public class UserSession {
     public void setInternalSessionToken(String internalSessionToken) {
         this.internalSessionToken = internalSessionToken;
     }
-    @JsonIgnore
     public String getReauthToken() {
         return reauthToken;
     }
--- a/app/org/sagebionetworks/bridge/models/accounts/UserSession.java
+++ b/app/org/sagebionetworks/bridge/models/accounts/UserSession.java
@@ -54,6 +54,7 @@ public class UserSession {
     public void setInternalSessionToken(String internalSessionToken) {
         this.internalSessionToken = internalSessionToken;
     }
+    @JsonIgnore
     public String getReauthToken() {
         return reauthToken;
     }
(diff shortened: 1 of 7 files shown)


app/org/sagebionetworks/bridge/models/accounts/UserSession.java
```diff
@@ -54,7 +54,6 @@ public String getInternalSessionToken() {
     public void setInternalSessionToken(String internalSessionToken) {
         this.internalSessionToken = internalSessionToken;
     }
-    @JsonIgnore
     public String getReauthToken() {
```
DwayneJengSage: IIRC the reason why we didn't include the reauth token in the session was so that we would produce it only in API calls where the user provides credentials. Would it be better to leave the JsonIgnore here and inject the reauth token into the session as part of reauth?

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -185,6 +202,8 @@ public void signOut(final UserSession session) {
         if (session != null) {
             AccountId accountId = AccountId.forId(session.getStudyIdentifier().getIdentifier(), session.getId());
             accountDao.deleteReauthToken(accountId);
+            String reauthCacheKey = getReauthCacheKey(session.getStudyIdentifier().getIdentifier(), session.getReauthToken());
+            cacheProvider.removeObject(reauthCacheKey);
```
DwayneJengSage, line 218: Should we add a check to the sign out unit test?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/models/accounts/UserSession.java
alxdarksage: Switched to storing session token and reauth token separately, and adding reauth token back to the session when it's recovered through this API's grace period.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage: Added

DwayneJengSage (approved):

PR #1688, opened 2018-04-06
Better cache key management

alxdarksage:
The cache key situation is pretty out of control. We're creating string keys all over the place with no centralization, and our oldest code uses a set of key classes that Eric created, that were difficult enough to use that we stopped using them. In addition, we're creating some keys that aren't namespaced at all. 

Centralized this and introduced some type-safety so it's easier to keep and ensure we maintain a centralized list of all keys, with an explicit whitelist of which keys are allowed in the cache API (as per BRIDGE-2135).

On the downside, this touches a lot of code.


app/org/sagebionetworks/bridge/cache/CacheKeys.java
```diff
@@ -0,0 +1,118 @@
...
+ * to store data in Redis. This makes it easier to determine which keys can be returned as 
+ * part of the administrative API so we can keep PII data out of the cache keys we return.
+ */
+public final class CacheKeys {
+    
+    private static final Joiner COLON_JOINER = Joiner.on(":");
+    private static final String[] PUBLIC_KEYS = new String[] { "emailVerificationStatus", "AppConfigList",
+            "channel-throttling", "lock", "study", "Subpopulation", "SubpopulationList" };
+    
+    public final static boolean isPublic(String key) {
```
DwayneJengSage: Unit test for isPublic()?

app/org/sagebionetworks/bridge/cache/CacheKeys.java
```diff
@@ -0,0 +1,118 @@
...
+        for (String suffix : PUBLIC_KEYS) {
+            if (key.endsWith(":"+suffix)) {
+                return true;
+            }
+        }
+        return false;
+    }
+    
+    public final static class CacheKey {
```
DwayneJengSage: Does this need to be an inner class? Would it be cleaner to have CacheKey be the outer class with a bunch of static util methods? (Or alternatively, split these two into two top-level classes?)

app/org/sagebionetworks/bridge/cache/CacheKeys.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+    public static final CacheKey lock(String value, Class<?> clazz) {
+        return new CacheKey(value, clazz.getCanonicalName(), "lock");
+    }
+    public static final CacheKey passwordResetForEmail(String sptoken, String studyId) {
+        return new CacheKey(sptoken, studyId); // no type, not great
+    }
+    public static final CacheKey reauthCacheKey(String reauthToken, String studyId) {
+        return new CacheKey(reauthToken, studyId, ":reauthCacheKey");
+    }
```
DwayneJengSage: Unit test for reauthCacheKey?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/cache/CacheKeys.java
alxdarksage: Reduced this to one class with static factory methods (CacheKey singular).

app/org/sagebionetworks/bridge/cache/CacheKeys.java
alxdarksage: Added

app/org/sagebionetworks/bridge/cache/CacheKeys.java
alxdarksage: Added

DwayneJengSage (approved):

PR #1691, opened 2018-04-12
Schedule on multiple event IDs

DwayneJengSage:
This is needed to model study bursts. Each study burst looks the same, so instead of creating 9 identical schedules, it would be better to create 1 schedule that can be scheduled off of 9 different event IDs. See https://sagebionetworks.jira.com/wiki/spaces/BRIDGE/pages/438665217/Activity+Burst+Notifications

### DIFF
--- a/app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
+++ b/app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
@@ -2,50 +2,66 @@ package org.sagebionetworks.bridge.models.schedules;
 
 import static org.sagebionetworks.bridge.models.schedules.ScheduleType.ONCE;
 
+import java.util.ArrayList;
 import java.util.List;
 
+import com.google.common.collect.ImmutableList;
 import org.joda.time.DateTime;
 import org.joda.time.LocalDate;
 import org.joda.time.LocalDateTime;
 import org.joda.time.LocalTime;
 
 import org.sagebionetworks.bridge.BridgeUtils;
+import org.sagebionetworks.bridge.models.RangeTuple;
 
 public abstract class ActivityScheduler {
     
     protected final Schedule schedule;
-    protected DateTime sequenceEndsOn;
-    
+
     ActivityScheduler(Schedule schedule) {
         this.schedule = schedule;
     }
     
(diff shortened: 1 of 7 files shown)


app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -125,28 +141,43 @@ private LocalDateTime getExpiresOn(LocalDate localDate, LocalTime localTime) {
...
             for (String thisEventId : eventIds) {
                 if (context.getEvent(thisEventId) != null) {
-                    eventDateTime = context.getEvent(thisEventId);
-                    break;
+                    eventDateTimeList.add(context.getEvent(thisEventId));
+
+                    if (!getAll) {
+                        // We only wanted one event, and we found it, so break.
+                        break;
+                    }
```
alxdarksage, line 170: A common event list is something like "enrollment,actionFinished" (not the exact syntax). For a one time event, this would schedule an initial one on enrollment, and then another one on the actionFinished event. Will this still work at enrollment. It seems like yes, there's no actionFinished event, but once actionFinished event occurs, it still works? I presume it will keep scheduling an enrollment event, finding it is finished, and not returning it?

alxdarksage (approved):
Approved, but see question. I believe it probably works if all the existing tests pass.

alxdarksage:
Note to self: we'll need to incorporate both the automatic events and the scheduling against multiple event keys to the BSM.

app/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
DwayneJengSage: I believe that ActivitySchedulerTest.willSelectFirstEventIdWithARecord() https://github.com/Sage-Bionetworks/BridgePF/blob/develop/test/org/sagebionetworks/bridge/models/schedules/ActivitySchedulerTest.java#L243

This test tests the scenario you describe. If both survey:event and enrollment are in the events map, it'll schedule the event off survey:event. If there's no survey:event, it'll schedule the event off enrollment. Note that one-time schedules always return only 1 event, because of the trim logic.

Note that there's no functional change for one-time or persistent schedules. This change only affects recurring schedules.

alxdarksage (approved):

PR #1695, opened 2018-04-25
Researcher password api

alxdarksage:
Allows a user to generate a random password for another account, and returns that password. Intended to be used for "lab code" accounts where only an external ID exists for the account. In this situation an app my have a researcher sign in to download an initial password to the phone, before giving the device to the participant. Then the password or reauth token is used locally to keep the device signed in.


app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -298,6 +310,56 @@ public void resetPassword(PasswordReset passwordReset) throws BridgeServiceExcep
...
+        }
+        if (StringUtils.isBlank(externalId)) {
+            throw new BadRequestException("External ID is required");
+        }
+        ExternalIdentifier externalIdentifier = externalIdService.getExternalId(
+                study.getStudyIdentifier(), externalId);        
+        if (externalIdentifier == null) {
+            throw new EntityNotFoundException(ExternalIdentifier.class);
+        }
+        String password = generatePassword(study.getPasswordPolicy().getMinLength());
```
DwayneJengSage: Minor: Maybe move the generatePassword() call to after the check for null account or createAccount

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -298,6 +310,56 @@ public void resetPassword(PasswordReset passwordReset) throws BridgeServiceExcep
...
+    };
+    
+    public String generatePassword(int policyLength) {
+        // as long as required, but not less than 32
+        int length = Math.max(32, policyLength); 
+        
+        // We add one of each type of character class to the end of the string. This is 
+        // sufficient to pass password validation. 
+        SecureTokenGenerator generator = new SecureTokenGenerator(length);
+        return generator.nextToken() + "Br4%";
```
DwayneJengSage: Maybe to increase entropy, we can choose a random capital letter, insert it into a random index in the string, repeat for lowercase, number, and symbol.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage: Moved

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage: Created PasswordGenerator class that ensures a password is generated that is random, but also has all four required character types in random positions in the string.

app/org/sagebionetworks/bridge/PasswordGenerator.java
```diff
@@ -0,0 +1,44 @@
...
+        replace(buffer, NUMERIC_ARRAY);
+        replace(buffer, SYMBOLIC_ARRAY);
+        return new String(buffer);
+    }
+    
+    /** Select a position at random and put a random character from the array there */
+    private void replace(char[] buffer, char[] charArray) {
+        char charAt = charArray[RANDOM.nextInt(charArray.length)];
+        int pos = RANDOM.nextInt(buffer.length);
+        buffer[pos] = charAt;
```
DwayneJengSage: There is a low probability that we might generate the same position twice, which makes the symbol overwrite potentially the only number (or uppercase or lowercase) in the password. We'd be better off inserting rather than replacing.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/PasswordGenerator.java
alxdarksage: You're right. Changed this to generate four unique indices before replacing.

DwayneJengSage (approved):

PR #1697, opened 2018-04-27
Adding templates to send signed consents via email or SMS.

alxdarksage:
A first step toward sending signed consents via SMS, we're adding a template for email as well so we can explain what we're sending, rather than just sending it directly to people (it'll be attached to the email, as it currently is). This gives us a place to explain *what* the consent is, so people are less likely to delete it or report it as spam. We need a template in the SMS case because all we can send is a link to download the consent.

### DIFF
--- a/app/org/sagebionetworks/bridge/validators/StudyValidator.java
+++ b/app/org/sagebionetworks/bridge/validators/StudyValidator.java
@@ -118,6 +118,9 @@ public class StudyValidator implements Validator {
         validateEmailTemplate(errors, study.getVerifyEmailTemplate(), "verifyEmailTemplate", "${url}", "${emailVerificationUrl}");
         validateEmailTemplate(errors, study.getResetPasswordTemplate(), "resetPasswordTemplate", "${url}", "${resetPasswordUrl}");
         // Existing studies don't have the template, we use a default template. Okay to be missing.
+        if (study.getSignedConsentTemplate() != null) {
+            validateEmailTemplate(errors, study.getSignedConsentTemplate(), "signedConsentTemplate");    
+        }
         if (study.getEmailSignInTemplate() != null) {
             validateEmailTemplate(errors, study.getEmailSignInTemplate(), "emailSignInTemplate", "${url}", "${emailSignInUrl}",
                     "${token}");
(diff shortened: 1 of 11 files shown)


test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
```diff
@@ -631,6 +631,16 @@ public void loadingStudyWithoutAccountExistsSmsTemplateAddsADefault() {
...
     }
     
+    @Test
+    public void loadingStudyWithoutSignedConsentSmsTemplateAddsADefault() {
+        Study study = TestUtils.getValidStudy(StudyServiceMockTest.class);
+        study.setSignedConsentSmsTemplate(null);
+        when(studyDao.getStudy("foo")).thenReturn(study);
+        
+        Study retStudy = service.getStudy("foo");
+        assertNotNull(retStudy.getAccountExistsSmsTemplate());
```
DwayneJengSage: Copy-paste error. This should be getSignedConsentSmsTemplate(), not getAccountExistsSmsTemplte(). (Here and in updateStudyWithoutSignedConsentSmsTemplateAddsADefault() and in createStudyWithoutSignedConsentSmsTemplateAddsADefault().

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
```diff
@@ -631,6 +631,16 @@ public void loadingStudyWithoutAccountExistsSmsTemplateAddsADefault() {
         assertNotNull(retStudy.getAccountExistsSmsTemplate());
     }
     
+    @Test
+    public void loadingStudyWithoutSignedConsentSmsTemplateAddsADefault() {
```
DwayneJengSage, line 668: Should we also have the same tests for the email template?

app/org/sagebionetworks/bridge/validators/StudyValidator.java
```diff
@@ -283,7 +287,7 @@ private void validateEmailTemplate(Errors errors, EmailTemplate template, String
             }
             if (StringUtils.isBlank(template.getBody())) {
                 errors.rejectValue("body", "cannot be blank");
-            } else {
+            } else if (templateVariables.length > 0){
```
DwayneJengSage: Do we have any test cases where there are no template vars? Should we add unit tests for this case? (Both email and SMS.)

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
alxdarksage: Ack! fixed these

test/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
alxdarksage: Verified we have create/load/update tests for a null email template

DwayneJengSage (approved):

PR #1700, opened 2018-05-02
An URL shortener service.

alxdarksage:
Pretty simple, the service will generate an URL you can substitute for your original (probably very long) URL. The URL provided points back to BridgePF to resolve and redirect to the original URL. Service will not issue conflicting short URLs, and it will not generate a different URL for the same long URL if it hasn't expired. URLs do expire, and are not intended to be permanent (e.g. like bit.ly). The use case right now is to shorten Amazon presigned URLs, and these expire.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/UrlShortenerService.java
+++ b/app/org/sagebionetworks/bridge/services/UrlShortenerService.java
@@ -1,0 +1,53 @@
+package org.sagebionetworks.bridge.services;
+
+import org.sagebionetworks.bridge.SecureTokenGenerator;
+import org.sagebionetworks.bridge.cache.CacheKey;
+import org.sagebionetworks.bridge.cache.CacheProvider;
+import org.sagebionetworks.bridge.config.BridgeConfigFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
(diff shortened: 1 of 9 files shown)


app/org/sagebionetworks/bridge/services/UrlShortenerService.java
```diff
@@ -0,0 +1,53 @@
...
+        // used by another URL. If it's being used by the same URL, don't reset the 
+        // expiration
+        String token = getToken();
+        CacheKey cacheKey = CacheKey.shortenUrl(token);
+        String foundValue = cacheProvider.getObject(cacheKey, String.class);
+        while(foundValue != null && !foundValue.equals(url)) {
+            token = getToken();
+            cacheKey = CacheKey.shortenUrl(token);
+            foundValue = cacheProvider.getObject(cacheKey, String.class);
+        }
```
DwayneJengSage: Minor: Would a do-while loop be better, so you don't have to repeat the code twice?

DwayneJengSage (changes requested):
One minor comment about while loop vs do-while loop. Otherwise, looks fine.

alxdarksage:
Changed to a do loop.

DwayneJengSage (approved):

PR #1701, opened 2018-05-04
Send signed consent agreements via SMS

alxdarksage:
I substantially cleaned up the ConsentServiceMockTest class. I would kind of like to combine it with ConsentServiceTest and get rid of that class, while I'm in there. So maybe don't review yet.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/ConsentService.java
+++ b/app/org/sagebionetworks/bridge/services/ConsentService.java
@@ -70,7 +70,7 @@ public class ConsentService {
     private ActivityEventService activityEventService;
     private SubpopulationService subpopService;
     private String xmlTemplateWithSignatureBlock;
-    private S3Helper s3ConsentWriter;
+    private S3Helper s3Helper;
     private UrlShortenerService urlShortenerService;
     
     @Value("classpath:study-defaults/consent-page.xhtml")
@@ -101,9 +101,9 @@ public class ConsentService {
     final void setSubpopulationService(SubpopulationService subpopService) {
         this.subpopService = subpopService;
     }
-    @Resource(name = "s3ConsentWriter")
-    final void setS3Helper(S3Helper s3ConsentWriter) {
-        this.s3ConsentWriter = s3ConsentWriter;
+    @Resource(name = "s3Helper")
+    final void setS3Helper(S3Helper s3Helper) {
+        this.s3Helper = s3Helper;
     }
     @Autowired
     final void setUrlShortenerService(UrlShortenerService urlShortenerService) {
@@ -381,8 +381,8 @@ public class ConsentService {
         try {
             String fileName = getSignedConsentUrl();
             DateTime expiresOn = getDownloadExpiration();
-            s3ConsentWriter.writeBytesToS3(USERSIGNED_CONSENTS_BUCKET, fileName, consentPdf.getBytes());
-            URL url = s3ConsentWriter.generatePresignedUrl(USERSIGNED_CONSENTS_BUCKET, fileName, expiresOn, HttpMethod.GET);
+            s3Helper.writeBytesToS3(USERSIGNED_CONSENTS_BUCKET, fileName, consentPdf.getBytes());
+            URL url = s3Helper.generatePresignedUrl(USERSIGNED_CONSENTS_BUCKET, fileName, expiresOn, HttpMethod.GET);
             shortUrl = urlShortenerService.shortenUrl(url.toString(), SIGNED_CONSENT_DOWNLOAD_EXPIRE_IN_SECONDS);
         } catch(IOException e) {
             throw new BridgeServiceException(e);
--- a/app/org/sagebionetworks/bridge/services/ConsentService.java
+++ b/app/org/sagebionetworks/bridge/services/ConsentService.java
@@ -391,7 +391,7 @@ public class ConsentService {
                 .withStudy(study)
                 .withPhone(participant.getPhone())
                 .withExpirationPeriod(EXPIRATION_PERIOD_KEY, SIGNED_CONSENT_DOWNLOAD_EXPIRE_IN_SECONDS)
-                .withPromotionType()
+                .withTransactionType()
                 .withSmsTemplate(study.getSignedConsentSmsTemplate())
                 .withToken(BridgeConstants.CONSENT_URL, shortUrl)
                 .build();
--- a/app/org/sagebionetworks/bridge/services/ConsentService.java
+++ b/app/org/sagebionetworks/bridge/services/ConsentService.java
@@ -3,16 +3,27 @@ package org.sagebionetworks.bridge.services;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static org.sagebionetworks.bridge.BridgeUtils.commaListToOrderedSet;
+import static org.sagebionetworks.bridge.BridgeConstants.EXPIRATION_PERIOD_KEY;
+import static org.sagebionetworks.bridge.BridgeConstants.SIGNED_CONSENT_DOWNLOAD_EXPIRE_IN_SECONDS;
 
 import java.io.IOException;
+import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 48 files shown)


test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
```diff
@@ -88,19 +106,27 @@
...
+    @Mock
+    private S3Helper s3Helper;
+    @Mock
+    private UrlShortenerService urlShortenerService; 
+    @Mock
     private Subpopulation subpopulation;
+    @Mock
+    private StudyConsentView studyConsentView;
+    @Spy
+    private GenericAccount account;
```
DwayneJengSage: We should use Account instead of GenericAccount. (If needed, we can create Account.create() which creates a new GenericAccount.)

Also, why do we need to spy the account? It's just a bunch of getters and setters, so maybe we can call the getter to make sure we have the right value?

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
```diff
@@ -112,92 +138,67 @@ public void before() throws IOException {
...
-                .withSignedOn(SIGNED_ON).build();
         
-        account = spy(new GenericAccount()); // mock(Account.class);
-        ((GenericAccount)account).setId(ID);
+        // Reset these values
+        Study study = TestUtils.getValidStudy(ConsentServiceMockTest.class);
+        STUDY.setConsentNotificationEmail(study.getConsentNotificationEmail());
+        STUDY.setConsentNotificationEmailVerified(study.isConsentNotificationEmailVerified());
+        STUDY.setSignedConsentTemplate(study.getSignedConsentTemplate());
+        STUDY.setSignedConsentSmsTemplate(study.getSignedConsentSmsTemplate());
```
DwayneJengSage: Creating a class-wide constant, only to reset it with every test seems to be redundant. It would be cleaner to simply create a new study instance with each test.

app/org/sagebionetworks/bridge/services/ConsentService.java
```diff
@@ -294,45 +340,78 @@ private void sendWithdrawEmail(Study study, String externalId, Account account,
...
+            URL url = s3Helper.generatePresignedUrl(USERSIGNED_CONSENTS_BUCKET, fileName, expiresOn, HttpMethod.GET);
+            shortUrl = urlShortenerService.shortenUrl(url.toString(), SIGNED_CONSENT_DOWNLOAD_EXPIRE_IN_SECONDS);
+        } catch(IOException e) {
+            throw new BridgeServiceException(e);
         }
+        SmsMessageProvider provider = new SmsMessageProvider.Builder()
+                .withStudy(study)
+                .withPhone(participant.getPhone())
+                .withExpirationPeriod(EXPIRATION_PERIOD_KEY, SIGNED_CONSENT_DOWNLOAD_EXPIRE_IN_SECONDS)
+                .withPromotionType()
```
DwayneJengSage: This should be Transactional, since this is a Governance requirement.

app/org/sagebionetworks/bridge/services/ConsentService.java
```diff
@@ -294,45 +340,78 @@ private void sendWithdrawEmail(Study study, String externalId, Account account,
...
+                .build();
+        notificationsService.sendSmsMessage(provider);
+    }
+    
+    protected String getSignedConsentUrl() {
+        return SecureTokenGenerator.INSTANCE.nextToken() + ".pdf";
+    }
+    
+    protected DateTime getDownloadExpiration() {
+        return DateUtils.getCurrentDateTime().plusSeconds(SIGNED_CONSENT_DOWNLOAD_EXPIRE_IN_SECONDS);
```
DwayneJengSage, line 405: Given weird things like Daylight Savings time, would it be cleaner to have the expiration in days and use plusDays() instead?

conf/bridge-server.conf
```diff
@@ -151,3 +151,8 @@ local.study.whitelist = api
 dev.study.whitelist = api
 uat.study.whitelist = api,ios-sdk-int-tests,asthma,cardiovascular
 prod.study.whitelist = api,asthma,breastcancer,cardiovascular,diabetes,fphs,fphs-lab,ohsu-molemapper,parkinson,parkinson-lux,lilly
+
+local.usersigned.consents.bucket = dummy-value
```
DwayneJengSage: Would it be worthwhile to create a default local bucket in our S3 account, to reduce the amount of required work to bootstrap a new dev environment?

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
alxdarksage: Added necessary methods, removed spy and the two places we were not checking the account with an assertion.

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
alxdarksage: converted to instance variable

app/org/sagebionetworks/bridge/services/ConsentService.java
alxdarksage: Changed to transaction

app/org/sagebionetworks/bridge/services/ConsentService.java
alxdarksage: I don't have any reason not to do this. I don't know how Amazon measures "seven days" as the limit on the expiration value. Changed.

alxdarksage: It's more complicated to use days because the expiration label and Redis both use seconds... I don't think people will notice if we're off by an hour because we're not giving an exact date and time at which the file will expire.

conf/bridge-server.conf
alxdarksage: Added local.usersigned.consents.bucket and reference it in the config

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
```diff
@@ -88,116 +106,99 @@
...
-    private StudyParticipant participant;
-    private ConsentSignature consentSignature;
+
     private Account account;
     
     @Before
     public void before() throws IOException {
         documentString = IOUtils.toString(new FileInputStream("conf/study-defaults/consent-page.xhtml"));
-                
+        account = new GenericAccount();
```
DwayneJengSage: We should create Account.create() instead of leaving a GenericAccount reference here.

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
```diff
@@ -88,116 +106,99 @@
...
         consentService.setConsentTemplate(new ByteArrayResource((documentString).getBytes()));
-
+        
         study = TestUtils.getValidStudy(ConsentServiceMockTest.class);
-        study.setConsentNotificationEmailVerified(true);
-        study.setSignedConsentTemplate(new EmailTemplate("Subject", "Body", MimeType.HTML));
+        study.setConsentNotificationEmail(study.getConsentNotificationEmail());
+        study.setConsentNotificationEmailVerified(study.isConsentNotificationEmailVerified());
+        study.setSignedConsentTemplate(study.getSignedConsentTemplate());
+        study.setSignedConsentSmsTemplate(study.getSignedConsentSmsTemplate());
```
DwayneJengSage: Since we're creating a new study instance every test, these setter calls are redundant now.

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
alxdarksage: I've done this in the other PR that's currently pending, so I will resolve this on merge. I could change it here too but it'll lead to more merge conflicts.

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
alxdarksage: Yes. :-1 removed.

DwayneJengSage (approved):

PR #1705, opened 2018-05-07
Consent fixes

alxdarksage:
BRIDGE-2182

1) always indicate that the date is in GMT, because it always is (we most likely don't know the user's time zone at this point);

2) If the account has a verified email, use it, else a verified phone, use it, else use an external ID if it exists, otherwise, just don't show an identifier on the final consent document

3) This uncovered a bug related to creating a consent record with a credential before it is verified; I've moved ITP processing from sign up to sign in to avoid this. There's more places it needs to be inserted, but works the same way... before throwing a consent exception, we check to see if there's a pending consent and process that first, then continue.

Also filed this issue based on functionality I uncovered working on this: https://sagebionetworks.jira.com/browse/BRIDGE-2191

### DIFF
--- a/app/org/sagebionetworks/bridge/services/AuthenticationService.java
+++ b/app/org/sagebionetworks/bridge/services/AuthenticationService.java
@@ -171,6 +171,10 @@ public class AuthenticationService {
         
         Account account = accountDao.authenticate(study, signIn);
 
+        if (intentService.registerIntentToParticipate(study, account)) {
+            account = accountDao.getAccount(AccountId.forId(study.getIdentifier(), account.getId()));    
+        }
+
         UserSession session = getSessionFromAccount(study, context, account);
         // Do not call sessionUpdateService as we assume system is in sync with the session on sign in
         cacheProvider.setUserSession(session);
@@ -235,7 +239,7 @@ public class AuthenticationService {
         } 
     }
 
-    public IdentifierHolder signUp(Study study, StudyParticipant participant, boolean checkForConsent) {
+    public IdentifierHolder signUp(Study study, StudyParticipant participant) {
         checkNotNull(study);
         checkNotNull(participant);
         
@@ -248,14 +252,7 @@ public class AuthenticationService {
 
         try {
             // Since caller has no roles, no roles can be assigned on sign up.
-            IdentifierHolder holder = participantService.createParticipant(study, NO_CALLER_ROLES, participant, true);
-            if (checkForConsent) {
-                // Check to see if this user has saved consent records for the study. Consent them now, so sign in  
-                // will not return 412 (consent required). Need to retrieve the full participant record (w/ healthCode).
-                StudyParticipant updatedParticipant = participantService.getParticipant(study, holder.getIdentifier(), false);
-                intentService.registerIntentToParticipate(study, updatedParticipant);
-            }
-            return holder;
+            return participantService.createParticipant(study, NO_CALLER_ROLES, participant, true);
         } catch(EntityAlreadyExistsException e) {
             // Suppress this and send an email to notify the user that the account already exists. From 
             // this call, we simply return a 200 the same as any other sign up. Otherwise the response 
@@ -372,6 +369,10 @@ public class AuthenticationService {
         accountDao.verifyChannel(channelType, account);
 
         Study study = studyService.getStudy(signIn.getStudyId());
+        
+        if (intentService.registerIntentToParticipate(study, account)) {
+            account = accountDao.getAccount(AccountId.forId(study.getIdentifier(), account.getId()));    
+        }
         UserSession session = getSessionFromAccount(study, context, account);
         cacheProvider.setUserSession(session);
         
--- a/app/org/sagebionetworks/bridge/services/AuthenticationService.java
+++ b/app/org/sagebionetworks/bridge/services/AuthenticationService.java
@@ -173,6 +173,12 @@ public class AuthenticationService {
 
         UserSession session = getSessionFromAccount(study, context, account);
         // Do not call sessionUpdateService as we assume system is in sync with the session on sign in
+        
+        if (!session.doesConsent() && intentService.registerIntentToParticipate(study, account)) {
+            AccountId accountId = AccountId.forId(study.getIdentifier(), account.getId());
+            account = accountDao.getAccount(accountId);
+            session = getSessionFromAccount(study, context, account);        
+        }
         cacheProvider.setUserSession(session);
         
         if (!session.doesConsent() && !session.isInRole(Roles.ADMINISTRATIVE_ROLES)) {
@@ -265,19 +271,7 @@ public class AuthenticationService {
 
         Validate.entityThrowingException(VerificationValidator.INSTANCE, verification);
         Account account = accountWorkflowService.verifyChannel(type, verification);
-        updateChannelAfterVerification(type, account);
-    }
-    
-    // Channel is verified to exist, either by token or by successful sign in, so adjust the 
-    // account and process any pending ITP consents.
-    protected Account updateChannelAfterVerification(ChannelType type, Account account) {
-        Study study = studyService.getStudy(account.getStudyIdentifier());
-        
         accountDao.verifyChannel(type, account);
-        if (intentService.registerIntentToParticipate(study, account)) {
-            account = accountDao.getAccount(AccountId.forId(study.getIdentifier(), account.getId()));    
-        }
-        return account;
     }
     
     public void resendVerification(ChannelType type, AccountId accountId) {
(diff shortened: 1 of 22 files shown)


app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -171,6 +171,10 @@ public UserSession signIn(Study study, CriteriaContext context, SignIn signIn) t
         
         Account account = accountDao.authenticate(study, signIn);
 
+        if (intentService.registerIntentToParticipate(study, account)) {
+            account = accountDao.getAccount(AccountId.forId(study.getIdentifier(), account.getId()));    
+        }
```
DwayneJengSage: It looks like this is non-trivial and it will be executed on every sign-in, which could be a lot. Would it be better to move this to verify channel, which is most likely only executed once?

(Similarly with channelSignIn())

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -151,7 +153,7 @@
 
     private Study study;
 
-    private Account account;
+    private GenericAccount account;
```
DwayneJengSage: We should avoid direct references to GenericAccount, since this bypasses our abstraction, and makes it more difficult to merge HibernateAccount and GenericAccount in the future (if that's the path we're planning on going). If there's something we need in GenericAccount outside of the AccountDao, we should probably move it to the interface.

test/org/sagebionetworks/bridge/services/ConsentPdfTest.java
```diff
@@ -125,13 +123,60 @@ public void legacyDocWithInvalidSig() throws Exception {
...
+        ConsentSignature sig = makeSignatureWithoutImage();
+        
+        StudyParticipant noChannelParticipant = new StudyParticipant.Builder().build();
+        
+        ConsentPdf consentPdf = new ConsentPdf(study, noChannelParticipant, sig, SharingScope.NO_SHARING,
+                NEW_DOCUMENT_FRAGMENT, consentBodyTemplate);
+        consentPdf.getFormattedConsentDocument();
+    }
+    
+    public void dateFormattedCorrectly() throws Exception {
```
DwayneJengSage: Missing Test annotation?

test/org/sagebionetworks/bridge/services/ConsentPdfTest.java
```diff
@@ -125,13 +123,60 @@ public void legacyDocWithInvalidSig() throws Exception {
...
+        ConsentSignature sig = makeSignatureWithoutImage();
+        
+        ConsentPdf consentPdf = new ConsentPdf(study, EMAIL_PARTICIPANT, sig, SharingScope.NO_SHARING,
+                NEW_DOCUMENT_FRAGMENT, consentBodyTemplate);
+        String output = consentPdf.getFormattedConsentDocument();
+        
+        assertTrue("Contains formatted date", output.contains(ConsentPdf.FORMATTER.print(DateTime.now()) + " (GMT)"));
```
DwayneJengSage: It would be better to use DateUtils to mock now and verify the output against a known string. Otherwise, if ConsentPdf.FORMATTER formats the date incorrectly, this test will never catch it.

app/org/sagebionetworks/bridge/services/IntentService.java
```diff
@@ -133,6 +140,27 @@ public void registerIntentToParticipate(Study study, StudyParticipant participan
         }
     }
     
+    public boolean registerIntentToParticipate(Study study, Account account) {
```
DwayneJengSage: The above registerIntentToParticipate() appears to no longer be used, and this appears to be nearly identical. Instead of creating an overload, would it be better to instead change the old one?

app/org/sagebionetworks/bridge/services/IntentService.java
```diff
@@ -133,6 +140,27 @@ public void registerIntentToParticipate(Study study, StudyParticipant participan
...
+        List<Subpopulation> subpops = subpopService.getSubpopulations(study.getStudyIdentifier());
+        for (Subpopulation subpop : subpops) {
+            CacheKey cacheKey = CacheKey.itp(subpop.getGuid(), study.getStudyIdentifier(), phone);
+            IntentToParticipate intent = cacheProvider.getObject(cacheKey, IntentToParticipate.class);
+            if (intent != null) {
+                StudyParticipant participant = participantService.getParticipant(study, account.getId(), true);
+                consentService.consentToResearch(study, subpop.getGuid(), participant, 
+                        intent.getConsentSignature(), intent.getScope(), true);
+                cacheProvider.removeObject(cacheKey);
+                return true;
```
DwayneJengSage: Could an ITP represent multiple consents? If so, instead of returning true and short-circuiting, would it be better to save the result somewhere and return it later?

test/org/sagebionetworks/bridge/services/AuthenticationServiceTest.java
```diff
@@ -161,7 +161,7 @@ public void signUpWithIntentToParticipate() throws Exception {
             
             study.setAutoVerificationEmailSuppressed(true);
             study.setAutoVerificationPhoneSuppressed(true);
-            holder = authService.signUp(study, participant, true);
+            holder = authService.signUp(study, participant);
```
DwayneJengSage, line 164: Does this test still succeed? If so, how?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage: Hmm I hadn't thought of that. Yes that's a good idea. I'll look at moving it to the verification step. We just need to be sure we verify through that method if people sign in through email or phone, without explicitly verifying.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: Added setId and setStudyId (which does not match getStudyIdentifier, but I'm not going to change that in this PR).

test/org/sagebionetworks/bridge/services/ConsentPdfTest.java
alxdarksage: Added

test/org/sagebionetworks/bridge/services/ConsentPdfTest.java
alxdarksage: Changed to use a fixed point in time

app/org/sagebionetworks/bridge/services/IntentService.java
alxdarksage: This was a mistake on my part, this is mid-refactor. The participant method can go away, removed and altered tests that were testing it to test the other method

app/org/sagebionetworks/bridge/services/IntentService.java
alxdarksage: Theoretically it could. Changing this, also keeping track as to whether or not we've loaded the study participant.

test/org/sagebionetworks/bridge/services/AuthenticationServiceTest.java
alxdarksage: Yes because it's really looking to see what happens when you sign in, and we've moved ITP consent processing from sign up to sign in (which has no flag). However, I just moved it to the verification step, so this test has to be changed again. However it's still looking to verify that sign in after this does not throw an exception.

alxdarksage:
Updated. Moved ITP processing to consent rather than sign up, after authentication and only if the user is not consented (and would receive a 412 response). The associated integration test passes, but only by enabling the account without enabling the phone number, then signing in with a password. This is how our system works though I'm not sure this is really what we want. If we change this, we'll need some way for a caller to verify phone/email for testing... maybe an administrative UI to manipulate these for testing.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -735,8 +738,107 @@ public void creatingExternalIdOnlyAccountSucceedsIfIdsManaged() {
...
+        service.signUp(study, participant);
         
         verify(participantService).createParticipant(study, NO_CALLER_ROLES, participant, true);
     }
-}
+    
+    @Test
+    public void signInWithIntentToParticipate() throws Exception {
+        account.setId(USER_ID);
+        Account consentedAccount = new GenericAccount();
```
DwayneJengSage: We should create a Account.create() so we don't have references to GenericAccount everywhere. (Here and several other places in this PR.)

app/org/sagebionetworks/bridge/services/IntentService.java
```diff
@@ -114,23 +121,29 @@ public void submitIntentToParticipate(IntentToParticipate intent) {
...
+        boolean consentsUpdated = false;
+        StudyParticipant participant = null;
         List<Subpopulation> subpops = subpopService.getSubpopulations(study.getStudyIdentifier());
         for (Subpopulation subpop : subpops) {
             CacheKey cacheKey = CacheKey.itp(subpop.getGuid(), study.getStudyIdentifier(), phone);
             IntentToParticipate intent = cacheProvider.getObject(cacheKey, IntentToParticipate.class);
             if (intent != null) {
+                if (participant == null) {
+                    participant = participantService.getParticipant(study, account.getId(), true);
+                }
```
DwayneJengSage, line 139: Is getParticipant() a particularly expensive call? If it's not, it might be simpler to call getParticipant() before the loop, instead of having these extra if statements.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/IntentService.java
alxdarksage: Yes, it's expensive here because we're loading all the consent records and we do use them in the consentToResearch call. I would be careful not to load it unless it's really necessary. Admittedly, the chance of their being two ITP consents (or more) is low and this call is a one-time deal, so we could remove this check.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: Done.

DwayneJengSage (approved):

PR #1711, opened 2018-05-14
Do not allow blank email addresses to slip through validation.

alxdarksage:
I found while testing through the BSM that I could create a user with a blank email address and a phone number, that triggered a 500 error (we try and send an email to the blank email address).

Fixed with stronger validation.

### DIFF
--- a/app/org/sagebionetworks/bridge/validators/IdentifierUpdateValidator.java
+++ b/app/org/sagebionetworks/bridge/validators/IdentifierUpdateValidator.java
@@ -1,5 +1,6 @@
 package org.sagebionetworks.bridge.validators;
 
+import org.apache.commons.validator.routines.EmailValidator;
 import org.sagebionetworks.bridge.models.accounts.ExternalIdentifier;
 import org.sagebionetworks.bridge.models.accounts.IdentifierUpdate;
 import org.sagebionetworks.bridge.models.accounts.Phone;
@@ -11,6 +12,7 @@ import org.springframework.validation.Validator;
 
 public class IdentifierUpdateValidator implements Validator {
 
+    private static final EmailValidator EMAIL_VALIDATOR = EmailValidator.getInstance();
     private Study study;
     private ExternalIdService externalIdService;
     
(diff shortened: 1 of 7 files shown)


app/org/sagebionetworks/bridge/validators/IdentifierUpdateValidator.java
```diff
@@ -44,9 +46,15 @@ public void validate(Object object, Errors errors) {
         int updateFields = 0;
         if (update.getPhoneUpdate() != null) {
             updateFields++;
+            if (!Phone.isValid(update.getPhoneUpdate())) {
+                errors.rejectValue("phoneUpdate", "does not appear to be a phone number");
+            }
```
DwayneJengSage, line 51: Should we add tests to IdentifierUpdateValidatorTest to make sure we catch invalid emails and phone numbers?

app/org/sagebionetworks/bridge/validators/StudyParticipantValidator.java
```diff
@@ -51,7 +50,7 @@ public void validate(Object object, Errors errors) {
                 errors.rejectValue("phone", "does not appear to be a phone number");
             }
             // If provided, email must be valid
-            if (isNotBlank(email) && !EMAIL_VALIDATOR.isValid(email)) {
+            if (email != null && !EMAIL_VALIDATOR.isValid(email)) {
                 errors.rejectValue("email", "does not appear to be an email address");
             }
```
DwayneJengSage, line 55: If we're treating blank emails as invalid emails rather than as no email specified, we should probably make the change further up in the "email, phone, or externalId is required" section. Either null out blank emails, or check email != null instead if isBlank(email).

test/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
```diff
@@ -126,6 +126,18 @@ public void emailPhoneOrExternalIdRequired() {
         assertValidatorMessage(validator, withEmail(null), "StudyParticipant", "email, phone, or externalId is required");
     }
     
+    @Test
+    public void emailCannotBeBlank() {
+        validator = new StudyParticipantValidator(externalIdService, study, true);
+        assertValidatorMessage(validator, withEmail(""), "StudyParticipant", "email, phone, or externalId is required");
+    }
```
DwayneJengSage: Also, to ensure EMAIL_VALIDATOR catches it, we should test both empty strings and strings of all whitespace.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/validators/IdentifierUpdateValidator.java
alxdarksage: Added test of email validation (neither empty nor junk value pass). Invalid phone test exists.

app/org/sagebionetworks/bridge/validators/StudyParticipantValidator.java
alxdarksage: Makes sense, changed this to null pointer check only. Blanks will receive a different message (invalid email) instead.

test/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
alxdarksage: Added

DwayneJengSage (approved):

PR #1714, opened 2018-05-15
Optional birthdate

alxdarksage:
Only verify the birthdate exists, and is old enough to be valid, if a minimum age is set for the study (0 or -1 = no min age set). Consolidated the ConsentAgeValidator and and ConsentValidator as both have to be created per study now.


app/org/sagebionetworks/bridge/validators/ConsentSignatureValidator.java
```diff
@@ -22,11 +30,28 @@ public boolean supports(Class<?> clazz) {
...
+
+                    if (period.getYears() < minAgeOfConsent) {
+                        String message = String.format(TOO_YOUNG, minAgeOfConsent);
+                        errors.rejectValue("birthdate", message);
+                    }
+                }
+            }
+        } else if (sig.getBirthdate() != null) {
+            // Just verify it parses because it wasn't required, but they supplied it anyway
+            parseBirthday(errors, sig.getBirthdate());
```
DwayneJengSage: Instead of parsing birthdate in 2 different branches, considering parsing the birthdate once, before the minAge check, and saving the result so you don't need to parse again.

test/org/sagebionetworks/bridge/validators/ConsentSignatureValidatorTest.java
```diff
@@ -0,0 +1,240 @@
...
+    @Test
+    public void minAgeLimitButNoBirthdate() {
+        validator = new ConsentSignatureValidator(18);
+        ConsentSignature sig = new ConsentSignature.Builder().withName("test name").build();
+        assertValidatorMessage(validator, sig, "birthdate", "cannot be missing, null, or blank");
+    }
+    
+    @Test
+    public void minAgeLimitButBirthdateTooRecent() {
+        String birthdate = new DateTime(SIGNED_ON_TIMESTAMP).minusYears(18).plusDays(1).toLocalDate().toString();
```
DwayneJengSage: This test will fail on Feb 28 2022. See https://sagebionetworks.jira.com/browse/BRIDGE-2136 (And possibly the opposite test will fail on Feb 29 2020. The tests in ConsentServiceTest currently using plusDays(2) for this exact reason.

I dunno if this code will still exist in 2-4 years time, but I think we should future-proof the test if not the code.

DwayneJengSage (changes requested):
A few minor notes.

app/org/sagebionetworks/bridge/validators/ConsentSignatureValidator.java
alxdarksage: Rearranged

test/org/sagebionetworks/bridge/validators/ConsentSignatureValidatorTest.java
alxdarksage: Removed the fixed time of test, then it's easy to calculate from current time (forever).

app/org/sagebionetworks/bridge/validators/ConsentSignatureValidator.java
```diff
@@ -22,11 +30,29 @@ public boolean supports(Class<?> clazz) {
...
         if (Strings.isNullOrEmpty(sig.getName())) {
             errors.rejectValue("name", CANNOT_BE_BLANK);
         }
-        if (Strings.isNullOrEmpty(sig.getBirthdate())) {
-            errors.rejectValue("birthdate", CANNOT_BE_BLANK);
+        LocalDate birthdate = parseBirthday(sig.getBirthdate());
+        // There was a value, but it didn't parse
+        if (birthdate == null && !Strings.isNullOrEmpty(sig.getBirthdate())) {
+            errors.rejectValue("birthdate", "is invalid (required format: YYYY-MM-DD)");
+        }
```
DwayneJengSage: Would it be cleaner to check that sig.getBirthdate() is specified and *then* try to parse it instead of the other way around?

test/org/sagebionetworks/bridge/validators/ConsentSignatureValidatorTest.java
```diff
@@ -0,0 +1,232 @@
...
+    public void minAgeLimitButNoBirthdate() {
+        validator = new ConsentSignatureValidator(18);
+        ConsentSignature sig = new ConsentSignature.Builder().withName("test name").build();
+        assertValidatorMessage(validator, sig, "birthdate", "cannot be missing, null, or blank");
+    }
+    
+    @Test
+    public void minAgeLimitButBirthdateTooRecent() {
+        String birthdate = DateTime.now().minusYears(18).plusDays(1).toLocalDate().toString();
+        validator = new ConsentSignatureValidator(18);
```
DwayneJengSage: Actually, I think we want this the other way around. On Feb 28 2022, this *definitely* will fail, because Joda thinks there are exactly 18 years between Feb 29 2004 and Feb 28 2022. I think using fixed timestamps will make the test more reliable.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/validators/ConsentSignatureValidator.java
alxdarksage: Changed

test/org/sagebionetworks/bridge/validators/ConsentSignatureValidatorTest.java
alxdarksage: I fixed the time to 2/28/2022, this does not seem to make any difference. If you do not remove a day, it succeeds, otherwise it fails as intended. It seems smart enough to deal with leap years.

test/org/sagebionetworks/bridge/validators/ConsentSignatureValidatorTest.java
```diff
@@ -0,0 +1,240 @@
...
```
DwayneJengSage: Are we sure we have the timezone correct? The fixed time is 2022-02-28T00:00Z, which is 2022-02-27 in the local time zone, so we never have to worry about daylight savings time. However, this test will fail anywhere east of UTC.

I tested this with the old ConsentServiceTest (by fixing the date to 2022-02-28T00:00-0800, and changing it to plus/minus 1 day), and the test fails. I'm pretty sure this test will fail as well.

test/org/sagebionetworks/bridge/validators/ConsentSignatureValidatorTest.java
```diff
@@ -0,0 +1,240 @@
...
+        ConsentSignature sig = new ConsentSignature.Builder().withName("test name").withBirthdate("15 May 2018").build();
+        assertValidatorMessage(validator, sig, "birthdate", "is invalid (required format: YYYY-MM-DD)");
+    }
+    
+    @Test
+    public void optionalBirthdateGarbled() {
+        ConsentSignature sig = new ConsentSignature.Builder().withName("test name").withBirthdate("15 May 2018").build();
+        assertValidatorMessage(validator, sig, "birthdate", "is invalid (required format: YYYY-MM-DD)");
+    }
```
DwayneJengSage, line 247: Should we add a test case for optional birthday not specified is valid?

DwayneJengSage (changes requested):

DwayneJengSage:
I pulled down your branch and doing some debugging. Here are my findings.

As written, this passes locally, because locally, we run in PDT, so "today" is still 2022-02-27, and birthdate is 2004-02-28, which is definitely fine. But on the build machine (or in any machine that is east of UTC), "today" will be 2022-02-28 and birthdate will be 2004-02-29, which according to Joda, is 18 years apart.

If you set UTC timezone in both the test and in ConsentSignatureValidator, it will fail locally, for the same reason as above.

If you set UTC timezone on just the test but not ConsentSignatureValidator (which sounds like what you did in changes not yet pushed), minAgeLimitBirthdateOK() fails locally because birthdate is correctly set to 2004-02-28, but "today" is 2022-02-27 locally.

test/org/sagebionetworks/bridge/validators/ConsentSignatureValidatorTest.java
alxdarksage: Added

DwayneJengSage (approved):

PR #1715, opened 2018-05-18
Fix a 500 error

alxdarksage:
Fix a 500 error that occurs when you remove the formerly required consent email, and you send an empty string. We really want that to fail validation so it doesn't become bad data in the system (later we'll try and use it to send email... this fails).

Also return a 400 instead of a 500 when you manually trigger a consent email verification through the API, and that field is null.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/StudyService.java
+++ b/app/org/sagebionetworks/bridge/services/StudyService.java
@@ -610,7 +610,7 @@ public class StudyService {
         if (!originalStudy.getSupportEmail().equals(study.getSupportEmail())) {
             emailVerificationService.verifyEmailAddress(study.getSupportEmail());
         }
-        if (consentHasChanged && StringUtils.isNotBlank(study.getConsentNotificationEmail())) {
+        if (consentHasChanged && study.getConsentNotificationEmail() != null) {
             sendVerifyEmail(study, StudyEmailType.CONSENT_NOTIFICATION);    
         }
         return updatedStudy;
--- a/app/org/sagebionetworks/bridge/services/StudyService.java
+++ b/app/org/sagebionetworks/bridge/services/StudyService.java
@@ -610,7 +610,7 @@ public class StudyService {
         if (!originalStudy.getSupportEmail().equals(study.getSupportEmail())) {
             emailVerificationService.verifyEmailAddress(study.getSupportEmail());
         }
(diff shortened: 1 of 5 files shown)


app/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -447,8 +447,9 @@ public Study createStudy(Study study) {
         
         emailVerificationService.verifyEmailAddress(study.getSupportEmail());
 
-        sendVerifyEmail(study, StudyEmailType.CONSENT_NOTIFICATION);
-
+        if (study.getConsentNotificationEmail() != null) {
+            sendVerifyEmail(study, StudyEmailType.CONSENT_NOTIFICATION);    
+        }
```
DwayneJengSage, line 452: Can we make a unit test that creates a study without a consent notification email, so that we can test this branch?

test/org/sagebionetworks/bridge/validators/StudyValidatorTest.java
```diff
@@ -160,15 +160,57 @@ public void acceptsMultipleValidTechnicalEmailAddresses() {
     @Test
     public void rejectsInvalidTechnicalEmailAddresses() {
         study.setTechnicalEmail("test@test.com,asdf,test2@test.com");
-        assertValidatorMessage(INSTANCE, study, "technicalEmail", "'asdf' is not a valid email address");
+        assertValidatorMessage(INSTANCE, study, "technicalEmail", StudyValidator.EMAIL_ERROR);
     }
     
     @Test
-    public void requiresMissingTechnicalEmail() {
```
DwayneJengSage, line 167: Instead of deleting requiresMissingTechnicalEmail(), should we replace it with a test that a study without a technical email is still valid?

DwayneJengSage (changes requested):
Looks good. Two notes about tests.

test/org/sagebionetworks/bridge/validators/StudyValidatorTest.java
alxdarksage: Added a separate test that technical email is optional.

app/org/sagebionetworks/bridge/services/StudyService.java
alxdarksage: Added to the mock test.

DwayneJengSage (approved):

PR #1719, opened 2018-05-23
IP-locking sessions

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2120

Privileged accounts (accounts with roles) are always IP-locked. Non-privileged accounts can be IP-locked based on a study setting (which defaults to false). Note that as a side effect, every API will fetch the study now, but this should be fine since study is already well cached.

Also note that IP Address can be spoofed. That said, it's still better to lock on IP address than to have no defense at all.

Testing done: added unit tests and integration tests


app/org/sagebionetworks/bridge/play/controllers/BaseController.java
```diff
@@ -164,16 +167,27 @@ UserSession getAuthenticatedSession(boolean consentRequired, Roles...roles) {
...
+        Study study = studyService.getStudy(session.getStudyIdentifier());
+        Set<Roles> userRoles = session.getParticipant().getRoles();
+        boolean userHasRoles = !userRoles.isEmpty();
+        if (study.isParticipantIpLockingEnabled() || userHasRoles) {
+            String sessionIpAddress = parseIpAddress(session.getIpAddress());
+            String requestIpAddress = parseIpAddress(getRemoteAddress());
+            if (!Objects.equals(sessionIpAddress, requestIpAddress)) {
+                throw new NotAuthenticatedException();
+            }
+        }
```
alxdarksage, line 182: What if we tested, and only recorded the user's IP address if we needed to check it (enabled for entire study or has roles)? Then if the session had an IP address, we'd check against the request (or otherwise skip it). Then the study loading would only occur once and we probably already have it.

test/org/sagebionetworks/bridge/play/controllers/BaseControllerTest.java
```diff
@@ -60,10 +62,13 @@
 @SuppressWarnings("unchecked")
 public class BaseControllerTest {
     private static final String DOMAIN = "ws-test.sagebridge.org";
+    private static final String HEALTH_CODE = "health-code";
+    private static final String IP_ADDRESS = "dummy IP address";
```
alxdarksage, line 66: Should this and the different IP address be in IP address format just to make sure periods and whatnot don't throw off your string parsing?

alxdarksage (approved):
Do what you want with the comments

alxdarksage:
Is it useful to put the IP address we extract into the Metrics object, so it gets logged?

DwayneJengSage:
Discussed the comments on Slack:

* getAuthenticatedSession() already loads the study in the consentRequired=true case which is a majority of calls, so we've agreed it's not worth the extra optimization.
* In prod, it's always two IP addresses separated by a comma and space, but in local, it's "0:0:0:0:0:0:0:1", and in the general case, it could be any arbitrary string. The test looks the same either way, and we don't do any IP address parsing, so we're okay with leaving the test as is.
* MetricsInterceptor already records the IP address.

PR #1731, opened 2018-06-06
Remove the cookie.

alxdarksage:
No longer send the session in a cookie. This further protects us from the security risks described in BRIDGE-2181.


DwayneJengSage:
Can we keep the cookie in local? This makes testing much easier.

DwayneJengSage (approved):
Otherwise, this looks fine.

alxdarksage:
I'm going to hold onto this because we're getting busy with mPower 2.0, but I may close out and redo this just suppressing the behavior when not in a local environment. I had forgotten we were using this for testing.

PR #1734, opened 2018-06-11
improved logging for emails

DwayneJengSage:
Contains the following changes:
* Adds Request ID to ThreadLocal storage. This allows us to tie certain log messages to MetricsInterceptor logs. Currently used by the SendMailService to track which requests send which emails.
* Adds the EmailType enum and plumbs it through all relevant classes. Used so we can easily determine the type of each email sent.
* Reduce email throttling from 2 per 5 min to 1 per 5 min. This reduces the number of duplicate emails and reduces the chance of users marking multiple emails as spam.

Testing done:
* Updated unit tests.
* Ran integ tests (no changes, mainly as a sanity test).
* Manually tested each email type and verified logs.

### DIFF
--- a/app/org/sagebionetworks/bridge/BridgeUtils.java
+++ b/app/org/sagebionetworks/bridge/BridgeUtils.java
@@ -60,13 +60,28 @@ public class BridgeUtils {
     private static final int ONE_MINUTE = 60;
     
     private static final SecureRandom SECURE_RANDOM = new SecureRandom();
-    
+
+    // ThreadLocals are weird. They are basically a container that allows us to hold "global variables" for each
+    // thread. This can be used, for example, to provide the request ID to any class without having to plumb a
+    // "request context" object into every method of every class.
+    private static final ThreadLocal<String> REQUEST_ID_THREAD_LOCAL = ThreadLocal.withInitial(() -> null);
+
     public static boolean isExternalIdAccount(StudyParticipant participant) {
         return (StringUtils.isNotBlank(participant.getExternalId()) && 
                 StringUtils.isBlank(participant.getEmail()) && 
                 participant.getPhone() == null);
     }
-    
+
+    /** Gets the request ID for the current thread. See also RequestIdInterceptor. */
+    public static String getRequestId() {
+        return REQUEST_ID_THREAD_LOCAL.get();
+    }
+
+    /** @see #getRequestId */
+    public static void setRequestId(String requestId) {
+        REQUEST_ID_THREAD_LOCAL.set(requestId);
+    }
+
     /**
      * Convert expiration measures in seconds to an English language explanation of
      * the expiration time. This is not intended to cover odd cases--our expirations 
--- a/app/org/sagebionetworks/bridge/services/email/MimeTypeEmail.java
+++ b/app/org/sagebionetworks/bridge/services/email/MimeTypeEmail.java
@@ -5,10 +5,9 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
 import java.util.List;
-
+import java.util.function.Function;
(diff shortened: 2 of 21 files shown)


app/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -60,13 +60,28 @@
     private static final int ONE_MINUTE = 60;
     
     private static final SecureRandom SECURE_RANDOM = new SecureRandom();
-    
+
+    // ThreadLocals are weird. They are basically a container that allows us to hold "global variables" for each
+    // thread. This can be used, for example, to provide the request ID to any class without having to plumb a
+    // "request context" object into every method of every class.
+    private static final ThreadLocal<String> REQUEST_ID_THREAD_LOCAL = ThreadLocal.withInitial(() -> null);
```
alxdarksage, line 67: Once you get started with this... we could store the caller's session this way, and the study, so we don't have to pass them in every single method everywhere.

alxdarksage: Are we sure that Play delegates one request to one thread? And then when blocked by I/O, Play doesn't reallocate the thread to another process? I think some systems achieve scalability by *not* delegating a thread to an entire request, and I thought Play was doing this under the hood. See the (to me incomprehensible) discussion of this here: https://www.playframework.com/documentation/2.6.x/ThreadPools

app/org/sagebionetworks/bridge/services/email/MimeTypeEmail.java
```diff
@@ -17,29 +16,28 @@
...
 public final class MimeTypeEmail {
-    
     private static final String QUOTE = "\"";
     private static final String QUOTED_QUOTE = "\\\\\"";
-    private static final Function<String,String> APPLY_EMAIL_ESCAPER = new Function<String,String>() {
-        @Override public String apply(final String address) {
-            return escapeEmailAddress(address);
-        }
-    };
+    private static final Function<String, String> APPLY_EMAIL_ESCAPER = MimeTypeEmail::escapeEmailAddress;
```
alxdarksage, line 21: Drastically better, thanks

conf/bridge-server.conf
```diff
@@ -1,4 +1,4 @@
-channel.throttle.max.requests = 2
+channel.throttle.max.requests = 1
```
alxdarksage, line 1: We might see bugs about this. It's mysterious that when I ask for it, I don't always get a message and it's not clear then if the old message is still valid or not.

alxdarksage (commented):
I would verify ThreadLocal will work in the Play environment in production, then this PR is approved.

app/org/sagebionetworks/bridge/BridgeUtils.java
DwayneJengSage: I'm about 80% sure that this works. The HTTP context is, itself, stored in a ThreadLocal, and the linked docs explicitly call out that Play doesn't automagically propagate the HTTP context to async threads and that you need to do it yourself. The rest of the docs indicate that there is no Play magic that automatically breaks up your Java execution into multiple threads.

That said, I've seen some evidence in New Relic that Play _does_ have some thread magic, in particular around DynamoDB calls, so I'm 20% unsure.

That said, it's tested and works locally, and the logs suggest that a single thread is used for the entire request.

conf/bridge-server.conf
DwayneJengSage: Email complaints suggest this is happening already. Someone will request sign-in, and then they will click the link several hours later when it already expires, which generates a new sign-in link, and then they mark both emails as spam.

Given that our complaint rate is dangerously high, I think we should err on the side of sending too few emails rather than too many emails.

If emails are in fact not being delivered, we have the logs to verify this.

alxdarksage (approved):

PR #1740, opened 2018-06-25
Add survey identifiers to survey references in an app config

alxdarksage:
When calling through the public-facing app config API, add survey identifiers. Clients use these interchangeably with GUIDs as the primary key for the survey.

See https://sagebionetworks.jira.com/browse/BRIDGE-2229 which specifically references the app configs. There are integration tests to go with this change.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/AppConfigService.java
+++ b/app/org/sagebionetworks/bridge/services/AppConfigService.java
@@ -108,13 +108,16 @@ public class AppConfigService {
         return matched;
     }
 
+    // Survey and schema references in an app config are "hard" references... they must reference a 
+    // version or createdOn timestamp of a version and this has been validated when creating/
+    // updating the reference. We're only concerned with adding survey identifier here.
     SurveyReference resolveSurvey(SurveyReference surveyRef) {
-        GuidCreatedOnVersionHolder surveyKeys = new GuidCreatedOnVersionHolderImpl(surveyRef);
-        // Survey references can be bogus, and should be return unmodified if they don't reference an existing survey
         try {
-            Survey survey = surveyService.getSurvey(surveyKeys);
+            GuidCreatedOnVersionHolder surveyKeys = new GuidCreatedOnVersionHolderImpl(surveyRef);
+            Survey survey = surveyService.getSurvey(surveyKeys, false);
             return new SurveyReference(survey.getIdentifier(), survey.getGuid(), new DateTime(survey.getCreatedOn()));
         } catch(EntityNotFoundException e) {
+            // Survey references can be bogus, and should be return unmodified if they don't reference an existing survey
             return surveyRef;
         }
     }
@@ -128,7 +131,7 @@ public class AppConfigService {
         Study study = studyService.getStudy(studyId);
         Validator validator = new AppConfigValidator(study.getDataGroups(), true);
         Validate.entityThrowingException(validator, appConfig);
-        
+
         long timestamp = getCurrentTimestamp();
 
         DynamoAppConfig newAppConfig = new DynamoAppConfig();
--- a/app/org/sagebionetworks/bridge/services/AppConfigService.java
+++ b/app/org/sagebionetworks/bridge/services/AppConfigService.java
@@ -112,6 +112,9 @@ public class AppConfigService {
     // version or createdOn timestamp of a version and this has been validated when creating/
     // updating the reference. We're only concerned with adding survey identifier here.
     SurveyReference resolveSurvey(SurveyReference surveyRef) {
+        if (surveyRef.getIdentifier() != null) {
+            return surveyRef;
+        }
         try {
             GuidCreatedOnVersionHolder surveyKeys = new GuidCreatedOnVersionHolderImpl(surveyRef);
             Survey survey = surveyService.getSurvey(surveyKeys, false);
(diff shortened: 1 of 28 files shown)


app/org/sagebionetworks/bridge/services/AppConfigService.java
```diff
@@ -87,9 +100,23 @@ public AppConfig getAppConfigForUser(CriteriaContext context, boolean throwExcep
...
+            .map(surveyReference -> resolveSurvey(surveyReference)).collect(Collectors.toList()));
+        return matched;
+    }
+
+    SurveyReference resolveSurvey(SurveyReference surveyRef) {
+        GuidCreatedOnVersionHolder surveyKeys = new GuidCreatedOnVersionHolderImpl(surveyRef);
+        // Survey references can be bogus, and should be return unmodified if they don't reference an existing survey
+        try {
+            Survey survey = surveyService.getSurvey(surveyKeys);
+            return new SurveyReference(survey.getIdentifier(), survey.getGuid(), new DateTime(survey.getCreatedOn()));
```
DwayneJengSage: Three optimizations:

1. Can we create a flag in SurveyService to not load SurveyElements? Otherwise, we end up loading a bunch of survey elements that we just throw away.
2. getSurvey() requires that guid and createdOn exist, so this method is really only used to get the identifier. Can we add a check to skip loading the survey if the identifier is already present?
3. Lastly, a SurveyReference without a createdOn is a perfectly valid SurveyReference, and this will crash this code with an IllegalArgumentExcecption. How do we handle those cases?

test/org/sagebionetworks/bridge/services/AppConfigServiceTest.java
```diff
@@ -135,6 +162,12 @@ public void getAppConfig() {
     
     @Test
     public void getAppConfigForUser() {
+        Survey survey = Survey.create();
+        survey.setIdentifier("theIdentifier");
+        survey.setGuid(SURVEY_REF_LIST.get(0).getGuid());
+        survey.setCreatedOn(SURVEY_REF_LIST.get(0).getCreatedOn().getMillis());
+        when(surveyService.getSurvey(SURVEY_KEY)).thenReturn(survey);
```
DwayneJengSage: Unit test for the case where the survey doesn't exist?

test/org/sagebionetworks/bridge/services/AppConfigServiceTest.java
```diff
@@ -60,9 +82,11 @@
     public void before() {
         service.setAppConfigDao(mockDao);
         service.setStudyService(mockStudyService);
+        service.setSurveyService(surveyService);    
         
         when(service.getCurrentTimestamp()).thenReturn(TIMESTAMP.getMillis());
         when(service.getGUID()).thenReturn(GUID);
+        //doReturn(referenceResolver).when(service).getReferenceResolver(any(), any());
```
DwayneJengSage: Minor: Commented out code.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/AppConfigService.java
alxdarksage: In the AppConfigs, we currently require all the schema and survey references to be "hard" references... that is, they reference an explicit version or createdOn timestamp (soft references are not allowed). I could check for that here in case we change this later, but right now #3 won't be an issue.

I will add the other two optimizations.

alxdarksage: We don't want the app configs to cache a survey identifier value that is later updated in the survey itself, so either we're going to remove the identifiers in the app config before we persist it, which will cause an update of the identifiers on every retrieval, OR we look up the identifiers on every retrieval and ignore any values in the record. Either way, we're looking these up on each request, however, app configs are cached at the controller tier.

DwayneJengSage: Can AppConfig vend non-published surveys? If so, that will cause some weirdness if the user ever attempts to submit a response for a non-published survey (such as the lack of a schema). (And since surveys are so intricately tied to schemas through the survey ID, I think we actually want survey IDs to be fixed for the lifetime of a given survey. See https://sagebionetworks.jira.com/browse/BRIDGE-1833)

Assuming AppConfig can only vend published schemas, the survey identifier is already fixed and cannot be changed.

alxdarksage: There's a lot to this, but I'll check the parts to make sure this is all true. Then as you say, it's save to save the identifier in the app config so we can do that and then skip looking it up again.

alxdarksage:
Okay, I added a check and if the identifier exists for a survey reference, we skip resolution.

Then when we resolve, I've added flags to two of the SurveyDao methods to skip loading survey elements, and used the flags anywhere I could find where we're resolving survey references (or loading a survey to get the guid/createdOn info alone).

You have to reference a specific, published survey in an app config, so not checking for the createdOn timestamp. We validate the createdOn timestamp exists during app config validation. I filed JIRA https://sagebionetworks.jira.com/browse/BRIDGE-2237 because we check that the timestamp exists, but we don't attempt to load the survey to verify that it exists and is published (same thing with the schemas). We should do that.

PR #1742, opened 2018-06-26
Prevent duplicates using detection feature in Jackson

alxdarksage:
I dislike this solution but it's the only one I have found so far.


syoung-smallwisdom (approved):
Looks good. Thanks!

And yes, it's too bad that `JsonGenerationException` doesn't use a code as well as a description string. Oh well.

PR #1744, opened 2018-06-29
More strictly validate that the schema and survey references point to valid entities on create and update.

alxdarksage:


app/org/sagebionetworks/bridge/validators/AppConfigValidator.java
```diff
@@ -36,33 +49,55 @@ public void validate(Object object, Errors errors) {
...
+                
+                for (int i=0; i < appConfig.getSchemaReferences().size(); i++) {
+                    SchemaReference ref = appConfig.getSchemaReferences().get(i);
                     errors.pushNestedPath("schemaReferences["+i+"]");
-                    errors.rejectValue("revision", "is required");
+                    if (ref.getRevision() == null) {
+                        errors.rejectValue("revision", "is required");
+                    } else {
+                        try {
+                            schemaService.getUploadSchema(studyId, ref.getId());
```
DwayneJengSage: The ID might be valid, but the revision invalid. Consider using getUploadSchemaByIdAndRev() instead.

test/org/sagebionetworks/bridge/validators/AppConfigValidatorTest.java
```diff
@@ -55,16 +84,94 @@ public void surveyReferencesHaveCreatedOnTimestamps() {
...
         SchemaReference schemaRef = new SchemaReference("guid",null);
         appConfig.getSchemaReferences().add(schemaRef);
         
-        assertValidatorMessage(NEW_VALIDATOR, appConfig, "schemaReferences[0].revision", "is required");
+        assertValidatorMessage(newValidator, appConfig, "schemaReferences[0].revision", "is required");
+    }
+    
+    @Test
+    public void schemaDoesNotExistOnCreate() {
+        when(schemaService.getUploadSchema(any(), eq("guid"))).thenThrow(new EntityNotFoundException(AppConfig.class));
```
DwayneJengSage: Is it worth using eq("whatever") instead of any() to make sure we're passing through the args correctly? (Here and in all the other tests.) This is probably more important for verifying survey guid and createdOn are passed correctly.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/validators/AppConfigValidator.java
alxdarksage: Switched to this method call

test/org/sagebionetworks/bridge/validators/AppConfigValidatorTest.java
alxdarksage: Definitely. Changed all of these to concrete arguments and cleaned up the redundancy so the tests are shorter and clearer.

DwayneJengSage (approved):

PR #1748, opened 2018-07-09
Do not recreate activities in the scheduler if they have clientData, regardless of their status.

alxdarksage:
Client developers noticed that if you added client data to an unstarted activity and saved it, you would not get the client data back. That's because we were recreating the activity, assuming it could be updated from the status alone. Now also checks that there's no client data... once there is, we no longer regenerate the DB activity, we return it to the user.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
+++ b/app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
@@ -228,11 +228,13 @@ public class ScheduledActivityService {
             ScheduledActivity activity = scheduledActivities.get(i);
             ScheduledActivity dbActivity = dbMap.remove(activity.getGuid());
             
-            if (dbActivity != null && !UPDATABLE_STATUSES.contains(dbActivity.getStatus())) {
-                // Once the activity is in the database and is in a non-updatable state, we should use the one from the
-                // database. Otherwise, either (a) it doesn't exist yet and needs to be persisted or (b) the user
-                // hasn't interacted with it yet, so we can safely replace it with the newly generated one, which may
-                // have updated schemas or surveys.
+            boolean updatableStatus = dbActivity != null && UPDATABLE_STATUSES.contains(dbActivity.getStatus());
+            
+            if (dbActivity != null && (!updatableStatus || dbActivity.getClientData() != null)) {
+                // Once the activity is in the database and is in a non-updatable state (and that includes attaching
+                // client data to the activity), we should use the one from the database. Otherwise, either (a) it
+                // doesn't exist yet and needs to be persisted or (b) the user hasn't interacted with it yet, so we 
+                // can safely replace it with the newly generated one, which may have updated schemas or surveys.
                 //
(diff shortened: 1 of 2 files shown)


app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
```diff
@@ -228,11 +228,13 @@ protected DateTime getDateTime() {
...
             ScheduledActivity dbActivity = dbMap.remove(activity.getGuid());
             
-            if (dbActivity != null && !UPDATABLE_STATUSES.contains(dbActivity.getStatus())) {
-                // Once the activity is in the database and is in a non-updatable state, we should use the one from the
-                // database. Otherwise, either (a) it doesn't exist yet and needs to be persisted or (b) the user
-                // hasn't interacted with it yet, so we can safely replace it with the newly generated one, which may
-                // have updated schemas or surveys.
+            boolean updatableStatus = dbActivity != null && UPDATABLE_STATUSES.contains(dbActivity.getStatus());
+            
+            if (dbActivity != null && (!updatableStatus || dbActivity.getClientData() != null)) {
```
DwayneJengSage: Minor: The dbActivity != null clause in updatableStatus is redundant with the dbActivity != null in the if statement. You could simplify the whole thing to

if (dbActivity != null && (!UPDATABLE_STATUSES.contains(dbActivity.getStatus()) || dbActivity.getClientData() != null)) {

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
```diff
@@ -600,11 +603,81 @@ public void taskNotStartedIsUpdatedV4() {
...
         assertEquals(1, returnedActivities.size());
         assertEquals(5678, returnedActivities.get(0).getActivity().getSurvey().getCreatedOn().getMillis());
+        
+        verify(activityDao).getActivityHistoryV2(any(), any(), any(), any(), any(), anyInt());
+    }
+    
+    @Test
+    public void taskNotStartedWithClientDataNotUpdated() {
+        // Activity in DDB has survey reference pointing to createdOn 1234. Newly created activity has survey reference
+        // pointing to createdOn 5678. Activity in DDB hasn't been started. We should update the activity in DDB.
```
DwayneJengSage: Minor: Update comment to reflect the new reality. (Here and in the next test.)

DwayneJengSage (approved):
Looks fine. Just two minor comments.

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
alxdarksage: Changed

test/org/sagebionetworks/bridge/services/ScheduledActivityServiceMockTest.java
alxdarksage: Updated

PR #1750, opened 2018-07-18
increase activity max date range to 32 days

DwayneJengSage:
This is useful for 2 reasons:

1. Study bursts are 19 days long, so being able to get all activities in a single call is useful.
2. It's super convenient to be able to get activities a month at a time instead of just 15 days at a time, especially if you need to bootstrap 2 years worth of activities.

Note that there's a weird quirk of Joda which means the max date range is an exclusive upper bound (ie, it needs to be less than 32 days, not less than or equal to 32 days). Not sure why that is, but I figured it was easier to just set it to 32 days than figure out why I can't make it <=31 days.


alxdarksage:
We'll need to update API and developer.sagebridge.org documentation and it's worth telling Erin & Josh about this change too, as they often iterate ahead to get these records well into the future (or even for the entire duration of the study).

alxdarksage (approved):

PR #1753, opened 2018-07-27
Moving a lot of business logic from controller and DAO into the service.

alxdarksage:
Moving a number of  business rules out of the controller (which should just deal with web tier translation) and the DAO (which should be as simple as possible and only concerned with persistence implementation). This makes it possible to allow the physical deletion of logically/virtually deleted surveys, and removes some try/catch blocks where we were converting EntityNotFoundExceptions being thrown by the DAO.

I would like to move functionality into one package so the DAO implementation is package private, ensuring only the service can access it (and allowing *all* the business rules to be in the service), but I'll leave that for another day.

### DIFF
--- a/app/org/sagebionetworks/bridge/play/controllers/SurveyController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/SurveyController.java
@@ -11,8 +11,8 @@ import javax.annotation.Resource;
 
 import org.sagebionetworks.bridge.Roles;
 import org.sagebionetworks.bridge.cache.ViewCache;
+import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
 import org.sagebionetworks.bridge.cache.CacheKey;
-import org.sagebionetworks.bridge.exceptions.UnauthorizedException;
 import org.sagebionetworks.bridge.time.DateUtils;
 import org.sagebionetworks.bridge.models.GuidCreatedOnVersionHolder;
 import org.sagebionetworks.bridge.models.GuidCreatedOnVersionHolderImpl;
@@ -55,7 +55,6 @@ public class SurveyController extends BaseController {
 
         List<Survey> surveys = surveyService.getAllSurveysMostRecentVersion(studyId,
                 Boolean.valueOf(includeDeletedString));
-        verifySurveyIsInStudy(session, surveys);
         return okResult(surveys);
     }
     
@@ -65,7 +64,6 @@ public class SurveyController extends BaseController {
 
         List<Survey> surveys = surveyService.getAllSurveysMostRecentlyPublishedVersion(studyId,
                 Boolean.valueOf(includeDeletedString));
-        verifySurveyIsInStudy(session, surveys);
         return okResult(surveys);
     }
 
@@ -121,7 +119,7 @@ public class SurveyController extends BaseController {
     private Result getSurveyForWorker(String surveyGuid, String createdOnString) {
         long createdOn = DateUtils.convertToMillisFromEpoch(createdOnString);
         GuidCreatedOnVersionHolder keys = new GuidCreatedOnVersionHolderImpl(surveyGuid, createdOn);
-        Survey survey = surveyService.getSurvey(keys, true);
+        Survey survey = surveyService.getSurvey(null, keys, true, true);
         return okResult(survey);
     }
 
@@ -163,12 +161,19 @@ public class SurveyController extends BaseController {
     public Result deleteSurvey(String surveyGuid, String createdOnString, String physical) throws Exception {
         UserSession session = getAuthenticatedSession(DEVELOPER, ADMIN);
         StudyIdentifier studyId = session.getStudyIdentifier();
-        Survey survey = getSurveyWithoutCacheInternal(surveyGuid, createdOnString, session);
         
+        long createdOn = DateUtils.convertToMillisFromEpoch(createdOnString);
+        GuidCreatedOnVersionHolder keys = new GuidCreatedOnVersionHolderImpl(surveyGuid, createdOn);
+        
+        // This call will return logically deleted surveys, which allows them to be physically deleted.
+        Survey survey = surveyService.getSurvey(studyId, keys, false, false);
+        if (survey == null) {
+            throw new EntityNotFoundException(Survey.class);
+        }
         if ("true".equals(physical) && session.isInRole(ADMIN)) {
             surveyService.deleteSurveyPermanently(studyId, survey);
         } else {
-            surveyService.deleteSurvey(survey);    
+            surveyService.deleteSurvey(studyId, survey);    
         }
         expireCache(surveyGuid, createdOnString, studyId);
         return okResult("Survey deleted.");
@@ -180,7 +185,6 @@ public class SurveyController extends BaseController {
         
         List<Survey> surveys = surveyService.getSurveyAllVersions(studyId, surveyGuid,
                 Boolean.valueOf(includeDeletedString));
-        verifySurveyIsInStudy(session, surveys);
         return okResult(surveys);
     }
     
@@ -200,9 +204,10 @@ public class SurveyController extends BaseController {
         UserSession session = getAuthenticatedSession(DEVELOPER);
         StudyIdentifier studyId = session.getStudyIdentifier();
         
-        Survey survey = getSurveyWithoutCacheInternal(surveyGuid, createdOnString, session);
-
-        survey = surveyService.versionSurvey(survey);
+        long createdOn = DateUtils.convertToMillisFromEpoch(createdOnString);
+        GuidCreatedOnVersionHolder keys = new GuidCreatedOnVersionHolderImpl(surveyGuid, createdOn);
+        
+        Survey survey = surveyService.versionSurvey(session.getStudyIdentifier(), keys);
         expireCache(surveyGuid, createdOnString, studyId);
         
         return createdResult(new GuidCreatedOnVersionHolderImpl(survey));
@@ -212,9 +217,6 @@ public class SurveyController extends BaseController {
         UserSession session = getAuthenticatedSession(DEVELOPER);
         StudyIdentifier studyId = session.getStudyIdentifier();
         
-        // Just checking permission to access
-        getSurveyWithoutCacheInternal(surveyGuid, createdOnString, session);
-        
         // The parameters in the URL take precedence over anything declared in 
         // the object itself.
         Survey survey = parseJson(request(), Survey.class);
@@ -222,7 +224,7 @@ public class SurveyController extends BaseController {
         survey.setCreatedOn(DateUtils.convertToMillisFromEpoch(createdOnString));
         survey.setStudyIdentifier(studyId.getIdentifier());
         
-        survey = surveyService.updateSurvey(survey);
+        survey = surveyService.updateSurvey(session.getStudyIdentifier(), survey);
         expireCache(surveyGuid, createdOnString, studyId);
         
         return okResult(new GuidCreatedOnVersionHolderImpl(survey));
@@ -233,21 +235,13 @@ public class SurveyController extends BaseController {
         UserSession session = getAuthenticatedSession(DEVELOPER);
         StudyIdentifier studyId = session.getStudyIdentifier();
          
-        Survey survey = getSurveyWithoutCacheInternal(surveyGuid, createdOnString, session);
-
-        survey = surveyService.publishSurvey(studyId, survey, Boolean.parseBoolean(newSchemaRev));
-        expireCache(surveyGuid, createdOnString, studyId);
-        
-        return okResult(new GuidCreatedOnVersionHolderImpl(survey));
-    }
-
-    private Survey getSurveyWithoutCacheInternal(String surveyGuid, String createdOnString, UserSession session) {
         long createdOn = DateUtils.convertToMillisFromEpoch(createdOnString);
         GuidCreatedOnVersionHolder keys = new GuidCreatedOnVersionHolderImpl(surveyGuid, createdOn);
         
-        Survey survey = surveyService.getSurvey(keys, true);
-        verifySurveyIsInStudy(session, survey);
-        return survey;
+        Survey survey = surveyService.publishSurvey(studyId, keys, Boolean.parseBoolean(newSchemaRev));
+        expireCache(surveyGuid, createdOnString, studyId);
+        
+        return okResult(new GuidCreatedOnVersionHolderImpl(survey));
     }
     
     private Result getCachedSurveyInternal(String surveyGuid, String createdOnString, UserSession session) {
@@ -258,7 +252,7 @@ public class SurveyController extends BaseController {
                 session.getStudyIdentifier().getIdentifier());
         
         String json = getView(cacheKey, session, () -> {
-            return surveyService.getSurvey(keys, true);
+            return surveyService.getSurvey(session.getStudyIdentifier(), keys, true, true);
         });
 
         return ok(json).as(JSON_MIME_TYPE);
@@ -277,28 +271,8 @@ public class SurveyController extends BaseController {
     
     private String getView(CacheKey cacheKey, UserSession session, Supplier<Survey> supplier) {
         return viewCache.getView(cacheKey, () -> {
-            Survey survey = supplier.get();
-            verifySurveyIsInStudy(session, survey);
-            return survey;
+            return supplier.get();
         });
-    }
-    
-    private void verifySurveyIsInStudy(UserSession session,List<Survey> surveys) {
-        if (!surveys.isEmpty()) {
-            verifySurveyIsInStudy(session, surveys.get(0));
-        }
-    }
-    
-    private void verifySurveyIsInStudy(UserSession session, Survey survey) {
-        // This can happen if the user has the right keys to a survey, but it's not in the user's study.
-        if (survey == null) {
-            throw new UnauthorizedException();
-        }
-        StudyIdentifier studyId = session.getStudyIdentifier();
-        if (!session.isInRole(ADMIN) && 
-            !survey.getStudyIdentifier().equals(studyId.getIdentifier())) {
-            throw new UnauthorizedException();
-        }
     }
     
     private void expireCache(String surveyGuid, String createdOnString, StudyIdentifier studyId) {
--- a/app/org/sagebionetworks/bridge/play/controllers/SurveyController.java
+++ b/app/org/sagebionetworks/bridge/play/controllers/SurveyController.java
@@ -166,14 +166,14 @@ public class SurveyController extends BaseController {
         GuidCreatedOnVersionHolder keys = new GuidCreatedOnVersionHolderImpl(surveyGuid, createdOn);
         
         // This call will return logically deleted surveys, which allows them to be physically deleted.
-        Survey survey = surveyService.getSurvey(studyId, keys, false, false);
+        Survey survey = surveyService.getSurvey(session.getParticipant().getRoles(), studyId, keys, false, false);
         if (survey == null) {
             throw new EntityNotFoundException(Survey.class);
         }
         if ("true".equals(physical) && session.isInRole(ADMIN)) {
-            surveyService.deleteSurveyPermanently(studyId, survey);
+            surveyService.deleteSurveyPermanently(session.getParticipant().getRoles(), studyId, survey);
         } else {
-            surveyService.deleteSurvey(studyId, survey);    
+            surveyService.deleteSurvey(studyId, survey);
         }
         expireCache(surveyGuid, createdOnString, studyId);
         return okResult("Survey deleted.");
--- a/app/org/sagebionetworks/bridge/services/SharedModuleMetadataService.java
+++ b/app/org/sagebionetworks/bridge/services/SharedModuleMetadataService.java
@@ -95,7 +95,8 @@ public class SharedModuleMetadataService {
             long createdOn = metadata.getSurveyCreatedOn();
 
             // Metadata does not have study information
-            Survey survey = surveyService.getSurvey(null, new GuidCreatedOnVersionHolderImpl(surveyGuid, createdOn), false, false);
+            Survey survey = surveyService.getSurvey(SHARED_STUDY_ID,
+                    new GuidCreatedOnVersionHolderImpl(surveyGuid, createdOn), false, false);
             if (survey == null) {
                 throw new BadRequestException("Survey " + surveyGuid + " referred does not exist.");    
             }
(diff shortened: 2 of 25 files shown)


app/org/sagebionetworks/bridge/services/SurveyService.java
```diff
@@ -291,6 +295,20 @@ public Survey getSurveyMostRecentlyPublishedVersion(StudyIdentifier studyIdentif
 
         return surveyDao.getAllSurveysMostRecentVersion(studyIdentifier, includeDeleted);
```
DwayneJengSage, line 306: Needs a isInStudy() check. (Here and getAllSurveysMostRecentlyPublishedVersion())

app/org/sagebionetworks/bridge/play/controllers/SurveyController.java
```diff
@@ -121,7 +119,7 @@ public Result getSurvey(String surveyGuid, String createdOnString) throws Except
     private Result getSurveyForWorker(String surveyGuid, String createdOnString) {
         long createdOn = DateUtils.convertToMillisFromEpoch(createdOnString);
         GuidCreatedOnVersionHolder keys = new GuidCreatedOnVersionHolderImpl(surveyGuid, createdOn);
-        Survey survey = surveyService.getSurvey(keys, true);
+        Survey survey = surveyService.getSurvey(null, keys, true, true);
```
DwayneJengSage, line 122: Oddly enough, I can't find this API anywhere in our routes. With that in mind, could we re-define this API as taking in a Study ID and pass that into the service, just for consistency's sake?

app/org/sagebionetworks/bridge/services/SurveyService.java
```diff
@@ -199,11 +193,12 @@ public void deleteSurvey(GuidCreatedOnVersionHolder keys) {
...
-    public void deleteSurveyPermanently(StudyIdentifier studyId, GuidCreatedOnVersionHolder keys) {
-        checkArgument(StringUtils.isNotBlank(keys.getGuid()), "Survey GUID cannot be null/blank");
-        checkArgument(keys.getCreatedOn() != 0L, "Survey createdOn timestamp cannot be 0");
-
-        checkConstraintsBeforePhysicalDelete(studyId, keys);
+    public void deleteSurveyPermanently(StudyIdentifier studyIdentifier, GuidCreatedOnVersionHolder keys) {
+        Survey existing = surveyDao.getSurvey(keys, false);
+        if (existing == null || !isInStudy(studyIdentifier, existing)) {
+            throw new EntityNotFoundException(Survey.class);
+        }
```
DwayneJengSage: Delete survey permanently is an admin API, and I believe we wanted to make admin APIs cross-study. With that in mind, do we still need the isInStudy() check here?

app/org/sagebionetworks/bridge/play/controllers/SurveyController.java
```diff
@@ -163,12 +161,19 @@ public Result getSurveyMostRecentlyPublishedVersion(String surveyGuid) throws Ex
     public Result deleteSurvey(String surveyGuid, String createdOnString, String physical) throws Exception {
         UserSession session = getAuthenticatedSession(DEVELOPER, ADMIN);
         StudyIdentifier studyId = session.getStudyIdentifier();
-        Survey survey = getSurveyWithoutCacheInternal(surveyGuid, createdOnString, session);
         
+        long createdOn = DateUtils.convertToMillisFromEpoch(createdOnString);
+        GuidCreatedOnVersionHolder keys = new GuidCreatedOnVersionHolderImpl(surveyGuid, createdOn);
+        
+        // This call will return logically deleted surveys, which allows them to be physically deleted.
+        Survey survey = surveyService.getSurvey(studyId, keys, false, false);
```
DwayneJengSage: getSurvey() will check isInStudy() if you pass in the studyID. This means that the admin can't delete surveys outside of its study.

app/org/sagebionetworks/bridge/play/controllers/SurveyController.java
```diff
@@ -163,12 +161,19 @@ public Result getSurveyMostRecentlyPublishedVersion(String surveyGuid) throws Ex
...
+        if (survey == null) {
+            throw new EntityNotFoundException(Survey.class);
+        }
         if ("true".equals(physical) && session.isInRole(ADMIN)) {
             surveyService.deleteSurveyPermanently(studyId, survey);
         } else {
-            surveyService.deleteSurvey(survey);    
+            surveyService.deleteSurvey(studyId, survey);
```
DwayneJengSage: deleteSurvey() has an isInStudy() check. For developer, this is desirable, but for admin, this prevents cross-study logical delete.

test/org/sagebionetworks/bridge/play/controllers/SurveyControllerTest.java
```diff
@@ -557,28 +455,12 @@ public void physicalDeleteAllowedForAdmin() throws Exception {
     public void deleteSurveyThrowsGoodExceptionIfSurveyDoesntExist() throws Exception {
         setupContext(API_STUDY_ID, DEVELOPER, UNCONSENTED, null);
         
-        when(service.getSurvey(KEYS, true)).thenThrow(new EntityNotFoundException(Survey.class));
+        when(service.getSurvey(TestConstants.TEST_STUDY, KEYS, false, false)).thenReturn(null);
         
         Result result = controller.deleteSurvey(SURVEY_GUID, CREATED_ON.toString(), "false");
         TestUtils.assertResult(result, 200, "Survey deleted.");
```
DwayneJengSage: Nitpick: controller.deleteSurvey() throws, so the assert will never be checked.

app/org/sagebionetworks/bridge/services/SurveyService.java
```diff
@@ -233,7 +228,11 @@ private void verifySharedModuleExistence(GuidCreatedOnVersionHolder keys) {
         checkNotNull(studyIdentifier, Validate.CANNOT_BE_NULL, "study");
         checkArgument(isNotBlank(guid), Validate.CANNOT_BE_BLANK, "survey guid");
 
-        return surveyDao.getSurveyAllVersions(studyIdentifier, guid, includeDeleted);
+        List<Survey> allVersions = surveyDao.getSurveyAllVersions(studyIdentifier, guid, includeDeleted);
```
DwayneJengSage, line 241: Needs a isInStudy() check.

app/org/sagebionetworks/bridge/services/SurveyService.java
```diff
@@ -291,6 +295,20 @@ public Survey getSurveyMostRecentlyPublishedVersion(StudyIdentifier studyIdentif
...
+     * shared studies (which are not in the admin's study). For backwards compatibility, do not enforce the same 
+     * study rule for shared study surveys. Eventually we want admins to be able to switch into the shared study in 
+     * order to delete items there, then this exception to the check can be removed. 
+     */
+    private boolean isInStudy(StudyIdentifier studyId, Survey survey) {
+        if (studyId == null || survey == null || survey.getStudyIdentifier() == null) {
+            return true;
+        }
+        return survey.getStudyIdentifier().equals("shared") || 
+               survey.getStudyIdentifier().equals(studyId.getIdentifier());
```
DwayneJengSage: This allows non-Sage developers to modify, update, and delete surveys in the shared study, which is not something we want to allow. Shared modules should be managed by Sage.

app/org/sagebionetworks/bridge/services/SurveyService.java
```diff
@@ -76,22 +76,17 @@ public final void setStudyService(StudyService studyService) {
...
-     * @return
      */
-    public Survey getSurvey(GuidCreatedOnVersionHolder keys, boolean includeElements) {
-        checkArgument(StringUtils.isNotBlank(keys.getGuid()), "Survey GUID cannot be null/blank");
-        checkArgument(keys.getCreatedOn() != 0L, "Survey createdOn timestamp cannot be 0");
-        
-        return surveyDao.getSurvey(keys, includeElements);
+    public Survey getSurvey(StudyIdentifier studyIdentifier, GuidCreatedOnVersionHolder keys, boolean includeElements, boolean throwException) {
+        Survey survey = surveyDao.getSurvey(keys, includeElements);
+        if (throwException && survey == null || !isInStudy(studyIdentifier, survey)) {
```
DwayneJengSage: I'm not sure whether && takes priority or ||. (I think it's &&.) If && takes priority, then if the isInStudy() check fails, this will always through. If || takes priority, then this method can be used to return surveys outside of the study. I don't believe either case is desirable.

app/org/sagebionetworks/bridge/services/SharedModuleMetadataService.java
```diff
@@ -93,10 +94,10 @@ private void validateSchemaOrSurveyExists(SharedModuleMetadata metadata) {
             String surveyGuid = metadata.getSurveyGuid();
             long createdOn = metadata.getSurveyCreatedOn();
 
-            try {
-                surveyService.getSurvey(new GuidCreatedOnVersionHolderImpl(surveyGuid, createdOn), false);
-            } catch (EntityNotFoundException e) {
-                throw new BadRequestException("Survey " + surveyGuid + " referred does not exist.");
+            // Metadata does not have study information
+            Survey survey = surveyService.getSurvey(null, new GuidCreatedOnVersionHolderImpl(surveyGuid, createdOn), false, false);
+            if (survey == null) {
```
DwayneJengSage: I believe the intent is that shared modules can only point at surveys in the shared study. (Based on a similar pattern in the previous branch for schemas.) I think the lack of study checking in the original version of the code was a bug.

app/org/sagebionetworks/bridge/services/SharedModuleService.java
```diff
@@ -120,7 +120,7 @@ private SharedModuleImportStatus importModule(StudyIdentifier studyId, SharedMod
             long sharedSurveyCreatedOn = metadata.getSurveyCreatedOn();
             GuidCreatedOnVersionHolder sharedSurveyKey = new GuidCreatedOnVersionHolderImpl(sharedSurveyGuid,
                     sharedSurveyCreatedOn);
-            Survey sharedSurvey = surveyService.getSurvey(sharedSurveyKey, true);
+            Survey sharedSurvey = surveyService.getSurvey(studyId, sharedSurveyKey, true, true);
```
DwayneJengSage: Similarly, based on the schema branch above, this code is supposed to get the survey from the shared study and then create it in the local study. So this should be BridgeConstants.SHARED_STUDY_ID.

DwayneJengSage (changes requested):
Still need to review SurveyService.java and all unit tests. Will resume reviewing after the break.

app/org/sagebionetworks/bridge/services/SurveyService.java
alxdarksage: All the list-based methods use study identifier to query for the surveys, rather than GUID, so they will always matched the supplied study identifier. (In the controller we get this study identifier from the user's session).

app/org/sagebionetworks/bridge/play/controllers/SurveyController.java
alxdarksage: getSurveyForWorker is called from getSurvey in the controller. This is different than other worker calls because we didn't use the study Id as part of survey keys (which in hindsight is a mistake, particularly now that everyone refers to them via identifiers). 

As far as admins go, I would like to pursue the approach we discussed before the break where an admin can associate themselves to any study (so their session would include that study ID and their roles would be researcher/developer). Then after switching, that admin can do anything in that study without separate endpoints.

However we'd need to do that before removing any APIs administrators currently rely on.

Updated JIRA to describe this approach which will not require a lot of additional URLs to support: https://sagebionetworks.jira.com/browse/BRIDGE-2115

app/org/sagebionetworks/bridge/services/SurveyService.java
alxdarksage: See comment above.

app/org/sagebionetworks/bridge/play/controllers/SurveyController.java
alxdarksage: See comment above

app/org/sagebionetworks/bridge/play/controllers/SurveyController.java
alxdarksage: See comment above.

app/org/sagebionetworks/bridge/services/SurveyService.java
alxdarksage: How should we secure shared modules? Should they only be accessible to admins (presumably all Sage employees)? If so I will check and throw an exception if this is not true in the controller.

app/org/sagebionetworks/bridge/services/SurveyService.java
alxdarksage: Tests were inadequate. Added a test to verify this will return null when throwException=false, and throws otherwise, then fixed the logic.

app/org/sagebionetworks/bridge/services/SurveyService.java
alxdarksage: I believe if we follow the model that only shared project devs/admins can change that study, it is secure, so I'll remove this exception for "shared" modules. The integration test will have to find another way to delete test modules (more likely, check and create a test module if it's not there, rather than creating/deleting each time.

alxdarksage: I've replaced this with passing the user's roles into the service for a couple of methods, and skipping the study comparison check for admins. This is temporary until we have an API for admins to adopt any study

app/org/sagebionetworks/bridge/services/SharedModuleMetadataService.java
alxdarksage: Passing in shared study ID so this is checked (added test).

app/org/sagebionetworks/bridge/services/SharedModuleService.java
alxdarksage: Now using the constant for the shared study

app/org/sagebionetworks/bridge/services/SurveyService.java
```diff
@@ -291,6 +305,30 @@ public Survey getSurveyMostRecentlyPublishedVersion(StudyIdentifier studyIdentif
...
+     * shared studies (which are not in the admin's study). For backwards compatibility, do not enforce the same 
+     * study rule for shared study surveys. Eventually we want admins to be able to switch into the shared study in 
+     * order to delete items there, then this exception to the check can be removed.
+     * 
+     * The use of an admin role to skip this check is temporary; eventually admins will be able to act on behalf 
+     * of any study, allowing them to pass all such study checks. Currently we're skipping this check for admins
+     * when they are deleting studies permanently.
+     */
+    private boolean isInStudy(Set<Roles> roles, StudyIdentifier studyId, Survey survey) {
+        if (studyId == null || (!roles.isEmpty() && roles.contains(Roles.ADMIN))) {
```
DwayneJengSage: Nitpick: Set.contains() will simply return false if called on an empty set, so the isEmpty() check is redundant.

app/org/sagebionetworks/bridge/services/SurveyService.java
```diff
@@ -172,15 +180,11 @@ public Survey versionSurvey(GuidCreatedOnVersionHolder keys) {
...
-    public void deleteSurvey(GuidCreatedOnVersionHolder keys) {
-        checkArgument(StringUtils.isNotBlank(keys.getGuid()), "Survey GUID cannot be null/blank");
-        checkArgument(keys.getCreatedOn() != 0L, "Survey createdOn timestamp cannot be 0");
-
+    public void deleteSurvey(StudyIdentifier studyIdentifier, GuidCreatedOnVersionHolder keys) {
         Survey existing = surveyDao.getSurvey(keys, true);
-        if (existing.isDeleted()) {
+        if (existing == null || existing.isDeleted() || !isInStudy(studyIdentifier, existing)) {
             throw new EntityNotFoundException(Survey.class);
         }
```
DwayneJengSage, line 187: Can admin logically delete surveys? This code prevents that from happening.

test/org/sagebionetworks/bridge/play/controllers/SurveyControllerTest.java
```diff
@@ -528,57 +426,40 @@ public void physicalDeleteOfSurveyNotAllowedForDeveloper() throws Exception {
...
     
     @Test(expected = EntityNotFoundException.class)
     public void deleteSurveyThrowsGoodExceptionIfSurveyDoesntExist() throws Exception {
         setupContext(API_STUDY_ID, DEVELOPER, UNCONSENTED, null);
         
-        when(service.getSurvey(KEYS, true)).thenThrow(new EntityNotFoundException(Survey.class));
+        when(service.getSurvey(ImmutableSet.of(DEVELOPER), TestConstants.TEST_STUDY, KEYS, false, false)).thenReturn(null);
         
         Result result = controller.deleteSurvey(SURVEY_GUID, CREATED_ON.toString(), "false");
         TestUtils.assertResult(result, 200, "Survey deleted.");
```
DwayneJengSage: Nitpick: controller.deleteSurvey() throws, so the assert will never be checked.

app/org/sagebionetworks/bridge/services/SurveyService.java
DwayneJengSage, line 187: Unit test for deleting a non-existent survey?

app/org/sagebionetworks/bridge/services/SurveyService.java
```diff
@@ -247,7 +256,11 @@ public Survey getSurveyMostRecentVersion(StudyIdentifier studyIdentifier, String
         checkNotNull(studyIdentifier, Validate.CANNOT_BE_NULL, "study");
         checkArgument(isNotBlank(guid), Validate.CANNOT_BE_BLANK, "survey guid");
 
-        return surveyDao.getSurveyMostRecentVersion(studyIdentifier, guid);
+        Survey survey = surveyDao.getSurveyMostRecentVersion(studyIdentifier, guid);
+        if (survey == null || !isInStudy(studyIdentifier, survey)) {
+            throw new EntityNotFoundException(Survey.class);
+        }
```
DwayneJengSage, line 262: Unit test for non-existent survey?

app/org/sagebionetworks/bridge/services/SurveyService.java
```diff
@@ -264,15 +277,16 @@ public Survey getSurveyMostRecentlyPublishedVersion(StudyIdentifier studyIdentif
         checkNotNull(studyIdentifier, Validate.CANNOT_BE_NULL, "study");
         checkArgument(isNotBlank(guid), Validate.CANNOT_BE_BLANK, "survey guid");
 
-        return surveyDao.getSurveyMostRecentlyPublishedVersion(studyIdentifier, guid, includeElements);
+        Survey survey = surveyDao.getSurveyMostRecentlyPublishedVersion(studyIdentifier, guid, includeElements);
+        if (survey == null || !isInStudy(studyIdentifier, survey)) {
+            throw new EntityNotFoundException(Survey.class);
+        }
```
DwayneJengSage, line 283: Unit test for non-existent survey?

app/org/sagebionetworks/bridge/services/SurveyService.java
DwayneJengSage: Unit test for isInStudy() check with null study?

test/org/sagebionetworks/bridge/play/controllers/SurveyControllerTest.java
```diff
@@ -951,18 +652,15 @@ private void assertCacheIsCleared(ExecuteSurvey executeSurvey) throws Exception
...
+        when(service.versionSurvey(eq(TestConstants.TEST_STUDY), any())).thenReturn(survey);
+        when(service.updateSurvey(eq(TestConstants.TEST_STUDY), any())).thenReturn(survey);
         
         // execute the test method, this should delete the cache
         executeSurvey.execute(SURVEY_GUID, CREATED_ON.toString());
-        verify(service).getSurvey(any(), eq(true));
         
         // This call now hits the service, not the cache, for a total of two calls to the service
         controller.getSurvey(SURVEY_GUID, CREATED_ON.toString());
-        verify(service, times(2)).getSurvey(any(), eq(true));
```
DwayneJengSage, line 965: Should we still verify caching logic in our unit tests?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/SurveyService.java
alxdarksage: Currently none of the integration tests try and do this so it's not causing anything to fail. I would prefer to implement a cross-study admin (see PR) and change the integration tests to use that, rather than fully implement a role-based exception here.

test/org/sagebionetworks/bridge/play/controllers/SurveyControllerTest.java
alxdarksage: Removed inaccessible code

app/org/sagebionetworks/bridge/services/SurveyService.java
alxdarksage: Added

app/org/sagebionetworks/bridge/services/SurveyService.java
alxdarksage: Added

app/org/sagebionetworks/bridge/services/SurveyService.java
alxdarksage: Added

app/org/sagebionetworks/bridge/services/SurveyService.java
alxdarksage: Added a test of this for getSurvey() where the studyIdentifier is not required (for some calls we must supply it, like version/publish survey).

test/org/sagebionetworks/bridge/play/controllers/SurveyControllerTest.java
alxdarksage: Added this back corrected (we now call surveyService once).

app/org/sagebionetworks/bridge/services/SurveyService.java
alxdarksage: changed

app/org/sagebionetworks/bridge/services/SurveyService.java
alxdarksage: Note for the list-based methods, study ID is part of the query so it can't differ from the user's study.

DwayneJengSage (approved):

PR #1754, opened 2018-08-07
Create means for admin user to sign in and switch between studies.

alxdarksage:
An API to authenticate as an administrator using an administrator account in the API study. It then switches the user to the study that was indicated in the sign in request. From that point on it should function like an admin in that study (though this has not been extensively tested and I have no UI to sign in this way at the moment, to flush out how well this will work). However, it should work well, as we almost always retrieve a study from the user's session. 

There is also an API to change study (without signing out). Again you must be an admin to do this... though if you sign in as an admin in a non-API study, you could then use this API to switch to another study. Eventually I envision deleting the admin accounts in all studies but the API study just to make management of these accounts easier.

### DIFF
--- a/test/org/sagebionetworks/bridge/play/controllers/UserManagementControllerTest.java
+++ b/test/org/sagebionetworks/bridge/play/controllers/UserManagementControllerTest.java
@@ -5,10 +5,10 @@ import static org.mockito.Mockito.any;
 import static org.mockito.Mockito.anyBoolean;
 import static org.mockito.Mockito.anyObject;
 import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.eq;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 import static org.sagebionetworks.bridge.Roles.ADMIN;
-import static org.sagebionetworks.bridge.TestConstants.TEST_STUDY;
 import static org.sagebionetworks.bridge.TestConstants.TEST_STUDY_IDENTIFIER;
 import static org.sagebionetworks.bridge.TestUtils.assertResult;
 import static org.sagebionetworks.bridge.TestUtils.mockPlayContextWithJson;
@@ -21,6 +21,8 @@ import com.google.common.collect.Sets;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
 import org.mockito.Mock;
 import org.mockito.Spy;
 import org.mockito.runners.MockitoJUnitRunner;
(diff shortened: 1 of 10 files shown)


test/org/sagebionetworks/bridge/play/controllers/UserManagementControllerTest.java
```diff
@@ -87,6 +116,38 @@ public void before() throws Exception {
         when(request.headers()).thenReturn(map);
         doReturn(null).when(controller).getMetrics();
     }
+
+    @Test
+    public void signInForAdmin() throws Exception {
```
DwayneJengSage, line 125: For branch coverage, we probably also want a test for the case where a non-admin calls this API.

test/org/sagebionetworks/bridge/play/controllers/UserManagementControllerTest.java
```diff
@@ -87,6 +116,38 @@ public void before() throws Exception {
...
+        TestUtils.mockPlayContextWithJson(signIn);
+
+        when(authService.signIn(eq(study), any(CriteriaContext.class), signInCaptor.capture())).thenReturn(session);
+        
+        Result result = controller.signInForAdmin();
+        TestUtils.assertResult(result, 200);
+        
+        // This isn't in the session that is returned to the user, but verify it has been changed
+        assertEquals("originalStudy", session.getStudyIdentifier().getIdentifier());
+        assertEquals("api", signInCaptor.getValue().getStudyId());
```
DwayneJengSage, line 140: Do we also want to verify sessionUpdateService was called and cookies are created?

test/org/sagebionetworks/bridge/play/controllers/UserManagementControllerTest.java
```diff
@@ -87,6 +116,38 @@ public void before() throws Exception {
...
+        SignIn signIn = new SignIn.Builder().withStudy("nextStudy").build();
+        TestUtils.mockPlayContextWithJson(signIn);
+        
+        Study nextStudy = Study.create();
+        nextStudy.setIdentifier("nextStudy");
+        when(studyService.getStudy("nextStudy")).thenReturn(nextStudy);
+        
+        controller.changeStudyForAdmin();
+        assertEquals("nextStudy", session.getStudyIdentifier().getIdentifier());
+        verify(cacheProvider).setUserSession(session);
```
DwayneJengSage, line 173: Do we also want to verify sessionUpdateService was called?

DwayneJengSage (changes requested):
Overall looks good. Just a few test comments.

test/org/sagebionetworks/bridge/play/controllers/UserManagementControllerTest.java
alxdarksage: Added

test/org/sagebionetworks/bridge/play/controllers/UserManagementControllerTest.java
alxdarksage: Cache provider is a mock dependency of the sessionUpdateService, which was injected, so verifying the cache is called and the session object is updated, verifies all the work of the sessionUpdateService. Added test that cookie is set during this sign in method, in the Local environment.

test/org/sagebionetworks/bridge/play/controllers/UserManagementControllerTest.java
alxdarksage: It's a real object that has cache provider set as a mock; verifying cache provider and the session are updated, verifies the work that sessionUpdateService does (there's also a test of the method in that service in SessionUpdateServiceTest).

If this is too wonky, I can make sessionUpdateService a mock and just verify that it is being called instead.

DwayneJengSage (approved):

PR #1759, opened 2018-08-15
SMS Notifications and Criteria-Managed Topics

DwayneJengSage:
This is used to support sending monthly SMS messages about self-tracking, using Data Groups to determine audience. See https://sagebionetworks.jira.com/browse/BRIDGE-2255

Testing done:
* updated unit tests
* updated integ tests
* manually tested end-to-end and verified receiving SMS message through SNS


app/org/sagebionetworks/bridge/dynamodb/DynamoNotificationTopicDao.java
```diff
@@ -98,13 +116,24 @@ public NotificationTopic createTopic(NotificationTopic topic) {
         topic.setGuid(BridgeUtils.generateGuid());
         
         String snsTopicName = createSnsTopicName(topic);
-        CreateTopicRequest request = new CreateTopicRequest().withName(snsTopicName);
-        CreateTopicResult result = snsClient.createTopic(request);
+        CreateTopicResult result = snsClient.createTopic(snsTopicName);
+
+        // Display name is required for SMS notifications. Strangely, there's no way to set display name in the
+        // create API, only the set attribute API.
+        // Display name has max length 10 chars.
```
alxdarksage: Do we validate this (tell the user there's an issue if the topic name is over 10 characters)? I think it's worth having name and shortName for a topic as we do for studies where sometimes, the study name has to be very short for display via push notifications or SMS.

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
```diff
@@ -171,7 +178,13 @@ private Result giveConsentForVersion(int version, SubpopulationGuid subpopGuid)
         Map<SubpopulationGuid,ConsentStatus> statuses = consentService.getConsentStatuses(context);
         
         sessionUpdateService.updateConsentStatus(session, statuses, sharing.getSharingScope(), false);
-        
+
+        // After consent, we should sign the user up for SMS notifications.
+        if (session.doesConsent()) {
+            notificationsService.createSmsRegistrationAfterConsent(session.getParticipant(),
+                    getCriteriaContext(session));
```
alxdarksage: We have a lot of studies where people don't consent now, including some users in mPower 2.0 who are being externally consented. Will this catch every case of SMS subscription that we want to catch? Should we make the client register manually?

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -175,7 +182,13 @@ public UserSession signIn(Study study, CriteriaContext context, SignIn signIn) t
         if (!session.doesConsent() && intentService.registerIntentToParticipate(study, account)) {
             AccountId accountId = AccountId.forId(study.getIdentifier(), account.getId());
             account = accountDao.getAccount(accountId);
-            session = getSessionFromAccount(study, context, account);        
+            session = getSessionFromAccount(study, context, account);
+
+            if (session.doesConsent()) {
```
alxdarksage: Same comment as above about consent

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -371,7 +384,13 @@ private UserSession channelSignIn(ChannelType channelType, CriteriaContext conte
         
         if (!session.doesConsent() && intentService.registerIntentToParticipate(study, account)) {
             account = accountDao.getAccount(accountId);
-            session = getSessionFromAccount(study, context, account);        
+            session = getSessionFromAccount(study, context, account);
+
+            if (session.doesConsent()) {
+                // After consent, we should sign the user up for SMS notifications.
+                CriteriaContext updatedContext = updateContextFromSession(context, session);
+                notificationsService.createSmsRegistrationAfterConsent(session.getParticipant(), updatedContext);
```
alxdarksage: Same.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -414,16 +433,21 @@ UserSession getSessionFromAccount(Study study, CriteriaContext context, Account
         session.setStudyIdentifier(study.getStudyIdentifier());
         session.setReauthToken(account.getReauthToken());
         
-        CriteriaContext newContext = new CriteriaContext.Builder()
-                .withContext(context)
+        CriteriaContext newContext = updateContextFromSession(context, session);
+        session.setConsentStatuses(consentService.getConsentStatuses(newContext, account));
```
alxdarksage, line 430: Same, you get the picture. It would be easier if the client manually registered for those studies that want it, or perhaps it's a study setting and we do this registration when we create/add a phone number to any account in the system for those studies that are going to send SMS messages (just brainstorming)

app/org/sagebionetworks/bridge/services/NotificationsService.java
```diff
@@ -84,27 +102,93 @@ public NotificationRegistration getRegistration(String healthCode, String guid)
...
+
+        return createdRegistration;
     }
-    
+
+    /**
+     * This API is called after consent (either manual consent or intent to participate) to set up SMS notification
+     * registrations. If the user doesn't have a verified phone number, or they already have a notification
+     * registration, this API does nothing.
+     */
```
alxdarksage, line 167: One thing we don't catch right now is a change in the version of the application (also in the context). I cannot think of a case where we'd register or unregister someone from a notification based on this, but we should document that this aspect of criteria isn't useful for a topic (I'll adjust the BSM UI accordingly).

alxdarksage (approved):
Approved, though I'm not sure about subscribing people to the SMS topic on consent. Or, I'm a little confused about it. Is there a Bridge Topic for SMS or are all these registrations directly through SNS? Can we use criteria to determine who gets an SMS message? Can we create multiple topics with different criteria, all of which are for sending SMS messages? It looks like yes, but I wanted to check.

app/org/sagebionetworks/bridge/dynamodb/DynamoNotificationTopicDao.java
DwayneJengSage: I've added a ShortName field to the Topic and have added validation to make sure it is present and not more than 10 characters.

Note that we'll need to backfill existing topics.

app/org/sagebionetworks/bridge/play/controllers/ConsentController.java
DwayneJengSage: I've removed the logic that creates registrations based on Consent and Intent to Participate.

alxdarksage:
One last comment... for criteria-based topics, we adjust the user's membership anytime the user changes a criterion, using the SessionUpdateService. But shouldn't we enroll a user on creation into these topics if they are applicable? Until we create a flag to indicate the user has declined membership, it's 100% based on the applicability of criteria, so if a user is created in an applicable state, they should go into a topic (create is just a special case of update).

DwayneJengSage:
We don't create registrations on user creation, because there is no single funnel point where we can guarantee that every user will go through during account creation.

We do apply topics when they create registrations. This is tested in integration tests.

PR #1761, opened 2018-08-22
Reauth logging, increase grace period

DwayneJengSage:
Increase reauth grace period to 5 min to account for mobile network connectivity.

Log 10 bits worth of the reauth token to verify if apps are sending the same reauth token. (The reauth token still has ~115 unknown bits.)


app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -2,9 +2,10 @@
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static org.sagebionetworks.bridge.BridgeConstants.NO_CALLER_ROLES;
-import static org.sagebionetworks.bridge.BridgeConstants.REAUTH_TOKEN_CACHE_LOOKUP_IN_SECONDS;
```
alxdarksage, line 5: This constant is now only used in HibernateAccountDao. Should we switch entirely to your new constant so the times are in sync between these two classes? Otherwise it seems like updateReauthToken in that class won't cache the token for long enough.

alxdarksage (changes requested):

app/org/sagebionetworks/bridge/services/AuthenticationService.java
DwayneJengSage: As per in-person conversation, these two caching layers cover two different use cases. The first one is to make sure we don't generate a reauth token more than once every 15 seconds, to avoid 409 concurrent modification errors and so that multiple reauths from the same app generate the same token. The second cache is to allow the same reauth token to be used, so that a network timeout on a successful reauth doesn't leave the client in a bad state.

If we changed the first caching layer to also be 5 minutes, it would have the side effect that using the new reauth token within 5 minutes wouldn't refresh the reauth token. In the real world, this never happens, but it breaks our integ tests. Specifically the integ tests that verify that reauth rotates the token.

As per in-person conversation, it's fine to leave this as is.

PR #1765, opened 2018-08-24
Accounts are defaulted to unverified, with specific pathways to verify external ID-only accounts.

alxdarksage:
Should not be able to sign in with an account that has external ID (it shouldn't be enabled, until there's a password). Also changing logic slightly to assume an account is unverified until something happens to verify it.


DwayneJengSage:
I'm assuming this is for https://sagebionetworks.jira.com/browse/BRIDGE-2268

alxdarksage:
Yes, this is for that JIRA item

DwayneJengSage (approved):

PR #1768, opened 2018-08-28
Publish an event when a user first accesses scheduled activities.

alxdarksage:
In addition, remove the "synthetic" events two_weeks_before_enrollment and two_months_before_enrollment because these are now implementable using automatic custom events, and in addition, they are not in use in any study right now (they were created for FHPS).

The event you can now schedule against is "activities_retrieved" and it's similar to enrollment in that it's immutable.

### DIFF
--- a/app/org/sagebionetworks/bridge/BridgeUtils.java
+++ b/app/org/sagebionetworks/bridge/BridgeUtils.java
@@ -30,8 +30,10 @@ import org.sagebionetworks.bridge.exceptions.BadRequestException;
 import org.sagebionetworks.bridge.exceptions.BridgeServiceException;
 import org.sagebionetworks.bridge.json.BridgeTypeName;
 import org.sagebionetworks.bridge.time.DateUtils;
+import org.sagebionetworks.bridge.models.Tuple;
 import org.sagebionetworks.bridge.models.accounts.AccountId;
 import org.sagebionetworks.bridge.models.accounts.StudyParticipant;
+import org.sagebionetworks.bridge.models.activities.ActivityEventObjectType;
 import org.sagebionetworks.bridge.models.schedules.Activity;
 import org.sagebionetworks.bridge.models.schedules.ActivityType;
 import org.sagebionetworks.bridge.models.studies.PasswordPolicy;
@@ -66,6 +68,15 @@ public class BridgeUtils {
     // "request context" object into every method of every class.
     private static final ThreadLocal<String> REQUEST_ID_THREAD_LOCAL = ThreadLocal.withInitial(() -> null);
 
+    public static Tuple<String> parseAutoEventValue(String automaticEventValue) {
+        // Will property split something like "custom:myEvent:P3W"
+        int lastIndex = automaticEventValue.lastIndexOf(":P");
+        if (lastIndex == -1) {
+            return new Tuple<>(ActivityEventObjectType.ENROLLMENT.name().toLowerCase(), automaticEventValue); 
+        }
+        return new Tuple<>(automaticEventValue.substring(0, lastIndex), automaticEventValue.substring(lastIndex+1));
+    }
+    
     public static boolean isExternalIdAccount(StudyParticipant participant) {
         return (StringUtils.isNotBlank(participant.getExternalId()) && 
                 StringUtils.isBlank(participant.getEmail()) && 
--- a/app/org/sagebionetworks/bridge/BridgeUtils.java
+++ b/app/org/sagebionetworks/bridge/BridgeUtils.java
@@ -33,7 +33,6 @@ import org.sagebionetworks.bridge.time.DateUtils;
 import org.sagebionetworks.bridge.models.Tuple;
 import org.sagebionetworks.bridge.models.accounts.AccountId;
 import org.sagebionetworks.bridge.models.accounts.StudyParticipant;
-import org.sagebionetworks.bridge.models.activities.ActivityEventObjectType;
 import org.sagebionetworks.bridge.models.schedules.Activity;
 import org.sagebionetworks.bridge.models.schedules.ActivityType;
 import org.sagebionetworks.bridge.models.studies.PasswordPolicy;
@@ -69,10 +68,10 @@ public class BridgeUtils {
     private static final ThreadLocal<String> REQUEST_ID_THREAD_LOCAL = ThreadLocal.withInitial(() -> null);
 
     public static Tuple<String> parseAutoEventValue(String automaticEventValue) {
-        // Will property split something like "custom:myEvent:P3W"
         int lastIndex = automaticEventValue.lastIndexOf(":P");
         if (lastIndex == -1) {
-            return new Tuple<>(ActivityEventObjectType.ENROLLMENT.name().toLowerCase(), automaticEventValue); 
+            // This will certainly not pass validation
+            return new Tuple<>(null, automaticEventValue); 
         }
         return new Tuple<>(automaticEventValue.substring(0, lastIndex), automaticEventValue.substring(lastIndex+1));
     }
--- a/app/org/sagebionetworks/bridge/services/ActivityEventService.java
+++ b/app/org/sagebionetworks/bridge/services/ActivityEventService.java
@@ -54,7 +54,11 @@ public class ActivityEventService {
                 .withObjectType(ActivityEventObjectType.CUSTOM)
                 .withObjectId(eventKey)
                 .withTimestamp(timestamp).build();
-        activityEventDao.publishEvent(event);
+        
+        if (activityEventDao.publishEvent(event)) {
+            // Create automatic events, as defined in the study
+            createAutomaticCustomEvents(study, healthCode, event);
+        }
     }
 
     /**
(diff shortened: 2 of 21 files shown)


app/org/sagebionetworks/bridge/services/ActivityEventService.java
```diff
@@ -157,4 +169,25 @@ public void deleteActivityEvents(String healthCode) {
...
+            }
+        }        
+    }
+    
+    private Tuple<String> parseAutoEventValue(String automaticEventValue) {
+        // though Period serialization does not contain semicolons, limit the split to the first occurrence
+        String[] valueElements = automaticEventValue.split(":", 2); 
+        return (valueElements.length == 1) ?
+            new Tuple<>(ActivityEventObjectType.ENROLLMENT.name().toLowerCase(), valueElements[0]) :
+            new Tuple<>(valueElements[0], valueElements[1]);
```
DwayneJengSage: Do we need to special case enrollment events? I thought we had decided to just move these wholesale to activities_retrieved. Enrollment only happens on consent, which means this is a less reliable way to create automatic custom events.

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
```diff
@@ -192,10 +195,14 @@ protected DateTime getDateTime() {
         List<ScheduledActivity> saves = performMerge(scheduledActivities, dbMap);
         activityDao.saveActivities(saves);
         
-        return orderActivities(scheduledActivities, V3_FILTER);
+        List<ScheduledActivity> orderedActivities = orderActivities(scheduledActivities, V3_FILTER);
+        String healthCode = context.getCriteriaContext().getHealthCode();
+        activityEventService.publishActivitiesRetrieved(study, healthCode, DateUtils.getCurrentDateTime());
```
DwayneJengSage: We need to publish the event _before_ we schedule them, since the schedule depends on this event. (Same with V4.)

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/ActivityEventService.java
alxdarksage: This was to ensure that the current configuration in mPower 2.0 works as currently specified (no event in the string). You are right that this is the only use of it so we could probably assume it will be fixed (I updated the editor to handle this origin event specification yesterday, it's in staging).

However I should validate that the automatic custom events when you save a study (to ensure they are the new format), and I'm not doing that. I will add it.

I don't see any reason to entirely avoid triggering off of enrollment if that makes sense for a study (or some other future event if we think of one besides these two, or a custom event that the study designers create and add as part of their study, etc.).

alxdarksage: Also realizing I need to parse this so that "custom:myEvent:P3W" works correctly (custom events will have two colons).

app/org/sagebionetworks/bridge/services/ScheduledActivityService.java
alxdarksage: Oh boy, very true! Moved.

app/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -66,6 +68,15 @@
     // "request context" object into every method of every class.
     private static final ThreadLocal<String> REQUEST_ID_THREAD_LOCAL = ThreadLocal.withInitial(() -> null);
 
+    public static Tuple<String> parseAutoEventValue(String automaticEventValue) {
+        // Will property split something like "custom:myEvent:P3W"
+        int lastIndex = automaticEventValue.lastIndexOf(":P");
+        if (lastIndex == -1) {
+            return new Tuple<>(ActivityEventObjectType.ENROLLMENT.name().toLowerCase(), automaticEventValue); 
+        }
```
DwayneJengSage: We shouldn't default to enrollment. mPower 2.0 configs are going to change anyway, so back-compat is a non-issue. Adding a special case for enrollment just adds complexity and confusion.

If we absolutely need a default for some reason, that default should be activities_retrieved.

app/org/sagebionetworks/bridge/services/ActivityEventService.java
```diff
@@ -53,7 +54,11 @@ public void publishCustomEvent(Study study, String healthCode, String eventKey,
                 .withObjectType(ActivityEventObjectType.CUSTOM)
                 .withObjectId(eventKey)
                 .withTimestamp(timestamp).build();
-        activityEventDao.publishEvent(event);
+        
+        if (activityEventDao.publishEvent(event)) {
+            // Create automatic events, as defined in the study
+            createAutomaticCustomEvents(study, healthCode, event);
+        }
```
DwayneJengSage, line 61: We have a circular dependency here. createAutomaticCustomEvents() calls publishCustomEvents() which calls createAutomaticCustomEvents(). It would be difficult to construct a custom event config that causes an infinite recursion here.

test/org/sagebionetworks/bridge/services/ActivityEventServiceTest.java
```diff
@@ -225,6 +233,131 @@ public void canPublishEnrollmentEventWithAutomaticCustomEvents() {
...
+        
+        activityEventService.publishEnrollmentEvent(study,"AAA-BBB-CCC", new ConsentSignature.Builder().build());
+        
+        // Only happens once, none of the other custom events are published.
+        verify(activityEventDao, times(1)).publishEvent(any());
+    }
+    
+    @Test
+    public void whenNoActivitiesRetrievedEventPublishNoCustomEvents() {
+        // Configure study with automatic custom events
```
DwayneJengSage: Is it worth adding an equivalent test for custom events publishing automatic custom events?

app/org/sagebionetworks/bridge/validators/StudyValidator.java
```diff
@@ -66,16 +73,39 @@ public void validate(Object obj, Errors errors) {
...
             if (study.getIdentifier().length() < 2) {
                 errors.rejectValue("identifier", "must be at least 2 characters");
             }
         }
         if (study.getActivityEventKeys().stream()
-                .anyMatch(k -> !k.matches(BridgeConstants.BRIDGE_IDENTIFIER_PATTERN))) {
-            errors.rejectValue("activityEventKeys", "must contain only lower-case letters and/or numbers with " +
-                    "optional dashes");
+                .anyMatch(k -> !k.matches(BridgeConstants.SYNAPSE_IDENTIFIER_PATTERN))) {
+            errors.rejectValue("activityEventKeys", SYNAPSE_IDENTIFIER_ERROR);
```
DwayneJengSage: Was this change intentional? Activity Event Keys have nothing to do with Synapse, so I'm not sure if the constants are correct.

app/org/sagebionetworks/bridge/validators/StudyValidator.java
```diff
@@ -66,16 +73,39 @@ public void validate(Object obj, Errors errors) {
...
+        }
+        if (study.getAutomaticCustomEvents() != null) {
+            for (Map.Entry<String, String> entry : study.getAutomaticCustomEvents().entrySet()) {
+                String key = entry.getKey();
+                String value = entry.getValue();
+                
+                // Validate that the key follows the same rules for activity event keys
+                if (!key.matches(BridgeConstants.SYNAPSE_IDENTIFIER_PATTERN)) {
+                    errors.rejectValue("automaticCustomEvents["+key+"]", SYNAPSE_IDENTIFIER_ERROR);
+                }
```
DwayneJengSage: Similarly, are we sure we're supposed to be using Synapse identifiers here?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/BridgeUtils.java
alxdarksage: Okay we don't default any longer, and the validator will return validation errors if the string is something along the lines of "P3D". You need to add either a system event or a custom event key to trigger off of. The existing mPower 2.0 config won't be valid after deployment.

app/org/sagebionetworks/bridge/services/ActivityEventService.java
alxdarksage: Now calling the DAO directly rather than back into the service for the creation of these automatic custom events.

test/org/sagebionetworks/bridge/services/ActivityEventServiceTest.java
alxdarksage: Added similar tests that automatic custom events don't fire for enrollment or custom events, if the DAO returns false.

app/org/sagebionetworks/bridge/validators/StudyValidator.java
alxdarksage: Yes. Your current study burst keys are camel-cased and there's no reason they can't be camel-cased, this validation seemed too stringent. I wanted it to be possible for the custom activity event keys and the automatic event keys to have the same format constraints. The synapse validation allows for upper-case letters and numbers and seemed more appropriate for this.

app/org/sagebionetworks/bridge/validators/StudyValidator.java
alxdarksage: The regex seemed appropriate for event keys: [a-zA-Z0-9_-]. Will discuss on Slack if we want to create a different constant to avoid confusion.

DwayneJengSage (approved):

PR #1776, opened 2018-09-14
Logical delete shared modules

alxdarksage:


test/org/sagebionetworks/bridge/play/controllers/SharedModuleMetadataControllerTest.java
```diff
@@ -87,33 +88,156 @@ public void create() throws Exception {
...
+    public void deleteByIdAndVersionLogically() throws Exception {
     }
 
+    @Test
+    public void deleteByIdAndVersionDefaultsToLogicalDelete() throws Exception {
+    }
+    
+    @Test
+    public void deleteByIdAndVersionByDeveloperDefaultsToLogicalDelete() throws Exception {
+    }
```
DwayneJengSage: There are a bunch of blank tests here? Are these meant to be filled in? Or are they redundant with other tests?

app/org/sagebionetworks/bridge/services/SharedModuleMetadataService.java
```diff
@@ -68,6 +70,7 @@ public SharedModuleMetadata createMetadata(SharedModuleMetadata metadata) {
         if (metadata.getVersion() == 0) {
             metadata.setVersion(oldVersion + 1);
         }
+        metadata.setDeleted(false);
```
DwayneJengSage, line 73: Should we add a unit test that we can't create a deleted module?

test/org/sagebionetworks/bridge/services/SharedModuleMetadataServiceTest.java
```diff
@@ -444,34 +574,101 @@ public void queryWithTags() {
...
         assertTrue(svcOutputMetadataList.contains(metadata2));
         assertTrue(svcOutputMetadataList.contains(metadata3));
         assertTrue(svcOutputMetadataList.contains(metadata4));
         assertTrue(svcOutputMetadataList.contains(metadata5));
     }
+    
+    @Test
+    public void queryWithTagsIncludeDeleted() {
+        // null query with 2 tags
+        Set<String> tags = ImmutableSet.of("foo", "bar");
```
DwayneJengSage: I don't think this is a sustainable test pattern. Every time we add a new feature, we double the number of test cases we have.

Instead of testing every combination of every feature, we should instead strive to make the features behave independently so we can test them independently.

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/play/controllers/SharedModuleMetadataControllerTest.java
alxdarksage: Sorry, these are redundant. Removing. The tests for the logical delete are there, they just don't say "logical" in the test name... the existing delete tests because the logical delete tests.

app/org/sagebionetworks/bridge/services/SharedModuleMetadataService.java
alxdarksage: Added

test/org/sagebionetworks/bridge/services/SharedModuleMetadataServiceTest.java
alxdarksage: I agree with this but was worried you would want the full matrix of options. Each of the clauses in the WHERE portion of the HQL query are separate from one another and ought to be testable independently. I can just add tests for a few scenarios with this one expression: 1) deleted not included, 2) include deleted alone, 3) include deleted with one other expression (just to make sure these are combining correctly). Will change.

alxdarksage: Added two tests to cover the includeDeleted=true case for queryAllMetadata and queryMetadataById each (the case where includeDeleted=false is tested over and over by all the existing tests).

DwayneJengSage (approved):

PR #1780, opened 2018-09-27
Merge update identifiers API into participant update

alxdarksage:
- Email, external ID and phone are now mutable.
- Cleaned up exception handling and reporting
- removed the updateIdentifiers API (redundant and not in use)

JIRAs:
https://sagebionetworks.jira.com/browse/BRIDGE-2211
https://sagebionetworks.jira.com/browse/BRIDGE-2309


DwayneJengSage:
I thought we decided against merging updateIdentifiers w/ updateParticipant because of security concerns? updateIdentifiers was originally put in place as a counter-measure against hijacking an account. If an attacker hijacks a session, they can permanently take over an account simply by changing email/phone number. updateIdentifier defends against that by requiring the user to enter their credentials.

DwayneJengSage:
Summary of offline discussion: There are actually two different tangentially-related feature requests here. The first is to allow updating external ID. The second is to allow updating email/phone. The first is relatively simple and non-controversial. The second has open security concerns that need to be addressed before we move forward with this. (These security concerns were raised on Sep 7 on Slack in the Bridge channel, in the context of discussing whether or not we want to do this, so these security concerns shouldn't be a surprise.)

Additionally, I'm not sure anyone has asked for the ability to update email/phone. (What people say they want and what people actually want aren't always the same thing.) Or if someone did ask for it, I'm not sure they fully understand what they're asking for and what the security risks are. Unfortunately, the JIRA tracking this issue only discusses External ID, and sprint board stickies aren't a very good way to track task details. Given that there's non-zero security risk involved in this change, and given that this change is much bigger than the External IDs change, if no one's asked for it, I don't think we should do it.

My recommendation: Split this PR (and the corresponding sprint task as needed) into two separate PRs, one for External ID, one for Email/Phone. This allows us to unblock the External ID work while we discuss Email/Phone separately.

PR #1782, opened 2018-09-28
Remove healthId<->healthCode mapping

alxdarksage:
Some overdue cleanup that removes unused code, reduces calls to DDB while generating a user, makes user generation more robust as a result.


DwayneJengSage (changes requested):
As per previous discussion, I don't think we can remove the healthCode to studyId reverse mapping. Really old uploads don't know their own study ID, and if we ever need to redrive them, we'd be SOL.

alxdarksage:
This code keeps the mapping and uses it where studyId is not in the health data table... let me know if this is not true somewhere

DwayneJengSage:
https://github.com/Sage-Bionetworks/BridgePF/pull/1782/files#diff-d58bc9332176157543538d471a2267f5R115

The code in question, now uses upload.studyId() instead of the healthcode mapping.

alxdarksage:
Okay, updated that so we will fall back to healthCodeDao if studyId is missing.

DwayneJengSage (approved):

PR #1785, opened 2018-10-04
Mutable external ID

alxdarksage:
Also removing the "allowIdentifierUpdates" in the account DAO because we always copy user input over from a StudyParticipant to an account from the database, so there's no danger of these being changed other than where we change them in our own code (by intention).

NOTE: I will put the exception handling fixes and improvements from the other PR into a separate PR.


app/org/sagebionetworks/bridge/services/ConsentPdf.java
```diff
@@ -106,11 +106,17 @@ private String createSignedDocument() {
...
+            // exposed this bug.
             String userEmail = signer.getEmail();
             String html = studyConsentAgreement.replace("@@name@@", username);
             html = html.replace("@@signing.date@@", signingDate);
-            html = html.replace("@@email@@", userEmail);
+            if (signer.getEmail() != null && Boolean.TRUE.equals(signer.getEmailVerified())) {
+                html = html.replace("@@email@@", userEmail);
+            } else {
+                html = html.replace("@@email@@", "");
+            }
```
DwayneJengSage, line 119: Was this intended to be part of this pull request? If so, should we add unit tests for this?

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -30,12 +30,12 @@
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.junit.Before;
+//import org.junit.Ignore;
```
DwayneJengSage: Nitpick: Remove commented import.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/ConsentPdf.java
alxdarksage: Yes, integration tests run against my consents as fixed locally, ran a text against a legacy document when there was no email address, and this generates an NPE. (I don't think any of these legacy documents are in production.) Added a test for this.

test/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Removed

DwayneJengSage (approved):

PR #1786, opened 2018-10-05
Improve constraint violation exception messages

alxdarksage:
And fixed the typing of some exception translations.

### DIFF
--- a/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
+++ b/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
@@ -547,15 +547,10 @@ public class BridgeSpringConfig {
         return metadataSources.buildMetadata().buildSessionFactory();
     }
 
-    @Bean(name = "accountPersistenceExceptionConverter")
-    public PersistenceExceptionConverter accountPersistenceExceptionConverter(AccountDao accountDao) {
-        return new AccountPersistenceExceptionConverter(accountDao);
-    }
-    
     @Bean(name = "accountHibernateHelper")
     @Autowired
     public HibernateHelper accountHibernateHelper(AccountDao accountDao, SessionFactory sessionFactory,
-            PersistenceExceptionConverter converter) {
+            AccountPersistenceExceptionConverter converter) {
         return new HibernateHelper(sessionFactory, converter);
     }
     
--- a/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
+++ b/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
@@ -77,7 +77,9 @@ import org.sagebionetworks.bridge.dynamodb.DynamoUpload2;
 import org.sagebionetworks.bridge.dynamodb.DynamoUploadDedupe;
 import org.sagebionetworks.bridge.dynamodb.DynamoUploadSchema;
 import org.sagebionetworks.bridge.dynamodb.DynamoUtils;
+import org.sagebionetworks.bridge.hibernate.AccountPersistentExceptionConverter;
 import org.sagebionetworks.bridge.hibernate.HibernateAccount;
+import org.sagebionetworks.bridge.hibernate.HibernateHelper;
 import org.sagebionetworks.bridge.hibernate.HibernateSharedModuleMetadata;
 import org.sagebionetworks.bridge.json.BridgeObjectMapper;
 import org.sagebionetworks.bridge.s3.S3Helper;
@@ -543,6 +545,11 @@ public class BridgeSpringConfig {
         return metadataSources.buildMetadata().buildSessionFactory();
     }
 
+    @Bean(name = "accountHibernateHelper")
+    public HibernateHelper accountHibernateHelper(SessionFactory sessionFactory) {
+        return new HibernateHelper(sessionFactory, new AccountPersistentExceptionConverter());
+    }
+    
     @Bean(name = "sessionExpireInSeconds")
     public int getSessionExpireInSeconds() {
         return BridgeConstants.BRIDGE_SESSION_EXPIRE_IN_SECONDS;
--- a/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
+++ b/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
@@ -82,7 +82,6 @@ import org.sagebionetworks.bridge.hibernate.AccountPersistenceExceptionConverter
 import org.sagebionetworks.bridge.hibernate.HibernateAccount;
 import org.sagebionetworks.bridge.hibernate.HibernateHelper;
 import org.sagebionetworks.bridge.hibernate.HibernateSharedModuleMetadata;
-import org.sagebionetworks.bridge.hibernate.PersistenceExceptionConverter;
 import org.sagebionetworks.bridge.json.BridgeObjectMapper;
 import org.sagebionetworks.bridge.s3.S3Helper;
 import org.sagebionetworks.bridge.upload.DecryptHandler;
--- a/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
+++ b/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
@@ -57,6 +57,7 @@ import org.sagebionetworks.bridge.crypto.AesGcmEncryptor;
 import org.sagebionetworks.bridge.crypto.BridgeEncryptor;
 import org.sagebionetworks.bridge.crypto.CmsEncryptor;
 import org.sagebionetworks.bridge.crypto.CmsEncryptorCacheLoader;
+import org.sagebionetworks.bridge.dao.AccountDao;
 import org.sagebionetworks.bridge.dynamodb.DynamoHealthDataRecord;
 import org.sagebionetworks.bridge.dynamodb.DynamoIndexHelper;
 import org.sagebionetworks.bridge.dynamodb.DynamoReportData;
@@ -77,10 +78,11 @@ import org.sagebionetworks.bridge.dynamodb.DynamoUpload2;
 import org.sagebionetworks.bridge.dynamodb.DynamoUploadDedupe;
 import org.sagebionetworks.bridge.dynamodb.DynamoUploadSchema;
 import org.sagebionetworks.bridge.dynamodb.DynamoUtils;
-import org.sagebionetworks.bridge.hibernate.AccountPersistentExceptionConverter;
+import org.sagebionetworks.bridge.hibernate.AccountPersistenceExceptionConverter;
 import org.sagebionetworks.bridge.hibernate.HibernateAccount;
 import org.sagebionetworks.bridge.hibernate.HibernateHelper;
 import org.sagebionetworks.bridge.hibernate.HibernateSharedModuleMetadata;
+import org.sagebionetworks.bridge.hibernate.PersistenceExceptionConverter;
 import org.sagebionetworks.bridge.json.BridgeObjectMapper;
 import org.sagebionetworks.bridge.s3.S3Helper;
 import org.sagebionetworks.bridge.upload.DecryptHandler;
@@ -545,9 +547,16 @@ public class BridgeSpringConfig {
         return metadataSources.buildMetadata().buildSessionFactory();
     }
 
+    @Bean(name = "accountPersistenceExceptionConverter")
+    public PersistenceExceptionConverter accountPersistenceExceptionConverter(AccountDao accountDao) {
+        return new AccountPersistenceExceptionConverter(accountDao);
+    }
+    
     @Bean(name = "accountHibernateHelper")
-    public HibernateHelper accountHibernateHelper(SessionFactory sessionFactory) {
-        return new HibernateHelper(sessionFactory, new AccountPersistentExceptionConverter());
+    @Autowired
+    public HibernateHelper accountHibernateHelper(AccountDao accountDao, SessionFactory sessionFactory,
+            PersistenceExceptionConverter converter) {
+        return new HibernateHelper(sessionFactory, converter);
     }
     
     @Bean(name = "sessionExpireInSeconds")
--- a/app/org/sagebionetworks/bridge/hibernate/AccountPersistenceExceptionConverter.java
+++ b/app/org/sagebionetworks/bridge/hibernate/AccountPersistenceExceptionConverter.java
@@ -1,22 +1,27 @@
 package org.sagebionetworks.bridge.hibernate;
 
-import java.util.Map;
-
 import javax.persistence.OptimisticLockException;
 import javax.persistence.PersistenceException;
 
+import org.sagebionetworks.bridge.dao.AccountDao;
 import org.sagebionetworks.bridge.exceptions.ConcurrentModificationException;
 import org.sagebionetworks.bridge.exceptions.ConstraintViolationException;
 import org.sagebionetworks.bridge.exceptions.EntityAlreadyExistsException;
 import org.sagebionetworks.bridge.models.accounts.Account;
+import org.sagebionetworks.bridge.models.accounts.AccountId;
 import org.springframework.stereotype.Component;
 
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
 
 @Component
-public class AccountPersistentExceptionConverter implements PersistenceExceptionConverter {
-
+public class AccountPersistenceExceptionConverter implements PersistenceExceptionConverter {
+    private final AccountDao accountDao;
+    
+    public AccountPersistenceExceptionConverter(AccountDao accountDao) {
+        this.accountDao = accountDao;
+    }
+    
     @Override
     public RuntimeException convert(PersistenceException exception, Object entity) {
         // The sequence of type-checking and unwrapping of this exception is significant as unfortunately, 
@@ -31,25 +36,25 @@ public class AccountPersistentExceptionConverter implements PersistenceException
             String message = cause.getMessage();
             if (message != null && entity != null) {
                 HibernateAccount account = (HibernateAccount)entity;
-                // These are the constraint violation messages. Instead of finding the userId with another 
-                // trip to the database... just include the credential that is conflicting. It makes more 
-                // sense and we're still able to find the pre-existing account with this information.
+                // These are the constraint violation messages. To ensure we don't log credentials, we look
+                // up the existing account and its userId to report in the EntityAlreadyExistsException.
+                // Messages:
                 // "Duplicate entry 'api-email@email.com' for key 'Accounts-StudyId-Email-Index'"
                 // "Duplicate entry 'api-+12064953700' for key 'Accounts-StudyId-Phone-Index'"
                 // "Duplicate entry 'api-ext' for key 'Accounts-StudyId-ExternalId-Index'" 
-                
+                EntityAlreadyExistsException eae = null;
                 if (message.matches("Duplicate entry.*for key 'Accounts-StudyId-ExternalId-Index'")) {
-                    return new EntityAlreadyExistsException(Account.class,
-                            "External ID has already been used by another account.",
-                            getEntityKey("externalId", account.getExternalId()));
+                    eae = createEntityAlreadyExistsException("External ID",
+                            AccountId.forExternalId(account.getStudyId(), account.getExternalId()));
                 } else if (message.matches("Duplicate entry.*for key 'Accounts-StudyId-Email-Index'")) {
-                    return new EntityAlreadyExistsException(Account.class,
-                            "Email address has already been used by another account.",
-                            getEntityKey("email", account.getEmail()));
+                    eae = createEntityAlreadyExistsException("Email address",
+                            AccountId.forEmail(account.getStudyId(), account.getEmail()));
                 } else if (message.matches("Duplicate entry.*for key 'Accounts-StudyId-Phone-Index'")) {
-                    return new EntityAlreadyExistsException(Account.class,
-                            "Phone number has already been used by another account.",
-                            getEntityKey("phone", account.getPhone()));
+                    eae = createEntityAlreadyExistsException("Phone number",
+                            AccountId.forPhone(account.getStudyId(), account.getPhone()));
+                }
+                if (eae != null) {
+                    return eae;
                 }
             }
             ConstraintViolationException.Builder cveBuilder = new ConstraintViolationException.Builder();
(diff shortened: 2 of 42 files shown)


app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -130,7 +136,11 @@ public void verifyChannel(AuthenticationService.ChannelType channelType, Account
                 hibernateAccount.setStatus(AccountStatus.ENABLED);
             }
             hibernateAccount.setModifiedOn(DateUtils.getCurrentMillisFromEpoch());
-            hibernateHelper.update(hibernateAccount);
+            try {
+                hibernateHelper.update(hibernateAccount);    
+            } catch(PersistenceException pe) {
+                throw convertPersistenceException(pe, hibernateAccount.getId());
+            }
```
DwayneJengSage: We add this error handling to a lot of places, but only test create() and update(). Is it worth adding unit tests for:

* verifyChannel()
* changePassword()
* authenticateInternal()
* getAccountAfterAuthentication()
* deleteReauthToken()
* getAccount()

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -379,7 +446,11 @@ public void updateAccount(Account account, boolean allowIdentifierUpdates) {
         accountToUpdate.setModifiedOn(DateUtils.getCurrentMillisFromEpoch());
 
         // Update
-        hibernateHelper.update(accountToUpdate);
+        try {
+            hibernateHelper.update(accountToUpdate);            
+        } catch(PersistenceException pe) {
+            throw convertPersistenceException(pe, persistedAccount.getId());    
+        }
```
DwayneJengSage: If you update an account to add an email/phone/extID, the error message will point to this account and not to the conflicting account. Is this the expected behavior?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: I feel like in terms of mock tests, no. The question is open as to the full set of actual errors that could be thrown here. It appears in all cases, these will be wrapped in a PersistenceException, so I could add some tests of other exceptions that are not currently specifically accounted for, to be sure an acceptable fallback occurs.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: Hmm... no. I will change this.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: I've added some test of unexpected exceptions (wrapped and not wrapped) to verify they are thrown as BridgeServiceExceptions. I have only caught these from update calls, but persistence exceptions could come from other calls. In this case, the exceptions being thrown would differ, so there's an argument to be made that we should just rethrow the exception if we don't want to translate it, rather than wrapping it in BridgeServiceException, which would make the behavior consistent if something like a get() call threw a PersistenceException. BTW I would have told you all of these runtime exceptions were converted by the ExceptionInterceptor to _look_ like BSEs, but it does not. The actually class name is used as the JSON type, etc. though they are all 500 responses.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -248,10 +270,69 @@ public void deleteReauthToken(AccountId accountId) {
...
+                    return new EntityAlreadyExistsException(Account.class, "Phone number has already been used by another account.", entityKeys);
+                }
+            }
+            ConstraintViolationException.Builder cveBuilder = new ConstraintViolationException.Builder();
+            if (message != null) {
+                cveBuilder.withMessage(cause.getMessage());
+            }
+            return cveBuilder.build();
+        }
+        return new BridgeServiceException(exception);
```
alxdarksage: here, I'm now thinking we should just rethrow the error if not translated. Consistence with non-update calls to HibernateHelper.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: See slack about an approach that would cover all HibernateHelper calls and exception translation

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
```diff
@@ -2147,6 +2158,220 @@ public void authenticateAccountEmailUnverifiedWithoutEmailVerificationOK() throw
...
+                "Duplicate entry ABC for key 'Accounts-StudyId-Email-Index'", null, null);
+        PersistenceException ex = new PersistenceException(cve);
+        doThrow(ex).when(mockHibernateHelper).update(any());
+        
+        GenericAccount account = makeValidGenericAccount();
+        account.setId("id");
+        
+        HibernateAccount hibernateAccount = makeValidHibernateAccount(false, false);
+        hibernateAccount.setId("id2");
+        when(mockHibernateHelper.getById(any(), any())).thenReturn(hibernateAccount);
```
DwayneJengSage: This test returns account2 when we're trying to update account1. This would never happen in prod (I hope), but it results in a nonsensical test. It's also probably why the incorrect IDs issue wasn't caught by tests.

This as well as phoneConstraintViolationOnUpdate() and externalIdConstraintViolationOnUpdate().

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/hibernate/HibernateHelper.java
```diff
@@ -5,55 +5,40 @@
...
 import org.hibernate.query.Query;
-import org.springframework.beans.factory.annotation.Autowired;
+import org.sagebionetworks.bridge.exceptions.BridgeServiceException;
 import org.springframework.stereotype.Component;
 
-import org.sagebionetworks.bridge.exceptions.ConcurrentModificationException;
-
 /** Encapsulates common scenarios for using Hibernate to make it easier to use. */
 @Component
 public class HibernateHelper {
```
DwayneJengSage, line 17: You should remove the Component annotation. Otherwise, Spring will try to instantiate a second one of these.

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
```diff
@@ -543,6 +545,11 @@ public SessionFactory hibernateSessionFactory() {
         return metadataSources.buildMetadata().buildSessionFactory();
     }
 
+    @Bean(name = "accountHibernateHelper")
+    public HibernateHelper accountHibernateHelper(SessionFactory sessionFactory) {
+        return new HibernateHelper(sessionFactory, new AccountPersistentExceptionConverter());
+    }
```
DwayneJengSage, line 554: You should take in the AccountPersistentExceptionConverter via autowiring. Otherwise, this creates a second instance of the converter.

app/org/sagebionetworks/bridge/hibernate/AccountPersistentExceptionConverter.java
```diff
@@ -0,0 +1,71 @@
...
+            Throwable cause = Throwables.getRootCause(exception);
+            String message = cause.getMessage();
+            if (message != null && entity != null) {
+                HibernateAccount account = (HibernateAccount)entity;
+                // These are the constraint violation messages. Instead of finding the userId with another 
+                // trip to the database... just include the credential that is conflicting. It makes more 
+                // sense and we're still able to find the pre-existing account with this information.
+                // "Duplicate entry 'api-email@email.com' for key 'Accounts-StudyId-Email-Index'"
+                // "Duplicate entry 'api-+12064953700' for key 'Accounts-StudyId-Phone-Index'"
+                // "Duplicate entry 'api-ext' for key 'Accounts-StudyId-ExternalId-Index'"
```
DwayneJengSage: This means we're logging personal identifying info (email addresses and phone numbers) into the logs. We shouldn't do this.

test/org/sagebionetworks/bridge/hibernate/HibernateHelperTest.java
```diff
@@ -249,8 +231,9 @@ public void execute() {
         when(mockSessionFactory.openSession()).thenReturn(mockSession);
 
         // un-spy HibernateHelper.execute()
-        doCallRealMethod().when(helper).execute(any());
-        helper.setHibernateSessionFactory(mockSessionFactory);
+        // doCallRealMethod().when(helper).execute(any());
+        //helper.setHibernateSessionFactory(mockSessionFactory);
```
DwayneJengSage: Nitpick: Remove commented code.

test/org/sagebionetworks/bridge/hibernate/HibernateHelperTest.java
```diff
@@ -41,14 +42,16 @@
...
+    
     @Before
     public void setup() {
         // mock session
         mockSession = mock(Session.class);
+        mockSessionFactory = mock(SessionFactory.class);
 
         // Spy Hibernate helper. This allows us to mock execute() and test it independently later.
-        helper = spy(new HibernateHelper());
+        helper = spy(new HibernateHelper(mockSessionFactory, new AccountPersistentExceptionConverter()));
```
DwayneJengSage: Would it be better to create a test converter (or a mock converter) instead of taking a hard dependency on AccountPersistentExceptionConverter?

app/org/sagebionetworks/bridge/hibernate/HibernateHelper.java
```diff
@@ -128,14 +113,22 @@ public int queryUpdate(String queryString, Map<String,Object> parameters) {
...
-                return obj;
-            });
-        } catch (OptimisticLockException ex) {
-            throw new ConcurrentModificationException("Row has the wrong version number; it may have been saved in " +
-                    "the background.");
+            return execute(function);
+        } catch(PersistenceException pe) {
+            RuntimeException ex = exceptionConverter.convert(pe, originalEntity);
+            if (ex == pe) {
+                throw new BridgeServiceException(ex);
```
DwayneJengSage, line 126: Test case for when the converter returns the original exception?

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -266,10 +266,21 @@ public IdentifierHolder signUp(Study study, StudyParticipant participant) {
...
+            if (e.getEntityKeys().get("email") != null) {
+                accountId = AccountId.forEmail(study.getIdentifier(), participant.getEmail());
+            } else if (e.getEntityKeys().get("phone") != null) {
+                accountId = AccountId.forPhone(study.getIdentifier(), participant.getPhone());
+            } else if (e.getEntityKeys().get("externalId") != null) {
+                accountId = AccountId.forExternalId(study.getIdentifier(), participant.getExternalId());
+            }
+            if (accountId != null) {
+                accountWorkflowService.notifyAccountExists(study, accountId);
+                LOG.info("Sign up attempt for existing credential '"+accountId+"'");
```
DwayneJengSage: Similarly, this logs personal identifying info.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -266,10 +266,21 @@ public IdentifierHolder signUp(Study study, StudyParticipant participant) {
...
-            LOG.info("Sign up attempt for existing email address in study '"+study.getIdentifier()+"'");
+            // reveals that the credential has been taken.
+            AccountId accountId = null;
+            if (e.getEntityKeys().get("email") != null) {
+                accountId = AccountId.forEmail(study.getIdentifier(), participant.getEmail());
+            } else if (e.getEntityKeys().get("phone") != null) {
+                accountId = AccountId.forPhone(study.getIdentifier(), participant.getPhone());
+            } else if (e.getEntityKeys().get("externalId") != null) {
+                accountId = AccountId.forExternalId(study.getIdentifier(), participant.getExternalId());
+            }
```
DwayneJengSage: Add unit tests for each of the 3 cases, as well as the case where no account is found.

test/org/sagebionetworks/bridge/hibernate/HibernateHelperTest.java
```diff
@@ -270,4 +253,84 @@ public void execute() {
...
+        PersistenceException ex = new PersistenceException(cve);
+        
+        reset(helper); // clear the doAnswer set up in @before 
+        // this isn't exactly when the exception is thrown, but it's close enough to simulate
+        doThrow(ex).when(mockSessionFactory).openSession();
+        
+        HibernateAccount account = new HibernateAccount();
+        account.setStudyId("testStudy");
+
+        helper.update(account);
```
DwayneJengSage: This test is called queryUpdate(), but it actually calls update().

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/hibernate/HibernateHelper.java
alxdarksage: Removed

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
alxdarksage: Okay, changed that and autowired it.

test/org/sagebionetworks/bridge/hibernate/HibernateHelperTest.java
alxdarksage: Removed

test/org/sagebionetworks/bridge/hibernate/HibernateHelperTest.java
alxdarksage: Created a simple passthrough converter under TestConstants and used that instead.

app/org/sagebionetworks/bridge/hibernate/HibernateHelper.java
alxdarksage: Added

app/org/sagebionetworks/bridge/hibernate/AccountPersistentExceptionConverter.java
alxdarksage: Okay, I've injected AccountDao into the converter so we can look up the account that is conflicting, and include its ID in this exception. This prevents any chance of a credential getting logged down the line.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage: Changed how this works so it is only returning userId now.

app/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage: Changed this to look for userId only, covered under the existing test.

test/org/sagebionetworks/bridge/hibernate/HibernateHelperTest.java
alxdarksage: Fixed

app/org/sagebionetworks/bridge/hibernate/AccountPersistenceExceptionConverter.java
```diff
@@ -0,0 +1,79 @@
...
+import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableMap;
+
+@Component
+public class AccountPersistenceExceptionConverter implements PersistenceExceptionConverter {
+    private final AccountDao accountDao;
+    
+    public AccountPersistenceExceptionConverter(AccountDao accountDao) {
+        this.accountDao = accountDao;
+    }
```
DwayneJengSage, line 25: Does the accountDao need to be autowired?

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
```diff
@@ -543,6 +547,18 @@ public SessionFactory hibernateSessionFactory() {
         return metadataSources.buildMetadata().buildSessionFactory();
     }
 
+    @Bean(name = "accountPersistenceExceptionConverter")
+    public PersistenceExceptionConverter accountPersistenceExceptionConverter(AccountDao accountDao) {
+        return new AccountPersistenceExceptionConverter(accountDao);
+    }
```
DwayneJengSage: It would be better to use the Component annotation over the Bean annotation to reduce the amount of boilerplate we have to write.

Also, since you use both, Spring will create 2 instances of AccountPersistenceExceptionConverter.

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
```diff
@@ -543,6 +547,18 @@ public SessionFactory hibernateSessionFactory() {
...
+    
+    @Bean(name = "accountHibernateHelper")
+    @Autowired
+    public HibernateHelper accountHibernateHelper(AccountDao accountDao, SessionFactory sessionFactory,
+            PersistenceExceptionConverter converter) {
```
DwayneJengSage: Is it possible to reference AccountPersistenceExceptionConverter specifically instead of PersistenceExceptionConverter? Thinking ahead, we'll want to reference that bean unambiguously so when we migrate shared modules to using HibernateHelper, there will be no ambiguity.

app/org/sagebionetworks/bridge/hibernate/HibernateHelper.java
```diff
@@ -128,14 +111,22 @@ public int queryUpdate(String queryString, Map<String,Object> parameters) {
 
     /** Updates a single object. */
     public <T> T update(T obj) {
+        return executeWithExceptionHandling(obj, session -> {
+            session.update(obj);
+            return obj;
+        });
```
DwayneJengSage, line 117: Unit test for update converts exception?

app/org/sagebionetworks/bridge/hibernate/AccountPersistenceExceptionConverter.java
```diff
@@ -0,0 +1,79 @@
...
+                } else if (message.matches("Duplicate entry.*for key 'Accounts-StudyId-Email-Index'")) {
+                    eae = createEntityAlreadyExistsException("Email address",
+                            AccountId.forEmail(account.getStudyId(), account.getEmail()));
+                } else if (message.matches("Duplicate entry.*for key 'Accounts-StudyId-Phone-Index'")) {
+                    eae = createEntityAlreadyExistsException("Phone number",
+                            AccountId.forPhone(account.getStudyId(), account.getPhone()));
+                }
+                if (eae != null) {
+                    return eae;
+                }
```
DwayneJengSage, line 60: The original exception may include the email address or phone number, which may get logged. I recommend converting it to a different exception with a generic message instead.

test/org/sagebionetworks/bridge/hibernate/HibernateHelperTest.java
```diff
@@ -41,14 +42,16 @@
...
+    
     @Before
     public void setup() {
         // mock session
         mockSession = mock(Session.class);
+        mockSessionFactory = mock(SessionFactory.class);
 
         // Spy Hibernate helper. This allows us to mock execute() and test it independently later.
-        helper = spy(new HibernateHelper());
+        helper = spy(new HibernateHelper(mockSessionFactory, TestConstants.PASSTHROUGH_PERSISTENCE_CONVERTER));
```
DwayneJengSage: Now we can't tell the difference between the converter not being called vs the converter returning the original exception, nor can we tell the difference between the converter returning the original exception vs the converter converting the exception.

Maybe a simple a mock would be better, so we can control it in our test.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/hibernate/AccountPersistenceExceptionConverter.java
alxdarksage: Constraint exceptions now have a generic error message.

test/org/sagebionetworks/bridge/hibernate/HibernateHelperTest.java
alxdarksage: Switched to mock

app/org/sagebionetworks/bridge/hibernate/AccountPersistenceExceptionConverter.java
alxdarksage: Added the annotation (all the integration tests pass... so apparently it doesn't, but still, that's consistent).

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
alxdarksage: AccountPersistenceExceptionConverter is now created through a @Component annotation and accountHibernateHelper is created in BridgeSpringConfig, where we can wire it up with other persistenceExceptionConverts under different names.

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
alxdarksage: Changed this.

app/org/sagebionetworks/bridge/hibernate/HibernateHelper.java
alxdarksage: Added

test/org/sagebionetworks/bridge/hibernate/HibernateHelperTest.java
```diff
@@ -270,4 +261,180 @@ public void execute() {
...
+        doThrow(iae).when(mockSessionFactory).openSession();
+        
+        HibernateAccount account = new HibernateAccount();
+        account.setStudyId("testStudy");
+        try {
+            helper.update(account);
+            fail("Should have thrown exception");
+        } catch(RuntimeException e) {
+            assertSame(iae, e);
+        }
```
DwayneJengSage, line 443: This test doesn't test what you think it does. executeWithExceptionHandling() only catches PersistenceExceptions, so the IllegalArgumentException is uncaught and being throw all the way up to the test, bypassing the code we're trying to test.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #1787, opened 2018-10-05
clean up S3 when cleaning up study consents

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2312

When deleting a subpop (or a study, which deletes all subpops), we'll delete the entries from DDB, but not from S3. This is a problem, because we suspect all these extra files in S3 is causing a slowdown.

This change refactors deleting study consents to also delete the S3 files.

Testing done:
* Updated unit tests
* Ran integ tests and verified S3 buckets are clean afterward

### DIFF
--- a/app/org/sagebionetworks/bridge/services/SubpopulationService.java
+++ b/app/org/sagebionetworks/bridge/services/SubpopulationService.java
@@ -221,7 +221,8 @@ public class SubpopulationService {
     public void deleteSubpopulationPermanently(StudyIdentifier studyId, SubpopulationGuid subpopGuid) {
         checkNotNull(studyId);
         checkNotNull(subpopGuid);
-        
+
+        studyConsentService.deleteAllConsentsPermanently(subpopGuid);
         subpopDao.deleteSubpopulationPermanently(studyId, subpopGuid);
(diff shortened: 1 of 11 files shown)


app/org/sagebionetworks/bridge/services/SubpopulationService.java
```diff
@@ -221,7 +221,8 @@ public void deleteSubpopulation(StudyIdentifier studyId, SubpopulationGuid subpo
     public void deleteSubpopulationPermanently(StudyIdentifier studyId, SubpopulationGuid subpopGuid) {
         checkNotNull(studyId);
         checkNotNull(subpopGuid);
-        
+
+        studyConsentService.deleteAllConsentsPermanently(subpopGuid);
```
alxdarksage: If you do this first, there will be a tiny window where loading the subpopulation might throw an error for lack of a consent... ? Do it after?

alxdarksage (approved):
Just one thing to consider

app/org/sagebionetworks/bridge/services/SubpopulationService.java
DwayneJengSage: Good point. Changed.

PR #1789, opened 2018-10-17
SMS Message Logs

DwayneJengSage:
Initial prototype. Still work in progress. Unit tests pass, but this hasn't been integ tested, nor has it been tested thoroughly with a real Twilio back-end.


app/org/sagebionetworks/bridge/dynamodb/DynamoSmsMessage.java
```diff
@@ -0,0 +1,101 @@
...
+    @Override
+    public String getNumber() {
+        return number;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void setNumber(String number) {
+        this.number = number;
+    }
```
alxdarksage, line 36: Is this the phone number? Can we call it phone or phoneNumber for clarity?

app/org/sagebionetworks/bridge/dynamodb/DynamoSmsOptOutSettings.java
```diff
@@ -0,0 +1,73 @@
...
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBHashKey;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBTable;
+
+import org.sagebionetworks.bridge.models.sms.SmsOptOutSettings;
+
+@DynamoThroughput(readCapacity=1, writeCapacity=1)
+@DynamoDBTable(tableName = "SmsOptOut")
+public class DynamoSmsOptOutSettings implements SmsOptOutSettings {
+    private String number;
+    private boolean globalPromotionalOptOut;
```
alxdarksage: FWIW I learned that if you annotate boolean fields with @DynamoDBTyped(DynamoDBAttributeType.BOOL), they will be saved as booleans in DDB and not as ints. The advantage of that is that our IndexHelper class can correctly deserialize these records from Item objects, using BridgeObjectMapper.

app/org/sagebionetworks/bridge/dynamodb/DynamoSmsOptOutSettingsDao.java
```diff
@@ -0,0 +1,35 @@
...
+
+    /** DynamoDB mapper for the opt-out settings table. */
+    @Resource(name = "smsOptOutDdbMapper")
+    public final void setMapper(DynamoDBMapper mapper) {
+        this.mapper = mapper;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public SmsOptOutSettings getOptOutSettings(String phoneNumber) {
```
alxdarksage: Personally I would convert from Phone --> String (of the phone in E.164 format) as far down the tiers as possible, maybe here at the DAO level.

app/org/sagebionetworks/bridge/models/sms/SmsOptOutSettings.java
```diff
@@ -0,0 +1,56 @@
...
+    boolean getGlobalPromotionalOptOut();
+
+    /** @see #getGlobalPromotionalOptOut */
+    void setGlobalPromotionalOptOut(boolean globalPromotionalOptOut);
+
+    /**
+     * Promotional opt-outs per study. This takes precedence over global promotional opt-outs. That is, if a user is
+     * opted out globally, but opted into study Foo, then the user is opted into study Foo.
+     */
+    Map<String, Boolean> getPromotionalOptOuts();
```
alxdarksage: just checking this is Map<StudyId,OptOutStatus>?

conf/routes
```diff
@@ -91,6 +94,9 @@ POST   /v3/participants/self/identifiers @org.sagebionetworks.bridge.play.contro
 GET    /v3/participants/:userId                                        @org.sagebionetworks.bridge.play.controllers.ParticipantController.getParticipant(userId: String, consents: Boolean ?= true)
 POST   /v3/participants/:userId                                        @org.sagebionetworks.bridge.play.controllers.ParticipantController.updateParticipant(userId: String)
 DELETE /v3/participants/:userId                                        @org.sagebionetworks.bridge.play.controllers.ParticipantController.deleteTestParticipant(userId: String)
+GET    /v3/participants/:userId/sms/optout                             @org.sagebionetworks.bridge.play.controllers.SmsController.getOptOutSettings(userId: String)
+POST   /v3/participants/:userId/sms/optout                             @org.sagebionetworks.bridge.play.controllers.SmsController.setOptOutSettings(userId: String)
+GET    /v3/participants/:userId/sms/recent                             @org.sagebionetworks.bridge.play.controllers.SmsController.getMostRecentMessage(userId: String)
```
alxdarksage, line 94: Because the opt-out setting is scoped to this user in this study, 1:1, I think the better approach would be to add an SMS opt-out flag to the StudyParticipant, similar to the notifyByEmail flag. Even if internally, that data is saved in the SmsOptOutSettings table. Otherwise, each bit of data related to a user starts to have its own API, we have to call multiple APIs to show the general status of a user, etc. (I think the permission of the GET method here is ADMIN, RESEARCHER, unless this bit is not included in a study manager.) We went down this road a little bit with ParticipantOptions and it turned into a hassle (a change sharing status API, a change data groups API, etc.).

The most recent SMS message would still be GET /v3/participants/:userId/sms/recent as it's a separate thing, always scoped to the user. Is it just ADMIN or is it ADMIN/RESEARCHER?

test/org/sagebionetworks/bridge/sms/TwilioHelperTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.mockito.ArgumentCaptor;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+import org.sagebionetworks.bridge.config.BridgeConfig;
+import org.sagebionetworks.bridge.models.accounts.Phone;
+
+@PrepareForTest({Message.class, Twilio.class})
+@RunWith(PowerMockRunner.class)
+public class TwilioHelperTest {
```
alxdarksage: I like PowerMock and I'm glad to see it make an appearance in BridgePF!

alxdarksage (changes requested):

app/org/sagebionetworks/bridge/dynamodb/DynamoSmsOptOutSettingsDao.java
DwayneJengSage: In some code paths (such as the Twilio Webhook) we don't have a Phone object. In those cases, would you prefer that I create a Phone object (and assume region code US), or would you prefer that I just pass the String phone number?

app/org/sagebionetworks/bridge/models/sms/SmsOptOutSettings.java
DwayneJengSage: Yes, this is a map from study ID to opt-out status.

conf/routes
DwayneJengSage: I think opt-out status should be a separate entity because (a) we may or may not be required to track global opt-outs (b) if a user changes their phone number or deletes their phone number (such as a withdrawal), we may or may not be required to remember the opt-outs for that phone number (c) researchers aren't allowed to change a user's opt-out status.

The reason why these APIs in particular use userId and not phone number is because we log URLs, so the URL can't contain a phone number.

For the get API, it's currently scoped to admins, because it's intended to be used only by integration tests. I don't know if researchers currently want this (at least not in this format, and not in the Bridge Study Manager), so I didn't make the API available to researchers.

app/org/sagebionetworks/bridge/dynamodb/DynamoSmsMessage.java
DwayneJengSage: Changed this to phoneNumber in both SmsMessage and SmsOptOutSettings.

app/org/sagebionetworks/bridge/dynamodb/DynamoSmsOptOutSettingsDao.java
alxdarksage: No. If it's the case that the string and Phone object code paths converge at this DAO, just leave it as is. Thanks!

conf/routes
alxdarksage: Hmm. If these are tied to phone number, maybe just /v3/smsoptouts ? Put the phone number in the body message? We can live with this API, but if we need to remember these settings even if we delete the phone in an account, or delete an account entirely, it's not even related to a user account. Yes we would normally do POST /v3/smsoptouts/:phoneNumber but we can skip that for security purposes.

alxdarksage: I do also wonder how we'll answer the question "why isn't this user seeing SMS messages?" We would want people to be able to look in the UI and see the user's opt out status, even if it's not changeable. If they haven't opted out, then we would assume there's a problem and investigate further, I believe.

DwayneJengSage: Unfortunately, you can't use a request body with an HTTP GET. (Well, you can, but it's considered an anti-pattern, and some tools (like Postman) don't support it.)

As for how researchers can view opt-outs, there's already a separate sprint story for this.

alxdarksage: What will we do if the account is deleted? Are we just off the hook to persist opt out status at that point? (I guess it doesn't matter... the record is there and if they ever re-enroll, we would know the opt out setting.)

DwayneJengSage: These APIs are only used by integration tests. The actual opt-out handling comes in through the Twilio webhook.

alxdarksage (approved):

alxdarksage:
Approved.

app/org/sagebionetworks/bridge/dynamodb/DynamoSmsOptOutSettings.java
DwayneJengSage: Added.

DwayneJengSage:
Pull request appears to be stuck and unable to be updated. Closing and re-opening a new one.

DwayneJengSage:
Github was down last night. Re-opening to see if this gets this PR back in a reasonable state.

alxdarksage (approved):

app/org/sagebionetworks/bridge/play/controllers/SmsController.java
```diff
@@ -0,0 +1,33 @@
...
+    @Autowired
+    public final void setSmsService(SmsService smsService) {
+        this.smsService = smsService;
+    }
+
+    /** Returns the most recent message sent to the phone number of the given user. Used by integration tests. */
+    public Result getMostRecentMessage(String userId) {
+        UserSession session = getAuthenticatedSession(Roles.ADMIN);
+        Study study = studyService.getStudy(session.getStudyIdentifier());
+        SmsMessage message = smsService.getMostRecentMessage(study, userId);
```
alxdarksage: I feel this would be cleaner if it went to ParticipantService, retrieved the account, and then called an SmsService that expects Accounts? Then the SmsService wouldn't need to reference ParticipantService or know about all that. E.g. SmsService. getMostRecentMessage() could just take a phone number at that point.

app/org/sagebionetworks/bridge/services/NotificationsService.java
```diff
@@ -250,18 +251,8 @@ public void deleteRegistration(StudyIdentifier studyId, String healthCode, Strin
...
-        if (request.getMessage().getBytes(Charset.forName("US-ASCII")).length > SMS_CHARACTER_LIMIT) {
-            throw new BridgeServiceException("SMS message cannot be longer than 140 UTF-8/ASCII characters.");
-        }
-        
-        PublishResult result = snsClient.publish(request);
-
-        LOG.debug("Sent SMS message, study=" + provider.getStudy().getStudyIdentifier().getIdentifier()
-                + ", message ID=" + result.getMessageId());
+        // This is now a call-through to SmsService.
+        smsService.sendSmsMessage(provider);
```
alxdarksage, line 255: I'm fine with ripping the bandaid off and rewriting call locations to use smsService directly. It's called from ParticipantService which has all the other service references, since it frequently loads an Account, then calls on to the appropriate service.

alxdarksage (changes requested):

app/org/sagebionetworks/bridge/play/controllers/SmsController.java
DwayneJengSage: Changed.

app/org/sagebionetworks/bridge/services/NotificationsService.java
DwayneJengSage: As per IM discussion, I'd prefer unnecessary code churn in an already large change list. I'd prefer to refactor this out of NotificationService at a later date.

alxdarksage (approved):

PR #1790, opened 2018-10-19
App config elements

alxdarksage:
Individual JSON elements that are versioned, publishable, and accessible via the API. This covers the CRUD cases for these elements, code to include in the app config file, and to create custom editors for some fragments like medication lists, is all forthcoming.

BRIDGE-2333


app/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigElementDao.java
```diff
@@ -0,0 +1,152 @@
...
+    
+    @Resource(name = "appConfigElementDdbMapper")
+    final void setMapper(DynamoDBMapper mapper) {
+        this.mapper = mapper;
+    }
+    
+    @Resource(name = "appConfigElementStudyIndex")
+    final void setIndexHelper(DynamoIndexHelper indexHelper) {
+        this.indexHelper = indexHelper;
+    }
```
DwayneJengSage: IndexHelper is being deprecated, because it returns results in an undefined order, and because it uses Jackson to convert between the Document API and the Mapper API, which is prone to problems if the JSON representation differs from the DDB representation (for example, for timestamps). See https://sagebionetworks.jira.com/browse/BRIDGE-1467

You can query the index directly using a mapper. For an example of how to populate all fields from the results of a GSI query, see https://github.com/Sage-Bionetworks/BridgePF/blob/develop/app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchemaDao.java#L165

app/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigElementDao.java
```diff
@@ -0,0 +1,152 @@
...
+import com.amazonaws.services.dynamodbv2.model.Condition;
+import com.amazonaws.services.dynamodbv2.model.ConditionalCheckFailedException;
+import com.google.common.collect.Maps;
+import com.newrelic.agent.deps.com.google.common.collect.Lists;
+
+/**
+ * The index "studyId-index" must be a GSI that includes the fields key, revision, studyId, 
+ * id, and deleted. Our DynamoDB introspection and table creation code does not currently support
+ * configuration of the fields included in a GSI, this needs to be created manually.
+ */
```
DwayneJengSage: I wouldn't recommend this. We only put keys in a GSI because additional fields means (a) more DDB storage and (b) more DDB writes. Also, we want to avoid manual bootstrap steps where possible, as this makes bootstrapping more difficult.

Instead, I'd recommend querying the index, and then re-querying the table for the rest of the fields. Reads are cheap. See my other comment for an example.

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
```diff
@@ -331,6 +332,19 @@ public DynamoDBMapper appConfigDdbMapper(DynamoUtils dynamoUtils) {
...
+    public DynamoDBMapper appConfigElementDdbMapper(DynamoUtils dynamoUtils) {
+        return dynamoUtils.getMapper(DynamoAppConfigElement.class);
+    }
+    
+    @Bean(name = "appConfigElementStudyIndex")
+    @Autowired
+    public DynamoIndexHelper appConfigElementStudyIndex(AmazonDynamoDBClient dynamoDBClient, DynamoUtils dynamoUtils,
+            DynamoNamingHelper dynamoNamingHelper) {
+        return DynamoIndexHelper.create(DynamoAppConfigElement.class, "studyId-index", dynamoDBClient, dynamoNamingHelper, dynamoUtils);
+    }
```
DwayneJengSage: Would not recommend using the IndexHelper. See other comments.

app/org/sagebionetworks/bridge/services/AppConfigElementService.java
```diff
@@ -0,0 +1,177 @@
...
+    
+    @Autowired
+    final void setAppConfigElementDao(AppConfigElementDao appConfigElementDao) {
+        this.appConfigElementDao = appConfigElementDao;
+    }
+    
+    // In order to mock time for tests
+    DateTime getDateTime() {
+        return DateTime.now();
+    }
```
DwayneJengSage: This is unnecessary. Use DateTimeUtils.setCurrentMillisFixed() instead.

app/org/sagebionetworks/bridge/services/AppConfigElementService.java
```diff
@@ -0,0 +1,177 @@
...
+        
+        Validate.entityThrowingException(AppConfigElementValidator.INSTANCE, element);
+        
+        AppConfigElement existing = getElementRevision(studyId, element.getId(), element.getRevision());
+        if (element.isDeleted() && existing.isDeleted()) {
+            throw new EntityNotFoundException(AppConfigElement.class);
+        }
+        if (existing.isPublished()) {
+            throw new EntityPublishedException("App config element cannot be changed, it is published.");
+        }
```
DwayneJengSage: I would not recommend a publish workflow for this, especially not a one-way publish workflow. In practice, developers will create a revision, publish it immediately to test with, and then create a new revision, resulting in a myriad of unused revisions. This is the same reason we ended up removing the publish workflow from schemas.

It would be better to just always allow elements to be mutable.

app/org/sagebionetworks/bridge/services/AppConfigElementService.java
```diff
@@ -0,0 +1,177 @@
...
+        checkNotNull(studyId);
+        checkNotNull(element);
+        
+        if (element.getRevision() == null) {
+            element.setRevision(1L);
+        }
+        // Validate that ID exists before you try and use it to set the key
+        Validate.entityThrowingException(AppConfigElementValidator.INSTANCE, element);
+        
+        element.setKey(studyId.getIdentifier() + ":" + element.getId());
```
DwayneJengSage: I wouldn't recommend this. Having a separate key in the element that we have to remember to set it in every service create or update and in a bunch of DAO getters. This is error prone and could lead to bugs if we forget to do this. I'd recommend having the element itself enforce the key formatting. See for example https://github.com/Sage-Bionetworks/BridgePF/blob/develop/app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchema.java#L82

app/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigElementDao.java
```diff
@@ -0,0 +1,152 @@
...
+
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBQueryExpression;
+import com.amazonaws.services.dynamodbv2.datamodeling.PaginatedQueryList;
+import com.amazonaws.services.dynamodbv2.model.AttributeValue;
+import com.amazonaws.services.dynamodbv2.model.ComparisonOperator;
+import com.amazonaws.services.dynamodbv2.model.Condition;
+import com.amazonaws.services.dynamodbv2.model.ConditionalCheckFailedException;
+import com.google.common.collect.Maps;
+import com.newrelic.agent.deps.com.google.common.collect.Lists;
```
DwayneJengSage: Nitpick: newrelic

(Is there anyway to exclude these from our IDE?)

app/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigElementDao.java
```diff
@@ -0,0 +1,152 @@
...
+        
+        // This only retrieves the keys of these elements and the deleted flag, which is included in the GSI
+        List<DynamoAppConfigElement> elements = indexHelper.queryKeys(DynamoAppConfigElement.class, "studyId",
+                studyId.getIdentifier(), null);
+        
+        // Looking for the highest revision of each ID
+        Map<String,Long> versionMap = Maps.newHashMap();
+        for (DynamoAppConfigElement oneElement : elements) {
+            if (includeDeleted || !oneElement.isDeleted()) {
+                String elementKey = studyId.getIdentifier() + ":" + oneElement.getId();
```
DwayneJengSage: Nitpick: AppConfig IDs are unique per study, so pre-pending the studyID to the hash key is unnecessary.

test/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigElementTest.java
```diff
@@ -0,0 +1,67 @@
...
+        assertNull(node.get("key"));
+        assertNull(node.get("studyId"));
+        assertEquals(1L, node.get("revision").longValue());
+        assertEquals("id", node.get("id").textValue());
+        assertTrue(node.get("published").booleanValue());
+        assertTrue(node.get("deleted").booleanValue());
+        assertEquals(TestUtils.getClientData(), node.get("data"));
+        assertEquals(CREATED_ON.toString(), node.get("createdOn").textValue());
+        assertEquals(MODIFIED_ON.toString(), node.get("modifiedOn").textValue());
+        assertEquals(2L, node.get("version").longValue());
```
DwayneJengSage, line 96: Type attribute?

conf/routes
```diff
@@ -243,6 +243,17 @@ POST   /v3/topics/:guid                  @org.sagebionetworks.bridge.play.contro
...
+DELETE /v3/appconfigs/elements/:id                             @org.sagebionetworks.bridge.play.controllers.AppConfigElementsController.deleteElementAllRevisions(id: String, physical: String ?= "false")
+GET    /v3/appconfigs/elements/:id/published                   @org.sagebionetworks.bridge.play.controllers.AppConfigElementsController.getMostRecentlyPublishedElement(id: String)
+GET    /v3/appconfigs/elements/:id/revisions/:revision         @org.sagebionetworks.bridge.play.controllers.AppConfigElementsController.getElementRevision(id: String, revision: String)
+POST   /v3/appconfigs/elements/:id/revisions/:revision         @org.sagebionetworks.bridge.play.controllers.AppConfigElementsController.updateElementRevision(id: String, revision: String)
+POST   /v3/appconfigs/elements/:id/revisions/:revision/publish @org.sagebionetworks.bridge.play.controllers.AppConfigElementsController.publishElementRevision(id: String, revision: String)
+DELETE /v3/appconfigs/elements/:id/revisions/:revision         @org.sagebionetworks.bridge.play.controllers.AppConfigElementsController.deleteElementRevision(id : String, revision: String, physical: String ?= "false")
```
DwayneJengSage, line 254: What happens if you define revision as an Int instead of String? UploadSchema APIs currently do this.

app/org/sagebionetworks/bridge/play/controllers/AppConfigElementsController.java
```diff
@@ -0,0 +1,122 @@
...
+    public Result getMostRecentlyPublishedElement(String id) {
+        UserSession session = getAuthenticatedSession(DEVELOPER);
+        
+        AppConfigElement element = service.getMostRecentlyPublishedElement(session.getStudyIdentifier(), id);
+        return okResult(element);
+    }
+
+    public Result getElementRevision(String id, String revisionString) {
+        UserSession session = getAuthenticatedSession(DEVELOPER);
+        Long revision = BridgeUtils.getLongOrDefault(revisionString, 0L);
```
DwayneJengSage: Should we be converting this to a default? Revision 0 is not a valid revision. Should we just throw in this case? Or pass it onto the service and have the service validate?

Do we need a test case for unspecified revision?

Similarly in updateElementRevision(), publishElementRevision(), and deleteElementRevision().

app/org/sagebionetworks/bridge/play/controllers/AppConfigElementsController.java
```diff
@@ -0,0 +1,122 @@
...
+        return okResult(element);
+    }
+
+    public Result updateElementRevision(String id, String revisionString) {
+        UserSession session = getAuthenticatedSession(DEVELOPER);
+        Long revision = BridgeUtils.getLongOrDefault(revisionString, 0L);
+        
+        AppConfigElement element = parseJson(request(), AppConfigElement.class);
+        element.setId(id);
+        element.setRevision(revision);
```
DwayneJengSage, line 89: Should this logic be in the controller or the service?

test/org/sagebionetworks/bridge/services/AppConfigElementServiceTest.java
```diff
@@ -0,0 +1,368 @@
...
+        
+        // These have been correctly reset
+        assertEquals(new Long(1), elementCaptor.getValue().getRevision());
+        AppConfigElement captured = elementCaptor.getValue();
+        assertNull(captured.getVersion());
+        assertFalse(captured.isDeleted());
+        assertEquals("api", captured.getStudyId());
+        assertEquals("api:id", captured.getKey());
+        assertEquals(TIMESTAMP.getMillis(), captured.getCreatedOn());
+        assertEquals(TIMESTAMP.getMillis(), captured.getModifiedOn());
```
DwayneJengSage, line 107: TIMESTAMP is DateTime.now() and TestUtils creates these timestamps with DateTime.now(). We should consider using different values to ensure that the test actually works. (Similarly in updateElementRevision(), publishElementRevision(), deleteElementRevision(), deleteElementAllRevisions().)

app/org/sagebionetworks/bridge/services/AppConfigElementService.java
```diff
@@ -0,0 +1,177 @@
...
+        return appConfigElementDao.getMostRecentElements(studyId, includeDeleted);
+    }
+    
+    public VersionHolder createElement(StudyIdentifier studyId, AppConfigElement element) {
+        checkNotNull(studyId);
+        checkNotNull(element);
+        
+        if (element.getRevision() == null) {
+            element.setRevision(1L);
+        }
```
DwayneJengSage, line 43: Do we want a test case that you can create an element with any revision?

test/org/sagebionetworks/bridge/services/AppConfigElementServiceTest.java
```diff
@@ -0,0 +1,368 @@
...
+        element.setPublished(true);
+        when(dao.getElementRevision(TEST_STUDY, element.getId(), element.getRevision())).thenReturn(element);
+        
+        service.publishElementRevision(TEST_STUDY, "id", 3L);
+    }
+    
+    @Test
+    public void deleteElementAllRevisions() {
+        when(service.getDateTime()).thenReturn(TIMESTAMP);
+        when(dao.getElementRevisions(TEST_STUDY, "id", false)).thenReturn(ELEMENTS);
```
DwayneJengSage: This test modifies the app config elements in the static member ELEMENTS. If this isn't the last test that runs, this could cause weird things to happen if another test tries to use ELEMENTS. Recommend creating clean elements instead. (Similarly in deleteElementAllRevisionsPermanently().)

DwayneJengSage (changes requested):

DwayneJengSage:
Also, how does the new AppConfig stuff interact with the old AppConfig stuff and Criteria?

conf/routes
alxdarksage: I believe if you provide a non-int, Play will return an error during parsing, which will not be our standard BadRequestException. By moving the parsing into the controller method, we get more consistent error response payloads. But I'll test this again.

app/org/sagebionetworks/bridge/play/controllers/AppConfigElementsController.java
alxdarksage: Controller, because we're enforcing the primacy of the URL over anything in the body of the message, and this is HTTP-related.

app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
alxdarksage: Removed.

app/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigElementDao.java
alxdarksage: Removed (I also Googled and found that the type filters are applied to the import support in Eclipse, so added new relic deps to this... it works.)

app/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigElementDao.java
alxdarksage: Changed to this approach, so the comment no longer true; removed. Deleted index and re-ran tests, it's created by our code and works.

app/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigElementDao.java
alxdarksage: Changed this

app/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigElementDao.java
alxdarksage: The app config element key is a studyId + identifier combination. This makes it possible to use identical ids (they are user assigned) in multiple studies without a conflict.

app/org/sagebionetworks/bridge/play/controllers/AppConfigElementsController.java
alxdarksage: Currently would return an InvaldEntityException because the number isn't valid, but the message would be confusing. Using null, the only way you can really get an exception is if you use a non-numerical value to match the route, e.g. /v3/appconfigs/elements/:id/revisions/NOT_A_NUMBER. Switched to null, and throwing bad request exception if this method returns null.

app/org/sagebionetworks/bridge/services/AppConfigElementService.java
alxdarksage: Removed.

app/org/sagebionetworks/bridge/services/AppConfigElementService.java
alxdarksage: As I understand the schema revision system (which I based this on for consistency), you can do this, so yes, test added.

app/org/sagebionetworks/bridge/services/AppConfigElementService.java
alxdarksage: Added a setter that takes the component parts instead.

app/org/sagebionetworks/bridge/services/AppConfigElementService.java
alxdarksage: Removed publication as discussed.

conf/routes
alxdarksage: Errors are handled not by the ExceptionInterceptor, but by a Play-specific class called the BridgeErrorHandler. Right now, it doesn't report numeric parsing errors correctly. While we could fix this, as we discussed, that's putting more code into the Play part of the system, whereas the exception interceptor is Spring-specific, centralizes our error handling, and would port to something like Spring MVC. So keeping the parsing within the controller.

test/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigElementTest.java
alxdarksage: Added

test/org/sagebionetworks/bridge/services/AppConfigElementServiceTest.java
alxdarksage: Changed it so the values are different.

test/org/sagebionetworks/bridge/services/AppConfigElementServiceTest.java
alxdarksage: Changed this to reinitialize after every test.

alxdarksage (commented):
Updated

app/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigElement.java
```diff
@@ -0,0 +1,131 @@
...
+    public String getKey() {
+        return key;
+    }
+    public void setKey(String key) {
+        this.key = key;
+    }
+    @DynamoDBIgnore
+    public void setKey(StudyIdentifier studyId, String id) {
+        this.setKey(studyId.getIdentifier() + ":" + id);
+    }
```
DwayneJengSage, line 54: I still wouldn't recommend this. If someone sets the studyId and Id but forgets to set the key, the app config element could get into a broken state.

Instead of having a separate key member variable and relying on callers to maintain the constraints, it would be better to not have a "key" and having the key being generated on the fly from the studyId and Id. See https://github.com/Sage-Bionetworks/BridgePF/blob/develop/app/org/sagebionetworks/bridge/dynamodb/DynamoUploadSchema.java#L82 for example.

Also, we should have unit tests that test the key is being generated correctly.

test/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigElementDaoTest.java
```diff
@@ -0,0 +1,281 @@
...
+    
+    @SuppressWarnings("unchecked")
+    @Test
+    public void getMostRecentElementsIncludeDeleted() {
+        when(mockMapper.batchLoad(any(List.class))).thenReturn(appConfigElementMapId1And2());
+        
+        List<AppConfigElement> returned = dao.getMostRecentElements(TestConstants.TEST_STUDY, true);
+        assertEquals(2, returned.size());
+        assertIdAndRevision(returned.get(0), ID_1, 3L);
+        assertIdAndRevision(returned.get(1), ID_2, 3L);
```
DwayneJengSage, line 73: Test should verify the DDB query. Also, we should probably return something "real" from the DDB query and pass it into the batchLoad() to verify that the data flow is working correctly.

test/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigElementDaoTest.java
```diff
@@ -0,0 +1,281 @@
...
+    }
+    
+    @Mock
+    private PaginatedQueryList<DynamoAppConfigElement> mockQueryList;
```
DwayneJengSage: Nitpick: Is this mock being used?

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigElement.java
alxdarksage: Removed key field; added tests that it can be generated and parsed into the studyId and id

test/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigElementDaoTest.java
alxdarksage: Verified query, verified the results returned from batchLoad are realistic for that query.

DwayneJengSage (approved):

PR #1791, opened 2018-10-25
Config elements in the app config

alxdarksage:


test/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigTest.java
```diff
@@ -66,12 +71,15 @@ public void canSerialize() throws Exception {
...
+        appConfig.setConfigReferences(CONFIG_REFS);
+        appConfig.setConfigElements(ImmutableMap.of("config1", TestUtils.getClientData()));
         appConfig.setClientData(clientData);
         appConfig.setVersion(3L);
         appConfig.setDeleted(true);
         
         Set<String> fields = Sets.newHashSet("criteria", "label", "createdOn", "modifiedOn", "clientData",
-                "surveyReferences", "schemaReferences", "version", "type", "deleted");
+                "surveyReferences", "schemaReferences", "configReferences", "configElements", "version", 
+                "type", "deleted");
```
DwayneJengSage: Should this test also verify that the JSON representation looks the way we expect it to?

app/org/sagebionetworks/bridge/dynamodb/DynamoAppConfig.java
```diff
@@ -144,11 +164,41 @@ public void setSurveyReferences(List<SurveyReference> references) {
...
         this.schemaReferences = references;
     }
     
+    @DynamoDBTypeConverted(converter=ConfigReferenceListMarshaller.class)
+    @Override
+    public List<ConfigReference> getConfigReferences() {
+        if (configReferences == null) {
+            configReferences = new ArrayList<>();
+        }
+        return configReferences;
```
DwayneJengSage, line 179: Is it worth adding a unit test that getConfigReferences() never returns null? (Similarly for getConfigElements().)

app/org/sagebionetworks/bridge/play/controllers/AppConfigElementsController.java
```diff
@@ -116,6 +125,8 @@ public Result deleteElementRevision(String id, String revisionString, String phy
         } else {
             service.deleteElementRevision(session.getStudyIdentifier(), id, revision);
         }
+        // App config elements are included in the app configs, so allow cache to update
+        cacheProvider.removeSetOfCacheKeys(CacheKey.appConfigList(session.getStudyIdentifier()));
```
DwayneJengSage, line 129: Unit test for deleteElementRevision()?

test/org/sagebionetworks/bridge/services/AppConfigServiceTest.java
```diff
@@ -255,13 +390,29 @@ public void getAppConfigForUserReturnsOldestVersion() {
...
+        
         assertEquals(TIMESTAMP.getMillis(), returnValue.getCreatedOn());
         assertEquals(TIMESTAMP.getMillis(), returnValue.getModifiedOn());
         assertEquals(GUID, returnValue.getGuid());
+        assertEquals(newConfig.getLabel(), returnValue.getLabel()); //
+        assertEquals(TEST_STUDY.getIdentifier(), returnValue.getStudyId()); //
+        assertNotNull(returnValue.getClientData());
+        assertNotNull(returnValue.getSurveyReferences());
+        assertNotNull(returnValue.getSchemaReferences());
+        assertNotNull(returnValue.getConfigReferences());
```
DwayneJengSage: Instead of checking that they exist, would it be better to check that they match the expected values?

test/org/sagebionetworks/bridge/validators/AppConfigValidatorTest.java
```diff
@@ -53,8 +68,45 @@ public void before() {
...
+        ConfigReference ref1 = new ConfigReference("id:1", 1L);
+        ConfigReference ref2 = new ConfigReference("id:2", 2L);
+        
+        when(appConfigElementService.getElementRevision(any(), any(), anyLong())).thenReturn(AppConfigElement.create());
+        
+        List<ConfigReference> references = ImmutableList.of(ref1, ref2);
+        appConfig.setConfigReferences(references);
+        
+        // This succeeds because the mock does not throw an exception
+        assertValidatorMessage(newValidator, appConfig, "criteria", "are required");
```
DwayneJengSage: This test semantically suggests that the validator should throw because criteria are missing. It also means if there's a bug and the validator throws anyway, we wouldn't catch that bug. It would be better to just call Validate.validateThrowException().

DwayneJengSage (changes requested):
Overall looks good. Comments on tests.

test/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigTest.java
alxdarksage: Changed this to assert against the exact JSON we want, and added an assert for the JSON that's in the configElements map (which was missing).

app/org/sagebionetworks/bridge/dynamodb/DynamoAppConfig.java
alxdarksage: Added tests of all four collections.

app/org/sagebionetworks/bridge/play/controllers/AppConfigElementsController.java
alxdarksage: Updated that test to also verify the cache is cleared

test/org/sagebionetworks/bridge/services/AppConfigServiceTest.java
alxdarksage: Changed these to test actual values

test/org/sagebionetworks/bridge/validators/AppConfigValidatorTest.java
alxdarksage: Yes, it's confusing. Changed this.

test/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigTest.java
```diff
@@ -66,27 +92,34 @@ public void canSerialize() throws Exception {
...
         
+        assertEquals(clientDataString, node.get("clientData").toString());
         assertEquals("AppConfig", node.get("type").asText());
         assertEquals(TIMESTAMP.toString(), node.get("createdOn").textValue());
         assertEquals(TIMESTAMP.toString(), node.get("modifiedOn").textValue());
         
         AppConfig deser = BridgeObjectMapper.get().treeToValue(node, AppConfig.class);
         assertNull(deser.getStudyId());
+        assertEquals(clientDataString, deser.getClientData().toString());
+        assertEquals(clientDataString, deser.getConfigElements().get("config1").toString());
```
DwayneJengSage: I think there was a misunderstanding here. We don't want to verify the exact string conversion of the JSON, since that depends on a lot of factors (key ordering, formatting). Both here and in the JSON above.

However, we _do_ want to verify that the JSON contains the fields we expect in the format we expect (as opposed to some weird different format).

test/org/sagebionetworks/bridge/services/AppConfigServiceTest.java
```diff
@@ -255,13 +391,29 @@ public void getAppConfigForUserReturnsOldestVersion() {
...
+        newConfig.setConfigReferences(CONFIG_REF_LIST);
         
         AppConfig returnValue = service.createAppConfig(TEST_STUDY, newConfig);
+        
         assertEquals(TIMESTAMP.getMillis(), returnValue.getCreatedOn());
         assertEquals(TIMESTAMP.getMillis(), returnValue.getModifiedOn());
         assertEquals(GUID, returnValue.getGuid());
+        assertEquals(newConfig.getLabel(), returnValue.getLabel()); //
+        assertEquals(TEST_STUDY.getIdentifier(), returnValue.getStudyId()); //
+        assertEquals(TestUtils.getClientData().toString(), returnValue.getClientData().toString());
```
DwayneJengSage: Minor: If these are the same object, then the .toString() will also be the same. That said, it might just be cleaner to test the objects themselves are the same instead of doing string comparison.

test/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigTest.java
```diff
@@ -56,6 +81,7 @@ public void canSerialize() throws Exception {
         criteria.setMaxAppVersion(ANDROID, 15);
         criteria.setLanguage("fr");
 
+        String clientDataString = TestUtils.createJson("{'booleanFlag':true,'stringValue':'testString','intValue':4}");
         JsonNode clientData = TestUtils.getClientData();
```
DwayneJengSage: We create clientDataString from hardcoded JSON, but we use TestUtils.getClientData(). If we ever change TestUtils.getClientData() for whatever reason, this test breaks.

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigTest.java
alxdarksage: Jackson serialization preserves order (unless you use @JsonPropertyOrder), and type is in the string (booleans, number and string values are different when serialized as JSON), but I can do this field by field if you want.

test/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigTest.java
```diff
@@ -66,32 +91,39 @@ public void canSerialize() throws Exception {
...
+                "surveyReferences", "schemaReferences", "configReferences", "configElements", "version", 
+                "type", "deleted");
                 
         JsonNode node = BridgeObjectMapper.get().valueToTree(appConfig);
         assertEquals(fields, Sets.newHashSet(node.fieldNames()));
         
+        TestUtils.assertNode(clientData, node.get("clientData"));
         assertEquals("AppConfig", node.get("type").asText());
         assertEquals(TIMESTAMP.toString(), node.get("createdOn").textValue());
         assertEquals(TIMESTAMP.toString(), node.get("modifiedOn").textValue());
```
DwayneJengSage, line 120: I think there's still a misunderstanding here. What we need is to verify the JSON representation of the AppConfig object as a whole, not just the the clientData.

test/org/sagebionetworks/bridge/TestUtils.java
```diff
@@ -538,12 +539,16 @@ public static JsonNode getOtherClientData() {
...
     
-    public static Set<String> getFieldNamesSet(JsonNode node) {
-        HashSet<String> set = new HashSet<>();
-        for (Iterator<String> i = node.fieldNames(); i.hasNext(); ) {
-            set.add(i.next());
+    public static void assertNode(JsonNode expected, JsonNode actual) {
+        assertEquals(expected.size(), actual.size());
+        
+        for (Iterator<Entry<String,JsonNode>> i = expected.fields(); i.hasNext();) {
+            Entry<String,JsonNode> expectedProperty = i.next();
```
DwayneJengSage: This is unnecessary. JsonNode.equals() does what you want it to do. Also, this method only verifies types, not values, doesn't work for non-object types, and doesn't do deep recursion.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #1792, opened 2018-10-25
Opt phone numbers back in when they create an account in a different study

DwayneJengSage:
This is the work around to when users opt out of SMS in one study, but we don't want to lock them out of other studies.

This has been unit tested. Unfortunately, there seems to be a propagation delay to the opt-out list in SNS, so I'm unable to test this at the moment.

### DIFF
--- a/app/org/sagebionetworks/bridge/services/ParticipantService.java
+++ b/app/org/sagebionetworks/bridge/services/ParticipantService.java
@@ -12,10 +12,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 
-import com.amazonaws.services.sns.AmazonSNSClient;
-import com.amazonaws.services.sns.model.CheckIfPhoneNumberIsOptedOutRequest;
-import com.amazonaws.services.sns.model.CheckIfPhoneNumberIsOptedOutResult;
-import com.amazonaws.services.sns.model.OptInPhoneNumberRequest;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 
@@ -92,7 +88,7 @@ public class ParticipantService {
 
     private ScheduledActivityService scheduledActivityService;
 
-    private AmazonSNSClient snsClient;
+    private SmsService smsService;
 
     private ActivityEventService activityEventService;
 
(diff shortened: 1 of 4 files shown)


app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -347,9 +360,32 @@ public IdentifierHolder createParticipant(Study study, Set<Roles> callerRoles, S
...
+
+                // User was previously opted out. They created a new account (almost certainly in a new study). We need
+                // to opt them back in. Note that according to AWS, this can only be done once every 30 days to prevent
+                // abuse.
+                OptInPhoneNumberRequest optInRequest = new OptInPhoneNumberRequest().withPhoneNumber(
+                        phone.getNumber());
+                snsClient.optInPhoneNumber(optInRequest);
+            }
+        }
+
```
alxdarksage, line 368: Cold we throw all this code into SmsService?

alxdarksage (changes requested):

app/org/sagebionetworks/bridge/services/ParticipantService.java
DwayneJengSage: I mean, I could, but at the time I wrote this, SmsService didn't exist yet, so...

alxdarksage: It makes a lot of sense to me to put it there, where there is already a reference to an SNS client, and there's no updating back of the participant or anything. Unless you're under time pressure, I think it would be cleaner.

DwayneJengSage: Moved to SmsService.

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -347,9 +356,19 @@ public IdentifierHolder createParticipant(Study study, Set<Roles> callerRoles, S
         if (sendEmailVerification && !study.isAutoVerificationEmailSuppressed()) {
             accountWorkflowService.sendEmailVerificationToken(study, accountId, account.getEmail());
         }
+
+        // If you create an account with a phone number, this opts the phone number in to receiving SMS. We do this
+        // _before_ phone verification / sign-in, because we need to opt the user in to SMS in order to send phone
+        // verification / sign-in.
+        Phone phone = account.getPhone();
+        if (phone != null) {
+            // Note that there is no object with both accuntId and phone, so we need to pass them in separately.
```
alxdarksage: trivial: accuntId --> accountId

alxdarksage (approved):

app/org/sagebionetworks/bridge/services/ParticipantService.java
DwayneJengSage: Fixed

PR #1796, opened 2018-11-06
SMS Log as Health Data, and Get Health Data By CreatedOn API

DwayneJengSage:
This allows us to export SMS Log to Synapse. Note that, as with all other forms of Health Data, this requires the participant to consent and set sharing status. This means that pre-consent messages like phone verification and sign-in will (generally speaking) not be exported to Synapse.

The Get Health Data By CreatedOn API is to enable integ tests.


app/org/sagebionetworks/bridge/play/controllers/HealthDataController.java
```diff
@@ -31,6 +36,21 @@ final void setHealthDataService(HealthDataService healthDataService) {
         this.healthDataService = healthDataService;
     }
 
+    /** Gets a list of records for the given healthCode between the specified createdOn times (inclusive). */
+    public Result getRecordsByCreatedOn(String createdOnStartString, String createdOnEndString) throws IOException {
+        UserSession session = getAuthenticatedAndConsentedSession();
```
alxdarksage, line 41: This is an authenticated user API?

app/org/sagebionetworks/bridge/services/HealthDataService.java
```diff
@@ -397,18 +400,42 @@ public HealthDataRecord getRecordById(String id) {
         return healthDataDao.getRecordsForUploadDate(uploadDate);
     }
 
-    public List<HealthDataRecord> getRecordsByHealthcodeCreatedOnSchemaId(String healthCode, Long createdOn, String schemaId) {
+    /** Gets a list of records for the given healthCode between the specified createdOn times (inclusive). */
+    public List<HealthDataRecord> getRecordsByHealthCodeCreatedOn(String healthCode, DateTime createdOnStart,
+            DateTime createdOnEnd) {
+        if (StringUtils.isBlank(healthCode)) {
+            throw new BadRequestException(String.format(Validate.CANNOT_BE_BLANK, "healthCode"));
+        }
```
alxdarksage: I think this should be a preconditions check. It should never be null if the system is working and it's not a 400 error to the user, it's a 500 error on us

alxdarksage (changes requested):
Should this be given to users with a forward cursor rather than, or in addition to, a time range limit? It it's client facing, they have preferred a forward cursor over having to search by time ranges.

app/org/sagebionetworks/bridge/play/controllers/HealthDataController.java
DwayneJengSage: As per IM conversation, I'm using a DateTimeRangeResourceList because this is intended for integ tests, and I didn't think the extra complexity of using a ForwardCursorPagedResourceList was worth it since no one was asking for it.

Also as per IM conversation, adding it as a user API is fine. Other alternatives considered are developer API or an internal admin API (but definitely not researcher API).

alxdarksage (approved):

PR #1800, opened 2018-11-08
Substudies

alxdarksage:
This is just the CRUD API for the substudy object. Requires the following SQL be run for the database:

```
CREATE TABLE `Substudies` (
  `id` VARCHAR(60) NOT NULL,
  `studyId` VARCHAR(60) NOT NULL,
  `name` VARCHAR(255) NULL,
  `version` int(10) unsigned NOT NULL DEFAULT '0',
  `deleted` tinyint(1) DEFAULT 0,
  `createdOn` bigint(20) NOT NULL,
  `modifiedOn` bigint(20) NOT NULL,
  PRIMARY KEY (`id`, `studyId`)
) DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
```


app/org/sagebionetworks/bridge/hibernate/HibernateSubstudyDao.java
```diff
@@ -0,0 +1,75 @@
...
+    public List<Substudy> getSubstudies(StudyIdentifier studyId, boolean includeDeleted) {
+        checkNotNull(studyId);
+        
+        Map<String,Object> parameters = ImmutableMap.of("studyId", studyId.getIdentifier());
+        String query = "from HibernateSubstudy as substudy where studyId=:studyId";
+        if (!includeDeleted) {
+            query += " and deleted != 1";
+        }
+        return hibernateHelper.queryGet(query, parameters, null, null, HibernateSubstudy.class)
+                .stream().map((oneSubstudy) -> (Substudy) oneSubstudy).collect(Collectors.toList());
```
DwayneJengSage: Minor: Would it be easier to use ImmutableList.copyOf()?

app/org/sagebionetworks/bridge/models/substudies/SubstudyId.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.io.Serializable;
+import java.util.Objects;
+
+import javax.persistence.Column;
+import javax.persistence.Embeddable;
+
+@Embeddable
+public final class SubstudyId implements Serializable {
+    private static final long serialVersionUID = 3414483917399974708L;
```
DwayneJengSage: I would recommend not including the serialVersionUID and suppressing the warning instead. AFAIK this is only used for Java Serialization (see ObjectOutputStream), which I'm 99% sure Hibernate isn't using.

The purpose of the serialVersionUID is to mark incompatible versions of classes w.r.t serialization. Java automatically generates one for us based on class details, so if we change the class, the auto-generated serialVersionUID will also be different.

In practice, I've found many cases where developers update the class but forget to update the serialVersionUID and break something and rarely the other way around. With that in mind, I believe the standard practice of always specifying a serialVersionUID is incorrect, and that we should, in fact, be relying on Java to auto-generate them for us.

test/org/sagebionetworks/bridge/models/substudies/SubstudyIdTest.java
```diff
@@ -0,0 +1,15 @@
...
+import org.junit.Test;
+
+public class SubstudyIdTest {
+    @Test
+    public void test() {
+        SubstudyId studyId = new SubstudyId("studyId", "id");
+        
+        assertEquals("studyId", studyId.getStudyId());
+        assertEquals("id", studyId.getId());
+    }
```
DwayneJengSage, line 24: EqualsVerifier test?

test/org/sagebionetworks/bridge/services/SubstudyServiceTest.java
```diff
@@ -0,0 +1,208 @@
...
+    @Test(expected = EntityAlreadyExistsException.class)
+    public void createSubstudyAlreadyExists() {
+        Substudy substudy = Substudy.create();
+        substudy.setId("oneId");
+        substudy.setName("oneName");
+        
+        when(substudyDao.getSubstudy(TestConstants.TEST_STUDY, "oneId")).thenReturn(substudy);
+        
+        service.createSubstudy(TestConstants.TEST_STUDY, substudy);
+    }
```
DwayneJengSage, line 131: Is it worth also adding a test for createSubstudy() when the input is invalid? (Also, updateSubstudy().)

DwayneJengSage (changes requested):
Overall looks good. A few minor comments and a few comments about test.

app/org/sagebionetworks/bridge/hibernate/HibernateSubstudyDao.java
alxdarksage: Works great, switched

app/org/sagebionetworks/bridge/models/substudies/SubstudyId.java
alxdarksage: Changed. I did not know that Java auto-generated this if not present

test/org/sagebionetworks/bridge/models/substudies/SubstudyIdTest.java
alxdarksage: Added (you have to suppress the non-final fields warning but still useful).

test/org/sagebionetworks/bridge/services/SubstudyServiceTest.java
alxdarksage: Added

DwayneJengSage (approved):

PR #1802, opened 2018-11-13
Associate substudies

alxdarksage:
Associate accounts to substudy. Administrative users can create users with any substudies (or no substudies) if they have no substudy tag. But if a user is tagged with a substudy, they cannot create users unless those users are tagged with one or more of their substudies.

We are not doing any filtering on this or anything in this checkin... currently substudy association can be completely ignored.

The following SQL needs to be run on the database before deploying:
```
    CREATE TABLE `AccountsSubstudies` (
      `studyId` VARCHAR(60) NOT NULL,
      `substudyId` VARCHAR(60) NOT NULL,
      `accountId`  VARCHAR(255) NOT NULL,
      `externalId` VARCHAR(255) NULL,
      PRIMARY KEY (`studyId`,`substudyId`,`accountId`),
      CONSTRAINT `fk_substudy` FOREIGN KEY (`substudyId`,`studyId`) REFERENCES `Substudies` (`id`,`studyId`),
      CONSTRAINT `fk_account` FOREIGN KEY (`accountId`) REFERENCES `Accounts` (`id`)
    ) DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
```

### DIFF
--- a/app/org/sagebionetworks/bridge/models/substudies/AccountSubstudyId.java
+++ b/app/org/sagebionetworks/bridge/models/substudies/AccountSubstudyId.java
@@ -17,15 +17,15 @@ public final class AccountSubstudyId implements Serializable {
     private String substudyId;
     
     @Column(name = "accountId")
-    private String id;
+    private String accountId;
 
     public AccountSubstudyId() {
     }
  
-    public AccountSubstudyId(String studyId, String substudyId, String id) {
+    public AccountSubstudyId(String studyId, String substudyId, String accountId) {
         this.studyId = studyId;
         this.substudyId = substudyId;
-        this.id = id;
+        this.accountId = accountId;
     }
     
     public String getStudyId() {
@@ -36,13 +36,13 @@ public final class AccountSubstudyId implements Serializable {
         return substudyId;
     }
     
-    public String getId() {
-        return id;
+    public String getAccountId() {
+        return accountId;
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(studyId, substudyId, id);
+        return Objects.hash(studyId, substudyId, accountId);
     }
 
     @Override
@@ -54,7 +54,7 @@ public final class AccountSubstudyId implements Serializable {
         AccountSubstudyId other = (AccountSubstudyId) obj;
         return Objects.equals(studyId, other.studyId) &&
                 Objects.equals(substudyId, other.substudyId) &&
-                Objects.equals(id, other.id);
+                Objects.equals(accountId, other.accountId);
     }    
     
 }
--- a/app/org/sagebionetworks/bridge/models/substudies/AccountSubstudyId.java
+++ b/app/org/sagebionetworks/bridge/models/substudies/AccountSubstudyId.java
@@ -1,0 +1,113 @@
+package org.sagebionetworks.bridge.models.substudies;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+import javax.persistence.Column;
+import javax.persistence.Embeddable;
+
+@Embeddable
+public final class AccountSubstudyId implements Serializable {
+    private static final long serialVersionUID = -2741510011134968409L;
+
+    @Column(name = "studyId")
+    private String studyId;
+
+    @Column(name = "substudyId")
+    private String substudyId;
+    
+    @Column(name = "accountId")
+    private String id;
+
+    public AccountSubstudyId() {
+    }
+ 
+    public AccountSubstudyId(String studyId, String substudyId, String id) {
+        this.studyId = studyId;
+        this.substudyId = substudyId;
+        this.id = id;
+    }
+    
+    public String getStudyId() {
+        return studyId;
+    }
+    
+    public String getSubstudyId() {
+        return substudyId;
+    }
+    
+    public String getId() {
+        return id;
+    }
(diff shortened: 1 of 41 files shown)


DwayneJengSage:
Can you also add the SQL command needed to bootstrap the table? Thanks!

app/org/sagebionetworks/bridge/models/substudies/AccountSubstudyId.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.io.Serializable;
+import java.util.Objects;
+
+import javax.persistence.Column;
+import javax.persistence.Embeddable;
+
+@Embeddable
+public final class AccountSubstudyId implements Serializable {
+    private static final long serialVersionUID = -2741510011134968409L;
```
DwayneJengSage: Minor: Remove serialVersionUID and suppress warning.

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -440,6 +466,11 @@ private void updateAccountAndRoles(Study study, Set<Roles> callerRoles, Account
             updateRoles(callerRoles, participant, account);
         }
     }
+    
+    private <T> void updatePersistedSet(Set<T> persisted, Set<T> updated) {
+        persisted.removeAll(Sets.difference(persisted, updated));
+        persisted.addAll(updated);
```
DwayneJengSage: Would it be easier just to call account.setAccountSubstudies()?

test/org/sagebionetworks/bridge/services/SubstudyServiceTest.java
```diff
@@ -54,15 +54,15 @@ public void getSubstudy() {
...
         assertEquals(substudy, returnedValue);
         
         verify(substudyDao).getSubstudy(TestConstants.TEST_STUDY, "id");
     }
     
     @Test(expected = EntityNotFoundException.class)
     public void getSubstudyNotFound() {
-        service.getSubstudy(TestConstants.TEST_STUDY, "id");
+        service.getSubstudy(TestConstants.TEST_STUDY, "id", true);
     }
```
DwayneJengSage, line 66: Unit test for throwsException=false?

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
```diff
@@ -206,6 +207,7 @@ public void before() throws Exception {
         
         participant = new StudyParticipant.Builder()
                 .withRoles(CALLER_ROLES)
+                .withSubstudyIds(ImmutableSet.of("substudyA"))
```
DwayneJengSage: Minor: This test inconsistently uses teh CALLER_SUBS constant or creates a new ImmutableSet on the fly.

test/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
```diff
@@ -39,6 +42,12 @@
     @Mock
     private ExternalIdService externalIdService;
     
+    @Mock
+    private SubstudyService substudyService;
+    
+    @Mock
+    private Substudy substudy;
```
DwayneJengSage, line 48: Minor: Does substudy need to be mocked? Would it be easier just to create a new one for every test?

DwayneJengSage (changes requested):
Overall looks good. Mostly minor notes.

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: This doesn't work. The list implementation being returned by Hibernate is "aware" of its member elements and will update when we update the account. But if you replace this list entirely with something like an ArrayList, this update doesn't happen correctly (SFAICT adding items to the array works, but removing them does not). I did realize yesterday evening that clear() and addAll() might be simpler and work just as well, I'll try it.

app/org/sagebionetworks/bridge/models/substudies/AccountSubstudyId.java
alxdarksage: Removed

test/org/sagebionetworks/bridge/services/SubstudyServiceTest.java
alxdarksage: Added

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
alxdarksage: Cleaned this up

test/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
alxdarksage: Removed mock annotation, instantiated in @Before now

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Yep that works fine, updating.

DwayneJengSage (approved):

PR #1803, opened 2018-11-20
Filter by substudies

alxdarksage:
Checking authorization by substudy, including filtering queries for accounts and blocking the retrieval of individual accounts. If you are assigned to no substudies, then you can see all accounts in the study; but if you are assigned to one or more substudies, you can only see accounts that have at least one substudy in common with you.

1) Replaced the request ID thread local variable with a RequestContext object that holds security-related information about a caller. Because we don't want to pick up the caller's context during unauthenticated/public calls, and an interceptor would accept context if a "Bridge-Session" header is present, we only set request ID in the interceptor and add session-related context information when we retrieve a session (BaseController);

2) All queries in HibernateAccountDao were rewritten to include a check for substudy association, if present;

3) A check was added after retrieving an individual account and the retrieval is blocked if substudy memberships do not align;

4) Create/update validation verifies substudy rules (so these are returned in an InvalidEntityException);

5) Removed the passing of roles/substudies via methods where this was occurring for security checks (such as checking what roles you can assign to a new account based on your own role)

6) Removed iterator-based methods of the AccountDao, rather than rewrite them to follow these security rules (they are no longer used).

I didn't do this for study partitioning... such a refactor would be massive at this point as we pass study in almost all method calls. But it's there in the request context and can be checked at any layer of the system.

https://sagebionetworks.jira.com/browse/BRIDGE-2339
https://sagebionetworks.jira.com/browse/BRIDGE-2320


app/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -83,14 +84,35 @@ public static boolean isExternalIdAccount(StudyParticipant participant) {
...
 
-    /** @see #getRequestId */
-    public static void setRequestId(String requestId) {
-        REQUEST_ID_THREAD_LOCAL.set(requestId);
+    /** @see #getRequestContext */
+    public static void setRequestContext(RequestContext context) {
+        REQUEST_CONTEXT_THREAD_LOCAL.set(context);
+    }
+    
+    public static HibernateAccount filterForSubstudy(HibernateAccount account) {
```
DwayneJengSage: Should this be Account rather than HibernateAccount? We want to avoid leaking implementation details higher up the stack.

app/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -83,14 +84,35 @@ public static boolean isExternalIdAccount(StudyParticipant participant) {
...
+        if (account == null) {
+            return account;
+        }
```
DwayneJengSage: Unit test for the null case?

test/org/sagebionetworks/bridge/BridgeUtilsTest.java
```diff
@@ -37,6 +42,58 @@
...
+    public void filterForSubstudyNoContextReturnsSubstudyAccount() {
+        BridgeUtils.setRequestContext(null);
+        assertNotNull(BridgeUtils.filterForSubstudy(getAccountWithSubstudy("substudyA")));
+    }
+    
+    @Test
+    public void filterForSubstudyEmptyContextReturnsNormalAccount() {
+        BridgeUtils.setRequestContext(RequestContext.NULL_INSTANCE);
+        assertNotNull(BridgeUtils.filterForSubstudy(getAccountWithSubstudy()));
+    }
```
DwayneJengSage: setRequestContext(null) and setRequestContext(RequestContext.NULL_INSTANCE) are equivalent, as tested in getRequestContext(). So these tests (filterForSubstudyEmptyContextReturnsNormalAccount() and filterForSubstudyEmptyContextReturnsSubstudyAccount()) are redundant with previous tests.

app/org/sagebionetworks/bridge/RequestContext.java
```diff
@@ -0,0 +1,77 @@
...
+import com.google.common.collect.ImmutableSet;
+
+public class RequestContext {
+    
+    public static final RequestContext NULL_INSTANCE = new RequestContext.Builder().build();
+
+    final String id;
+    final String callerStudyId;
+    final Set<String> callerSubstudies;
+    final Set<Roles> callerRoles;
```
DwayneJengSage: Should these be private?

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -413,91 +415,103 @@ private HibernateAccount getHibernateAccount(AccountId accountId) {
...
-            HibernateAccount hibernateAccount = getHibernateAccount(accountId);
-            userId = hibernateAccount.getId();
+        Set<String> callerSubstudies = context.getCallerSubstudies();
+        if (!callerSubstudies.isEmpty()) {
+            builder.append("AND acctSubstudy.substudyId IN (:substudies)", "substudies", callerSubstudies);
         }
-        hibernateHelper.deleteById(HibernateAccount.class, userId);    
+        if (!isCount) {
+            builder.append("GROUP BY acct.id");        
+        }
```
DwayneJengSage, line 488: If an account has multiple substudies, what happens? Does Hibernate create a HibernateAccount object with the aggregate of the substudies? Or does this flatten it into a single row and return a random substudy?

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
```diff
@@ -1439,22 +1447,55 @@ public void getPagedWithOptionalParams() throws Exception {
...
+        Set<String> substudyIds = ImmutableSet.of("substudyA", "substudyB");
+        try {
+            RequestContext context = new RequestContext.Builder()
+                    .withCallerSubstudies(substudyIds).build();
+            BridgeUtils.setRequestContext(context);
+            
+            AccountSummarySearch search = new AccountSummarySearch.Builder().build();
+            dao.getPagedAccountSummaries(study, search);
+            
+            verify(mockHibernateHelper).queryCount(any(), paramCaptor.capture());
```
DwayneJengSage: Should this test verify the query to make sure we're adding the substudies clause correctly?

app/org/sagebionetworks/bridge/play/controllers/AuthenticationController.java
```diff
@@ -173,6 +175,9 @@ public Result signOutV4() throws Exception {
     }
     
     public Result signUp() throws Exception {
+        // Ignore Bridge-Session cookie if it exists. This is not an authenticated request.
+        BridgeUtils.setRequestContext(RequestContext.NULL_INSTANCE);
+
```
alxdarksage: Remove this

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -576,21 +543,23 @@ private boolean validateHealthCode(HibernateAccount hibernateAccount) {
...
             studyId = new StudyIdentifierImpl(hibernateAccount.getStudyId());
         }
+        // Hibernate will not load the collection of substudies once you use the constructor form of HQL 
+        // to limit the data you retrieve from a table. May need to manually construct the objects to 
+        // avoid this 1+N query.
+        List<HibernateAccountSubstudy> accountSubstudies = hibernateHelper.queryGet(
+                "FROM HibernateAccountSubstudy WHERE accountId=:accountId",
+                ImmutableMap.of("accountId", hibernateAccount.getId()), null, null, HibernateAccountSubstudy.class);
+        Set<String> substudyIds = accountSubstudies.stream()
+                .map(AccountSubstudy::getSubstudyId).collect(BridgeCollectors.toImmutableSet());
```
DwayneJengSage: Should we add a check to the unit test unmarshallAccountSummarySuccess() to verify that we're loading and populating substudies into the account summary?

app/org/sagebionetworks/bridge/play/controllers/BaseController.java
```diff
@@ -168,7 +170,16 @@ UserSession getAuthenticatedSession(boolean consentRequired, Roles...roles) {
...
-
+        
+        // Update request context with security-related information about the user. This will be
+        // immediately removed from the thread local if an exception is thrown.
+        String requestId = RequestUtils.getRequestId(request());
+        RequestContext.Builder builder = new RequestContext.Builder().withRequestId(requestId);
+        builder.withCallerStudyId(session.getStudyIdentifier());
+        builder.withCallerSubstudies(session.getParticipant().getSubstudyIds());
+        builder.withCallerRoles(session.getParticipant().getRoles());
+        BridgeUtils.setRequestContext(builder.build());
```
DwayneJengSage, line 181: Do we have any tests that verify the request context is being populated from the session?

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
```diff
@@ -1573,17 +1613,21 @@ public void unmarshallAccountSummarySuccess() {
 
     // branch coverage, to make sure nothing crashes.
     @Test
-    public void unmarshallAccountSummaryBlankAccount() {
-        AccountSummary accountSummary = HibernateAccountDao.unmarshallAccountSummary(new HibernateAccount());
+    public void unmarshallAccountSummaryBlankAccount() throws Exception {
+        AccountSummary accountSummary = dao.unmarshallAccountSummary(makeValidHibernateAccount(false, false));
```
DwayneJengSage: The purpose of this test is to make sure that unmarshallAccountSummary is robust against an account where everything is (for some reason) null. With that in mind, we want to pass in new HibernateAccount(), not makeValidHibernateAccount().

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
```diff
@@ -1752,6 +1847,196 @@ public void authenticateAccountUnverifiedPhoneSucceedsForLegacy() throws Excepti
...
+    @Test
+    public void getAccountAfterAuthenticationSucceedsOnSubstudyMatch() throws Exception { 
+        testSubstudyMatch(ImmutableSet.of(SUBSTUDY_A), ACCOUNT_SUBSTUDIES, (accountId) -> {
+            Account account = dao.getAccountAfterAuthentication(accountId);
+            assertNotNull(account);
+        });
+    }
+    
+    @Test
+    public void getAccountAfterAuthenticationRespectsSubstudy() throws Exception {
```
DwayneJengSage: Are these next several tests redundant with the tests above?

test/org/sagebionetworks/bridge/services/AuthenticationServiceTest.java
```diff
@@ -255,6 +256,9 @@ public void canResendEmailVerification() throws Exception {
 
     @Test
     public void createResearcherAndSignInWithoutConsentError() {
+        BridgeUtils.setRequestContext(new RequestContext.Builder()
+                .withCallerRoles(ImmutableSet.of(Roles.ADMIN)).build());
+        
         testUser = helper.getBuilder(AuthenticationServiceTest.class)
                 .withConsent(false).withSignIn(false).withRoles(Roles.RESEARCHER).build();
```
DwayneJengSage, line 264: The request context says admin, but the user says researcher.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/BridgeUtils.java
alxdarksage: Yes. All the calls to this don't need to work with HibernateAccount as the interface is comprehensive. Changed this.

app/org/sagebionetworks/bridge/BridgeUtils.java
alxdarksage: Added test

test/org/sagebionetworks/bridge/BridgeUtilsTest.java
alxdarksage: Removed these NULL_INSTANCE tests

app/org/sagebionetworks/bridge/RequestContext.java
alxdarksage: Yes, made these private

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: Because this query has the HibernateAccount() constructor in it, substudies are not retrieved at all. You cannot include a collection in the constructor apparently (it's not supported by JPA), and I didn't find a way around this (except to load the whole record perhaps, but since the record could include a large blob of JSON, I'm still avoiding that for the summary query). For the query of an individual record by an identifier... the GROUP BY clause doesn't do anything as there's only one account ID

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
alxdarksage: Query verification added

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: Added this to the happy case getPaged() test, it now calls for and loads AccountSubstudy records.

app/org/sagebionetworks/bridge/play/controllers/BaseController.java
alxdarksage: Added a test to verify this.

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
alxdarksage: You are right, I changed this in a bad way. Restored it and added null-pointer check for ID so the test will pass

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
alxdarksage: You are right, these are redundant. After refactoring I did not remove all the tests that were covered with testSubstudyMatch(). Removed.

test/org/sagebionetworks/bridge/services/AuthenticationServiceTest.java
alxdarksage: This is correct. The AuthenticationService will now verify the caller is an ADMIN, so RequestContext must be set. I've added a comment to the two tests that set this.

DwayneJengSage (approved):

PR #1805, opened 2018-11-27
Changes to ensure concurrent requests modify the same session

alxdarksage:
JIRA https://sagebionetworks.jira.com/browse/BRIDGE-2342

I am able to reproduce BRIDGE-2342 now with a JMeter script that runs the two requests concurrently. Interestingly, the changes mean that the second sign in, which is retrieving activities, will now return 401, "Not signed in," once the first consent succeeds, because it will overwrite the session token that was issued to request two. Whereas before it would keep returning 412. That's despite the fact that the login for request #2 just succeeded.

This behavior is an improvement because it will retriever a re-sign in rather than a consent flow of some kind.


app/org/sagebionetworks/bridge/cache/CacheProvider.java
```diff
@@ -127,24 +114,92 @@ public void setUserSession(final UserSession session) {
...
+        UserSession session = getUserSessionNewVersion(sessionToken);
+        if (session == null) {
+            session = getUserSessionOldVersion(sessionToken);
+        }
+        return session;
+    }
+    
+    private UserSession getUserSessionNewVersion(String sessionToken) {
+        checkNotNull(sessionToken);
+        try {
```
DwayneJengSage, line 129: Do we need unit tests for the new get methods? Should include the following cases:

* session token not found
* user ID has no session
* session ID mismatch
* session successfully retrieved

app/org/sagebionetworks/bridge/cache/CacheProvider.java
```diff
@@ -127,24 +114,92 @@ public void setUserSession(final UserSession session) {
...
+    public UserSession getUserSessionByUserId(String userId) {
+        UserSession session = getUserSessionByUserIdNewVersion(userId);
+        if (session == null) {
+            session = getUserSessionByUserIdOldVersion(userId);
+        }
+        return session;
+    }
+    
+    private UserSession getUserSessionByUserIdNewVersion(final String userId) {
+        checkNotNull(userId);
```
DwayneJengSage, line 185: Similarly, we should add unit tests for this as well, including the following cases:

* user ID has no session
* user ID has session

test/org/sagebionetworks/bridge/cache/CacheProviderMockTest.java
```diff
@@ -171,39 +162,52 @@ public void testGetUserSessionByUserId() throws Exception {
...
+        session.setSessionToken(DECRYPTED_SESSION_TOKEN);
         
         cacheProvider.removeSession(session);
-        cacheProvider.getUserSession(SESSION_TOKEN);
+        cacheProvider.getUserSession(DECRYPTED_SESSION_TOKEN);
         
-        verify(transaction).del(SESSION_KEY.toString());
-        verify(transaction).del(USER_SESSION_KEY.toString());
+        verify(transaction).del(TOKEN_TO_USER_ID.toString());
+        verify(transaction).del(USER_ID_TO_SESSION.toString());
```
DwayneJengSage, line 222: Test should verify that we delete all 4 keys.

test/org/sagebionetworks/bridge/cache/CacheProviderMockTest.java
```diff
@@ -171,39 +162,52 @@ public void testGetUserSessionByUserId() throws Exception {
...
+        String ser = BridgeObjectMapper.get().writeValueAsString(session);
+
+        when(jedisOps.get(USER_ID_TO_SESSION.toString())).thenReturn(ser);
+        
         cacheProvider.removeSessionByUserId(USER_ID);
         
-        verify(transaction).del(SESSION_KEY.toString());
-        verify(transaction).del(USER_SESSION_KEY.toString());
+        verify(transaction).del(TOKEN_TO_USER_ID.toString());
+        verify(transaction).del(USER_ID_TO_SESSION.toString());
```
DwayneJengSage, line 238: Similarly, this should verify we delete all 4 keys.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/cache/CacheProvider.java
alxdarksage: I added tests for these specifically against the new getSession method alone. Will adjust once we cleanup the old methods.

test/org/sagebionetworks/bridge/cache/CacheProviderMockTest.java
alxdarksage: Added

test/org/sagebionetworks/bridge/cache/CacheProviderMockTest.java
alxdarksage: Added

DwayneJengSage (approved):

PR #1808, opened 2018-12-04
Do not return substudy IDs for substudies the caller is not a member of.

alxdarksage:
Ultimately, if you are a partner contributing users, you shouldn't know the other partners contributing users to that study (unless for some reason we want to share that information). So after doing the search, we remove substudies that the caller is not associated to.


DwayneJengSage:
Code-wise, this looks fine.

Design-wise, how do we protect from an updateParticipant() call accidentally removing the participant from every other substudy?

alxdarksage:
This has been augmented: after creation, you cannot change substudy associations when you update the participant record. The only exception to this (right now) is admins, who can still do this. In fact, since this association will eventually be created by assigning external IDs and/or signing specific consents, we don't want people to be able to change these mappings directly.

DwayneJengSage (approved):

PR #1813, opened 2018-12-10
add userId to Metrics for request channel signIn

DwayneJengSage:
So we can look up sign-in attempts for specific users. Also adds request ID to SMS logs so we can trace requests. (This already exists for emails.)

Testing done:
* unit tests
* integ tests
* manually tested and verified logs


alxdarksage (approved):
Do we also get the userId on a normal sign in? If not, we should.

PR #1819, opened 2019-01-14
Update to Mockito 2.

alxdarksage:
- Many changes are just a mechanical change, between packages for example

- The Mockito JUnit tests now throw an error if a mock is created that is not used. I removed these everywhere they occurred in order to get tests to pass but I did not attempt to figure out if the failure to use a mock indicated that the test was in error. Our test coverage is the same as it was before, for better or worse.

- One serious complication was the use of TestUtils mocking methods, which overmocked the calls being made by any individual test. I replaced this with a builder-like mocking tool on the same utilities class, and simplified some request handling (e.g. always using getHeader() rather than headers()) to make it easier to mock Play's HTTP environment without errors or inconsistencies.

- A handful of tests failed because of a null or incorrect type being passed to the mocks that Mockito 2 now catches and correctly rejects. I fixed these where they occurred.

The end result is a huge number of trivial changes and a few more serious ones.


test/org/sagebionetworks/bridge/dynamodb/DynamoNotificationTopicDaoTest.java
```diff
@@ -387,44 +264,102 @@ public void deleteTopicPermanentlyWithCriteria() {
...
+        when(mockQueryResultPage.getResults()).thenReturn(topicList);
+        when(mockMapper.queryPage(eq(DynamoNotificationTopic.class), any())).thenReturn(mockQueryResultPage);
+    }
     
     // If the SNS topic creation fails, no DDB record will exist.
     // (okay if SNS topic exists, but DDB record doesn't, so don't need to test that path)
     @Test
     public void noOrphanOnPartialCreate() {
-        doThrow(new RuntimeException()).when(mockSnsClient).createTopic(any(CreateTopicRequest.class));
-
```
DwayneJengSage: We should probably leave this throw just to make it clear that we're testing the SNS client throwing an exception.

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
```diff
@@ -1924,7 +1897,6 @@ public void editAccountFailsAcrossSubstudies() throws Exception {
         
         HibernateAccount hibernateAccount = makeValidHibernateAccount(false, false);
         hibernateAccount.setAccountSubstudies(ACCOUNT_SUBSTUDIES);
-        when(mockHibernateHelper.getById(any(), any())).thenReturn(hibernateAccount);
```
DwayneJengSage, line 1961: I think this changes the test. This now tests that the account doesn't exist, rather than testing that the account is in the wrong substudy.

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
```diff
@@ -498,13 +494,13 @@ private IdentifierHolder setUpCreateParticipant() throws Exception {
...
                 "'attributes':{'phone':'123456789'},"+
-                "'languages':['en','fr']}"));
+                "'languages':['en','fr']}")).mock();
         return holder;
     }
 
     @Test
     public void updateParticipantWithMismatchedIdsUsesURL() throws Exception {
-        mockPlayContextWithJson(createJson("{'id':'id2'}"));
+        TestUtils.mockPlay().withJsonBody(createJson(createJson("{'id':'id2'}"))).mock();
```
DwayneJengSage: Minor: You call createJson() twice. Is this intentional?

test/org/sagebionetworks/bridge/play/controllers/SubpopulationControllerTest.java
```diff
@@ -112,10 +112,10 @@ public void getAllSubpopulationsExcludeDeleted() throws Exception {
 
     @Test
     public void getAllSubpopulationsIncludeDeleted() throws Exception {
-        TestUtils.mockPlayContext();
+        TestUtils.mockPlay().mock();
         
-        List<Subpopulation> list = createSubpopulationList();
-        when(subpopService.getSubpopulations(study.getStudyIdentifier(), false)).thenReturn(list);
+        // */List<Subpopulation> list = createSubpopulationList();
+        // */when(subpopService.getSubpopulations(study.getStudyIdentifier(), false)).thenReturn(list);
```
DwayneJengSage: I think this is just a faulty test. The test is supposed to be for includeDeleted=true, but we mock includeDeleted=false. We also don't verify the output, so we just never noticed the test was broken.

test/org/sagebionetworks/bridge/services/AccountWorkflowServiceTest.java
```diff
@@ -273,7 +271,7 @@ public void sendPhoneVerificationTokenNoPhone() {
     @Test
     public void sendPhoneVerificationTokenThrottled() {
         // Throttle limit is 2. Make 3 requests, and send only 2 emails.
-        when(service.getNextToken()).thenReturn(TOKEN);
+     // */when(service.getNextToken()).thenReturn(TOKEN);
```
DwayneJengSage: This shouldbe getNextPhoneToken(). It looks like the code was refactored at some point and somehow didn't break the test.

test/org/sagebionetworks/bridge/services/AppConfigServiceTest.java
```diff
@@ -372,7 +369,6 @@ public void getAppConfigForUserReturnsOldestVersion() {
         survey.setIdentifier("theIdentifier");
         survey.setGuid(SURVEY_REF_LIST.get(0).getGuid());
         survey.setCreatedOn(SURVEY_REF_LIST.get(0).getCreatedOn().getMillis());
-        when(mockSurveyService.getSurvey(TestConstants.TEST_STUDY, SURVEY_KEY, false, true)).thenReturn(survey);
```
DwayneJengSage, line 375: I think this test was another victim of a refactor. We changed from throwException=true to throwException=false, but this test doesn't verify the survey in the result, so no one noticed this test was broken.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -225,7 +225,6 @@ public void adminSignInWithEmail() {
                 .withRoles(Sets.newHashSet(Roles.DEVELOPER)).build();
         doReturn(account).when(accountDao).authenticate(study, EMAIL_PASSWORD_SIGN_IN);
         doReturn(participant).when(participantService).getParticipant(study, account, false);
-        doReturn(UNCONSENTED_STATUS_MAP).when(consentService).getConsentStatuses(any());
```
DwayneJengSage, line 229: Another victim of refactoring. We refactored getConsentStatuses to take 2 arguments instead of one. This test only succeeds because an empty consent status map is treated the same as a consent status map with an unconsented result. (Looks like several of the tests here have the same problem.)

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
```diff
@@ -333,8 +333,7 @@ public void withdrawConsentWithAccount() throws Exception {
     @Test
     public void withdrawFromStudyWithEmail() throws Exception {
         setupWithdrawTest();
-        account.getAccountSubstudies().iterator().next().setExternalId("asExternalId");
```
DwayneJengSage, line 336: Was this deleted accidentally? iirc, this was added as part of the external IDs stuff.

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
```diff
@@ -593,8 +591,6 @@ public void emailConsentAgreementNoConsentAdministratorEmail() {
     public void emailConsentAgreementDoesNotSuppressEmailNotification() {
         account.setConsentSignatureHistory(SUBPOP_GUID, ImmutableList.of(CONSENT_SIGNATURE));
         
-        when(subpopulation.isAutoSendConsentSuppressed()).thenReturn(true);
-
```
DwayneJengSage, line 597: This defeats the purpose of the test, which is to verify that turning on the suppression flag doesn't actually suppress if the user explicitly asks for the consent email.

Maybe it would be better to merge with emailConsentAgreementNoConsentAdministratorEmail() and verify that subpopulation.isAutoSendConsentSuppressed() is never called?

test/org/sagebionetworks/bridge/services/SurveyServiceMockTest.java
```diff
@@ -195,7 +193,6 @@ public void successfulDeletePermanently() {
                 .withSurvey("Survey", "otherGuid", SURVEY_CREATED_ON).build();
         getActivityList(plans).set(0, activity);
         
-        doReturn(plans).when(mockSchedulePlanService).getSchedulePlans(ClientInfo.UNKNOWN_CLIENT, TEST_STUDY, false);
```
DwayneJengSage, line 198: I think this is a bug in the test. We call getSchedulePlans() with includeDeleted=true, but we mock includeDeleted=false. (Here and several other places in this class.)

test/org/sagebionetworks/bridge/services/SurveyServiceMockTest.java
```diff
@@ -560,7 +550,6 @@ public void getSurveyReturnsNull() {
     public void updateSurveyInOtherStudy() {
         Survey survey = Survey.create();
         survey.setStudyIdentifier(TestConstants.TEST_STUDY_IDENTIFIER);
-        when(mockSurveyDao.updateSurvey(survey)).thenReturn(survey);
```
DwayneJengSage, line 563: This is a bug in the test. This should be getSurvey(), not updateSurvey(). Similarly for publish, version, and delete.

test/org/sagebionetworks/bridge/services/SurveyServiceMockTest.java
```diff
@@ -627,7 +614,6 @@ public void deleteSurveyPermanentlyInOtherStudyThrowsException() {
     public void getSurveyMostRecentlyPublishedVersionInOtherStudy() {
         Survey survey = Survey.create();
         survey.setStudyIdentifier(TestConstants.TEST_STUDY_IDENTIFIER);
-        when(mockSurveyDao.getSurveyMostRecentlyPublishedVersion(TestConstants.TEST_STUDY, SURVEY_GUID, true)).thenReturn(survey);
```
DwayneJengSage, line 630: This test is actually nonsensical, since getSurveyMostRecentlyPublishedVersion takes study as an argument, you can't actually get a survey from another study.

Similarly with getSurveyMostRecentVersion().

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/dynamodb/DynamoNotificationTopicDaoTest.java
alxdarksage: This wasn't called and then a NPE was making the test succeed (it wasn't called because the method is overloaded and it's looking for the wrong method overload). Restored, fixed overload so it is called, test passes.

test/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
alxdarksage: Mock wasn't called because it was mocking the wrong thing. Fixing this caused test to fail because we're not filtering in editAccount (or AccountDao) anymore.

I added this filtering back in to fix the text. This seems reasonable because sometimes we want to retrieve accounts across substudies (email notifications that someone has signed up with your credentials) but we never want to write across substudies unless you're an admin. That's a normal update operation.

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
alxdarksage: No, fixed

test/org/sagebionetworks/bridge/play/controllers/SubpopulationControllerTest.java
alxdarksage: Fixed the mock so it is called, and verify the list is returned.

test/org/sagebionetworks/bridge/services/AccountWorkflowServiceTest.java
alxdarksage: This fixes the test but since it's testing throttling, and not the content of the sms message provider, I don't think it needs to be fixed (but I do need to remove the commented out code).

test/org/sagebionetworks/bridge/services/AppConfigServiceTest.java
alxdarksage: Fixed the mock, then added a test to verify it is called (though Mockito 2 now enforces that).

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: Yes, restored and fixed mock so it is matched and added a test the content is being returned in the session.

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
alxdarksage: I'm not sure why this would be here. We don't remove AccountSubstudy records when you withdraw from a study. We need to continue to ask the record to provide consent, and so do subsidy administrators. So this external ID isn't changed at all. So I think removing this makes sense.

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
alxdarksage: Ah, yes it's better to verify it is never called.

test/org/sagebionetworks/bridge/services/SurveyServiceMockTest.java
alxdarksage: Changed this to true and test now passes with it in there. In this test we do not set a schedule plan that would block the delete.

test/org/sagebionetworks/bridge/services/SurveyServiceMockTest.java
alxdarksage: Changed and added in other tests.

test/org/sagebionetworks/bridge/services/SurveyServiceMockTest.java
alxdarksage: I've fixed these two tests so they test what they are supposed to be testing, however, as you note, the DAO should never return a survey with a study ID that's different than the study ID you supply to the DAO, because the study ID is part of the DDB query. Filed BRIDGE-2366 to remove these checks entirely (the comment about this references admins but doesn't make any sense to me... I can't think of any reason to enforce this again in the service).

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -264,7 +265,6 @@ public void adminSignInWithPhone() {
                 .copyOf(PARTICIPANT).withRoles(Sets.newHashSet(Roles.RESEARCHER)).build();
         doReturn(account).when(accountDao).authenticate(study, PHONE_PASSWORD_SIGN_IN);
         doReturn(participant).when(participantService).getParticipant(study, account, false);
-        doReturn(UNCONSENTED_STATUS_MAP).when(consentService).getConsentStatuses(any());
```
DwayneJengSage, line 268: Is it worth applying the same fix to the other tests in this class?

* adminSignInWithPhone()
* emailSignInThrowsConsentRequired()
* emailSignInAdminOK()
* reauthTokenRequired()
* phoneSignInThrowsConsentRequired()

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
DwayneJengSage, line 336: I think adding this was specifically to demonstrate that external IDs _aren't_ removed when we withdraw. I think the verification on this just slipped through the cracks.

test/org/sagebionetworks/bridge/services/SurveyServiceMockTest.java
```diff
@@ -228,9 +226,8 @@ public void logicallyDeleteSurveyNotEmptySharedModules() {
     @Test(expected = BadRequestException.class)
     public void physicallyDeleteSurveyNotEmptySharedModules() {
         when(mockSharedModuleMetadataService.queryAllMetadata(anyBoolean(), anyBoolean(), anyString(), any(),
-                anySetOf(String.class), anyBoolean())).thenReturn(ImmutableList.of(makeValidMetadata()));
+                any(), anyBoolean())).thenReturn(ImmutableList.of(makeValidMetadata()));
 
-        doReturn(ImmutableList.of()).when(mockSchedulePlanService).getSchedulePlans(ClientInfo.UNKNOWN_CLIENT, TEST_STUDY, false);
```
DwayneJengSage, line 233: Should we fix this for all the other tests as well?

* physicallyDeleteSurveyNotEmptySharedModules()
* deleteSurveyPermanentlyNotConstrainedByScheduleWithMultiplePublishedSurveys()
* deleteSurveyPermanentlyNotConstrainedByCompoundScheduleWithMultiplePublishedSurveys()

DwayneJengSage (changes requested):
Mostly good. A few fixes need to be applied to other tests, and one open question about withdraw consent and external ID.

test/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
alxdarksage: I've added a test after the call to verify that the account/substudy relationship still exists and hasn't been changed, including leaving the external ID as is.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: Restored this with the correct number of arguments in all tests except for reauthTokenRequired. This method is never called in that test, an exception is thrown before it gets to that mock. If it was there to begin with, I think that was an error.

test/org/sagebionetworks/bridge/services/SurveyServiceMockTest.java
alxdarksage: Added to these tests along with mock of getSurveyAllVersions() which is required when the mocks return values correctly.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -264,7 +265,7 @@ public void adminSignInWithPhone() {
...
+        doReturn(UNCONSENTED_STATUS_MAP).when(consentService).getConsentStatuses(any(), any());
```
DwayneJengSage, line 269: In adminSignInWithEmail(), we verify the test by fetching the consent map from the session. Should we do the same in these tests?

* adminSignInWithPhone()
* emailSignInThrowsConsentRequired()
* emailSignInAdminOK()
* phoneSignInThrowsConsentRequired()

DwayneJengSage, line 269: reauthTokenRequired() should actually be reauthThrowsUnconsentedException().

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: Added

alxdarksage: Updated with the unconsented status map and a catch clause to verify it is there and is the cause of the exception being thrown.

app/org/sagebionetworks/bridge/play/controllers/ExternalIdController.java
```diff
@@ -0,0 +1,95 @@
...
+import org.sagebionetworks.bridge.models.ForwardCursorPagedResourceList;
+import org.sagebionetworks.bridge.models.accounts.ExternalIdentifier;
+import org.sagebionetworks.bridge.models.accounts.ExternalIdentifierInfo;
+import org.sagebionetworks.bridge.models.accounts.GeneratedPassword;
+import org.sagebionetworks.bridge.models.accounts.UserSession;
+import org.sagebionetworks.bridge.models.studies.Study;
+import org.sagebionetworks.bridge.services.ExternalIdService;
+
+@Controller("externalIdController")
+public class ExternalIdController extends BaseController {
```
DwayneJengSage: It looks like due to a merge error, you accidentally re-created this file.

test/org/sagebionetworks/bridge/play/controllers/ExternalIdControllerTest.java
```diff
@@ -0,0 +1,240 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+
+import play.mvc.Http;
+import play.mvc.Result;
+import play.test.Helpers;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ExternalIdControllerTest {
```
DwayneJengSage: Similarly here.

DwayneJengSage (changes requested):
Overall looks good, but it looks like the merge accidentally brought back ExternalIdControllerV3.

DwayneJengSage (approved):

PR #1820, opened 2019-01-24
Fixing ability to lookup user by external ID, adding API to migrate managed external IDs to have substudy association.

alxdarksage:
Also adding the annotations to create a GSI although this table already exists, and our DDB initialization code will not add the index, so we need to add it manually on initial deployment. Will check with Khai if this needs to be added to stack scripts.

### DIFF
--- a/app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
+++ b/app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
@@ -450,7 +450,7 @@ public class HibernateAccountDao implements AccountDao {
                         "number", unguarded.getPhone().getNumber(),
                         "regionCode", unguarded.getPhone().getRegionCode());
             } else {
-                builder.append("AND acct.externalId=:externalId", "externalId", unguarded.getExternalId());
+                builder.append("AND (acctSubstudy.externalId=:externalId OR acct.externalId=:externalId)", "externalId", unguarded.getExternalId());
             }
         }
         if (search != null) {
--- a/test/org/sagebionetworks/bridge/services/ExternalIdServiceTest.java
+++ b/test/org/sagebionetworks/bridge/services/ExternalIdServiceTest.java
@@ -160,8 +160,11 @@ public class ExternalIdServiceTest {
         
         Account account = Account.create();
         account.setId(USER_ID);
-        // This was already a managed external ID, and will be updated correctly
-        account.setExternalId(ID);  
+        // This starts as a different value and is changed to ID in the test. This would only 
+        // happen in the real world if we migrate an account associated to multiple external 
+        // IDs. We intend to finish this migration (and export) before allowing multiple 
+        // external IDs in production.
+        account.setExternalId("differentExternalId");
         
         AccountId accountId = AccountId.forExternalId(TestConstants.TEST_STUDY_IDENTIFIER, ID);
         when(accountDao.getAccount(accountId)).thenReturn(account);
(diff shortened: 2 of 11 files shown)


test/org/sagebionetworks/bridge/services/ExternalIdServiceTest.java
```diff
@@ -63,13 +71,137 @@ public void before() {
...
+        
+        AccountId accountId = AccountId.forExternalId(TestConstants.TEST_STUDY_IDENTIFIER, ID);
+        when(accountDao.getAccount(accountId)).thenReturn(account);
+        
+        // execute
+        externalIdService.migrateExternalIdentifier(study, ID,  SUBSTUDY_ID);
+        
+        // verify
+        assertEquals(SUBSTUDY_ID, extId.getSubstudyId());
+        assertEquals(1, account.getAccountSubstudies().size());
```
DwayneJengSage, line 103: We should also verify that we set the external ID on the account.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -450,7 +450,7 @@ QueryBuilder makeQuery(String prefix, String studyId, AccountId accountId, Accou
                         "number", unguarded.getPhone().getNumber(),
                         "regionCode", unguarded.getPhone().getRegionCode());
             } else {
-                builder.append("AND acct.externalId=:externalId", "externalId", unguarded.getExternalId());
+                builder.append("AND (acctSubstudy.externalId=:externalId OR acct.externalId=:externalId)", "externalId", unguarded.getExternalId());
```
DwayneJengSage, line 453: We should update integ tests to verify that this new query works.

test/org/sagebionetworks/bridge/services/ExternalIdServiceTest.java
```diff
@@ -63,13 +71,137 @@ public void before() {
...
+            .thenReturn(Substudy.create());
+        
+        ExternalIdentifier extId = ExternalIdentifier.create(TestConstants.TEST_STUDY, ID);
+        when(externalIdDao.getExternalId(TestConstants.TEST_STUDY, ID))
+            .thenReturn(extId);
+        
+        Account account = Account.create();
+        account.setId(USER_ID);
+        // This was already a managed external ID, and will be updated correctly
+        account.setExternalId(ID);
```
DwayneJengSage: Would it be better if this starts as a different external ID so we can verify it got replaced?

DwayneJengSage (changes requested):
Mostly looks good. 2 comments about unit tests and 1 about integ tests.

app/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: Checked in some new PRs to verify this. Created an account with two external IDs and verified both can be used to retrieve the account. Only possible if it looks at the associative records.

test/org/sagebionetworks/bridge/services/ExternalIdServiceTest.java
alxdarksage: Added.

test/org/sagebionetworks/bridge/services/ExternalIdServiceTest.java
alxdarksage: Changed this along with a comment about how this would be remarkable in production (it would be ok, but we aren't exporting multiple external IDs yet, so if this method found an account under ID #1, and yet it had ID #2 in the account.externalId field, this call is probably fixing something that's broken).

DwayneJengSage (approved):

PR #1824, opened 2019-02-13
S3 encryption for upload attachments.

alxdarksage:


app/org/sagebionetworks/bridge/services/HealthDataService.java
```diff
@@ -189,7 +190,11 @@ public HealthDataRecord submitHealthData(StudyIdentifier studyId, StudyParticipa
         String rawDataValue = BridgeObjectMapper.get().writerWithDefaultPrettyPrinter().writeValueAsString(
                 healthDataSubmission.getData());
         byte[] rawDataBytes = rawDataValue.getBytes(Charsets.UTF_8);
-        s3Helper.writeBytesToS3(ATTACHMENT_BUCKET, rawDataAttachmentId, rawDataBytes);
+        
+        ObjectMetadata metadata = new ObjectMetadata();
+        metadata.setSSEAlgorithm(ObjectMetadata.AES_256_SERVER_SIDE_ENCRYPTION);
```
zaro0508, line 195: Wondering if this should be parameterized because AWS buckets support both AES_256 and KMS encryption.

alxdarksage: The only encryption constant made available in the SDK is the one I'm using... I don't believe we'll ever really need to control or rotate the keys for this since the encryption happens transparently and AWS rotates the keys.

DwayneJengSage (approved):

PR #1825, opened 2019-02-18
Use the token <none> instead of an empty string so the substudy membership map properly serializes.

alxdarksage:


app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
```diff
@@ -307,7 +307,8 @@ public void setUserDataGroups(Set<String> userDataGroups) {
     /** @see #getUserSubstudyMemberships() */
     @Override
     public void setUserSubstudyMemberships(Map<String, String> userSubstudyMemberships) {
-        this.userSubstudyMemberships = userSubstudyMemberships;
+        this.userSubstudyMemberships = (userSubstudyMemberships != null && !userSubstudyMemberships.isEmpty())
+                ? userSubstudyMemberships : null;
```
DwayneJengSage, line 311: Should we add a test to HealthDataRecordTest similar to emptyDataGroupSetConvertedToNull()?

DwayneJengSage (commented):
Overall looks good. One comment about adding one test, which should be pretty straightforward.

app/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
alxdarksage: Updated

DwayneJengSage (approved):

PR #1831, opened 2019-03-07
Check consent status before returning cached session on reauth call.

alxdarksage:


DwayneJengSage (changes requested):
I'm not sure this fixes the issue. If the client reauths with the old app version, the old session will be consented. When they then reauth with the new app version, we get the session from the cache, but the session doesn't see the new app version.

DwayneJengSage (approved):

PR #1834, opened 2019-03-08
Expose consent sig timestamps

alxdarksage:
See https://sagebionetworks.jira.com/browse/BRIDGE-2408


test/org/sagebionetworks/bridge/models/subpopulations/ConsentSignatureTest.java
```diff
@@ -33,36 +33,39 @@ public void after() {
...
+        assertEquals(SIGNED_ON_TIMESTAMP.toString(), 
+                node.get("signedOn").textValue());
+        assertEquals("ConsentSignature", node.get("type").textValue());
         
         ConsentSignature deser = ConsentSignature.fromJSON(node);
         assertEquals("Dave Test", deser.getName());
         assertEquals("1970-01-01", deser.getBirthdate());
-        assertTrue(deser.getSignedOn() > 0L); // this is set in the builder
+        assertEquals(SIGNED_ON_TIMESTAMP.getMillis(), deser.getSignedOn()); // this is set in the builder
         assertEquals(0L, deser.getConsentCreatedOn());
```
DwayneJengSage, line 58: What about withdrewOn?

app/org/sagebionetworks/bridge/services/ConsentService.java
```diff
@@ -261,6 +261,7 @@ public void consentToResearch(Study study, SubpopulationGuid subpopGuid, StudyPa
             ConsentStatus status = new ConsentStatus.Builder().withName(subpop.getName())
                     .withGuid(subpop.getGuid()).withRequired(subpop.isRequired())
                     .withConsented(hasConsented).withSignedMostRecentConsent(hasSignedActiveConsent)
+                    .withSignedOn(hasConsented ? signature.getSignedOn() : null)
```
DwayneJengSage, line 264: Unit test?

test/org/sagebionetworks/bridge/models/subpopulations/ConsentSignatureTest.java
```diff
@@ -125,22 +128,24 @@ public void existingSignatureJsonDeserializesWithoutSignedOn() throws Exception
...
-        ConsentSignature updated = new ConsentSignature.Builder().withConsentSignature(sig).withSignedOn(SIGNED_ON_TIMESTAMP).build();
+        ConsentSignature updated = new ConsentSignature.Builder().withConsentSignature(sig)
+                .withSignedOn(SIGNED_ON_TIMESTAMP.getMillis()).build();
         assertEquals("test name", updated.getName());
         assertEquals("1970-01-01", updated.getBirthdate());
-        assertEquals(SIGNED_ON_TIMESTAMP, updated.getSignedOn());
+        assertEquals(SIGNED_ON_TIMESTAMP.getMillis(), updated.getSignedOn());
         
-        json = "{\"name\":\"test name\",\"birthdate\":\"1970-01-01\",\"signedOn\":-10}";
+        String timestamp = new DateTime(SIGNED_ON_TIMESTAMP, DateTimeZone.UTC).toString();
```
DwayneJengSage: Wouldn't it be better to just call SIGNED_ON_TIMESTAMP.toString()?

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/models/subpopulations/ConsentSignatureTest.java
alxdarksage: It's null, but that's wrong since I just added it to the SDK. Fixed here, will add a test to the integration tests.

alxdarksage: The ConsentSignature.fromJSON() method deserializes signatures sent from the client, and we don't allow the client to submit the timestamps. So they will not be deserialized. This looks like 0L for the consentCreatedOn timestamp because it's required, and null for withdrewOn because it can be null in the object we serialize and send to the client.

Having said that... I don't think we need the forJSON method, but that's more change in this PR to remove it.

app/org/sagebionetworks/bridge/services/ConsentService.java
alxdarksage: Added

test/org/sagebionetworks/bridge/models/subpopulations/ConsentSignatureTest.java
alxdarksage: Changed.

DwayneJengSage (approved):

PR #1837, opened 2019-03-20
Ensure consent status is correct on User-Agent dependent requests

alxdarksage:
This arose from BRIDGE-2413 where the Android client, on an update, was attempting to send a consent signature to the server even before the server session had been updated to show the consent was required.

While the client fixed this by signing in and out as part of the first run of the updated version, we would like server session to reflect the correct state given the app version, which is expressed in the User-Agent header. There are two areas where this is an issue:

- when send a consent or withdrawing from a subpopulation;
- when requesting a up-to-date StudyParticipant record for a study participant (/v3/participants/self).

This PR alters the order of operations to ensure the consent statuses are up-to-date when we look at them; it no longer looks in the session (which is cached) when it should regenerate consent status objects; and it alters the ParticipantService to be more flexible without adding any additional getParticipant(...) overloaded methods.


app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -280,40 +320,29 @@ public StudyParticipant getParticipant(Study study, Account account, boolean inc
...
+        }
+        builder.withConsentHistories(consentHistories);
+        return builder;
+    }
+    
+    private StudyParticipant.Builder addConsentStatus(StudyParticipant.Builder builder, Account account, CriteriaContext context) {
+        Map<SubpopulationGuid, ConsentStatus> consentStatuses = consentService.getConsentStatuses(context, account);
+        boolean isConsented = ConsentStatus.isUserConsented(consentStatuses);
+        builder.withConsented(isConsented);
+        return builder;
```
alxdarksage, line 341: This is called a second time when we create a session; it would be good to avoid that.

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -233,13 +233,25 @@ public void createSmsRegistration(Study study, String userId) {
...
+        AccountId accountId = BridgeUtils.parseAccountId(study.getIdentifier(), userId);
         Account account = getAccountThrowingException(accountId);
         return getParticipant(study, account, includeHistory);
     }
-
-    public StudyParticipant getParticipant(Study study, String id, boolean includeHistory) {
-        return getParticipant(study, AccountId.forId(study.getIdentifier(),  id), includeHistory);
+    
+    public StudyParticipant getSelfParticipant(Study study, CriteriaContext context, boolean includeHistory) {
+        AccountId accountId = AccountId.forId(study.getIdentifier(),  context.getUserId());
```
DwayneJengSage, line 244: Where are the unit tests for getSelfParticipant?

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -280,40 +320,29 @@ public StudyParticipant getParticipant(Study study, Account account, boolean inc
...
-            }
-        }
-        return builder.build();
+        return builder;
+    }
+    
+    private StudyParticipant.Builder addHistory(StudyParticipant.Builder builder, Account account, CriteriaContext context) {
+        Map<String,List<UserConsentHistory>> consentHistories = Maps.newHashMap();
+        // The history includes all subpopulations whether they match the user or not.
+        List<Subpopulation> subpopulations = subpopService.getSubpopulations(context.getStudyIdentifier(), false);
```
DwayneJengSage: Context is not needed, except for studyId. Would it be better to just take studyId instead?

app/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -261,6 +273,34 @@ public StudyParticipant getParticipant(Study study, Account account, boolean inc
...
         StudyParticipant.Builder builder = new StudyParticipant.Builder();
+        SubstudyAssociations assoc = BridgeUtils.substudyAssociationsVisibleToCaller(account.getAccountSubstudies());
+        addAccount(builder, assoc, account);
+    
+        CriteriaContext.Builder contextBuilder = new CriteriaContext.Builder();
+        contextBuilder.withStudyIdentifier(study.getStudyIdentifier());
+        contextBuilder.withUserId(account.getId());
+        contextBuilder.withHealthCode(account.getHealthCode());
+        contextBuilder.withUserDataGroups(account.getDataGroups());
+        contextBuilder.withUserSubstudyIds(assoc.getSubstudyIdsVisibleToCaller());
```
DwayneJengSage: Since addHistory() doesn't actually need the context, would it be better to include this in the second if (requestInfo != null) below? This keeps the code simpler and reduces the amount of code churn.

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
```diff
@@ -1109,10 +1110,8 @@ public void updateIdentifierRequiresAuthentication() throws Exception {
     
     @Test
     public void getParticipantByHealthCode() throws Exception {
-        AccountId accountId = AccountId.forHealthCode(study.getIdentifier(), "hc");
-        
         StudyParticipant studyParticipant = new StudyParticipant.Builder().withFirstName("Test").build();
-        when(mockParticipantService.getParticipant(study, accountId, true)).thenReturn(studyParticipant);
+        when(mockParticipantService.getParticipant(study, "healthCode:hc", true)).thenReturn(studyParticipant);
```
DwayneJengSage: There's probably no need to keep these "get by health code" and "get by external ID" tests, since the logic has been moved to the service.

DwayneJengSage (changes requested):

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Added

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Removed this. Once you do this, you can reorganize the code to avoid creating CriteriaContext unless you are trying to calculate the actual consent status.

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: I moved this around so we only create the context if we have the RequestInfo to get the consent status.

test/org/sagebionetworks/bridge/play/controllers/ParticipantControllerTest.java
alxdarksage: Removed

app/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: I'm going to need to refactor this so consent statuses are set, not the flag, and then the flag can be calculated from within the session. As a consequent, this expensive calculation should not need to be redone in the session code. Filed BRIDGE-2419 on this.

DwayneJengSage (approved):

PR #1843, opened 2019-03-26
[Test Refactor] Remove AuthenticationServiceTest

alxdarksage:


test/org/sagebionetworks/bridge/services/AuthenticationServiceTest.java
```diff
@@ -1,528 +0,0 @@
...
-
-    @Test(expected = BridgeServiceException.class)
-    public void signInNoEmail() throws Exception {
-        authService.signIn(study, TEST_CONTEXT, new SignIn.Builder().withStudy(study.getIdentifier()).withPassword("bar").build());
-    }
-
-    @Test(expected = BridgeServiceException.class)
-    public void signInNoPassword() throws Exception {
-        authService.signIn(study, TEST_CONTEXT, new SignIn.Builder().withStudy(study.getIdentifier()).withEmail("foobar").build());
-    }
```
DwayneJengSage, line 143: Should we add a test to AuthenticationServiceMockTest which tests invalid signIn?

test/org/sagebionetworks/bridge/services/AuthenticationServiceTest.java
```diff
@@ -1,528 +0,0 @@
...
-                .withConsent(true).withLanguages(LANGS).withSignIn(true).build();
-        
-        authService.signOut(testUser.getSession());
-        
-        Study study = studyService.getStudy(testUser.getStudyIdentifier());
-        CriteriaContext context = testUser.getCriteriaContext();
-        
-        UserSession session = authService.signIn(study, context, testUser.getSignIn());
-        assertEquals(LANGS, session.getParticipant().getLanguages());
-    }
```
DwayneJengSage, line 445: Should these language tests be moved to unit tests?

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -261,6 +263,58 @@ public void signIn() {
...
+            .withIpAddress("127.1.1.11").build();
+        
+        doReturn(account).when(accountDao).authenticate(study, EMAIL_PASSWORD_SIGN_IN);
+        doReturn(PARTICIPANT_WITH_ATTRIBUTES).when(participantService).getParticipant(study, account, false);
+        doReturn(UNCONSENTED_STATUS_MAP).when(consentService).getConsentStatuses(contextCaptor.capture(), any());
+        doReturn(REAUTH_TOKEN).when(service).generateReauthToken();
+        doReturn("SESSION_TOKEN").when(service).getGuid();
+        doReturn(Environment.PROD).when(config).getEnvironment();
+        
+        UserSession session = service.signIn(study, context, EMAIL_PASSWORD_SIGN_IN);
```
DwayneJengSage: This test almost certainly doesn't do what you expect it to do. If signIn() throws, none of the asserts after this line will be executed.

DwayneJengSage (changes requested):

test/org/sagebionetworks/bridge/services/AuthenticationServiceTest.java
alxdarksage: Added signInWithAccountNotFound() where an account isn't found for whatever reason, and signInWithBadCredentials() for a case where the SignIn object is not valid.

test/org/sagebionetworks/bridge/services/AuthenticationServiceTest.java
alxdarksage: Added a unit test to verify that persisted language preferences are used over changed language preferences in the context object.

test/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: Catching the exception so we can test that session and secret are still handled property.

DwayneJengSage (approved):

PR #1844, opened 2019-03-27
Schema and Metadata Width Limits

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-1721


app/org/sagebionetworks/bridge/validators/UploadSchemaValidator.java
```diff
@@ -89,6 +94,17 @@ public void validate(Object target, Errors errors) {
...
+
+            // fieldDef max size
+            UploadFieldSize fieldSize = UploadUtil.calculateFieldSize(fieldDefList);
+            if (fieldSize.getNumBytes() > MAX_BYTES) {
+                errors.rejectValue("fieldDefinitions",
+                        "cannot be greater than 50000 bytes combined");
+            }
+            if (fieldSize.getNumColumns() > MAX_COLUMNS) {
+                errors.rejectValue("fieldDefinitions",
+                        "cannot be greater than 100 columns combined");
```
alxdarksage, line 106: 152 cols Synapse limitation - (100 schema cols + 20 shared metadata cols) = 32 columns that are left for what?

test/org/sagebionetworks/bridge/upload/UploadUtilTest.java
```diff
@@ -32,6 +34,151 @@
...
+                .withType(UploadFieldType.TIMESTAMP).build();
+        List<UploadFieldDefinition> fieldDefList = ImmutableList.of(stringField, multiChoiceField, timestampField);
+
+        // Execute and validate.
+        UploadFieldSize fieldSize = UploadUtil.calculateFieldSize(fieldDefList);
+        assertEquals(3195, fieldSize.getNumBytes());
+        assertEquals(6, fieldSize.getNumColumns());
+    }
+
+    // todo sum of multiple fields
```
alxdarksage: Still needs these tests or should comment be removed?

alxdarksage (approved):
Are there any existing studies where you will not be able to save a study or update a schema as a result of these validation restrictions? What will we do about that? I'm a little worried because if multi answer questions in a survey generate N columns, I'm guessing we have some surveys that will throw an error if people try and change them. We might need to double-check that if that happens, the survey is not published.

DwayneJengSage:
Discussed Alx's comments on IM. In summary:

> Are there any existing studies where you will not be able to save a study or update a schema as a result of these validation restrictions? What will we do about that?

I went through every study last week and set the metadata limits to be above what's currently in use. It's possible that someone may have added more metadata fields between then and now, and if so, we'll have to manually clean those up.

As for schemas, we're pretty sure some of them are close to the Synapse limits, and the Bridge limits are intentionally lower to give us wiggle room to evolve the platform. As such, I fully expect some schemas to be no longer editable because of this. This should be fine, since you can always bump the schema rev.

> I'm a little worried because if multi answer questions in a survey generate N columns, I'm guessing we have some surveys that will throw an error if people try and change them. We might need to double-check that if that happens, the survey is not published.

Schemas are only created when surveys are published, so unpublished surveys are still editable and published surveys that work will continue to work.

That said, the rawData form for iOS-style surveys is far from ideal, and we can't support writing every survey question as a schema field. See https://sagebionetworks.jira.com/browse/BRIDGE-2416 for changing survey uploads to write a JSON file as a key-value pair of questions and answers and uploading that to Synapse as a LargeText.

app/org/sagebionetworks/bridge/validators/UploadSchemaValidator.java
DwayneJengSage: See https://sagebionetworks.jira.com/wiki/spaces/BRIDGE/pages/763166825/Bridge+Table+Width+Limits for Bridge limits. This will be moved to external-facing documentation once this has been pushed out to production.

test/org/sagebionetworks/bridge/upload/UploadUtilTest.java
DwayneJengSage: Tests were added, but I forgot to remove the comment. Comment is now removed.

PR #1851, opened 2019-04-05
Schemas can be empty, Survey Schema new format

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2416

In light of the Synapse table width limitations issue, auto-generated survey schemas will now have a single field "answers", is key-value pairs representing the survey answers. This has been written in a backwards-compatible way, such that this field will be seamlessly inserted into existing survey schemas. New survey schemas will only have the "answers" field.

### DIFF
--- a/app/org/sagebionetworks/bridge/validators/UploadSchemaValidator.java
+++ b/app/org/sagebionetworks/bridge/validators/UploadSchemaValidator.java
@@ -89,9 +89,7 @@ public class UploadSchemaValidator implements Validator {
 
             // fieldDefinitions
             List<UploadFieldDefinition> fieldDefList = uploadSchema.getFieldDefinitions();
-            if (fieldDefList.isEmpty()) {
-                errors.rejectValue("fieldDefinitions", "requires at least one definition");
-            } else {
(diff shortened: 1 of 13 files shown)


app/org/sagebionetworks/bridge/validators/UploadSchemaValidator.java
```diff
@@ -89,9 +89,7 @@ public void validate(Object target, Errors errors) {
 
             // fieldDefinitions
             List<UploadFieldDefinition> fieldDefList = uploadSchema.getFieldDefinitions();
-            if (fieldDefList.isEmpty()) {
-                errors.rejectValue("fieldDefinitions", "requires at least one definition");
-            } else {
+            if (!fieldDefList.isEmpty()) {
                 UploadFieldDefinitionListValidator.INSTANCE.validate(fieldDefList, errors, "fieldDefinitions");
             }
```
alxdarksage, line 94: Is this changing the behavior so that you can save a schema with no fields? If so I approve! It seems like such a schema might just save the rawData column, which might be all researchers want to start.

alxdarksage (approved):

PR #1858, opened 2019-05-01
Slight change so a timezone can be added in bridge-server.conf

alxdarksage:
BridgeServer2 throws errors around the timezone, which you can add to the jdbc connection string, but when you do that in the local file, you need to test and append ? or & as appropriate.

### DIFF
--- a/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
+++ b/app/org/sagebionetworks/bridge/config/BridgeSpringConfig.java
@@ -544,13 +544,13 @@ public class BridgeSpringConfig {
         BridgeConfig config = bridgeConfig();
         props.put("hibernate.connection.password", config.get("hibernate.connection.password"));
         props.put("hibernate.connection.username", config.get("hibernate.connection.username"));
-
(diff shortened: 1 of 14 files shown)


DwayneJengSage (changes requested):
This looks like a red flag. Can we find out why this works in BridgePF and not in BridgeServer2? It seems like in BridgeServer2, we're not connecting to the database properly.

alxdarksage:
We're fixing timezone to UTC. Updated to reflect where Metrics are now stored (RequestContext).

The timezone issue comes up (I believe) because Spring Boot automatically substitutes our declared JDBC driver, which is obsolete, with a newer driver. I do not see a way to prevent this as we're already configuring the JDBC driver at a very low level, and it just gets replaced.

DwayneJengSage (changes requested):
I think we should avoid making significant code changes to BridgePF. It's causing a lot of extra code churn and a lot of maintenance work over something that's going to be deleted very shortly anyway.

Can you close this PR and make a new one with just minimal changes?

PR #1859, opened 2019-05-03
Change nothing in BridgePF.

alxdarksage:


DwayneJengSage (commented):
Is the SQL timezone issue still an issue?

DwayneJengSage (approved):

alxdarksage:
No, it was an issue with Spring Boot, that change was only to keep the two Spring, Java-based configurations in sync with one another.

